Issue #15, opened 2020-12-22
Hosted GTFS validator


e-lo:
**Source**: https://github.com/MobilityData/gtfs-validator

**Timing**: We want to wait for v2.0 (likely Jan/Feb), but want it to be spun up by the spring.

**Dependencies**: This is a dependency for the following 2021 MSD milestones:
1. 2021 Q2 Launch DIY GTFS Assessment Toolkit

**Priority**: Fairly high

**Interested Parties**: MSD Team @mcplanner @e-lo

machow:
Working on this now, as part of #14. Going to try using a KubernetesPodOperator to run the validator. A big piece is making sure that it plays well in our dev environment. Otherwise, if getting the operator up takes more than a day or two, putting up a server and using the SSHOperator is a surprisingly robust solution (that can be replaced later with the right way :).

machow:
Here are some notes on using the KubernetesPodOperator. The key challenge is getting google cloud credentials to work on dev, so that we can access the data in our buckets.

## Background

**Configuring to use secrets on local dev**

* Google cloud tutorial on [adding authentication secret to kubernetes](https://cloud.google.com/kubernetes-engine/docs/tutorials/authenticating-to-cloud-platform)
* Google composer tutorial on [mounting the secret in KubernetesPodOperator](https://cloud.google.com/composer/docs/how-to/using/using-kubernetes-pod-operator#secret-config)
* Astronomer.io on [running a pod operator locally](https://www.astronomer.io/docs/cloud/stable/develop/kubepodoperator-local)
* Source code of [GKEPodOperator](https://github.com/apache/airflow/blob/5786dcdc392f7a2649f398353a0beebef01c428e/airflow/contrib/operators/gcp_container_operator.py)

**Making sure the pod doesn't hog resources**

* [See this article](https://medium.com/grensesnittet/best-practises-for-kubernetespodoperator-in-cloud-composer-91760f848a39)

**Things that seem like a bad idea**

* Attempting to [mount a bucket](https://medium.com/grensesnittet/mounting-a-gcp-bucket-as-nfs-in-kubernetes-8f6d3faf4da3). Also see this [mailing list entry](https://groups.google.com/g/cloud-composer-discuss/c/gcoaUPSAn1k), where nobody seems to know how to do it.

## Solutions

[...]

Issue #6, opened 2020-12-15
Dataset: GTFS Static Validators


hunterowens:

hunterowens:
@hunterowens to talk to @mcplanner

e-lo:
Clarification question: are the validators the dataset? or do you mean validation scores for feeds?

hunterowens:
in  progress: #51

hunterowens:
done!

PR #19, opened 2021-01-29
WIP: GTFS-Static Downloader

hunterowens:
This PR adds a script and some structure for a static GTFS downloader. It also adds code formatting and linting as a pre-commit hook.

As of now, the deployment strategy isn't figured out. I see two routes 

1) [Kubernetes Cron](https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/)
2) [Cloud Composer](https://cloud.google.com/composer) / Airflow 

Each of these have pro / cons that would be good to discuss as a team. 

TODOs

- [ ] Dockerfile / Requirements TXT 
- [ ] Deployment Strategy 
- [ ] Save to gcs or similar, not `/tmp` 
- [ ] notify when GTFS fails to parse / run and if the URL is down / down-ish 

Once w have this done, we need to pipe the collected data at the GTFS-Validator. 

Commits: 
- adding src folder
- adding catalogs folder
- Adding readme
- rename file, fix paths

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,15 @@
+name: Lint and style checks
+
+on:
+  push:
+    branches: master
+  pull_request:
+    branches: master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - uses: pre-commit/action@v2.0.0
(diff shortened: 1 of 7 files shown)


hunterowens:
ref #15 #6. 

cc @unreasonableman  @e-lo @mcplanner

hunterowens:
fixed linter!

hunterowens:
merging this as WIP to make the linter play nice.

PR #22, opened 2021-02-25
fix datetime

hunterowens:
fix a datetime.datetime.today() call to replace with datetime.datetime.now()


thekaveman (commented):
Airflow runs in UTC right? So you'll get the UTC `now`, which will be 7-8 hours ahead of California, which could be the "next day" depending on what time this runs. Just something to be aware of.

hunterowens:
Yeah, intentional for now. should read the next day to match with the
timestamps on the GCS bucket, even if not even if local midnight.

thekaveman (approved):

PR #23, opened 2021-03-11
Create Initial GTFS "Catalog" 

hunterowens:
This PR converts the existing GTFS spreadsheet from the drive and creates rows for every agency with a GTFS URL that starts with http(s). Missing agencies that have "in google" or "not in google" filled out and I haven't loaded RT feeds yet. 

I think, to load RT feeds, the best option is to join on transitland atlas via NTP ID, but curious if we've been tracking those URL(s) elsewhere.

### DIFF
--- a/data/agencies.yml
+++ b/data/agencies.yml
@@ -1,0 +1,896 @@
+ac-transit:
+  agency_name: AC Transit
+  gtfs_schedule_url:
+  - https://api.actransit.org/transit/gtfs/current?token=2512B8179D2DC44895CDDC6D42
+  itp_id: 4
+alhambra-community-transit:
+  agency_name: Alhambra Community Transit
+  gtfs_schedule_url:
+  - https://data.trilliumtransit.com/gtfs/alhambra-ca-us/alhambra-ca-us.zip
+  itp_id: 6
+altamont-corridor-express:
+  agency_name: Altamont Corridor Express
+  gtfs_schedule_url:
+  - https://transitfeeds.com/p/altamont-corridor-express/823/latest/download
+  itp_id:
+amador-regional-transit-system:
+  agency_name: Amador Regional Transit System
+  gtfs_schedule_url:
+  - http://data.trilliumtransit.com/gtfs/amador-ca-us/amador-ca-us.zip
+  itp_id: 11
+anaheim-resort-transportation:
+  agency_name: Anaheim Resort Transportation
+  gtfs_schedule_url:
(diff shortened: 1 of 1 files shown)


e-lo (commented):
LGTM in general.
Two thoughts:
1. Create an entry for all transit providers even if they don't currently have a feed (empty list).
2. Add documentation [ somewhere ] including any implied meaning from ordering of feeds in the list.

hunterowens:
1. I thought about this but decided against it since the master "list" remains the spreadsheet, and therefore, to determine something like % listing GTFS feeds, you should do a left join. Additionally, there's the open question that invites on  should we have "empty list but only for agencies that should have GTFS (ie, operate fixed route, etc) 
2. Once the list has implied meaning, I'll add docs for it. For now, we haven't incorporate the MTC subfeeds etc

e-lo:
> 1. I thought about this but decided against it since the master "list" remains the spreadsheet, and therefore, to determine something like % listing GTFS feeds, you should do a left join. Additionally, there's the open question that invites on  should we have "empty list but only for agencies that should have GTFS (ie, operate fixed route, etc)

machow:
Hey--I'm noticing the gtfs_schedule_url in the yaml config is a list, but a single column in the sheet. Is this by design? What should happen when there are two schedule urls?

(guessing there is a good reason for it, so this question is more for my understanding :)

hunterowens:
Yep! You might want to take a look at #21, which a very long thread about the nature of feeds. 

at least one agency in CA, LA metro produces two feeds (one rail / one bus), so rather than store that with a , separating (as in th sheet) I split into a list. 

additionally, every agency in the MTC region (bay area) often produces their own feed plus participates in the [Bay Area Regional GTFS](https://www.interline.io/blog/mtc-regional-gtfs-feed-release/) that _should_ be the same, but we should list both in the catalog

machow:
Ah, thanks--I've been crawling down #21, and think it is slowly coming together. Guessing that this means the "primary key" for each url entry is <agency id> x <url entry index>?

Should work okay in most circumstances, but I wonder if weird things could happen if there were two url entries, and the first ended up being removed.

hunterowens:
Yep, since for storage each run is the {run_time}/{itp_id}/{url}/(zip
contents), if we remove one from the list, it should just not be there for
future runs./

hunterowens:
gonna merge this knowing that it gets moved and updated in #25

Issue #24, opened 2021-03-16
Update GTFS Downloader to use the new `yml` based file


hunterowens:
Currently, the GTFS downloader uses a Google Sheet and then uses intake / pandas to read the data in at runtime, establishing a list of urls that should contain zipfiles that represent a single agencies GTFS feed. 

With #23 being ready to merge, we are now storing that information in a yml file that for now, lives in this repo (which I think makes sense until the Mobility Database is relaunched). 

This issue to update the DAG to use the yml file rather than google sheets. 

FYI @machow, when I was originally working on this, I was unable to determine a way to dynamically generate a list of tasks based on the Google Sheets in a DAG, because ideally I wanted each task to represent downloading one agency. Instead, I used `try / except` blocks to email at the end of the DAG a list of agencies that failed to parse or download. I think if possible to dynamically generate tasks would be good to have 1 task = 1 agency but take a look at the code and let me know what you think. 

some useful background reading will be #23 #21 #20

hunterowens:
this is done!

PR #25, opened 2021-03-16
Use agencies.yml in airflow tasks

machow:
Addresses #24. Note that I moved `data` into the airflow folder, so it could be volume mounted.

TODO:

- [x] : ~~generate one task per provider~~
- [x] : handle case of multiple urls per agency(?)

Note that airflow 2.0 introduced [task groups](https://airflow.apache.org/blog/airflow-two-point-oh-is-here/#task-groups-aip-34), which will help us to sanely view the ~180 separate tasks in a dag! (Might also work with a subdag, but afaik they have some weird limitations; mentioned in the task group article).

## Key changes

* **refactored to use https://github.com/chriscardillo/gusty**. In this approach, each dag is a folder with a file named `METADATA.yml` for configuration. Each file in the folder is a task. When a file is a .py, then gusty creates a PythonOperator from it, using a yaml header for configuration. This makes it easy to separate out tasks, and add new ones.
* made sending an email of failed agency urls its own task.
* now save a csv that is essentially a table representation of `agencies.yml`, with a column for whether the url download succeeded or how it failed.

## Misc notes

* I set the pre-commit hook to be less strict with yaml, since this allows gusty to do full dag configuration in metadata.yml
* do we need `dags/catalogs`? Can it be moved into the data folder? (don't have strong opinions here)
* I'm still getting a feel for how `gcsfs` works, and am guessing that the full picture will come together once we get some extract + load code set up!


hunterowens:
:+1: NB, [Google Cloud Composer](https://cloud.google.com/composer/docs/concepts/versioning/composer-versions), which we use for SAAS manged deploys as our production env, only supports airflow 1.10.14 and lower right now so we are still on the old version

airflow/dags/gtfs_static_downloader.py
```diff
@@ -22,22 +24,23 @@ def make_gtfs_list():
...
+    df["gtfs_schedule_url"] = df["gtfs_schedule_url"].str.get(0)
+
     # TODO: Figure out what to do with Metro
     # For now, we just take the bus.
 
     # TODO: Replace URLs with Zip ones.
     # For now we filter, and then remove ones that don't contain
     # zip filters
-    df = df[(df.GTFS.str.contains("zip")) & (df.GTFS.notnull())]
+    df = df[(df.gtfs_schedule_url.str.contains("zip")) & (df.gtfs_schedule_url.notna())]
```
hunterowens, line 43: we should probably go through the list and identify which ones are not returning zip files and replacing the urls as need.

hunterowens: For example, Glendale Beeline doesn't have `.zip` in the URL but does successfully return a zip file when you go to the following url http://glendaleca.gov/Home/ShowDocument?id=29549

machow: That makes sense--if it works for you, I can create a quick notebook tomorrow that checks the urls and chuck it on the notebooks repo!

hunterowens: perf! Also adding to the agenda for the MSD meeting at 2:30 (check your calendar for an invite!)

machow:
Ah, okay--thanks for the heads RE airflow. I can take a stab at using a subdag for it tomorrow

hunterowens:
I'm also okay implementing it as a single task, just want to make sure we
are capturing which agencies fail to downtown or fail to be a single zip
file at URL

machow:
That makes sense--let me quickly test out what some of the views in the UI look like with subdags, since they might give a nice quick glimpse of success for agency x date.

Similar to this heat map, but with agency on the rows

image

hunterowens:
:+1:

machow:
Alright--after reading / discussing a bit more with an old colleague, it sounds like SubDAGs have some weird properties (see [this article](https://www.astronomer.io/guides/subdags); also they're deprecated in airflow).

I refactored the tasks to use gusty. This makes each task its own file, so it's a bit easier to add / modify them. It seems like the play here is just to store a file next to the extracted data in gcs (e.g. a csv with agency_id, url, and status columns) that reports on the result.

machow:
Dumped a list of urls with issues here: https://github.com/cal-itp/notebooks/pull/3, will wrap this up tomorrow morning

machow:
Alright, I think it's ready for review--added notes to the top

airflow/dags/gtfs_downloader/email_failures.py
```diff
@@ -0,0 +1,28 @@
...
+
+    # email out error agencies
+    email_template = (
+        "The follow agencies failed to have GTFS a GTFS feed at"
+        "the URL or the Zip File Failed to extract:"
+        f"{error_agencies}"
+        "{{ ds }}"
+    )
+    send_email(
+        to=["ruth.miller@dot.ca.gov", "hunter.owens@dot.ca.gov"],
```
hunterowens, line 23: @machow add yourself to this list

hunterowens:
This LGTM! One small comment and let's schedule a time to merge

hunterowens:
/ make sure I document / share with you access to the production assets

PR #34, opened 2021-03-23
fix: google cloud dag dir is not in AIRFLOW_HOME

machow:
Quick fix. The gist is that in our dev environment, the dags folder lives inside AIRFLOW home. However, google cloud uses `/etc/airflow` for home, and has dags in `/home/airflow/gcs/dags`.


hunterowens:
ah nice catch!

machow:
your response time was incredible..!

Issue #35, opened 2021-03-23
DAGs bucket adds new files, but does not remove old ones on deploy


machow:
When shelled in, I noticed that files deleted in this repo still seem to be in the dags folder:

![image](https://user-images.githubusercontent.com/2574498/112075907-664a0d00-8b47-11eb-9302-ff1ecae088c7.png)

I think google-github-actions/upload-cloud-storage action only adds files. Will open a PR in a sec with a proposed fix, then swing back and double check tomorrow.

PR #36, opened 2021-03-23
remove old files from dags folder on gcloud

machow:
addresses #35. Will double check in the morning. Pulled the sync command out of [this article](https://www.mickaelvieira.com/blog/2020/01/29/deploying-a-static-website-to-google-cloud-storage-with-github-actions.html
):

## Notes

* the bucket the dag dir is being uploaded to also has the whole data-infra repo in it (from some point in time?)
* tested by creating a new bucket `gtfs-data-test` and syncing a folder to it, deleting file, re-syncing.
* the dag bucket currently has a task called `airflow_monitoring.py`. Looks like it is [automatically added](https://stackoverflow.com/questions/56496779/why-is-there-a-dag-named-airflow-monitoring-automatically-generated-in-cloud-c) for health checks.
* for some reason on the 19th a job used the full datetime-stamp when saving data, rather than only the date (see image below). Guessing it's related to having two files in the dags generating dags with the same ID or something? Will be easier to diagnose once this is merged, and hopefully merging fixes.

image

### DIFF
--- a/.github/workflows/push_to_gcloud.yml
+++ b/.github/workflows/push_to_gcloud.yml
@@ -21,9 +21,4 @@ jobs:
       with:
         service_account_key: ${{ secrets.GCP_SA_KEY }}
         export_default_credentials: true
-    - id: upload-files
-      uses: google-github-actions/upload-cloud-storage@main
-      name: upload-the-file
-      with:
-        path: airflow/dags
-        destination: us-west2-calitp-airflow-pro-332827a9-bucket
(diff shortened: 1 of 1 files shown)


machow:
Note that apparently the gsutil command breaks because a package compiled w/ google's sdk uses code that's not python 3 compatible (`sys.maxint`). Going to fix on my machine, but guessing based on the tutorial it's not an issue in the gh action?

https://github.com/GoogleCloudPlatform/gsutil/issues/1052

machow:
@hunterowens I edited notes into the top comment!

hunterowens:
this LGTM

hunterowens:
this worked 
image

Issue #37, opened 2021-03-23
Cleaning out copy of this repo from the bucket holding dags folder


machow:
Currently, the bucket holding DAG files also has a copy of this repo from some time around Feb 19-24th (see image below). If seems like we remove everything that is not the dags folder, but cc @hunterowens in case it is used in some way. I'm not super familiar with cloud composer, so could be totally wrong here!

![image](https://user-images.githubusercontent.com/2574498/112180768-d8633600-8bc9-11eb-9fb8-363bec7e5fca.png)

machow:
Noticed in this [google cloud doc](https://cloud.google.com/composer/docs/how-to/using/using-kubernetes-pod-operator#troubleshoot)

> Output of the Airflow scheduler and workers:
>   1. Go to the Cloud Composer environment's Cloud Storage bucket. This is bucket where the DAGs are located.
>  2. Review the logs at logs/DAG_NAME/TASK_ID/EXECUTION_DATE.

hunterowens:
I think this is probably from my testing and futzing around w/ gcloud and gh actions

machow:
I finally stumbled on a list of folders composer adds to the bucket (dags, data, logs, plugins)--

https://cloud.google.com/composer/docs/concepts/cloud-storage#folders_in_the_bucket

This made me wonder when things like our data folder is uploaded. Since it seems like the airflow server is only updated when requirements.txt changes. I wonder if that's when the data folder is also synced? I tried re-running the deploy job, but it failed because the requirements.txt file hadn't actually changed..

hunterowens:
the main [deploy](https://github.com/cal-itp/data-infra/blob/main/.github/workflows/push_to_gcloud.yml) job should sync everything in `dags` on the repo to `dags`. hence why `data/agencies.yml` needs to live inside `dags`

The secondary job is run the update requirements.txt command. might be good to change names

machow:
This is what I'm a bit surprised about--`agencies.yml` lives in `airflow/data`. We can sync that to the data folder in the bucket, and it seems like it will be mounted alongside dags (in the `$AIRFLOW_HOME` directory) for worker nodes (screenshot from worker via gcloud shell below). Or maybe it syncs when re-deploying airflow?

![image](https://user-images.githubusercontent.com/2574498/112346362-e2ea0200-8c93-11eb-9265-dd5c9cf83758.png)

In any event, I'll dig a bit deeper to get a feel for how these pieces fit together..

[...]

PR #40, opened 2021-03-24
sync dags, plugins, and data

machow:
addresses #37


hunterowens:
this looks good to me? basically everything should sync between plugins, data, and dags folders?

hunterowens:
looks like the precommit hooks are failing on thje `.gitkeep` file @machow. Let me know if you need help

machow:
Ah, thanks--I made the mistake of trying to add .gitkeep via github :/. I noticed a task is failing to load the agencies file on prod, even though I see it in data. Added a couple temporary lines to print out info on where the task is running, just to sanity check whether the prod and dev setups are running differently.

PR #41, opened 2021-03-26
WIP: Daily Blackcat Downloader 

hunterowens:
This is an initial stub of a daily blackcat ETL function that creates a daily backup on GCS. Still a bit unsure on how to make it a task using gusty so that is the todo.

### DIFF
--- a/airflow/dags/blackcat/download_to_gcs.py
+++ b/airflow/dags/blackcat/download_to_gcs.py
@@ -1,0 +1,58 @@
+import requests
+import gcsfs
+import os
+import logging
+import datetime
+import pandas as pd
+from sqlalchemy import create_engine
+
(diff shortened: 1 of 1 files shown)


machow:
In case it's useful, I think the two gusty steps are...

1. update dags.py to include a path to the `blackcat` folder [here](https://github.com/cal-itp/data-infra/blob/blackcat-etl/airflow/dags/dags.py#L21)
2. copy out the yaml header from another task, and set `download_and_save_blackcat` in the python_callable field.

hunterowens:
depends on #43

machow:
@hunterowens what's the priority on this? (Is this a thing that becomes important for payments data?)

machow:
Closing since this is 5 months old. Happy to work on this if it's still relevant (and we lay out in an issue)

PR #50, opened 2021-04-05
add MobilityData list of initial GTFS-RT urls

hunterowens:
This PR starts the process of tracking GTFS RT urls. Based on a dump from Maxime at Mobility Data, this is the valid-ish URLs I could find. 

Next batches to process
 
 - TransitLand
 - GRaSS Clients

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -252,8 +252,8 @@ desert-roadrunner:
   gtfs_schedule_url:
     - http://data.trilliumtransit.com/gtfs/paloverde_valley-ca-us/paloverde_valley-ca-us.zip
   itp_id: 238
-dinuba-area-regional-transit-:
-  agency_name: "Dinuba Area Regional Transit "
+dinuba-area-regional-transit:
+  agency_name: "Dinuba Area Regional Transit"
   gtfs_schedule_url:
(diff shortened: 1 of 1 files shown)


thekaveman:
Incoming commit with some yaml formatting cleanup.

airflow/data/agencies.yml
```diff
@@ -1,698 +1,752 @@
...
   itp_id: 91
 desert-roadrunner:
   agency_name: Desert Roadrunner
   gtfs_schedule_url:
-  - http://data.trilliumtransit.com/gtfs/paloverde_valley-ca-us/paloverde_valley-ca-us.zip
+    - http://data.trilliumtransit.com/gtfs/paloverde_valley-ca-us/paloverde_valley-ca-us.zip
   itp_id: 238
 dinuba-area-regional-transit-:
-  agency_name: 'Dinuba Area Regional Transit '
+  agency_name: "Dinuba Area Regional Transit "
```
thekaveman, line 256: Is the `-` at the end of the agency key / space at the end of the name correct?

hunterowens:
Nope, I think that is a processing from the spreadsheet artifact, AFAIK

airflow/data/agencies.yml
thekaveman: Fixed

PR #51, opened 2021-04-05
add gtfs validator via pod operator

machow:
This PR adds a new task for running the gtfs-validator on our data stored in google cloud, and saving the results.

* uses GKEPodOperator with https://github.com/cal-itp/gtfs-validator-api container
* the previous task returns via xcom the paths of gtfs data to validate
* results are stored in the same folder as the data, and named `validation.json` (e.g. `{bucket}/schedule/{execution_date}/{itp_id}/{url_number}/validation.json`)
* running locally still runs the pod in our production cluster

The implementation is rough, since it does not specify resource constraints, etc.., for the pod. I'm going to sync with @lottspot from Jarvus on looking things down on the kubernetes side. But since our airflow pipeline does have many tasks right now, I'd vote for us merging this PR early, so we can work on loading to bigquery at the same time.


hunterowens:
:+1: to merging early, the cluster is not very resource constrained ATM.

Issue #59, opened 2021-04-13
investigate loading validation (JSON) data into bigquery


machow:
Right now the gtfs-validator returns a JSON payload. [This notebook](https://github.com/cal-itp/notebooks/blob/main/notebooks/loading-gtfs-schedules.ipynb) shows how it can be turned into a table, with 1 row per notice (see `tidy_notice_details` table), and either..

* a column of JSON entries
* many columns

A big question though is how BigQuery likes to deal with these kinds of tables. AFAIK there are three potential options:

* many columns (with mostly NULL values)
* 1 column of nested data
* separate tables for details generated by different error codes
  - (I think bigquery queries can treat many tables like one in some cases..)

I'm guessing one of the first two makes most sense.

hunterowens:
it's also more important to track validator results over time 

so I think we might want something like 

{each file} -> {check if same SHAHASH of zip compared to prior day run ->if new GTFS , save validator with run ID

machow:
Ah--that's helpful to hear! Maybe a place to start is a `validator_latest` table, and then a `gtfs_schedule_change` table with one row per agency x change in zip (that could later be unpacked into more details).

(I can run checks on data changes in notebooks first to get a feel for how often it's changing)

machow:
Alright, maybe there are two stages to do this in. Will edit in a bit more detail, but wanted to put here for now.

## 1. low-hanging fruit

Questions answered:

* validation reporting: validator is currently reporting these X serious messages
* validation sanity check: validator is currently reporting X, why does the data look like Y?
* specific aggregate validation history: your agency had N severity level "ERROR" codes on Y date

Technical:

* upload validator latest tables, so we've got latest everything
* run change detection on tables, store as a csv
  - create a table summarizing when datasets changed
  - create a table summarizing when validation changed
* upload some aggregate validation history table

## 2. full table history

Questions answered:

* validator report changed for X agency, what changed in their data?

Technical:

* will likely need to use an external table in bigquery

hunterowens:
@machow let's close this?

PR #60, opened 2021-04-13
Load gtfs schedule data into bigquery

machow:
This PR adds the loading of the 5 required GTFS static tables for agencies: agency, stops, stop_times, routes, trips. More can be added by adding simple yaml files to the `gtfs_loader` DAG.

Notes:

* data loaded to `gtfs_schedule` dataset at the moment (e.g. `gtfs_schedule.stops`, `gtfs_schedule.agency`)
* tables columns are loaded with mostly STRING types, so analysts can diagnose misspec'd agencies.
* wrangling airflow's BigQuery operators took a lot of work; we should move away from it down the road!
* changed the paths to be {itp_id}_{url_number}, instead of {itp_id}/{url_number}. This makes it easy to use with bigquery, which only allows a single wildcard in paths.

## TODO before merging

- [x] use environment variable to toggle bigquery dataset (its term for schema?) being loaded to.
- [x] quickly toss up `calitp_status`, `calitp_files`, and `calitp_load_errors` tables (using `calitp` here to prefix data that reports warehouse internals / details)

Going to punt loading validation data to https://github.com/cal-itp/data-infra/issues/59.

## Follow-up tasks

* prepare to package calitp.py. I would just package it in the root of this repo for now--will be possible to pip install. Lets us keep moving it quickly, but also share it outside the pipeline. Once things stabilize, seems useful to make its own package.
* add bigquery fuctionality to calitp.py (and break it into modules)
  * loading into bigquery
  * analysts querying bigquery

### DIFF
--- a/airflow/dags/gtfs_loader/METADATA.yml
+++ b/airflow/dags/gtfs_loader/METADATA.yml
@@ -1,0 +1,17 @@
+description: "Download the state of CA GTFS files, async version"
+schedule_interval: "0 0 * * *"
+tags:
+  - all_gusty_features
+default_args:
+    owner: airflow
+    depends_on_past: False
+    start_date: !days_ago 1
+    email:
+      - "hunter.owens@dot.ca.gov"
+    email_on_failure: True
+    email_on_retry: False
+    retries: 1
+    retry_delay: !timedelta 'minutes: 2'
+    concurrency: 50
+    #sla: !timedelta 'hours: 2'
+latest_only: False
--- a/airflow/dags/gtfs_loader/calitp_files_process.py
+++ b/airflow/dags/gtfs_loader/calitp_files_process.py
@@ -1,0 +1,42 @@
+# ---
+# python_callable: main
+# provide_context: true
+# external_dependencies:
+#   - gtfs_downloader: download_data
+# ---
+
+from calitp import read_gcfs, get_bucket, save_to_gcfs
+import pandas as pd
+import gcsfs
+
+
+def main(execution_date, **kwargs):
+    # TODO: remove hard-coded project string
+    fs = gcsfs.GCSFileSystem(project="cal-itp-data-infra")
+
+    bucket = get_bucket()
+
+    f = read_gcfs(f"schedule/{execution_date}/status.csv")
+    status = pd.read_csv(f)
+
+    success = status[lambda d: d.status == "success"]
+
+    gtfs_files = []
+    for ii, row in success.iterrows():
+        agency_folder = f"{row.itp_id}_{row.url_number}"
+        gtfs_url = f"{bucket}/schedule/{execution_date}/{agency_folder}/*"
+
+        gtfs_files.append(fs.glob(gtfs_url))
+
+    res = (
+        success[["itp_id", "url_number"]]
+        .assign(gtfs_file=gtfs_files)
+        .explode("gtfs_file")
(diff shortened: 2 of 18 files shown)


hunterowens:
playing around with the dataset, it doesn't look like all the tables have calitp ids attached? only looks like the agency table so far. 

since agency_ids  are set by the agency, they are non-unique, which prevents joins. make sure to capture the ITP ID in all the tables

machow:
For putting things up on dev--I wonder if we should start a separate project for holding dev data on bigquery. I can add a prefix/suffix to dataset names, but it seems nice to be able to see only the production tables (similar to being able to switch buckets for storage). WDYT?

Not urgent--I'll add prefixes to dev table names for now, but wanted to queue up discussion

image

machow:
@hunterowens--alright, should be ready! The subdags are pretty gnarly, and within the next couple PRs, it seems worth breaking the operators into their own folders (maybe when making calitp packageable).

machow:
DAG image for reference
image

airflow/dags/gtfs_loader/METADATA.yml
```diff
@@ -0,0 +1,17 @@
+description: "Download the state of CA GTFS files, async version"
+schedule_interval: "0 0 * * *"
+tags:
+  - all_gusty_features
+default_args:
+    owner: airflow
+    depends_on_past: False
+    start_date: !days_ago 1
+    email:
+      - "hunter.owens@dot.ca.gov"
```
hunterowens, line 10: Add @machow

airflow/dags/gtfs_loader/calitp_files_process.py
```diff
@@ -0,0 +1,42 @@
...
+
+    f = read_gcfs(f"schedule/{execution_date}/status.csv")
+    status = pd.read_csv(f)
+
+    success = status[lambda d: d.status == "success"]
+
+    gtfs_files = []
+    for ii, row in success.iterrows():
+        agency_folder = f"{row.itp_id}_{row.url_number}"
+        gtfs_url = f"{bucket}/schedule/{execution_date}/{agency_folder}/*"
```
hunterowens, line 27: how does this handle multiple files per agency?

machow: Ah, sorry--one thing I should have included in the notes is that the BigQuery can only handle a single wildcard in a bucket path. To avoid the need to constantly use `fs.glob` (which can do multiple wildcards), I changed the paths to be {itp_id}_{url_number}. 

I think it's only an issue because we have 2 calitp specific table keys in this case. Shouldn't be an issue for the GTFS-RT data, since it only has 1 key (itp_id), and 3 tables / blobs (e.g. vehicle_positions).

[Relevant bigquery docs](https://cloud.google.com/bigquery/docs/batch-loading-data#load-wildcards)

(edited this detail in to the top post just now, so it's easy to find)

hunterowens: ahhh makes sense

hunterowens: just for any joins etc need to do a combo of {key * ID number * url number}?

hunterowens: for now this is just Metro too so not a big issues.

Issue #56, opened 2021-04-12
Investigate HTTP connection pool error w/ Humbolt / Trinity feeds


hunterowens:

hunterowens:
This appears to be taking down the entire task 
```
[2021-04-13 00:07:17,520] {taskinstance.py:1152} ERROR - HTTPSConnectionPool(host='data.trilliumtransit.com', port=443): Max retries exceeded with url: /gtfs/cudahy-ca-us/cudahy-ca-us.zip (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7fc67a748400>: Failed to establish a new connection: [Errno 110] Connection timed out',))
Traceback (most recent call last)
  File "/opt/python3.6/lib/python3.6/site-packages/urllib3/connection.py", line 160, in _new_con
    (self._dns_host, self.port), self.timeout, **extra_k
  File "/opt/python3.6/lib/python3.6/site-packages/urllib3/util/connection.py", line 84, in create_connectio
    raise er
  File "/opt/python3.6/lib/python3.6/site-packages/urllib3/util/connection.py", line 74, in create_connectio
    sock.connect(sa
TimeoutError: [Errno 110] Connection timed ou

During handling of the above exception, another exception occurred

Traceback (most recent call last)
  File "/opt/python3.6/lib/python3.6/site-packages/urllib3/connectionpool.py", line 677, in urlope
    chunked=chunked
  File "/opt/python3.6/lib/python3.6/site-packages/urllib3/connectionpool.py", line 381, in _make_reques
    self._validate_conn(conn
  File "/opt/python3.6/lib/python3.6/site-packages/urllib3/connectionpool.py", line 978, in _validate_con
    conn.connect(
  File "/opt/python3.6/lib/python3.6/site-packages/urllib3/connection.py", line 309, in connec
    conn = self._new_conn(
  File "/opt/python3.6/lib/python3.6/site-packages/urllib3/connection.py", line 172, in _new_con
    self, "Failed to establish a new connection: %s" % 
urllib3.exceptions.NewConnectionError: <urllib3.connection.HTTPSConnection object at 0x7fc67a748400>: Failed to establish a new connection: [Errno 110] Connection timed ou

During handling of the above exception, another exception occurred

Traceback (most recent call last)
  File "/opt/python3.6/lib/python3.6/site-packages/requests/adapters.py", line 449, in sen
    timeout=timeou
[...]

PR #62, opened 2021-04-13
Fix download data failing due to http retries

machow:
Addresses #56, in the sense that the task now succeeds. I'm not sure what's going on with the urls. I tried one on my laptop, and it worked okay--but they may be blocking our pipeline's IP / user-agent?


hunterowens:
interesting! I wonder if we are getting IP blocked? The user agent IIRC I futzed with before. Let's ask Aaron inside of the slack maybe?

Issue #48, opened 2021-04-01
Fix template: timestamp in agency failure email is showing {{ds}}


machow:
Looks like it's this line: https://github.com/cal-itp/data-infra/blob/main/airflow/dags/gtfs_downloader/email_failures.py#L20

TODO:

* add spaces at end of each line in template string
* fix {{ds}} showing up raw

PR #63, opened 2021-04-15
feat: add table of urls, errors to email

machow:
addresses #48, and adds an html table of results. Note that sendgrid isn't configured in the container, so I have tested that it attempts to send (and viewed the contents). Since this is the only email we're sending, I'd vote for merging, and wait to get things set up on dev if/when we start sending more emails.


hunterowens (approved):

hunterowens:
Sounds good! @machow can you trigger a test job so we can see before 4pm if this works?

PR #64, opened 2021-04-15
[WIP] add Michael to task failure emails

machow:
@hunterowens are you also getting emailed on dev failures? If so, I can disable that before merging, so we don't get blasted with failure emails (sorry it took me getting added to the emails for me to ask this question ;).


hunterowens:
I haven't seen any dev related ones!

Issue #66, opened 2021-04-15
validator failing intermittently


machow:
May be related to an addressed gcsfs issue https://github.com/dask/gcsfs/issues/315. Going to bump gtfs-validator-api to use a newer gcsfs version. Otherwise, I'll investigate whether its a kubernetes resource issue. 

From looking at the airflow logs, it seems to fail when either using fs.get, or fs.pipe (depending on how far it gets).

hunterowens:
ok cool. saw that you cut the new release

hunterowens:
man it is wild how much less mature the GCSFS compared to s3fs / boto tooling is.

machow:
yeah--for sure, big ups to dask for maintaining gcsfs. I'm still pretty curious what the best route forward is for pulling out the main bigquery ops we'll want to do from airflow is... (or whether we should just use their hooks outside of airflow...)

machow:
Hmm.. no luck with upgrading gcsfs. It works when we run kube pods via the dev pipeline, and when I run locally. I'm going to check w/ @lottspot to see if we can pair on checking kube stuff next week. Even though dev and prod should be running in the same cluster, the task seems to run longer on prod (and then fail).

In the meantime, I'll..

* do a run on dev and prod that doesn't delete the pods, so we can take a closer look at them.
* Use docker on my laptop each morning, to run the validator over the prod bucket.

machow:
Reviewed w/ @lottspot yesterday--and he pointed out that the airflow workers are using a high proportion of their memory. I've been taking a closer look at metrics, and wanted to dump some findings.

**tl;dr: we should either beef up our workers, or cut the number of concurrent celery workers (e.g. from 6 to 2). I don't know how to stop the growing memory usage. Once @lottspot is done with the GTFS-RT work, maybe we can dig in to this + the pricing story (#45) further.**

* our custom pod task (running the gtfs-validator) doesn't much memory (~200mb), BUT..
* there seems to be growing memory usage in worker pods, caused by a container called gcs-syncd.

Before I show graphs, here are some resources:

* [upgrading gke nodes](https://cloud.google.com/composer/docs/how-to/managing/updating#upgrading_the_machine_type_for_gke_nodes)

I'll break this into two sections:

* overview: lots of graphs on memory usage
* possible causes

## Overview

| graph |  |
| --- | --- |
[...]

PR #67, opened 2021-04-16
Fix validator

machow:
Should address #66. I tested locally by running gtfs-validator-api manually, and on the dev airflow. Unless there's some issue specific to the KubernetesPodOperator that we run on composer only, then should resolve!


machow:
Since the validator is currently error'ing out on prod, going to merge so I can make sure it works...

hunterowens:
LGTM!

PR #71, opened 2021-04-19
Update agencies.yml

machow:
An entry was named `Agency Name`, rather than `agency_name`. Normally this would cause the pipe to fail, only to create a new column of data, but a sort of complex set of events happened...

* it created an `Agency Name` column of data in status.csv, in addition to `agency_name`
* BigQuery automatically wanted to rename it to `agency_name`
* BigQuery complained that there were two `agency_name` columns, causing an error

(merging so I can re-run airflow)


hunterowens:
Whoops, my bad while doing copy/ paste to get additional D4 agencies on
there

machow:
No worries! It was the bigquery autoschema renaming columns that stopped things up

Issue #70, opened 2021-04-16
create golden-record-style "merged GTFS table"


hunterowens:
Building on #69, we should create a golden record/  merged GTFS table that has all of the foreigns into a main wide table that can be queried, summarized and aggegrated. 

the foreign keys can be found at https://developers.google.com/transit/gtfs/reference

machow:
What grain are you thinking of for the table? One row per stop, including its trip and route info (w/ stop ids duplicated if stops can have multiple trips)? It seems like the three grain sizes (from smallest to largest) are...

* stops
* trips
* routes

(I'm still wrapping my head around the data, so could be totally off! putting this partridge table here, since I keep checking it)


![image](https://user-images.githubusercontent.com/2574498/115077619-5ca89080-9ecc-11eb-8ade-a77416cee709.png)

machow:
Looking into joining agencies, routes, and trips ("agency_trips"). One thing I'm noticing is that some routes have NULL agency_ids.

From the GTFS spec, it looks like for routes.txt, agency_id is conditionally required (when there is only a single agency). It seems like a good opportunity for to transform in the implicit agency_id in this case! WDYT? I'm curious if there are other classic cases of this happening in the GTFS spec!

machow:
Complicating this, some agencies don't specify an agency id, so we will need to make one for them (for example, the agency_name, or a number like 0, or -999 as a small cue we put the id in?)

e-lo:
I think the defacto `agency_id` would be 0 or 1 (depending on indexing).  

Best practice is to add it and keep it stable even though it is only conditionally required.   That said, it could, in-fact, change.

Another thing to consider is that `agency_id` is supposed to represent a customer-facing "brand" not necessarily the agency operating or funding the service.

e-lo:
> I'm surprised this is in the spec, since it is a big nuisance for cross-dataset analyses 😬

--> totally agree...but not what GTFS was designed for. Feel free to put in a PR to the spec to see if it passes a vote! ;-)

machow:
@cvc5185 I have a kind of nuts query that creates on big history table. Do you mind putting it into gtfs_schedule_type2, and then querying the latest data from it (e.g. `calitp_extracted_at IS NULL`)?

This would involve creating two tasks

**gtfs_schedule_type2/route_trip_stops.sql**:

* depends on merge_updates
[...]

Issue #69, opened 2021-04-16
UPSERT gtfs schedule data


machow:
Should need these steps:

* add a calitp_extracted_at column
* primary key fields in the yaml schemas
* use a merge statement for moving in staging tables

machow:
Since we are not doing a rowwise operation, but an "agency-wise" one, afaik we'll need to do two steps:

* drop rows matching agencies w/ data for the day
  - (select distinct calitp_itp_id, calitp_url_number from the staging table)
* insert those agencies' full table of data

Could also just copy data into a new bucket (e.g. schedule/processed/{itp_id}_{url_number}/{table_name}). New data will re-write old ones. Would need to remove an agencies' folder first, in case they removed files.

hunterowens:
I think this might be simpler, which is to say

create a new table from most recent run, look to see which agencies are
missing, copy their data from the most recent "good" run.

machow:
I've been reading into bigquery's external tables, and think we might be able to take advantage of...

* a historical table of all agency files, including gcs's calculation of file md5 hashes for them
* an external table (sitting on top of gcs bucket) holding every changed agency's data

On a given run, we can filter the external table to be only the latest data. If we filter using the special `_FILE_NAME` column with exact file names, then it only queries the relevant files. A big benefit of this is that we could also look up specific instances from history of agency datasets / validator results (I can set the table so it only allows queries using efficient `_FILE_NAME` filters).

PR #74, opened 2021-04-23
feat: load validation report, and unnest

machow:
Addresses #59:

### Features

* loads the validation reports into bigquery (1 row per itp_id)
* create a views dag, to handle transformation, and two convenient tables:
  - `views.validation_codes`: unnesting validation into 1 row per agency x code
  - `views.validation_details`: unnesting validation into 1 row per agency x code x individual code violation
 * creates custom airflow macros, so we can handle loading into our test datasets by...
   - writing `SELECT * FROM { "gtfs_schedule.agency" | table }`
   - generating `SELECT * FROM cal-itp-data-infra.test_gtfs_schedule.agency`.

### Notes

* bigquery requires an aggregation in its materialized views, so I've implemented a MaterializedViewOperator, but don't use it 😭.
* bigquery's api feels very low level, so I'm still leaning on the airflow bigquery hook
* I'm keen on knocking out the UPSERTING issue (#69), and creating a "golden" table or two in the views dag (#70), but once those are done it seems like maybe a good time to queue up some gentle refactoring of calitp.py and operators.py.

### DIFF
--- a/airflow/dags/gtfs_loader/validation_report_process.py
+++ b/airflow/dags/gtfs_loader/validation_report_process.py
@@ -1,0 +1,69 @@
+# ---
+# python_callable: validator_process
+# provide_context: true
+# external_dependencies:
+#   - gtfs_downloader: validate_gcs_bucket
+# ---
+
+import pandas as pd
+import json
+
+from calitp import save_to_gcfs, read_gcfs
+
+PANDAS_TYPES_TO_BIGQUERY = {"O": "STRING", "i": "INTEGER", "f": "NUMERIC"}
+
+
+def process_notices(itp_id, url_number, validation):
+    raw_codes = validation["data"]["report"]["notices"]
+
+    # one row per code. includes a notices column with nested details
+    df_codes = pd.DataFrame(raw_codes)
+    df_codes.insert(0, "calitp_itp_id", itp_id)
+    df_codes.insert(1, "calitp_url_number", url_number)
+
+    return df_codes
+
+    return df_codes.explode("notices")
+
+
+def infer_notice_schema(notice_entries):
+    # note that convert_dtypes enables integer cols w/ missing values
+    wide = pd.DataFrame(notice_entries).convert_dtypes()
+    schema = []
+    for k in wide:
+        schema.append(dict(name=k, type=PANDAS_TYPES_TO_BIGQUERY[wide[k].dtype.kind]))
+
+    return schema
+
+
+def validator_process(execution_date, **kwargs):
+    base_path = f"schedule/{execution_date}"
+
+    status = pd.read_csv(read_gcfs(f"{base_path}/status.csv"))
+    success = status[lambda d: d.status == "success"]
+
+    # hold on to notices, so we can infer schema after
+    # notice_entries = []
+    for k, row in success.iterrows():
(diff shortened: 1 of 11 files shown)


airflow/dags/gtfs_loader/validation_report_process.py
```diff
@@ -0,0 +1,72 @@
...
+# external_dependencies:
+#   - gtfs_downloader: validate_gcs_bucket
+# ---
+
+import pandas as pd
+import json
+
+from calitp import save_to_gcfs, read_gcfs
+
+PANDAS_TYPES_TO_BIGQUERY = {"O": "STRING", "i": "INTEGER", "f": "NUMERIC"}
```
hunterowens, line 13: this is gonna be something we might want to pull out in the refactor

airflow/dags/gtfs_loader/validation_report_process.py
```diff
@@ -0,0 +1,72 @@
...
+
+    status = pd.read_csv(read_gcfs(f"{base_path}/status.csv"))
+    success = status[lambda d: d.status == "success"]
+
+    # hold on to notices, so we can infer schema after
+    # note that I've commented out the code for inferring schema below,
+    # but it was usefule for generating, then hand-tweaking to load
+    # into bigquery
+    # notice_entries = []
+    for k, row in success.iterrows():
```
hunterowens, line 50: we should be storing the validator runtime, schedule version next  to the validator rows? this might be doing it but not sure as I read through. 

ie. 

`gtfs_downloaded_at`: TS 
`gtfs_validated_by`: `validator_version`

airflow/dags/gtfs_views/METADATA.yml
```diff
@@ -0,0 +1,17 @@
+description: "Download the state of CA GTFS files, async version"
+schedule_interval: "0 0 * * *"
+tags:
+  - all_gusty_features
+default_args:
+    owner: airflow
+    depends_on_past: False
+    start_date: !days_ago 1
+    email:
+      - "hunter.owens@dot.ca.gov"
```
hunterowens, line 10: add yourself?

hunterowens:
@machow made a few small questions / comments but otherwise looks good to me.

hunterowens:
I think the strategy of knocking out those two issues and then refactoring makes sense. Coming down the pipe will be the reports based on this data from @mcplanner next week hopefully as well

hunterowens (approved):

airflow/dags/gtfs_loader/validation_report_process.py
machow: The validator version is stored in validation.json, but not in the data pulled out here. I can pull it in. The date the data was extracted will need to be added to everything for #69, and I'll use the strategy you list there to upsert validation as well.

`validation.json` looks like:

```
{
  version: "2.0.0", # added by gtfs-validator-api,
  data: {
    report: {
      # this is what we pull out
      ...
    },
    system_errors: {
      # I have never seen this non-empty, since I think it populates when the process fails
    },
  }
} 
```

airflow/dags/gtfs_views/METADATA.yml
machow: Ah, whoops--thanks

PR #78, opened 2021-04-26
fix: unpack notices in views.validator_notices

machow:
Needed a star on column selection: `notices.*`


machow:
Going to close, since PR #81 also implemented this

PR #81, opened 2021-04-30
Load historical data, keep agencies latest data in gtfs_schedule dataset

machow:
## New

* gtfs_schedule: now keeps the most recent dataset for an agency
* gtfs_schedule_history dataset: an external table (on top of cloud storage) that holds all distinct agency datasets
  - calitp_files: a list of every file in a agencies' gtfs static data (and our validation.json file), along w/ an md5 hash per file
  - calitp_files_updates: augments calitp_files with data on whether a file has changed
  - calitp_feed_updates: aggregates the above table to be one entry per itp_id x url_number x extracted_at, for feeds where data has changed.
  - calitp_feed_latest: same as above, but only rows for an agency's latest data
  - calitp_included_gtfs_tables: lists tables that will be loaded into gtfs_schedule (used throughout pipeline)
 * using pybigquery+sqlalchemy behind the scenes. We can use this to remove the need for airflow's operators. A big advantage of this approach is that all our warehouse operations will be using SQL, which we can test in the bigquery console (vs the http API). We could also swap out warehouses pretty easily (not that we should!).
   - note we had to use a pre-release branch of pybigquery, which contains a critical fix.
 
## TODO for deploying

* backfill `gtfs_loader` tasks
  - skip tasks depending on latest, run them at the end, then backfill gtfs_schedule_history_load.
* delete any old tables that might be lingering around

### gtfs_loader:

Note gtfs_schedule_history_load only copies over data extracted on the day its executed for, so we can backfill to repopulate historical tables.

image

### gtfs_schedule_history

Note this DAG only runs one time, and contains mostly external tables (which sit on top of a GCS bucket)

image


machow:
Quick note--I noticed this mornin that downloading gtfs-data didn't complete last Friday (due to an airflow database connection error? The task seems to have run for a long time). While backfilling, I can fill it in using data from the test bucket.

Leaving this snippet for myself:

```python
import gcsfs

fs = gcsfs.GCSFileSystem(project="cal-itp-data-infra")
fs.glob("gs://gtfs-data-test/schedule/2021-04-29T00:00:00+00:00/status.csv")
```

hunterowens:
@machow you can skip the backfill, I think, we don't need to have perfect upserting records

machow:
For the test data, I just backfilled from April 16th, which was pretty quick! It'll be useful to have a bit of history, so we can check the tables tracking file updates. The convenient thing here is  that It's all file processing in the buckets, so no database merges / etc..

hunterowens (approved):
:fireworks:

machow:
Thanks for looking over! Looking at the stops data has been helpful for sanity checking--let me try to get a notebook up for it, so I can be sure everything is in the right place, and then will merge!

hunterowens:
:+1:

PR #93, opened 2021-05-06
Added entry for PresidioGo Shuttle

mcplanner:
With newly discovered GTFS link


hunterowens:
@machow looks like your force push to `main` needed linter to run? can you debug this here?

PR #94, opened 2021-05-07
Fix "GTFS_RT" keys

hunterowens:
Turns out I was inconsistent of my use of `gtfs_rt_url` vs `gtfs_rt_urls`. This settles on the plural because their are multiple types


machow:
Is the correspondence between feeds in the GTFS schedule section and GTFS-RT section meaningful? For example, in this feed, it looks like the **second** GTFS schedule entry corresponds to the **first** and only RT entry (starting with `api.511`). 

If we will do anything involving "the RT feed for this GTFS schedule", then we need a way to make the links explicit.

image

machow:
This goes back to whether [the url position in the list is being used as a primary key](https://github.com/cal-itp/data-infra/pull/23#issuecomment-800584134). Another upshot of this would be that to accurately track file changes, AFAICT we should never re-order URLs. So want to be very careful and explicit with checking this again!

machow (approved):
This seems okay with me--the comments I left about url order and correspondence can always be sorted out downstream, once we're ready to extract / load the RT data in the pipeline!

PR #95, opened 2021-05-08
Removed duplicate entry for Menlo Park Shuttles

mcplanner:


machow:
Looks like this was covered in #116

PR #111, opened 2021-05-11
fix: vendor pybigquery into repo

machow:
Not the most exciting PR, but beats copying pybigquery into the cloud storage bucket ;)


machow:
Closing since there's a new pybigquery release--will open a PR to update depedencies.txt

PR #115, opened 2021-05-12
Corrected all Cal-ITP IDs

mcplanner:


machow:
Thanks! Do you mind linking to where these ids were pulled from? If I ingest it daily into the pipeline, I can keep an eye out for updates / removals..!

thekaveman:
@mcplanner the list that you linked to, is different from the one @e-lo and Juliet have been maintaining / pointing to, that we grabbed for using to generate the data for MobiMart. Do these lists correlate and should we be using so many different ones?

@machow this should already be in the warehouse under `transitstacks.providers_overview` (or `official_list`) - again sourced from the "Official List" spreadsheet https://docs.google.com/spreadsheets/d/105oar4q_Z3yihDeUlP-VnYpJ0N9Mfs7-q4TnribqYLU/edit

machow:
Cool--thanks both of you! If the sheet that's loaded into the warehouse is the official one, I can plan to add creating that external table to the pipeline (so we have the option tear down and put back up full warehouse). I'll go through tomorrow and try merging all these `agencies.yml` updates.

PR #130, opened 2021-05-24
[WIP] split calitp plugin into modules, prep for packaging

machow:
This PR splits calitp.py into a few modules, to make it a bit easier to understand. I've created a new repo https://github.com/cal-itp/calitp-py to move it to, and was thinking of going in this order...

* improve calitp in this PR to make function names consistent, etc..
* test and merge into pipeline
* move into package, release on PYPI, and set it as a pipeline dependency

The big benefit should be letting people pull down tables from the warehouse / work on pipeline tasks without having to worry about airflow particulars.

(note that in the submodules I do full path imports, e.g. `from calitp...` since the plugins folder is not technically on the python path, but custom imported by airflow or something)


machow:
closing in favor of #186

Issue #133, opened 2021-05-27
ExternalTaskSensor fails daily and requires clearing on gtfs_load dag


machow:
Submitting a fix for this, but wanted to document. Parts of the gtfs_load dag needs to wait on the validator to finish running. However, because gusty sets the default timeout for waiting for a dependency to 60 seconds (with a number of retries), and the validator takes ~20 minutes to run, the sensor kept failing.

Gusty has options for setting the timeout to be longer, which should fix.

Issue #75, opened 2021-04-23
GTFS Reports


hunterowens:
I suspect this is gonna become several issues, but as discussed on the Monday call, we'd like to generate a monthly "GTFS" report, capturing elements we can automate from the grading schema and either post on the web or generate PDFs to email or both. @mcplanner is working (tracking in teamwork) what the the report will look like, but let's have  astub here for discussion

mcplanner:
Draft monthly report
<https://docs.google.com/document/d/1-iXNdl4f2EbYayKg3e7jkTALAcNzht9yGQuyIOtPLc4/edit#>
.

hunterowens:
deployment gameplan: 

generate out a static monthly report, deploy via netlify

hunterowens:
PDFs can be generate from the print option in css

machow:
@mcplanner (cc @ryon, who is styling the report from the jarvus side), what do you think of a lineplot for service hours over a given month? 

![image](https://user-images.githubusercontent.com/2574498/121097526-f5f77300-c7a8-11eb-8a50-9fe43686e9cc.png)

This will let people see weekly trends (e.g. maybe more service hours on weekends?), and might help us report hours for agencies where they are fairly large numbers, without them crowding out each other on a calendar.

(Note: I generated fake data for this, which is why it is a saw shape :)

mcplanner:
A chart is better than a calendar, definitely. I'm inclined to suggest a
bar chart, to show that this data is discrete rather than continuous, but
it's easier to see minor changes in the line chart. Line chart it is! Thank
you.

machow:
Ah, nice--thanks. We could also add points, if it helps with discreteness!

![image](https://user-images.githubusercontent.com/2574498/121243065-4b945400-c86b-11eb-85b6-69dc054e28f1.png)

mcplanner:
love it!

machow:
Okay--great! @hunterowens, copying here the task description I sent to @ryon, so we can sync on this issue if needed. Note that if calitp can get reports done before June 15th, then they can send them in their newsletter. (may be able to use the uglier notebook version as a fallback on the 15th, it the timeline is too tight?)

**Existing report draft**

* Demo report:  https://github.com/cal-itp/reports/tree/feat-papermill/demo/output/256_0
* (note that all the data used in tables and figures is in the data folder; happy to reshape into nice formats)
[...]

Issue #83, opened 2021-05-03
improve dataset documentation + developing on data pipelines


hunterowens:

machow:
Breadcrumb: document running backfill commands. E.g.

```
gcloud beta composer environments run calitp-airflow-prod --location us-west2 backfill -- -s '2021-05-08' -e '2021-05-08' -x -y --rerun_failed_tasks gtfs_loader
```

machow:
Breadcrumb: have Chris explain what holo is, and how it is used here (for generating docs).

https://github.com/JarvusInnovations/hologit

machow:
Screencasts to record:

* using `debug.py` and importing / running a task from a dag
* creating a new view in `gtfs_views` DAG
* running queries in the bigquery web UI
  - checking how many bytes they processed
* basic siuba queries against warehouse

machow:
Going to close, opened issue for documenting debug.py (https://github.com/cal-itp/data-infra/issues/699), and have epics for fleshing out tools like bigquery in the analytics docs

PR #137, opened 2021-05-29
Feat snapshots

machow:
This PR builds out the tables needed for #75, and also fixed a few small things along the way. Main features:

* daily snapshots of gtfs schedule data (merge_updates.py)
* added views for describing daily service schedules (e.g. on a specific date, what does the most recent feed before that date say service is like in the calendar)
* fixed an issue with gusty configuration that was causing wait_for tasks to fail everyday #133 
* added a `query_yaml` function, which can take the name of a task file
* add a debug.py file that makes it easy to import from airflow plugins, dag tasks

TODO (but likely not urgent): 

* Change loading in `gtfs_loader/gtfs_schedule_tables_load.py` to load the latest data from the snapshots, move to history2 dags
* I'd go for renaming the current gtfs_schedule_history dataset to gtfs_schedule_external or something.

### Debugging example

e.g.

```python
import debug
from calitp import query_yaml

# these can be uncommented if you need to test a query against the production tables
# which is maybe not ideal, but useful until we put in a quick way to copy the prod
# bigquery tables to dev
# os.environ["AIRFLOW_ENV"] = "cal-itp-data-infra"
# os.environ["AIRFLOW_VAR_EXTRACT_BUCKET"] = "gs://gtfs-data"

query_yaml("dags/gtfs_views/gtfs_schedule_service_daily_metrics.yml", write_as="views.gtfs_schedule_servce_daily_metrics", replace=True)
```

Note that I run the new SQL views over production, so I could punt on figuring out easy ways to copy datasets to dev. Once I get back, it's probably about a good time to add in functionality for creating SQL table / column comments, and having them generate docs. There's also nice SQL testing functionality via https://github.com/sodadata/soda-sql!

### DIFF
--- a/airflow/dags/gtfs_views/gtfs_schedule_history_calendar_long.yml
+++ b/airflow/dags/gtfs_views/gtfs_schedule_history_calendar_long.yml
@@ -1,0 +1,26 @@
+operator: operators.SqlToWarehouseOperator
+dst_table_name: "views.gtfs_schedule_history_calendar_long"
+dependencies:
+  - warehouse_loaded
+
+sql: |
+  # Note that you can unnest values easily in SQL, but getting the column names
+  # is weirdly hard. To work around this, we just UNION ALL.
+  {% for dow in ["monday", "tuesday", "wednesday", "thursday", "friday", "saturday", "sunday"] %}
+
+  {% if not loop.first %}
+  UNION ALL
+  {% endif %}
+
+  SELECT
+    calitp_itp_id
+    , calitp_url_number
+    , service_id
+    , PARSE_DATE("%Y%m%d", start_date) AS start_date
+    , PARSE_DATE("%Y%m%d", end_date) AS end_date
+    , calitp_extracted_at
+    , calitp_deleted_at
+    , "{{dow | title}}" AS day_name
+    , {{dow}} AS service_indicator
+  FROM `{{ "gtfs_schedule_type2.calendar" | table }}`
+  {% endfor %}
--- a/airflow/dags/gtfs_views/gtfs_schedule_service_daily_metrics.yml
+++ b/airflow/dags/gtfs_views/gtfs_schedule_service_daily_metrics.yml
@@ -1,0 +1,14 @@
+operator: operators.SqlToWarehouseOperator
+dst_table_name: "views.gtfs_schedule_service_daily_metrics"
+dependencies:
+  - warehouse_loaded
+
+sql: |
+  SELECT
+    calitp_itp_id
+    , calitp_url_number
+    , service_date
+    , service_id
+    , SUM(service_hours) AS ttl_service_hours
+  FROM `{{ "views.gtfs_schedule_service_daily_trips" | table }}`
+  GROUP BY 1, 2, 3, 4
--- a/airflow/debug.py
+++ b/airflow/debug.py
@@ -1,0 +1,8 @@
+import os
+import sys
+
+import dotenv
+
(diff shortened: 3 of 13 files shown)


machow:
hmm.. it's too bad this check is failing now, since that is a very old piece of code.

thekaveman:
It's failing because of a newer `pre-commit` hook added as part of the security scanning work in #105:

```console
>> Issue: [B108:hardcoded_tmp_directory] Probable insecure usage of temp file/directory.
   Severity: Medium   Confidence: Medium
   Location: airflow/dags/gtfs_downloader/download_data.py:22
   More Info: https://bandit.readthedocs.io/en/latest/plugins/b108_hardcoded_tmp_directory.html
21	
22	SRC_DIR = "/tmp/gtfs-data/schedule"
23	DST_DIR = "schedule"
```

airflow/dags/gtfs_views/gtfs_schedule_history_calendar_long.yml
```diff
@@ -0,0 +1,26 @@
+operator: operators.SqlToWarehouseOperator
+dst_table_name: "views.gtfs_schedule_history_calendar_long"
+dependencies:
+  - warehouse_loaded
+
+sql: |
```
hunterowens, line 6: how does this templating get processed? Jinja?

airflow/dags/gtfs_views/gtfs_schedule_service_daily_metrics.yml
```diff
@@ -0,0 +1,14 @@
+operator: operators.SqlToWarehouseOperator
```
hunterowens, line 1: :+1:

airflow/debug.py
```diff
@@ -0,0 +1,8 @@
+import os
```
hunterowens, line 1: note to self to write more "developing" documentation

hunterowens:
some quick notes: 

* rather than using query yaml, might want to implement a [intake catalog](https://intake.readthedocs.io/en/latest/catalog.html) inside of calitp to take advantage of the library
* Docs! we should write them now that #138 has landed. 


Otherwise, looks good to merge and iterate on this next week!

thekaveman:
@machow regarding the [failing pre-commit check](https://github.com/cal-itp/data-infra/pull/137#issuecomment-852308855): I added 8db0899 to fix the syntax of the skip rules. It looks like a new rule was added as part of this PR, and it was overwriting the existing skip rule, causing the failure that had previously been ignored.

airflow/dags/gtfs_views/gtfs_schedule_history_calendar_long.yml
machow: yeah--airflow operators have a `template_fields` argument, that allows the listed arguments to be jinja template strings! It's pretty similar to how https://www.getdbt.com/ works!

airflow/debug.py
machow: Added some notes in the docs issue for some useful and quick screencasts--

https://github.com/cal-itp/data-infra/issues/83#issuecomment-855996337

machow:
Using intake looks super interesting! This is my first time looking at it, but it seems like it can also support things like return all validator notices for agency X (where agency X is defined using a parameter?). I wonder if it could be a quick way to do essentially what something in cal-itp/nootbooks might do, but when we just need to throw out a piece of data quickly (or w/o a UI).

hunterowens (approved):

Issue #84, opened 2021-05-03
rename repo


hunterowens:
juniper here we come!

mcplanner:
@hunterowens, what's keeping this from getting closed?

hunterowens:
AFAIK it was a back and forth between, you, me and @thekaveman on "public facing names" 

since we now have docs.calitp.org, etc, there's something to be said for names that are descriptive vs ones that you have to learn the various codenames, etc in terms of onboarding folks and being transparent.

thekaveman:
Furthermore... isn't [Juniper the front-end User Interface](https://docs.google.com/document/d/1lyM9w3gdPsn4yog-8nlHEKoczwc30N-vnEsniQWr6Ho/edit) e.g. [reports](cal-itp/reports)? This repository is all the back-end stuff that makes the data and metrics available, right? I don't agree that this repo should be renamed to Juniper.

mcplanner:
I put this into "Waiting" until we have a better sense of our suite of
tools and where this name will fit in among them.

hunterowens:
SG.

PR #138, opened 2021-06-01
docs: add documentation features references

themightychris:
Adds links and notes for using documentation features. Also shortens title of main page to "Welcome"

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,6 +1,8 @@
-# Juniper documentation
+# Welcome
 
-Welcome to the docs!
+This website provides technical documentation for the Juniper codebase.
+
+Documentation for the `main` branch is available online at: <https://cal-itp.github.io/data-infra/>
 
 ## Editing documentation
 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -1,6 +1,8 @@
-# Juniper documentation
+# Welcome
 
-Welcome to the docs!
+This website provides technical documentation for the Juniper codebase.
```
thekaveman, line 3: I suggest we avoid usage of internal code names like "Juniper" for things that are public-facing or publicly available.

thekaveman (changes requested):

docs/README.md
themightychris: @thekaveman when we initially discussed the ticket for setting up docs, @hunterowens had also resolved it was time to rename this repo to juniper (#84) so I aligned towards that

thekaveman (approved):

PR #139, opened 2021-06-01
style(ci): rename lint workflow to match  filename

themightychris:
The `lint.yml` workflow is named `build` -- seems like maybe some copy-pasta. I renamed it to `lint` which shows while the lint workflow is initializing and is the name used to require the check. Looks like github repo branch settings will need to be updated when this gets merged to make `lint` the required check instead of `build`

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,7 +7,7 @@ on:
     branches: [ main ]
 
 jobs:
-  build:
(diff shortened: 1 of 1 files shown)


hunterowens (approved):

thekaveman:
Updated branch protection on `main` to require the `lint` check instead.

PR #141, opened 2021-06-04
Kubernetes & Metabase Deployment

lottspot:


themightychris (changes requested):
@lottspot looks like the `.pre-commit-config.yaml` file supports hook-level exclude filters, so you should be able to configure the `check-yaml` hook to exclude `kubernetes/app/charts/*/templates/*.yml` (would be good to keep the top-level non-template chart yaml linted though)

https://pre-commit.com/#regular-expressions

doc/kubernetes.md
```diff
@@ -0,0 +1,114 @@
+kubernetes
```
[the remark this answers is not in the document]
lottspot: How entrenched is this convention? I think that in a monorepo structure where directories map to system components, there is a strong benefit to mapping documentation to a corresponding tree structure. Adding arbitrarily named top level containers serves really only to obfuscate the purpose/scope of each doc.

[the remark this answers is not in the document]
themightychris: None at all at this point, but I think in general there's two ways to go in organizing documentation at the top level: functional or by component

I don't have a super strong opinion either way at this point. My initial draft of the above comment suggested moving it to `docs/kubernetes/README.md` to create a top-level Kubernetes section, but then one thing we identified in our call this morning w/ @machow was that an important counterbalance to the monorepo could be having the docs organized more around having different entrypoints for each functional role someone might be coming at the project from.

I've historically found the functional role delineation more valuable, as certain components might have different ways you want to write about and organize them for users of their output, people that develop in "userland", people that operate them on the backend, etc. So in adding metabase we might add an article under "Operations" that talks about how its configured, how to backup/restore it, how to monitor its health etc and then an article under "Analysis" that documents the same components from a different entrypoint

I'd say merge it under either for now and we can take more time to see how the docs come together and reorganize them easily enough if needed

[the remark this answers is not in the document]
themightychris: looks good

kubernetes/apps/charts/metabase/templates/database.yaml
```diff
@@ -0,0 +1,142 @@
...
+        readinessProbe:
+          tcpSocket:
+            port: {{ .Values.workloads.database.port }}
+          initialDelaySeconds: 5
+          periodSeconds: 10
+        livenessProbe:
+          tcpSocket:
+            port: {{ .Values.workloads.database.port }}
+          initialDelaySeconds: 15
+          periodSeconds: 20
```
themightychris, line 57: @lottspot I've found that it can be important for metabase to define a `startupProbe`, as metabase can take a long while to start up when cold sometimes and get stuck in a crash loop with k8s killing it for not being live fast enough. Defining a startupProbe lets you provide a longer threshold for initial readiness

See: https://github.com/urbanSpatial/LetsPlanOrg/blob/901dc868b44cc743dfc1a948820946034b80ded9/helm-chart/templates/metabase.yaml#L30

hunterowens:
also this one?

machow:
Syncing on this w/ @lottspot to plan next steps. I think we put this on hold to prioritize getting GTFS RT up.

kubernetes/apps/charts/metabase/templates/database.yaml
lottspot: Can you define "important"? What negative consequences are we concerned with if a metabase startup is in a crash loop until the database comes online? To me, this effect actually seems desirable-- if the app crashes when a DB connection isn't up, then it is 100% in a failed state, and it seems better for us to be aware of that than to engage in any attempt to mask it.

lottspot: furthermore, we're talking about a minuscule edge case here-- the only time metabase will be coming up when a DB is not up is in a fresh bootstrap scenario. This seems hardly worth holding up a merge over.

themightychris: @lottspot this isn't about metabase crashing until the database comes online, metabase is a fat cow of a java app that can take 30-60s plus to start up. On a slow or taxed enough node, metabase can get into a crash loop even with all its resources available because it doesn't start up fast enough to not get killed by k8s over and over before it can finish starting

themightychris: oh a case where this happens a LOT is when metabase has migrations to run (i.e. initial startup of a new metabase internal DB or during version bumps)

themightychris: feel free to merge and then add the startupProbe in a follow-up

themightychris (approved):

kubernetes/apps/charts/metabase/templates/database.yaml
lottspot: @themightychris gotcha-- I opened up #217 for continuation

Issue #123, opened 2021-05-17
add [API_KEY] template in `agencies.yml`


hunterowens:
We should create a secrets file that contains the API key for the agencies compose of API keys

ie, 

`secrets.yml`
```
api_key:
  itp_id: 7
  rt_key: {SOME_KEY}
  schedule_key: {SOME_KEY}
```

and then add logic in the `agencies.yml` to handle it using templating. Jinja might be useful here but curious for thoughts @lottspot @machow @e-lo

hunterowens:
I've also used `git-crypt` to allow you to have the file stored inside git and versioned for public repos, works pretty nicely.

thekaveman:
`git-crypt` is a great tool, but the repo is unusable locally for anyone that doesn't have it installed on their system. Can be somewhat of a barrier for interacting with the non-sensitive areas of the repo.

machow:
I've heard good things about git-crypt, but haven't used it. @hunterowens are you thinking something like this?

* jinja templating the agencies.yml, to fill in API keys
* a separate, encrypted file with the keys themselves

What happens if someone without git crypt tries to clone the repo? (guessing they just see the file as encrypted?). If that's the case, seems okay with me!

hunterowens:
going the inject at runtime approach via "secret spreadsheet or file". @machow to document / approach

machow:
Here's a quick proposal for the secret spreadsheet approach. Happy to tweak!

### Proposal

* we keep a spreadsheet with 2 columns: api_name, api_token.
  - e.g. api_name might be AC_TRANSIT, api_key might be `2512B81107A09D2DC44895CDDC650D47`.
* in agencies.yml, people can include api keys into URLs using `{{API_KEY}}`
  - e.g. `https://api.actransit.org/transit/gtfs/download?token={{AC_TRANSIT}}`
* in the data pipeline, we will...
  - download keys before downloading feed data, and save them in our bucket
  - when downloading feeds: render each url with the secrets included as jinja context

hunterowens:
Thanks! I think that also makes, let me know what you think about the
approach I took in the PR I opened.

hunterowens:
this is done!

PR #144, opened 2021-06-07
Use Templates rather than API keys in agencies.yml

hunterowens:
This PR adds `{{ api_key }}` rather than storing API keys in the public file. 

Currently, API keys are stored in a google sheet on the team drive inside the Juniper folder. 

the schema of the sheet is as follows: 

| calitp_id | api_key | url_number_schedule | url_number_rt | 
| --------- | ------- | ------------------- | ------------- |
| the ITP id | the API key to be used | which schedule URL this confirms to. if null, no schedule URL requires this API key | the RT number URL, same as before | 

Logic will need to implemented in both the loaders to use. I think you can just process this using jinja templating?

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -1,14 +1,14 @@
 ac-transit:
   agency_name: AC Transit
   gtfs_schedule_url:
-    - https://api.actransit.org/transit/gtfs/download?token=2512B81107A09D2DC44895CDDC650D47
+    - https://api.actransit.org/transit/gtfs/download?token={{ api_key }}
   gtfs_rt_urls:
     vehicle_positions:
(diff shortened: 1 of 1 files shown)


hunterowens:
tagging @lottspot and @machow for your thoughts. 

Also, @mcplanner, once this lands we should be unblocked on adding more RT urls to the list.

machow:
@hunterowens the one place I'm wondering about is this feed, which uses separate tokens for different RT feeds:

https://github.com/cal-itp/data-infra/pull/144/files#diff-504fdf3d4b2c7b3ed2d908d6440733807cf977552333d6140bae36cf5488924eL309-L313

Is this a case where we could use the same key?

machow:
Linking this PR to #123

hunterowens:
@machow the original different keys is is a result of an excel error, I think. All 511 feeds and such should use the same secret key.

hunterowens:
closed in favor of #172

PR #155, opened 2021-06-09
GTFS-RT Data Archiving

### LINKED ISSUE #38: Setup Storage and Archiving of GTFS-RT data
once we complete #26, we need to begin the path of capturing all the GTFS-RT data in the state. 

Caltrans' GTFS guidelines suggest that we expect updates to GTFS-RT feed at least every 20 seconds, so I recommend a 20 second sample rate and archive on all the feeds.

Each [GTFS-RT Feed](https://developers.google.com/transit/gtfs-realtime) should contain the following elements. 
* Vehicle Positions 
* Trip Updates
* Alerts 

Some prior work in GTFS-RT processing can be found in @unreasonableman's [GTFU](https://github.com/cal-itp/gtfu) and the [following notebook](https://github.com/cal-itp/notebooks/blob/main/notebooks/gtfs-rt-performance.ipynb). 

Essentially, what I think we want to do is setup a cloud function or similar to do the following procedure on a schedule. It might make sense to overlap these triggers, but essentially each minute, you watch a feed for a minute (or more). credit to @thekaveman for the idea. 

```python
for agency in agencies: 
  download_gtfs_rt_and_save_to_gcs_with_timestamp(agency)
  sleep(20)
  download_gtfs_rt_and_save_to_gcs_with_timestamp(agency)
  sleep(20)
  download_gtfs_rt_and_save_to_gcs_with_timestamp(agency)

def   download_gtfs_rt_and_save_to_gcs_with_timestamp(agency):
  # snapshot function for an agency gcs bridge 
  r = requests.get(vehicle_pos)
  data  = process_into_compressed_parquet(r.data)
  save_to_gcs(data, timestamp=now()) 
```

Constraints: Every GTFS-RT feed is stateless, ie, we cannot go back in time. This makes using an airflow job to monitor them unappealing, as it means you need 100% uptime. Perhaps a set of long running overlapping tasks would work, but I kinda think we should outsource this to the high availability of lambda / cloud functions /etc

lottspot:
Resolves #38

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,5 +18,5 @@ repos:
     rev: 1.7.0
     hooks:
       - id: bandit
-        args: ["-ll", "--skip=B108,B608"]
+        args: ["-ll", "--skip=B108,B608,B310"]
         files: .py$
--- a/docs/services/gtfs-rt-archive.md
+++ b/docs/services/gtfs-rt-archive.md
@@ -1,0 +1,11 @@
+# gtfs-rt-archive
+
+## Get gcp bucket access token
+
+This method is suitable for testing
+
+```
+gcloud auth print-access-token
+```
+
+For production service account tokens, see [cloud storage docs](https://cloud.google.com/storage/docs/authentication)
--- a/docs/services/gtfs-rt-archive.md
+++ b/docs/services/gtfs-rt-archive.md
@@ -1,11 +1,31 @@
 # gtfs-rt-archive
 
-## Get gcp bucket access token
+## Quickstart
 
-This method is suitable for testing
+See dependencies section below for details on prerequisites
 
-```bash
-gcloud auth print-access-token
+```
+export CALITP_LOG_LEVEL=debug
+export CALITP_DATA_DEST=gs://gtfs-data/rt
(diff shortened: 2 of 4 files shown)


machow:
Nice! I've submitted a PR for the updated agencies.yml format in #157. From the pipeline side, I'm guessing the two big cases will be...

* how does it handle updates to agencies.yml?
  - (note that this file gets pushed to a bucket when we merge into master, which is used by the pipeline, but the rt extraction could do w/e it wants for updating)
* how does it handle updates to `gtfs-rt-archive.py`?

hunterowens:
Awesome! One additional question to add - 

How does it support handling changes to the raw protobuf file? My understanding of PB is that you need to read it with a schema, so given version changes in the pb schema, should we be saving to a more appropriate archival format? (ie, avro or parquet)?

hunterowens:
This PR is probably ready to merge?

lottspot:
This should be good to merge, pending a review

machow:
Let's review at Jarvus sync at 11am, since it'd help me to hear a bit about deployment / re-deployment

machow (approved):
LGTM--thanks for adding local instructions for running the script. I wonder if down the road if we want to pin requirements for deterministic builds. But tbh the requirements are so few / well-maintained libs that it seems like it may not be an issue.

PR #156, opened 2021-06-09
views: fix service_daily related tables

machow:
Main issues were..

* gtfs_schedule_service_daily needed to take into account calendar start_date and end_date
* service hours calculation for report is the service window available on a given day (ie. 24 hours or less); named service_window for now in data, since previous calculations are still in there.

(added some extra comments)


machow:
Ah, forgot to flag @hunterowens. But going to merge, so we can go over in a few minutes

Issue #146, opened 2021-06-08
Changing structure of agencies.yml


machow:
cc @themightychris @hunterowens @lottspot, does this format for `agencies.yml` work for you?

Currently, `agencies.yml` is structured as 1 entry per agency, with 3 kinds of info

* their name
* their itp_id
* fields for different kinds of feeds (e.g. gtfs_schedule_url, gtfs_rt_urls)

E.g.

```yaml
ac-transit:
  agency_name: AC Transit
  gtfs_schedule_url:
    - https://api.actransit.org/transit/gtfs/download?token=2512B81107A09D2DC44895CDDC650D47
  gtfs_rt_urls:
    vehicle_positions:
      - https://api.actransit.org/gtfsrt/vehicles?token=2512B81107A09D2DC44895CDDC650D47
    trip_updates:
      - https://api.actransit.org/gtfsrt/tripupdates?token=2512B81107A09D2DC44895CDDC650D47
    service_alerts:
      - https://api.actransit.org/gtfsrt/alerts?token=2512B81107A09D2DC44895CDDC650D47
  itp_id: 4
```

However, there are 2 limitations

* **nesting**: there are 3 different gtfs_rt feed types (vehicle_positions, trip_updates), nested under gtfs_rt_urls
* **multiple feeds within agency are represented by lists**: note that gtfs_schedule_url corresponds to a list. This is because a couple agencies have multiple feeds. This means we have to make sure that if an agency has feed 1 and feed 2, that all urls for feed 1 are the first list item everywhere. (currently being [violated in agencies.yml](https://github.com/cal-itp/data-infra/blob/main/airflow/data/agencies.yml#L323-L327)).

### Proposal

**One solution suggested by @themightychris** is to have agency entries be assumed to only contain a single url by default. I would suggest in addition that we unnest gtfs_rt_urls. In this case, the yaml above would become...

```yaml
ac-transit:
  agency_name: AC Transit
  gtfs_schedule_url: https://api.actransit.org/transit/gtfs/download?token=2512B81107A09D2DC44895CDDC650D47
  gtfs_rt_vehicle_position_url: https://api.actransit.org/gtfsrt/vehicles?token=2512B81107A09D2DC44895CDDC650D47
  gtfs_rt_trip_updates_url: https://api.actransit.org/gtfsrt/tripupdates?token=2512B81107A09D2DC44895CDDC650D47
[...]

Issue #117, opened 2021-05-12
Duplicate feed keys (itp_id, url_number) should raise error on extraction


machow:
See https://github.com/cal-itp/data-infra/issues/113. Currently, we read `agencies.yml` into a DataFrame that includes the columns `itp_id` and `url_number`. These columns are primary keys, in the sense that they are used to name subdirectores. E.g. `gtfs-data/schedule/{execution_date}/{itp_id}_{url_number}`.

These columns are also used throughout the pipeline to perform calculations (e.g. has this feed's files changed? how many routes does this feed have?). It would be ideal to use an independent key, so we could separate pipeline data feeds from agency specific info, but at the very least we need to ensure itp_id and url_number are unique, or fail before downloading data.

machow:
This is being done now in our github action that validates agencies.yml

https://github.com/cal-itp/data-infra/blob/main/script/agencies_validate.py#L30-L31

PR #157, opened 2021-06-09
Refactor agencies yml

machow:
This PR addresses #146, by changing agencies.yml to use a `feeds` field. I also addressed #117 by adding an assertion that checks for duplicate itp_id keys (and removed one case of this from the data; duplicate top-level entries for presidio shuttle).

Note that each feed has entries for all rt urls, with null values when they weren't specified. Happy to remove, if feels cleaner to not keep these null entries in.

Here's the code I used to migrate, in case it's useful:

```python
# +
import yaml

from itertools import zip_longest

old_agencies = yaml.safe_load(open("data/agencies.yml"))

# +
headers = ["gtfs_schedule_url", "gtfs_rt_vehicle_positions_url", "gtfs_rt_service_alerts_url", "gtfs_rt_trip_updates_url"]

new_agencies = {}
for k_agency, entry in old_agencies.items():
    rt_urls = entry.get("gtfs_rt_urls", {})
    feeds_headless = zip_longest(
        entry.get("gtfs_schedule_url", []),
        rt_urls.get("vehicle_positions", []),
        rt_urls.get("service_alerts", []),
        rt_urls.get("trip_updates", [])
    )
    
    feeds = [dict(zip(headers, entry)) for entry in feeds_headless]
    
    new_agencies[k_agency] = dict(
        agency_name = entry["agency_name"],
        feeds = feeds,
        itp_id = entry["itp_id"]
    )
        
    

# +
# Dumper allows us to consistently indent lists by two spaces,
# similar to the original file
# see: https://github.com/yaml/pyyaml/issues/234

class Dumper(yaml.Dumper):
    def increase_indent(self, flow=False, *args, **kwargs):
        return super().increase_indent(flow=flow, indentless=False)

# sort_keys keeps order of entries as close to original as possible, since
# pyyaml sorts everything alphabetically by default
yaml.dump(new_agencies, open("data/agencies.yml", "w"), sort_keys=False, Dumper=Dumper)
```

For testing that the airflow task loading `agencies.yml` works I used (from the airflow folder)...

```python
import debug

from dags.gtfs_downloader.generate_provider_list import make_gtfs_list
make_gtfs_list("data/agencies.yml")
```

### DIFF
--- a/airflow/dags/gtfs_downloader/generate_provider_list.py
+++ b/airflow/dags/gtfs_downloader/generate_provider_list.py
@@ -9,7 +9,7 @@ import pandas as pd
 from calitp import pipe_file_name
 
 
-def make_gtfs_list():
+def make_gtfs_list(fname=None):
     """
     Read in a list of GTFS urls
     from the main db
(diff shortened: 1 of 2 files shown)


machow:
I'm noticing a case where an agency had multiple feeds, but their entries were mismatched in the original data (so are also mismatched in this data). I can make a quick correction to it before merging!

https://github.com/cal-itp/data-infra/pull/157/files#diff-504fdf3d4b2c7b3ed2d908d6440733807cf977552333d6140bae36cf5488924eR466

hunterowens:
I would think that we shouldn't have null GTFS RT feeds, but otherwise SGTM

hunterowens (approved):

Issue #160, opened 2021-06-10
Allow views to add comments to warehouse tables


machow:
This is weirdly bespoke in bigquery (especially compared to [postgres, snowflake, etc.. COMMENT](https://www.postgresql.org/docs/9.1/sql-comment.html)). I think we have to use the web API (e.g. via their python library).

Essentially we have 3 cases of creating tables

1. from a csv file
2. from a SQL query (CREATE TABLE AS)
3. logical views (though we can probably ignore for now)

I have been looking through the bigquery python client and it seems pretty quirky.

machow:
RE quirks, if you run this code...

```python
from google.cloud import bigquery
client = bigquery.Client()
tbl_test_comments = client.get_table("test_views.test_comments")
```

Then it's no problem to set the table description...

```python
tbl_test_comments.description = "some description"
```

But no setters exist on the SchemaEntry object...

```python
schema_entry = tbl_test_comments.schema[0]
# errors
schema_entry.description = "some description"
```

Issue #118, opened 2021-05-12
Add task for loading external table of "official list" spreadsheet


machow:
See https://github.com/cal-itp/data-infra/pull/115. Should go into a transitstacks DAG, so we can reconstruct warehouse.

thekaveman:
This is great, thank you @machow. Just a note there are 4 different sheets, each loaded as an external table, then joined together in a (broken) view.

thekaveman:
Also, it sounds like from talking with @e-lo and Juliet yesterday - there are still some updates/changes to come to that spreadsheet. There may be another couple of sheets/tables needed, and some modifications to the columns.

machow:
Ah, thanks for the heads up--I can have it tear down the old table daily and put up a new w/ the latest column setup

hunterowens:
@thekaveman / @machow to work on this week

machow:
@hunterowens @thekaveman, I'm not able to query the external tables. It looks like maybe the range that was set no longer works (maybe column names were changed?). Can one of provide some explanation of the range you set (or the correct range now)? Thanks!

![image](https://user-images.githubusercontent.com/2574498/119873996-2ad71080-bef3-11eb-89c1-b6b3a1533bfb.png)

thekaveman:
@machow yeah all the sheet names changed, and some of the column names were cleaned up recently as well. We're just waiting on one more column rename in the *Fares* sheet, but the other four sheets have been re-mapped into BQ tables.

One note, I didn't recreate that view - it was failing anyway, and was just a lazy `LEFT JOIN` of all the external tables. Not sure if it's useful or not.

machow:
Thanks! I'll look over the tables today while looking over the pipeline

machow:
@thekaveman I can create the tables in the web UI, but there may be one small piece to add to load external tables in the pipeline: sharing any spreadsheet loaded with the pipeline's service account. I'm going to try giving the account view privileges on a sheet quickly to check.

machow:
Hmm.. that wasn't it--but it's weird that the failing SQL command, and the web UI are using the same user (michael.c@jarv.us). Lemme look around a bit more!

machow:
Ah, good discussion here--https://github.com/fishtown-analytics/dbt/issues/3040#issuecomment-789223055.

I'll likely need to add scope to looking credentials (and double check that the service account has access to drive)

machow:
[...]

Issue #99, opened 2021-05-10
calitp function to ingest CSVs or sheets -> cloud storage and bigquery


machow:
This can also be wrapped in an operator, so we can add csvs to our "views" DAG. The big upside here is that if we save the data to a bucket while we are uploading to bigquery, then we can always rollback (if something surprising / breaking happens on the sheet).

This sheet of nice names for error codes is a good first candidate!:

* https://docs.google.com/spreadsheets/d/1GDDaDlsBPCYn3dtYPSABnce9ns3ekJ8Jzfgyy56lZz4/edit#gid=0

machow:
Implemented as...

* CsvToWarehouseOperator: load a sheet from a URL
  - e.g. https://github.com/cal-itp/data-infra/blob/main/airflow/dags/transitstacks_loader/fares.yml
* PythonToWarehouseOperator: use python to do some processing, then upload to warehouse
  - e.g. https://github.com/cal-itp/data-infra/blob/main/airflow/dags/gtfs_views/validation_code_descriptions.py

PR #161, opened 2021-06-11
Feat external spreadsheets

machow:
This PR addresses...
* #118 and #99 by implementing operators for loading data from URLs to CSVs, and from SQL queries.
* #160 by adding a fields argument to two operators for adding table comments.
* #162 by adding human readable errors sheet to warehouse (though it may need to be transformed)
* refactors operators so each is in its own file.

Additional notes:

* adds two new functions to calitp --
  - `to_snakecase`
  - `sql_patch_comments` - change a table's column comments
* three operators introduced
  - CsvToWarehouseOperator
  - PythonToWarehouseOperator
  - SqlToWarehouseOperator


hunterowens:
Can we add to the docs on how to setup one of the yaml files/

machow:
Ah, yeah--good point! I'll add docs on using the main user-facing operators

machow:
From pairing yesterday--I'm going to merge in now, so I can do a couple screencasts with the updated prod dags.

PR #165, opened 2021-06-14
Added Tulare InterModal Express

mcplanner:


hunterowens:
@mcplanner before we merge this, can you have @julieteldred add in an entry for this provider into the primary list and assign an ITP ID?

mcplanner:
Assigning to me so I follow up with Juliet.

mcplanner:
Its ID is 345

PR #172, opened 2021-06-15
Use Named Keys for API keys

hunterowens:
This PR supplants #144 basedon the new agencies.yml file, but uses named keys rather logic so it is easier to template.

### DIFF
--- a/.github/workflows/push_to_gcloud.yml
+++ b/.github/workflows/push_to_gcloud.yml
@@ -17,6 +17,15 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: 3.8
+    - name: Install python packages
+      run: |
+        pip install jinja2==2.11.3 PyYAML==5.4.1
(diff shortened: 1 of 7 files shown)


.github/workflows/push_to_gcloud.yml
```diff
@@ -17,6 +17,15 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: 3.8
+    - name: Install python packages
```
hunterowens, line 20: this is so cool

machow:
@themightychris it looks like the markdown linter has a false positive for a dead link--have you seen it run into this much?

opening_a_page

machow:
@hunterowens if this works for you, I can loop around and merge tomorrow (may quickly set the action to fill in API keys in `agencies.yml` on every branch, so people get feedback on missing keys when updating)

hunterowens:
sg

PR #173, opened 2021-06-15
More RT feeds

mcplanner:


airflow/data/agencies.yml
```diff
@@ -530,9 +530,11 @@ gold-coast-transit:
   agency_name: Gold Coast Transit
   feeds:
     - gtfs_schedule_url: http://www.goldcoasttransit.org/images/GTFS/GTFS.zip
-      gtfs_rt_vehicle_positions_url: null
-      gtfs_rt_service_alerts_url: null
-      gtfs_rt_trip_updates_url: null
+    - gtfs_schedule_url: https://www.goldcoasttransit.org/images/GTFS/GCTD_GTFS.zip
+    -gtfs_schedule_url: https://govcbus.com/gtfs
```
hunterowens, line 534: does govcbus == gold coast transit? 

additionally the lack of a space after the `-` will result in a yaml error.

mcplanner:
Good catch on the spaces.

I think it's like Go Ventura County: https://govcbus.com/map

e-lo:
Do we have a precommit yaml filter?

machow:
PR #172 attempts to load the yaml file after substituting in API keys, so should catch basic errors (e.g. misformatted yaml)! We could enhance the github action there to do validation w/ something like [jsonschema](https://json-schema.org/).

hunterowens:
@e-lo we do have a yaml linter in the `lint_and_style_checks` action which you can see, failed for this.

hunterowens:
It's in a pre-commit but I think it doesn't get run since @mcplanner is using the web editor AFAIK

hunterowens:
I updated this PR, and yet again, it's an interesting case. 

goVCbus.org == Ventura County Transportation Commission, provides a countywide / all operators GMV licenses and installs. 

GMV then has a corresponding GTFS. Yet, Gold Coast Transit provides their own GTFS Schedule. I have yet to diff the 2, but I suspect the 2nd one is the entire county, while the first is just Gold Coast Transit. I kept both in the URL and then linked the RT twice, but am open to other approaches. 

image

hunterowens:
@e-lo to take a look at prototyping splitting agencies <-> feeds in a google sheet, @hunterowens to meet next week 

this is non blocking for getting out reports.

machow:
It sounds like the relationship here will be resolved once feeds and agencies are split. Can this PR be closed?

PR #182, opened 2021-06-23
fix(airflow): add more validator field types

machow:
More validator report field types. Going to merge and check on pipeline..


machow:
@themightychris the md link checker is failing erroneously. Is this because it cannot follow re-directs?

PR #191, opened 2021-06-30
feat(views): add reports_gtfs_schedule_index

machow:
This new view contains one row for each agency on each date a new report would be generated, along with the start and end dates need for a report. See `description` and `fields` in the yaml for more info.

Critically, it uses agency info as it existed on each publishing date, so that changes to `agencies.yml` will not affect report generation for previous months.

cc @Nkdiaz, let's go over when pairing today


machow:
Merging, since this is running as part of reports build--will review with Natalya at 10am today

PR #199, opened 2021-07-12
GTFS Schedule Views StopId Changes Grouped By Month 

Nkdiaz:
After testing in Bitquery, sql script was then formatted for yml. It has been added to the view in Bigquery and the filter by month is available in Metabase:
https://metabase.k8s.calitp.jarv.us/question/35

TODO- have the aggregated number of stop_ids calculated as a total percentage to better match the chart previously done in jupyter notebooks


machow:
@Nkdiaz I've edited the metabase question to show a plot for a single date and feed. Note that on some dates (e.g. if you filter for July 1st) only a single bar shows up. This is because if a specific status is 0 (e.g. no added or removed ids), then there is not a row with 0 for that status, just a missing row.

machow:
rebased on main

cvc5185 (approved):
Pairing with @machow, made some quick tweaks (eg converted to new .sql file format)

Issue #183, opened 2021-06-24
Setting up cal-itp-data-infra-staging project


machow:
@thekaveman (cc @hunterowens), can we set up a new cloud project called `cal-itp-data-infra-staging`? This will allow us to make a copy of our SQL tables on bigquery when developing. For other parts of the pipeline--e.g. GCS--we just change the name, so this is mainly for bigquery.

Do you mind setting up the new project? (I'm happy to take a stab at it, but think someone from compiler might be needed!)

thekaveman:
@machow I will take a look! I haven't set up a new project before but it can't be too bad.

thekaveman:
I imagine this new project will incur some additional costs? @machow @hunterowens do we have sign-off from Jeremy et. al on the additional spend?

machow:
Right now the tables our duplicated into our current project, so the costs should stay the same! (Right now, we're not exceeding the free resources BigQuery provides per project :)

thekaveman:
Oh I understand now 👍 Moving those `staging` tables into a separate project.

thekaveman:
@machow Created the project and added you as an Owner. You should have received an email notification:

![image](https://user-images.githubusercontent.com/1783439/123331000-8f39bf80-d4f3-11eb-89d5-1fd0ae560bff.png)

/cc @hunterowens

machow:
Thanks! Can we add the same billing account to the new project? It looks like transferring datasets is disabled for free projects.

I think we should still be able to keep an eye on it from the same billing page as cal-itp-data-infra, but will double check to make sure!

thekaveman:
Done @machow

machow:
Looks like copying datasets across project works--so we're all squared away, thanks!

PR #201, opened 2021-07-16
refactor(views): converted yml files to sql files and modified brackets

cvc5185:
This refactor simplifies the code as a result of  https://github.com/cal-itp/data-infra/issues/183
Tested locally to confirm that the views worked


machow (approved):
LGTM--thanks

PR #184, opened 2021-06-24
Payments Data ELT Docs and Operators 

hunterowens:
This PR starts a docs section for payments data. 

Ref #131

PR #203, opened 2021-07-19
fix(airflow): do not split table name in CsvToWarehouse

machow:
This is a small fix, to stop pulling project_id out of the table name in one of the operators. I tested the sandbox locally before merging before, so not sure what happened....


machow:
going to merge, so we can rebase / re-check #184, then have @cvc5185 merge it

PR #207, opened 2021-07-20
fix: combined work from previous (unmerged) branch to to fix the issu…

cvc5185:
…es with payments-loader dag error


machow (approved):
If everything ran locally, LGTM!

machow:
going to merge, since I'm pretty sure we walked through these locally (I'll revert if anything goes foul).

PR #211, opened 2021-07-20
change DASH to LADOT

hunterowens:
changes DASH names to LADOT. cc @thekaveman @machow

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -342,8 +342,8 @@ culver-citybus:
       gtfs_rt_service_alerts_url: null
       gtfs_rt_trip_updates_url: null
   itp_id: 87
-dash:
-  agency_name: DASH
+ladot:
+  agency_name: Los Angeles Department of Transportation
   feeds:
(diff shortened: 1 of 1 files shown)


machow:
@hunterowens what is the purpose of the top-level field names? I thought these were essentially immutable keys to identify a set of feeds. (They're currently unused by the warehouse, so changing them won't have any effect.)

I'm trying to understand--if they can be changed--are they equivalent to using a top-level list like this?...

```yaml
- short_name: ladot
  agency_name: Los Angeles Department of Transportation
  feeds: ...
```

The reason I ask, is because it seems very likely that at some point in the future it will be better to identify feeds using a single immutable ID, and top-level feeds tend to be heading in that direction. (e.g. if ladot had mulitple feeds, we could identify them using "ladot__1", etc.., and we know other feed entries can't use "ladot")

hunterowens:
No real purpose, actually, aside from basically being the short name or key
for use in display names, etc. but in the new structure I think will want
to do something similar to ^

machow (approved):

Issue #49, opened 2021-04-05
analyze feeds that fail to download


hunterowens:
Now that @mcplanner is back, we should try and analyze why the following feeds are failing. @machow can also pull some downloader log stats to help inform this. 

The follow agencies failed to have GTFS a GTFS feed atthe URL or the Zip File Failed to extract:['e-Tran', 'Foothill Transit', 'Manteca Transit', 'Menlo Park Shuttles', 'Paso Express', 'San Luis Obispo Regional Transit Authority', 'South County Area Transit', 'Tri Delta Transit', 'Tri-Valley Wheels']{{ ds }}

machow:
Hey--dumping a more complete table for feeds that failed to download, so we can reference. Happy to investigate any rows :)

|                                             | agency_name                                 | status           | gtfs_schedule_url                                                                                                                            |   itp_id |
|:--------------------------------------------|:--------------------------------------------|:-----------------|:---------------------------------------------------------------------------------------------------------------------------------------------|---------:|
| ac-transit                                  | AC Transit                                  | HTTP error: 401  | https://api.actransit.org/transit/gtfs/current?token=2512B8179D2DC44895CDDC6D42                                                              |        4 |
| b-line                                      | B-Line                                      | HTTP error: 404  | http://www.blinetransit.com/documents/google_transit.zip                                                                                     |       48 |
| clean-air-express                           | Clean Air Express                           | HTTP error: 404  | http://www.cleanairexpress.com/GTFS7.24%COVID-19%CAE%GTFS.zip                                                                                |      nan |
| culver-citybus                              | Culver CityBus                              | HTTP error: 404  | https://www.culvercity.org/home/showdocument?id=169                                                                                          |       87 |
| dinuba-area-regional-transit-               | Dinuba Area Regional Transit                | HTTP error: 404  | https://tularecog.org/tcag/data-gis-modeling/gis-project-and-data/gtfs-data-as-of-june-19-/dart-dinuba-area-regional-transit-gtfs/           |       93 |
| grapeline                                   | Grapeline                                   | HTTP error: 404  | http://www.mjcaction.com/MJC_GTFS_Public/lodi_google_transit.zip                                                                             |      168 |
| long-beach-transit                          | Long Beach Transit                          | HTTP error: 404  | https://lbtransit.box.com/shared/static/aoyeskwmsa9g7pyg78q3xuiolgqe4f.zip                                                                   |        1 |
[...]

PR #214, opened 2021-07-21
fixed errors flagged in daily reports

mcplanner:
As mentioned in issue #49 
* Updated Foothill, Manteca
* Removed E-Tran (they're part of SacRT now)
* South County and Paso Express are part of SLORTA, so I removed those entries as duplicates

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -481,7 +481,7 @@ folsom-stage-line:
 foothill-transit:
   agency_name: Foothill Transit
   feeds:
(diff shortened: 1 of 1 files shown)


hunterowens:
I don't actually see the e-tran getting removed here?

Confirming that South County and Paso express trips are represented in the main SLORTA feed?

mcplanner:
E-Tran is now removed. South County, Paso, and SLORTA all have the same URL.

hunterowens (approved):

PR #223, opened 2021-07-26
Metabase hostname dashboards.calitp.org

### LINKED ISSUE #222: Add dashboards.calitp.org to metabase deployment
This additional domain has been added to metabase by hand. The change needs to be captured permenently in the deployment values.

lottspot:
Resolves #222


machow (approved):
LGTM Thanks! (From discussion it sounds like this change is deployed)

PR #230, opened 2021-07-28
attempted to resolve missing feed issues

mcplanner:
* removed the broken link from Tri Delta Transit, since it was a duplicate anyway.
* corrected Burbank's to their GMV links.
* replaced the regional MTC feed links with the URLs that hide the API key.

SLORTA is still broken, and I contacted the agency to ask for the correct link.

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -1297,7 +1297,7 @@ torrance-transit-system:
 tri-delta-transit:
   agency_name: Tri Delta Transit
   feeds:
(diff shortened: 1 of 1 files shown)


machow:
@mcplanner, I'm noticing that service alerts does not have a parameter for agency, but other RT feed urls do. How will this work? Does it also need an agency parameter?

image

mcplanner:
Service alerts don't for 511, no. It's one regional alerts feed.

machow:
Ah, okay--thanks!

machow (approved):

Issue #227, opened 2021-07-26
July Reports


hunterowens:
Creating a tracking issue for the July reports, once july has ended (in 4 days? Time flies)

machow:
Read my mind--thanks! When do we want to send the reports out? Does the 6th (or even 9th) work? If so, I can...

* check @ryon's capacity to do some small style updates
* collect / review / write-up some of the improvements that have been proposed into one place
* estimate which pieces we could fit in by then (e.g. using feed_info to calculate whether a feed is expired is not in the warehouse, but was calculated for #192, so could be slotted in fairly quickly).

mcplanner:
Would August 2 be feasible? (The first workday of the month).

This is the only bug I see with the current reports that I know to be
blocking. What else?

https://github.com/cal-itp/data-infra/issues/219

machow:
Some of these can be handled async. My main goal in listing here is to check whether it's worth collecting these kinds of issues / addressing some before next release.

* reports do not get generated if an agency does not have a feed_info.txt
* you and Hunter prefer different metrics for the service hours plot (switched to sum-total of hours across running trips at his request)
* days feed expired calculation is currently this: sum days where service hours are 0
* may be useful to review this [human readable validator error spreadsheet](https://docs.google.com/spreadsheets/d/1GDDaDlsBPCYn3dtYPSABnce9ns3ekJ8Jzfgyy56lZz4/edit#gid=0) to ensure we summarize in the way you want. For example, are you okay with these 200 details being individually reported?

![image](https://user-images.githubusercontent.com/2574498/127189853-05065737-0da6-4b42-a241-2dcb76a2e6d0.png)

mcplanner:
- I was not aware that reports weren't getting generated if there's no
   feed_info.txt. Thank you for flagging that. Should that be a separate issue
   we can track?
   - I defer to Hunter's metric for service hours. If that's the only issue
   on this one, we can close this one.
   - What can I do to help clarify the "days feed expired" item? A short
   term fix would be to change the name to accurately reflect what's being
   shown, but longer term it should reflect days when there wasn't a valid
   feed based on start/end date. Also, would this be helpful to track with its
   own ticket?
   - I don't think I understand this last point. What 200 details? Is there
   one that looks especially unwieldy?

Thanks,

machow:
Ah, I'm not sure if it came through on email, but I linked an image of validation errors in this report: https://reports.calitp.org/gtfs_schedule/2021/06/269

[...]

Issue #90, opened 2021-05-06
dump mobility gtfs-validator error code schemas as json or yaml


machow:
Currently, the pipeline loads the json report from gtfs-validator, which includes about 100 code-specific fields. However, it has 2 limitations:

* when new fields are added by the validator, they will need to be added manually (a quick task)
* we can't say--for this code, these are the fields it uses (e.g. shapeDistTraveled, etc).

The second would be useful for analysis, since it will let us pull code-specific columns out of the very wide validation_notices table. Note that we pull this info from the reports themselves, but in the long run it'd be useful to have a way to generate it from the gtfs-validator repo.

**edit:** here's a quick script for pulling out fields based on the data

```python
import gcsfs
import json

fs = gcsfs.GCSFileSystem(project="cal-itp-data-infra")
reports = fs.glob("gs://gtfs-data-test/schedule/processed/*/validation_report.json")
reports_json = [json.load(fs.open(fname)) for fname in reports]

code_fields = defaultdict(lambda : set())
for report in reports_json:
    for notice in report["notices"]:
        for entry in notice["notices"]: code_fields[notice["code"]].update(entry.keys())
```

Note that one notice field, `suggestedExpirationDate` is a dictionary with subentries (of form `{"localDate": {"year": int, "month": int, "day": int}}`), which are not captured by the script (and maybe should be named in our flat table `suggestedExpirationDate_localDate_year`, etc..?).

e-lo:
> The second would be useful for analysis, since it will let us pull code-specific columns out of the very wide validation_notices table. Note that we pull this info from the reports themselves, but in the long run it'd be useful to have a way to generate it from the gtfs-validator repo.

I agree that this should be an issue in https://github.com/MobilityData/gtfs-validator

machow:
Thanks for the nudge--was going to focus on getting a table up the quick way before filing an issue, but tossed one up to stew in the meantime!

https://github.com/MobilityData/gtfs-validator/issues/899

hunterowens:
@machow to close when #129

machow:
Running in pipeline--will keep an eye on the mobilitydata issue, in case we can pull out the schema more easily  later on!

PR #239, opened 2021-08-02
fix(airflow): coerce specified validation fields to strings

machow:
Blocking #227 , related to loading validation schema #90. The gist is that there is a validation field "fieldValue" that can either be a string or an int. However, bigquery does not like this. Even though we code it as a string in the schema, bigquery complains that it is getting an integer.

The change below fixes this by coercing specified fields to string, when processing validation results. Will merge, then clear processing for July 31, when this issue occured in the results.

(tested by running on dev)


machow:
merging so I can get reports out the door

PR #240, opened 2021-08-02
Correct SLORTA feed

mcplanner:
Based on the link from their website here: https://www.slocity.org/government/department-directory/public-works/slo-transit/general-service-information


mcplanner:
Actually, this isn't the right URL either. From Olivia:

We have already had our meeting with them where we learned this link is very old. They switched vendors back in 2018 and now use Connexions. They are working on getting the link to their current feed- their director of operations, Omar McPherson, was going to track it down. Once he does so, they agreed to put it on the website and submit it to the feed aggregator sites. I don't have this link yet, though.

So moving this to Waiting while we get more info.

machow:
Looks like this is tracked in issue #256 -- which makes sense to me! Going to close this, so we can keep outstanding PRs down

PR #242, opened 2021-08-04
document warehouse

machow:
Creating demo tables for the sandbox dataset to show basic analyses, dependencies, and metabase.


machow:
pairing w/ @cvc5185 . Merging, so we can go over metabase w/ these in-progress tables.

PR #263, opened 2021-08-09
feat-schedule-type-2: checked out to proper branch, kept parse_Date f…

Nkdiaz:
…or calendar, feed, info, routes, and trips


machow (approved):
LGTM. There's some small fixes, e.g. no key on calendar_dates_clean, but I can toss them in when integrating with the golden view work!

PR #268, opened 2021-08-10
GTFS schedule feed snapshot table

Nkdiaz:
WIP for creating a gfts schedule feed snapshot table. Still need to add columns and add md5 hashes to track changes


machow (approved):
LGTM

PR #286, opened 2021-08-18
feat: moved payments SQL files and metadata to their own payments_vie…

cvc5185:
…ws dag folder, added timestamp to  mst_transactions_gtfs_enhanced to fix error, deleted old payemnts SQL files from gtfs_views


machow (approved):
LGTM thanks!

PR #290, opened 2021-08-18
fix(kubernetes/metabase): prevent kill loops on long startups

### LINKED ISSUE #217: Add startupProbe to metabase deployment
In order to prevent metabase from entering a crash loop in the event of a slow startup (e.g., should DB migrations need to be performed before startup completiion) we need to add a startupProbe to the metabase helm chart.

Example: https://github.com/urbanSpatial/LetsPlanOrg/blob/901dc868b44cc743dfc1a948820946034b80ded9/helm-chart/templates/metabase.yaml#L30

lottspot:
resolves #217


machow (approved):
I am a kubernetes barbarian, but on it's face LGTM

PR #305, opened 2021-08-23
Added all agency-specific feeds for the MTC region

mcplanner-zz:
Also confirmed that all agencies with two real time options have those options properly grouped.

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -1,14 +1,14 @@
 ac-transit:
   agency_name: AC Transit
   feeds:
-    - gtfs_schedule_url: http://api.511.org/transit/datafeeds?api_key={{ MTC_511_API_KEY}}&operator_id=AM
-      gtfs_rt_vehicle_positions_url: http://api.511.org/transit/vehiclepositions?api_key={{ MTC_511_API_KEY}}&agency=AM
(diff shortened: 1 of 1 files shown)


machow:
@mcplanner thanks. Just a heads up--indentation matters in this yaml file. I've fixed the indentation, and it appears some entries are either missing url fields (there need to be 4 fields, one for fixed schedule and 3 for RT related data), or the fields may be misnamed. Can you check the error report and correct?

I pulled this from the error check, so something around that entry is misspecified.

```
'gtfs_schedule_url': 'https://gtfs.vta.org/gtfs_vta.zip'
```

image

machow:
Per early discussion with @hunterowens, the format chosen for agencies.yml requires that people put new feeds later in the feeds list, so if the gtfs_vta link is being added, we need to put it at the end of the feeds list (or the pipeline will think the original 511 feed is now at a new url)

image

mcplanner:
I was really sure I lined up all the left hand spaces before I committed
this. Is there a subtle trick I'm missing?

machow:
I'm not sure--it's pretty easy for me to run the formatting automatically on my laptop, so I can handle that part, if you can check when it flags fields as missing!

machow:
@nkdiaz -- do you mind going through and trying to resolve the error in this agencies.yml file? The details from the failed GH action should give some clues (see my comment above).

machow (approved):
Based on visual scanning the changes--LGTM, thanks

PR #328, opened 2021-08-30
fields_from implementation

### LINKED ISSUE #325: Complete column description implementation in SqlToWarehouseOperator
Key pieces

* calitp-py [sql_patch_comments](https://github.com/cal-itp/calitp-py/blob/main/calitp/sql.py#L165)
* [get_table](https://github.com/cal-itp/calitp-py/blob/main/calitp/sql.py#L58) - fetch a table and its metadata as a sqlalchemy Table. E.g. use `table_obj.columns.keys()` to get column names.
* See this [example of a task with fields](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/gtfs_views/validation_stg_code_descriptions.py#L4-L6)

Let's add examples to sandbox that capture...

* adding a single field description
* using fields_from
  - with any
  - with specific fields

## Spec

```yaml
fields:
  column_a: some description
  column_b: another one
fields_from:
  table_a:
    - column_c
    - column_d
  table_b: any
  # or
  table_b:
    - any
```

ccjarvus:
Fixes #325

Adds functionality for a `fields_from` block in the `SqlToWarehouseOperator`, and includes usage example in sandbox DAG task `op_sql_to_warehouse.sql`.

Two ways designate fields from:

```
fields_from:
    table_a:
        -  individual
        -  fields
        -  here
    table_b: any
```

Where "any" is a special keyword.

`fields_from` checks that fields listed exist in the destination table and prioritizes fields listed from top to bottom. So if there are two tables that feature "any", the one listed first will get preference. Fields listed in the operator's `fields` blocked are prioritized over any fields listed in `fields_from`.

### DIFF
--- a/airflow/dags/sandbox/op_csv_to_warehouse.sql
+++ b/airflow/dags/sandbox/op_csv_to_warehouse.sql
@@ -3,8 +3,8 @@ operator: operators.CsvToWarehouseOperator
 table_name: "sandbox.csv_to_warehouse"
 src_uri: "https://docs.google.com/spreadsheets/d/1Ed62uU-SJYoV7ecEQ61aT-9FzF1R5W2w-V1D7Bd_Ib4/export?gid=0&format=csv"
 fields:
-  g: The g field
-  x: The x field
+  g: The g field csv
+  x: The x field csv
 
 dependencies:
   - create_dataset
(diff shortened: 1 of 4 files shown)


machow (approved):
Thanks--I found it a bit difficult to track dict comprehensions with 4 conditions + their reference to a variable being updated in a loop, so re-wrote it to..

* grab all columns specified from a parent
* at the end: unpack each parent's col descriptions into a single dictionary in reverse order, so earlier column definitions take precedent over later ones (which appears to be the original spec, and that makes sense to me)

PR #332, opened 2021-08-31
ops(kubernetes/metabase): metabase was becoming unavailable due to load

lottspot:


machow (approved):
LGTM thanks! Do you know if it was running out of memory, or was CPU related?

PR #349, opened 2021-09-07
fix(content): revert daily feed files

machow:
interpolating files for missing days requires a bit more finessing. I grouped by agency, which was the wrong action. Going to revert.


machow:
We ended up re-implementing this when putting in tests

PR #353, opened 2021-09-08
Origin/caltrans geoportal export

Nkdiaz:
added requests.raise_for_status for each request so that if it fails it prints the exception


machow:
LGTM thanks!

PR #374, opened 2021-09-14
fix: update LADOT GTFS url, remove commuter express

hunterowens:
This updates the LADOT gtfs url to avoid us their feed.


machow (approved):
Thanks--it looks like the gtfs url got duplicated so I did a quick fix! (looks like the URL works, so going to merge :).

PR #378, opened 2021-09-14
feat: add amtrak

hunterowens:
Adds a GCS url with the Amtrak feed for temporary use.


machow (approved):
Thanks. I moved Amtrak to be in schedule-assets, since sandbox is a folder many people interact with as part of airflow onboarding via the [sandbox DAG](https://github.com/cal-itp/data-infra/tree/main/airflow/dags/sandbox), and it's useful to delete sandbox from time to time..!

PR #381, opened 2021-09-17
ParseRtOperator

ccjarvus:
Creates a `ParseRtOperator` that converts rt vehicle_positions files into parquet and stores them in the `gtfs-data/rt-processed/vehicle_positions` bucket.

Some notes:

- All rt subfolders were posix timestamps, so I had to convert Airflow date context to posix in order to get things working properly. I am aware that we were supposed to receive a fix from ops to have dates/timestamps instead of posix but I did not want to let that stop me from proceeding.
- Had trouble retrieving the date context using a `.py`, so I made an operator instead to unblock myself. In Airflow 2, I can just include a context key (e.g. `execution_date`) in my python_callable and it comes out the other end, but wasn't working here.
- Though it was initial discussed to use the path `rt-processed/vehicle_positions/vp_{date*}_{itp_id}_{url_number}.parquet`, due to the folder structure, it was a lot simpler to just parse the top level posix folder's subfolders (itp_ids and urls) once and generate a single parquet across all itp_ids and urls. So the format is now `rt-processed/vehicle_positions/{posix timestamp}.parquet`. The file path is located in each parquet file, where itp_id and url can easily be extracted.

Potential optimizations:

- ~~We could have a posix date folder that contains all posix_timestamp.parquet files for a given posix date. Instead of many many timestamps.~~ (Done)
- If we can pull the execution_date from context into a .py, there wouldn't be a need to have a ParseRtOperator.
- In contradiction to the above, the ParseRtOperator is probably just one refactor away from being able to parse non-vehicle_positions rt protobuff data.

Notes on size:

- Each parquet so far is around 48 KB, so at 500 files per day I would expect this to become roughly 9 GB worth of data per year (48 KB * 500 files daily * 365 days / 1000 for MB / 1000 for GB)


ccjarvus:
@machow - I went ahead and adjusted the `gtfs-data/rt-processed/vehicle_positions` bucket and the `ParseRtOperator` to use posix date subfolders instead of keeping all parquet files on the same level in `gtfs-data/rt-processed/vehicle_positions`.

I also added a `limit` arg to `ParseRtOperator` so that the sandbox job doesn't run the full job.

airflow/dags/rt_loader/parse_rt_vehicle_positions.py
```diff
@@ -0,0 +1,143 @@
...
+def parse_pb(path, open_with=None):
+    """
+    Convert pb file to Python dictionary
+    """
+    open_func = open_with if open_with is not None else open
+    feed = gtfs_realtime_pb2.FeedMessage()
+    try:
+        feed.ParseFromString(open_func(path, "rb").read())
+        d = json_format.MessageToDict(feed)
+    except DecodeError:
```
machow, line 24: Why is this very lenient except block here? Did you encounter files that didn't parse?

airflow/dags/rt_loader/parse_rt_vehicle_positions.py
```diff
@@ -0,0 +1,143 @@
...
+        print("Creating " + file_name)
+
+        if len(files) > 0:
+            parsed_positions = [parse_pb(f, open_with=fs.open) for f in files]
+            positions_dfs = [*map(rectangle_positions, parsed_positions)]
+            positions_dfs = [df for df in positions_dfs if df is not None]
+            if len(positions_dfs) > 0:
+                positions_rectangle = pd.concat(positions_dfs)
+                with tempfile.TemporaryDirectory() as tmpdirname:
+                    fname = tmpdirname + "/" + r.split("/")[-1] + ".parquet"
```
machow, line 139: Instead of re-using `r`, a final loop scope variable 30 lines after it's for loop has concluded, please define as a variable what you want r.split to represent much earlier. (I am not sure what it's intended to be).

machow: It seems like this is a bug?

machow (changes requested):
@ccjarvus this task seems to run very slowly. I ran it locally, and it took about 10 mins per feed (and then it got sig killed).

I'm going to make some changes / kick the tires a bit more--please hold off on pushing commits here in the meantime!

airflow/dags/rt_loader/parse_rt_vehicle_positions.py
machow: Ah, I think I get it--this file could be named anything..

airflow/dags/rt_loader/parse_rt_vehicle_positions.py
ccjarvus: I encountered files that didn't parse.

machow:
Alright--merging in changes we paired on. Thanks for doing all the leg work to get this set up!

PR #385, opened 2021-09-21
fix ladot

hunterowens:
Fixes the LADOT swapped URLS that @e-lo caught 
- fix: update LADOT GTFS url, remove commuter express
- feat: add gtfs-rt
- fix: ladot swapped urls


machow:
Thanks--opened another issue to make sure we fix any names. I think the swapped names won't break our loaders (they should produce NULL values instead), but may be useful to go back and correct.

https://github.com/cal-itp/data-infra/issues/386

Issue #372, opened 2021-09-14
Payments Research: Prevalence of riders hitting fare caps


mjumbewu:
## Goal(s)

* Reduce burden on riders
* Increase equity

## Question

For people using contactless, how many people are being auto-fare capped?

## Metrics

* % of unique customers who (1) are using contactless payment, and (2) hit the fare capping limit **(daily? weekly? should those be considered separately? together?)** compared to total unique customers (MONTHLY) 

## Data sources

* Littlepay (micropayments) which shows fare type
* Littlepay (micropayment adjustments) which shows fare caps
* Littlepay (device transactions) which shows route/bus
* **No source for total _unique_ riders at the moment?**

## Presentation(s)/Visualization(s) needed

machow:
Because this is % of unique riders, this requires customer funding source

mjumbewu:
This is now available in the https://metabase.k8s.calitp.jarv.us/question/260-percent-of-riders-with-adjustments-per-week-grouped-by-adjustment-type chart on the dashboard.

Issue #361, opened 2021-09-09
Load a new payments table


machow:
See the payments docs on [DAG Maintenance and adding new tables](https://docs.calitp.org/data-infra/datasets/mst_payments/#dags-overview). Let's load a new data source, to make sure the documentation is accurate. We can definitely change how the process works to make it easier if needed.

PR #393, opened 2021-09-23
Ingest micropayments_adjustments Littlepay data

mjumbewu:
The miropayments_adjustments data has fare cap values, which will be used in reporting metrics. This work is in relation to:
* #361 
* #372 

Ran the `payments_loader` DAG locally to check that everything flows correctly. New table is loaded into staging.


machow (approved):
LGTM, thanks!

PR #397, opened 2021-09-23
feat(gtfs-schedule): add days from last extract to feed daily

machow:
This PR adds a column that tracks how many days its been since the feed updated, so we can add last published at to the gtfs schedule reports.

TODO: document fields of this view


evansiroky:
@machow, if you have any time left can you check whether this is still worth implementing?

machow:
Hey--I'll run through this with @lauriemerrell, since it's a good introduction to the gtfs schedule data!

lauriemerrell:
@mjumbewu - Ready for substantive review. The logic here was implemented by Michael, and I have done some testing and verified that I believe it is working as intended. As we discussed last week, the new column is tracking when files in the feed are changed. 

@charlie-costanzo - Wondering if you can review the new field descriptions here and give any thoughts? Trying to make use of the new/upcoming docs support since we were already touching this.

charlie-costanzo (commented):
@lauriemerrell these field descriptions look excellent, thank you so much for taking care of these while you're in here!

mjumbewu (approved):
Ok, I agree it seems to work as intended. And the column descriptions are very helpful.

Regarding the `null` data, I'm ok with investigating that in a follow-on. And then we can add a validation check for it too. Otherwise, LGTM!

lauriemerrell:
@evansiroky I think this is ready for your review if you want to take a look before we merge!

evansiroky (changes requested):
@lauriemerrell, can you please modify this PR description to add a screenshot of your local airflow that shows that this DAG task passed?

lauriemerrell:
@evansiroky This specific DAG task and its downstream tasks ran successfully, a few unrelated tasks failed -- `trip_stops_latest` said I hit a query limit and I think `validation_notices` just essentially timed out. I am not too concerned about those because they are unrelated to this change and I think may just be hitting resource limits on my local Airflow/Docker. Are you comfortable proceeding with these as is?

image 
image

evansiroky (approved):
LGTM and thanks for posting the screenshot. That's fine that the others don't work locally, since they are working in production.

Issue #210, opened 2021-07-20
add SacRT, SBCG, 


hunterowens:

hunterowens:
waiting on the data sharing agreement, then littlepay to provision

machow:
@hunterowens is this resolved? If not can you say more about what this is for?

hunterowens:
currently the data is in a separate S3 bucket (1 bucket for MST, 1 for all three others). We are waiting on LP to provision the bucket and then need to duplicate the data transfer job and load into the warehouse to setup dashboards for all three other agencies

hunterowens:
@machow @mjumbewu etc. We now have the keys! Sent details to your email Michael, but we can setup a duplicate ETL to create tables. Let's discuss monday

machow:
See "SBMTD Data Feed access" email with instructions for how to add sources and rotate AWS keys.

mjumbewu:
Ran into a slight hiccup while going in to rotate keys. Sent an email to Littlepay about the issues. Email is pasted below:

> Hi Jakir,
> 
> I'm having some issues with the credentials that have been sent over. I went through the emails with the AWS credentials in them and added each to a profile in my local `~/.aws/credentials` file. I then ran `aws iam get-user` for each set of credentials that we have. I ended up with the following:
> 
> * `aws iam get-user --profile littlepay_mst`: user is `mst`
> * `aws iam get-user --profile littlepay_sbcag`: user is `cal-itp`
> * `aws iam get-user --profile littlepay_sacrt`: (invalid credentials)
> * `aws iam get-user --profile littlepay_sbmtd`: (invalid credentials)
> 
> I was confused about the SBMTD result as this key had been working for us before, but then looked at our GCP Data Transfer jobs and saw that the last run of our “Littlepay AWS -> GCS for SBMTD” failed because of invalid credentials. So, my questions:
> 
> * By any chance, did you set the access key for SBMTD to expire? That wouldn’t have been unreasonable, since we should have rotated the keys a few days ago, but I wanted to check.
> * Do you know why the SacRT credentials might be invalid? I double-checked that they're entered as we received them, and they're paired with the secret key that we received as well.
> * Lastly, it looks like the SBCAG credentials with the user `cal-itp` brought two additional `participant_id` values into our data: `sacrt`(!) and `clean-air-express`. Is that expected? Notably absent is any `sbcag` data.
> 
> Thank you!
> - Mjumbe

Issue #394, opened 2021-09-23
Bug: littlepay data sync needs to remove files from destination, copy to subfolder


machow:
**Describe the bug**

Currently, when we sync littlepay data for MST via a [data transfer job](https://console.cloud.google.com/transfer/cloud/jobs/transferJobs%2F14397376673459736076/configuration?authuser=0&project=cal-itp-data-infra). However, this job currently does not delete files from the destination that are not in source. This means causes issues with us keeping data that no longer exists (or was shuffled into a different file).

@hunterowens can you change delete "never", to remove unmatched files in destination? I don't think we have the keys for it (or if you send to me I can update the job).

![image](https://user-images.githubusercontent.com/2574498/134537629-f7a2fabd-28cd-40e6-a764-116740e03180.png)

hunterowens:
also gave michael credentials

machow:
@mjumbewu when you get the chance, can you make a new job that transfers littlepay MST data to a subfolder of littlepay-data-extract-prod, and then delete the original job?

PR #398, opened 2021-09-24
Update the proprocess_columns task to support multiple providers

mjumbewu:
* Introduce a new function in the `preprocess_columns` task to handle preprocessing of a given littlepay data source. If need be, we can move this into a new operator in the future.
* Add SBMTD Littlepay data to be preprocessed. The SBMTD data is being synced to the same bucket as the MST data, but into a different folder. ~~The MST data is still being synced to the root, but should also be changed to use a different folder. We may have to create a new sync task to do so, as there doesn't appear to be a way to modify the bucket/folder being synced to. After merging this code, we should:~~
  1. ~~Delete the old transfer task for MST~~
  2. ~~Create a new transfer task for MST pointing to a subfolder of the Littlepay data bucket~~
  3. ~~Submit a new PR changing the `preprocess_columns` task to not override the default `src_url_template` for MST~~
  
  **A new MTS transfer task has been created and the old one has been disabled for now.**

This code has been run locally to ensure functionality. This PR addresses #210

### DIFF
--- a/airflow/dags/payments_loader/preprocess_columns.py
+++ b/airflow/dags/payments_loader/preprocess_columns.py
@@ -99,10 +99,5 @@ def preprocess_littlepay_provider_bucket(
 
 
 def main(execution_date, **kwargs):
-    preprocess_littlepay_provider_bucket(
-        execution_date,
-        provider_name="mst",
-        src_url_template="gs://littlepay-data-extract-prod/mst/{table_name}/*.psv",
-    )
-
+    preprocess_littlepay_provider_bucket(execution_date, provider_name="mst")
     preprocess_littlepay_provider_bucket(execution_date, provider_name="sbmtd")
--- a/airflow/dags/payments_loader/preprocess_columns.py
+++ b/airflow/dags/payments_loader/preprocess_columns.py
@@ -20,22 +20,19 @@ DATASET = "payments"
 # However, if we move to processing only the data for a given day each task,
 # we should change the SRC_DIR glob to use a datetime
 
-SRC_DIR = "gs://littlepay-data-extract-prod/mst/{table_name}/*.psv"
-STAGE_DIR = "mst/{table_name}"
-DST_DIR = "mst/processed/{table_name}"
+DEFAULT_SRC_URL_TEMPLATE = "gs://littlepay-data-extract-prod/{provider_name}/{provider_name}/{table_name}/*.psv"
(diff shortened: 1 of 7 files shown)


machow:
Thanks! Looks like this also takes care of https://github.com/cal-itp/data-infra/issues/394

mjumbewu:
Yup, that's correct!

airflow/dags/payments_loader/preprocess_columns.py
```diff
@@ -50,24 +50,41 @@ def main(execution_date, **kwargs):
...
                 table_name=table_name.replace("_", "-"),
                 date_string_narrow=date_string_narrow,
             )
         )
 
         print(f"\n\nTable {table_name} has {len(src_files)} new files =========")
 
+        # remove previously processed data, in case they remove any data files ---
+
+        dst_old_url = f"{get_bucket()}/{dst_dir}/*_{provider_name}_*"
```
machow, line 71: Ah, this is way better than my previously deleting the whole directory..!

airflow/dags/payments_loader/preprocess_columns.py
```diff
@@ -20,23 +20,23 @@
...
-
-
-def main(execution_date, **kwargs):
-
+DEFAULT_SRC_URL_TEMPLATE = (
+    "gs://littlepay-data-extract-prod/{provider_name}/"
+    "{provider_name}/{table_name}/*.psv"
+)
+DEFAULT_STG_DIR_TEMPLATE = "{table_name}"
+DEFAULT_DST_DIR_TEMPLATE = "processed/{table_name}"
```
machow, line 28: Can we change the top-level folder from processed to `littlepay-processed`, and the staging dir to `littlepay-staging/{table_name}`? We'll have an `rt-processed` folder for realtime, so this should keep everything matched up!

edit: (or could see using `payments-processed` instead)

machow (changes requested):
Thanks! The one quick piece--do you mind changing the top-level folder names? So we'd have `littlepay-processed` and `littlepay-staging` (or could use `payments` instead of the term `littlepay`)

Right now it's a bit tricky to tell which ones are payments related folders

image

airflow/dags/payments_loader/preprocess_columns.py
mjumbewu: That sounds good. I'll make that update.

machow (approved):

PR #403, opened 2021-09-28
Add dependency on `payments_feeds` for the transactions view

mjumbewu:
Small change: when there are changes to the `payments_feeds` table, we want those to run before doing anything with the `mst_transactions_gtfs_enhanced` view.


machow (approved):

machow:
Ah, thanks!

PR #412, opened 2021-10-01
Split Littlepay processing into separate for each source

mjumbewu:
Pre-processing one source takes enough time that it makes sense to have all the processes run in parallel, especially for testing. Instead of creating an operator, I added a `plugins/libs/` package and created `littlepay.py` to hold the common code. I don't have strong opinions about the names, though I do appreciate the non-operator approach.


machow (approved):
This looks great--thanks! The splitting out processing data sources looks really useful. Down the road, if we want, we could turn the tasks calling preprocess_littlepay_bucket into an operator, but given how few sources this will be could see it not being worth it..

PR #418, opened 2021-10-04
feat(push_to_gcloud): push agencies.yml changes to kubernetes cluster apps

lottspot:
This changeset automates the rollout of any changes to the agencies.yml file into the apps cluster and restarts the gtfs-rt-archive service upon such changes being pushed. Any future services deployed into the apps cluster which are dependent on the agencies.yml file will need to have lines added to the workflow to ensure they are also restarted after a change is pushed.

### DIFF
--- a/.github/workflows/push_to_gcloud.yml
+++ b/.github/workflows/push_to_gcloud.yml
@@ -32,14 +32,20 @@ jobs:
           './airflow/data/agencies.filled.yml' \
           './airflow/data/schema_agencies_entry.yml'
 
-    # The following only runs on the main branch ----
-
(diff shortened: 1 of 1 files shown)


lottspot:
This action has been [tested in preprod](https://github.com/cal-itp/data-infra/actions/runs/1303842636)

machow:
@lottspot thanks for this! Do you know how long a restart takes?

lottspot:
Not exactly, but in my test restart the gap between the last fetch of the old container and the first fetch of the new container was 19 seconds

machow (approved):
LGTM thanks!

PR #421, opened 2021-10-04
fix(kubernetes/gtfs-rt-archive): use recreate strategy to deploy changes

lottspot:
By default, when updating deployments kubernetes will create new
containers before terminating old ones. This behavior is undesirable
for the archiver since we want to ensure there is never more than one
instance running at a time.


machow (approved):
Thanks! I wasn't sure which order to merge the two PRs in. Do you mind merging them? :)

PR #426, opened 2021-10-04
Use ISO timestamp format for gtfs-rt feed filenames

lottspot:
This timestamp format is more human readable while still being easily parsed by machines


lottspot:
This is a merge-back of a change which has already been deployed into production

machow:
LGTM Thanks!

PR #430, opened 2021-10-07
New view: Reports weekly file checks

addiebundren:
This is the first step to resolve https://github.com/cal-itp/reports/issues/38


addiebundren:
@machow Found and fixed an issue with the way interpolation was implemented. Spot checked the data and it looks a lot better. LMKWYT.

Two questions: 
1. does `gtfs_schedule_history.calitp_files_updates` need to be listed as a dependency?
2. Is `views.reports_weekly_file_checks` in the format you are expecting?

PR #474, opened 2021-10-12
[WIP] jupyterbook -documentation migration 

Nkdiaz:
ongoing WIP for migrating existing pages to jupyterbook


machow:
Thanks for pushing up this WIP PR! Just a heads up, we should ignore the `_build` subdirectory. This is in our .gitignore file as `docs/_build`, but because you're using another name for the docs folder while swapping over (which makes sense) it is getting committed

PR #493, opened 2021-10-14
updated Culver City's URL

mcplanner:
The link is not stable. It is referred to on this page: https://www.culvercity.org/City-Hall/Departments/IT/Download-GIS-Data


machow:
thanks!

PR #494, opened 2021-10-14
[DO NOT MERGE] Removed Plumas Transit's expired GTFS link

mcplanner:
The feed was had is no longer supported by the provider.


evansiroky:
Marking the feed as null is probably not the way to go and this will need some kind of software development. To be addressed in #899.

PR #518, opened 2021-10-18
Feat airtable operator

machow:
Adds a proof-of-concept AirtableToWarehouseOperator that is able to extract data from airtable and load it into a BigQuery dataset or save as a CSV to Google Cloud Storage.


evansiroky:
Rebased so this works with airflow v2. I also modified the airflow operator to be able to upload a raw csv file to gcs.

evansiroky:
Just added some docs, so I think this is good to go.

holly-g:
Paging @mjumbewu and @Nkdiaz for review

holly-g:
@lauriemerrell, for reference

airflow/plugins/operators/airtable_to_warehouse.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def airtable_to_df(
+    air_base_id,
+    air_table_name,
+    id_name="__id",
+    rename_fields=None,
+    column_prefix=None,
+    api_key=None,
+):
```
mjumbewu, line 17: **thought:** This feels like a good candidate for a simple automated test. It doesn't look like we have any automated code test patterns set up in data-infra; something to add to the list.

airflow/docker-compose.yaml
```diff
@@ -54,14 +54,16 @@ x-airflow-common:
     AIRFLOW__CORE__PLUGINS_FOLDER: /opt/airflow/gcs/plugins
     AIRFLOW__WEBSERVER__RELOAD_ON_PLUGIN_CHANGE: 'true'
     CALITP_BQ_MAX_BYTES: 10000000000
-    # connections
-    # see https://stackoverflow.com/a/55064944/1144523
-    AIRFLOW_CONN_GOOGLE_CLOUD_DEFAULT: "google-cloud-platform://:@:?extra__google_cloud_platform__project=cal-itp-data-infra"
```
mjumbewu, line 59: **question:** This is just clean-up, right? I.e., this change has nothing to do with this PR, but we aren't using this Airflow connection anywhere, correct?

airflow/plugins/operators/airtable_to_warehouse.py
```diff
@@ -0,0 +1,124 @@
...
+                airtable base
+            table_name (str): The name of the table to save to in Big Query
+            id_name (str, optional): The name to give the ID column. Defaults to "__id".
+            gcs_path (str, optional): A GCS path prefix. If not provided, no uploading
+                to GCS will occur. The resulting item gets saved to GCS at the following
+                path:
+                    `{base_calitp_bucket}/{gcs_path}/{execution_date}/{table_name}.csv`.
+            rename_fields (list, optional): A list of objects where each object consists
+                of a string key and value representing the new desired column name and
+                airtable column name respectively. Defaults to None.
```
mjumbewu, line 86: **issue:** Is this correct? Seems like it's a `dict` above in `airtable_to_df`.

```suggestion
            rename_fields (dict, optional): A mapping new desired column name (string) to
                current airtable column name (string) respectively. Defaults to None.
```

**nitpick:** On a related note, why are we using `new_name: old_name`? I feel like I've more commonly seen `old_name: new_name`, and would probably expect that in the absence of instructions. I don't think I have a strong preference about it, but am curious why this order.

mjumbewu, line 86: **nitpick:** On a related note, why are we using `new_name: old_name`? I feel like I've more commonly seen `old_name: new_name`, and would probably expect that in the absence of instructions. I don't think I have a strong preference about it, but am curious why this order.

airflow/plugins/operators/airtable_to_warehouse.py
```diff
@@ -0,0 +1,124 @@
...
+            self.api_key,
+        )
+
+        if self.table_name:
+            print(f"Writing table with shape: {df.shape}")
+            write_table(df, self.table_name)
+
+        if self.gcs_path:
+            gcs_file = (
+                f"{self.gcs_path}/{context['execution_date']}/{self.table_name}.csv"
```
mjumbewu, line 121: **suggestion:** I'm not sure how GCS feels about double slashes in path names, but it might be good to either trim slashes from the right side of the `gcs_path`, or (maybe better) use `os.path.join` or something to build the file path.

mjumbewu (changes requested):
Just a few comments on this one. Tried using [conventional comments](https://conventionalcomments.org/) to communicate intent. Overall, looks good so far.

evansiroky:
> Just a few comments on this one. Tried using [conventional comments](https://conventionalcomments.org/) to communicate intent. Overall, looks good so far.

@mjumbewu I love the idea of conventional comments! I desire to get around to taking a closer look at your comments soon, but need to do other non-software-coding stuff for a bit. 😐

airflow/plugins/operators/airtable_to_warehouse.py
evansiroky: Yeah, you're right. I went ahead and made my own commit (1d27b283d0e5b091ad3d2a0e0b20a65c649c2b7e) to improve your wording a bit.

evansiroky: I prefer it that way too and went ahead and changed it to be that way in 1d27b283d0e5b091ad3d2a0e0b20a65c649c2b7e.

airflow/plugins/operators/airtable_to_warehouse.py
evansiroky: I opted for the first approach in 1d27b283d0e5b091ad3d2a0e0b20a65c649c2b7e.

evansiroky:
@mjumbewu, this is ready for review again.

mjumbewu (approved):
Thanks for making these changes; LGTM!

Issue #277, opened 2021-08-12
Track / load remaining files included in GTFS schedule spec


machow:
I'm not sure if these were missed when I was loading 6 months ago, or were recently added to the docs, but there are 3 files we currently don't load into the warehouse.

* attributions.txt
* pathways.txt
* translations.txt

It looks like relatively few feeds use these files right now (but analyzing e.g. pathways will likely be increasingly important).

![image](https://user-images.githubusercontent.com/2574498/129260087-410f0458-5fc2-49ab-89ff-f5b56d9df94c.png)

In order to load these, we need to...

* add them to the gtfs_schedule_history DAG (see e.g. the agencies task there; [gtfs schedule spec](https://developers.google.com/transit/gtfs/reference) for column types)
* add them to the [calitp_included_gtfs_tables](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/gtfs_schedule_history/calitp_included_gtfs_tables.py) task
* delete our gtfs_schedule_type2 tables
  - `gtfs_schedule_type2.{table_name}`
* delete gs://gtfs-schedule-data/schedule/processed
* re-run the gtfs_loader DAG across history (only run a couple days at a time, to not choke airflow)
* re-run gtfs_schedule_history2.merge_updates

Completing this would resolve https://github.com/cal-itp/data-infra/issues/180

machow:
Feel free to move all the file tracking tasks into their own dag (that depends on past)

![image](https://user-images.githubusercontent.com/2574498/131529477-d1a24308-fe1c-4775-92db-cfb385b99f77.png)

PR #539, opened 2021-10-21
Add additional files within GTFS static spec, improve security, create pathways dim and fact tables, make local development work easier to do

### LINKED ISSUE #238: Airflow ExternalTable operator should replace existing table
This basically only affects `gtfs_schedule_history.validator_report`, which has to use the google-bigquery python API. Currently, it's okay if the table is already created, but I don't think it replaces it in this case:

https://github.com/cal-itp/data-infra/blob/main/airflow/plugins/operators/external_table.py#L38

(not urgent: the solution right now is to delete the table before resetting the task)

### LINKED ISSUE #247: fix running generate_provider_list on dev
Currently, we...

* fill in API keys in airflow/data/agencies.yml
* upload the filled file to a gcs airflow bucket, as data/agencies.yml
* upload the original unfilled file as data/agencies_raw.yml

However, as a relic from when we didn't have to fill in agencies.yml,  the dev airflow looks for agencies.yml and agencies_raw.yml on the local disk (which is volume mounted). This means that you have to manually copy these files into airflow/data in the local repo.

We should either...

* add instructions for running this task (e.g. run something in `/script` etc...), OR
* modify the task so that it just downloads from the prod bucket (e.g. with an option to use the current local behavior)

### LINKED ISSUE #622: Security vulnerability: some API keys for downloading GTFS files could be accessed without authentication/authorization
**Describe the bug**

A link to a google sheet that lists some "secret" API keys for downloading GTFS files is exposed in plain text in code.

**To Reproduce**

See discussion on slack due to sensitive nature of this issue.

**Expected behavior**

Environment variables should be used to fill out the keys instead of a publicly-available google sheet.

**Additional context**

The API keys should probably be rotated.

### LINKED ISSUE #641: ExternalTable operator does not delete and replace existing table when re-run
Currently, the ExternalTable operator sometimes uses google's python bigquery library (e.g. to accommodate the nested field structure in the gtfs_schedule_history.validation_report task). This is because--at the time that the task was implemented--sqlalchemy support was still emerging. When we moved to sqlalchemy (see https://github.com/cal-itp/data-infra/pull/200), this was left in the ExternalTable operator because of the challenge of converting the specification for a nested schema between formats (see comments below).

image

Based on comments from @evansiroky about validation_report not updating after clearing a task, it appears that another argument is needed to tell google python bigquery to "REPLACE OR CREATE" this table. Without this, a user would need to first delete the table before clearing the task.

AFAIK this only affects the validation_report table (but others may also be affected)

evansiroky:
I apologize how bloated this PR has become, but hopefully it is well received at this point. This PR has 4 purposes:

1. Add more static GTFS tables
2. Create dimension and daily fact tables for pathways
3. Properly secures some API secrets used to generate download URLs for GTFS files
4. Makes it possible for local development to occur

### Add more static GTFS tables

This PR adds the relevant data for the static GTFS loader to be able to process the following files:

- attributions.txt
- levels.txt
- pathways.txt
- translations.txt

This closes https://github.com/cal-itp/data-infra/issues/277.

### Create dimension and daily fact tables for pathways

This PR now closes https://github.com/cal-itp/data-infra/issues/615 by creating dimension and fact tables for pathways.

### Properly secures some API secrets used to generate download URLs for GTFS files

The previous way of creating the filled agencies.yml file used a Google sheet open to the public. This PR refactors this to instead obtain a CSV file by creating it from a repository secret in GitHub Actions.

Fixes #247
Fixes #622

### Makes it possible for local development to occur

This PR fixes a number of snags that were hit when trying to do local development to run the GTFS static DAGs in a development environment. This includes:

1. Using a proper filled agencies.yml file
2. Uses the production Google Cloud kubernetes cluster to run validation reports (the output is still written to staging though)
3. Refactor all tasks that use BigQueryOperators to use SqlToWarehouseOperator instead
4. Change stop lat/lon fields in gtfs_schedule_history to be FLOAT types
5. Properly recreates the gtfs_schedule_history validation report table
6. Skips some email failures if in development mode

Fixes #238
Fixes #247
Fixes #641


.github/workflows/push_to_gcloud.yml
```diff
@@ -24,12 +24,20 @@ jobs:
     - name: Install python packages
       run: |
         pip install jinja2==2.11.3 PyYAML==5.4.1 jsonschema==3.2.0 pandas==1.1.4
+    # Create the secrets.csv from the contents of the `AGENCY_SECRETS` environment variable / GitHub
```
machow, line 27: Can we separate this into its own PR? AFAIK the API key spreadsheet is private (done under the rationale that the API keys are read keys, so need to be inaccessible outside the org, but can be widely accessible inside, if it makes filling the spreadsheet out)

This is a big enough change that it probably should be done via a quick sync + an issue

machow (changes requested):
This looks incredible! The one thing I might suggest pushing into a separate PR is the API keys change. I can't do a proper review on that, since I'm not sure the goals / constraints of the API key process (e.g. who should be able to update them, etc..; if @evansiroky / data service developers will own all key updates, then I'm on board with any changes to the process cc @hunterowens).

Let's pair on deploying, since it will take a few steps:

* merge into prod
* clear the latest `gtfs_schedule_history.calitp_included_gtfs_tables`, so it reflects your changes.
* clear the latest task `gtfs_loader.calitp_file_updates_loads` recursively, so it reflects which data feeds need to be loaded each day.
* clear non-recursively **all history** for `gtfs_loader.gtfs_schedule_history_load`. Let's do this together, since as I recall one time I did this and the DAG stopped running (seems like an airflow v1 bug)
* sometime late at night, delete all tables in gtfs_schedule_type2, and clear all of `gtfs_loader.gtfs_schedule_history2.merge_updates`. This will take several hours to complete, as it snapshots GTFS data for each day.

.github/workflows/push_to_gcloud.yml
evansiroky: I have removed this logic and will create another PR for this.

machow (approved):
LGTM thanks

PR #599, opened 2021-10-26
charts/metabase: inject default MB_SITE_URL from ingress

lottspot:
Metabase will use the internal container IP by default to form its
configured site url. This causes issues such as users receiving broken
links in account activation emails. This changeset adds the capability
to the chart to set the site url to whatever hostname is configured in
the ingress configuration.

### DIFF
--- a/kubernetes/apps/charts/metabase/templates/metabase.yaml
+++ b/kubernetes/apps/charts/metabase/templates/metabase.yaml
@@ -88,5 +88,8 @@ metadata:
   name: metabase-config
   namespace: {{ .Release.Namespace }}
 data:
-  {{- toYaml .Values.configs.metabase | nindent 2 }}
+  {{- $defaultHost := get (index (default (list (dict "host" "")) .Values.ingress.hosts) 0) "host" }}
+  {{- $emptyDict := dict }}
+  {{- $siteUrlDict := dict "MB_SITE_URL" (print "https://" $defaultHost) }}
+  {{- ne $defaultHost "" | ternary $siteUrlDict $emptyDict | merge .Values.configs.metabase | toYaml | nindent 2 }}
 {{- end }}
--- a/kubernetes/apps/charts/metabase/templates/metabase.yaml
+++ b/kubernetes/apps/charts/metabase/templates/metabase.yaml
@@ -88,6 +88,8 @@ metadata:
   name: metabase-config
   namespace: {{ .Release.Namespace }}
 data:
+  {{- /* select first hostname from the ingress hosts list (if it exists; empty string otherwise) */}}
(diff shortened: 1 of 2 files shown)


machow:
bump @themightychris do you mind reviewing?

kubernetes/apps/charts/metabase/templates/metabase.yaml
```diff
@@ -88,5 +88,8 @@ metadata:
   name: metabase-config
   namespace: {{ .Release.Namespace }}
 data:
-  {{- toYaml .Values.configs.metabase | nindent 2 }}
+  {{- $defaultHost := get (index (default (list (dict "host" "")) .Values.ingress.hosts) 0) "host" }}
+  {{- $emptyDict := dict }}
+  {{- $siteUrlDict := dict "MB_SITE_URL" (print "https://" $defaultHost) }}
+  {{- ne $defaultHost "" | ternary $siteUrlDict $emptyDict | merge .Values.configs.metabase | toYaml | nindent 2 }}
```
themightychris, line 94: this is...beyond anything I thought was possible in helm templates

**suggestion (non-blocking):** could it make sense for readability to bury some of this complexity under a descriptively named helper?

themightychris (approved):

kubernetes/apps/charts/metabase/templates/metabase.yaml
lottspot: My feeling is that hiding templating underneath abstractions is best reserved for two main cases:

1. When the snippet has a sufficiently negative impact on overall template readability
2. When the snippet is re-used in multiple places

(2) is obviously not the case here, but (1) it sounds like is the case you're alluding to. My thoughts hinge on that one contingent word-- "sufficiently"-- because the trade off to abstracting this snippet is that it becomes a little less explicit, and adds one more place future readers have to look to fully understand what's going on. I don't believe that the readability impact of this snippet is negative enough to justify this trade off. Though my opinion here isn't vehement-- I'm perfectly happy to refactor this as a helper if yours is moreso.

themightychris: @lottspot come to think of it, my suggestion was rooted in the line of thinking that well-factored code doesn't need comments. I agree with your analysis that refectoring here just to put a name on what we're doing is premature though. So--we just need a comment maybe. This is a heavy block that took some hard thinking for me to decode even knowing ahead of time what the goal was so maybe let's just add a quick comment `extract first hostname listed in ingress configuration for as canonical hostname for Metabase to use when generating links`

lottspot: Have to agree re: the non-obvious nature of the snippet. Comment added in `55f8998`

PR #621, opened 2021-10-28
Update overview.md

charlie-costanzo:
Added cell block to docs as example


machow:
closing, since I think we used this to test out docs

Issue #618, opened 2021-10-28
Bug: code errors in docs are not failing the docs build


machow:
cc @Nkdiaz https://docs.calitp.org/data-infra/warehouse/07_tutorial.html#query-examples

![image](https://user-images.githubusercontent.com/2574498/139307054-1d09094b-f5a6-44a2-87b9-f4470cd2ef9d.png)

Nkdiaz:
also fixed by @machow updating the requirements

Issue #614, opened 2021-10-27
Screencast adding entry to agencies.yml


machow:
Users who want to add an entry to agencies.yml are not told how they can do it. We have some instructions on https://docs.calitp.org/data-infra/warehouse/01_agencies.html, but they don't take you end-to-end.

* see https://docs.calitp.org/data-infra/warehouse/01_agencies.html
* show making a new entry
* address that all 4 fields are needed
* if no GTFS RT field, then set the 3 RT fields to null
* a github action will check and report back whether we are able to merge!

machow:
https://docs.calitp.org/data-infra/warehouse/01_agencies.html#walkthrough-of-adding-a-feed

Issue #548, opened 2021-10-22
Screencast manually adding data to GCS bucket for now to docs


machow:
* opening in web browser
* directory structure
* uploading a file for a analysis in data-analyses

machow:
https://docs.calitp.org/data-infra/warehouse/jupyterhub.html#uploading-data

PR #624, opened 2021-10-28
Doc screencasts

machow:
* updates dependency on calitp, so we can use sql magic in the docs (related to https://github.com/cal-itp/data-infra/issues/618; should make current doc build successfully run code)
* adds screencasts for using jupyterhub (#548 ), and adding entries to agencies.yml (#614 )


machow:
Going to merge, since I threw in a fix for running SQL in the docs--but happy to queue up any doc changes

PR #631, opened 2021-10-29
feat:documentation- change config file to fail on sphinx warning

Nkdiaz:
added -warnings is error and --keep going flags so sphinxs warnings will fail in github action. Second commit will be the fix to resolve the errors


machow:
Thanks cc @charlie-costanzo

PR #637, opened 2021-11-02
Update agencies.yml

o-ram:
2 updates to be made to agencies.yml:

- Schedule feed for TRACER (city of Tracy) needs to be added
- The Schedule feed for Turlock transit needs to be updated


machow (approved):
LGTM-- @evansiroky it looks like my most recent edits to the gh action also didn't work as intended :/. This should be okay to merge. If that sounds good to you, do you want to merge this? Tracking other issue here: https://github.com/cal-itp/data-infra/issues/638

PR #648, opened 2021-11-03
Enrich with payments export account and datetime info

mjumbewu:
This will be helpful (and necessary) for deduplicating the `customer_funding_source` data, and is generally useful. It allows us to see the source of each record even downstream of the raw ingested tables.


machow (approved):
LGTM thanks

PR #649, opened 2021-11-03
feat:documentation added section on how to push and pull to github

### LINKED ISSUE #547: Screencast in docs showing how to pull and push to github from jupyterhub
* Getting an authentication token
* Go through pulling data-analyses repo
* Go through pushing to the repo

Nkdiaz:
from jupyterhub using terminal commands, while this is more of a section of analytics, it was suggested to be added to the existing documentation on jupyterhub within the Kubernetes section. Will fold it in here until we do the big documentation reorganization. Chose not to do a screencast because I couldn't find a way to do it without stormy snoring in the background

### DIFF
--- a/docs/kubernetes/JupyterHub.md
+++ b/docs/kubernetes/JupyterHub.md
@@ -187,3 +187,21 @@ Apply these chart changes with:
 ```
 helm upgrade jupyterhub kubernetes/apps/charts/jupyterhub -n jupyterhub
 ```
+
+
+### How to pull and push to github from jupyterhub
+
+#### Working with git-versioned scripts/notebooks in jupyter lab:
+Clone your repository to your home directory on Jupyter Lab
+  - Open Terminal window: Files -> New Launcher -> Terminal
+  - type `git clone <url of your repository>` (get the url from the CALITP Github/GitLab repository page)
+  - In terminal, type cd <name_of_your_repository> tp mavigate into your repository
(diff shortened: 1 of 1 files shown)


machow (changes requested):
From pairing with @Nkdiaz, this is a good start. Once she goes through the process from end-to-end and updates docs, then we can merge. (For example, these docs do not specify generating an authentication token, so a user would not be able to pull and push to github on jupyterhub)

lottspot:
I don't think this belongs embedded in the kubernetes documentation. The kubernetes Jupyterhub document is purposed for an administrative persona who would be responsible for deploying Jupyterhub and supporting the underlying infrastructure rather than a user of Jupyterhub who is logging in and using the application. I would propose a new docs area, e.g., `docs/guides`, as a location for grouping together end-user documentation.

lottspot:
Also just quickly wanted add that this feedback isn't meant as an exercise in pedantry-- I think there are tangible downsides to putting this information here:

1. non-obvious place for those who need it to discover it
2. superfluous content to parse for those who need the existing contents of that doc

machow:
cc @charlie-costanzo let's go over this process Tues, and figure out where it could goes

machow:
Hey--sorry, I think we doubled up efforts on documenting this. This PR is super helpful, and this is more about arbitrarily focusing in on one of these "pushing to github PRs". Let's close this one. Thanks again for the time that you put into this

PR #650, opened 2021-11-03
Retrieve agency secrets from repository secret

### LINKED ISSUE #247: fix running generate_provider_list on dev

### LINKED ISSUE #622: Security vulnerability: some API keys for downloading GTFS files could be accessed without authentication/authorization

evansiroky:
Following-up on the conversation in https://github.com/cal-itp/data-infra/pull/539#discussion_r741987162, I have moved the discussion about using repository secrets to obtain API keys.

Fixes #247
Fixes #622

### DIFF
--- a/.github/workflows/push_to_gcloud.yml
+++ b/.github/workflows/push_to_gcloud.yml
@@ -24,17 +24,26 @@ jobs:
     - name: Install python packages
       run: |
         pip install jinja2==2.11.3 PyYAML==5.4.1 jsonschema==3.2.0 pandas==1.1.4
+    # Create the secrets.csv from the contents of the `AGENCY_SECRETS` environment variable / GitHub
+    # repository secret. The contents of this environment variable should be directly copied from
+    # the contents of a filled-out ./airflow/data/example-secrets.csv file.
+    - name: Create csv file with API keys
+      run: echo -e "$AGENCY_SECRETS" > ./airflow/data/secrets.csv
+      env:
+        AGENCY_SECRETS: ${{ secrets.AGENCY_SECRETS }}
+    # Populate the secrets within the agencies.yml file
     - name: Fill in agencies.yml API keys
       run: |
         python script/agencies_convert.py \
           './airflow/data/agencies.yml' \
           './airflow/data/agencies.filled.yml' \
-          'https://docs.google.com/spreadsheets/d/1Fp7sesSMS6VOIndTLTKcyyN2qHj6najKAaN-I9_3SFo/export?gid=0&format=csv'
+          './airflow/data/secrets.csv'
     - name: Validate agencies.yml
       run: |
         python script/agencies_validate.py \
           './airflow/data/agencies.filled.yml' \
           './airflow/data/schema_agencies_entry.yml'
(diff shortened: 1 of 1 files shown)


hunterowens:
Sorry for being late to this conversation, anyways, here's my 2 :dollar:: 

Thankfully, these are read only credentials (ie, only allow us to read the API) so the risk of exposure is fairly low. Moving to github secrets for now SGTM, as long as @evansiroky is good plugging / managing the process to update change etc. 

coming down the pipe as we add more realtime feeds is many more API keys to manage. Unfortunately for each swiftly client we will need our own key. 

longer term, I suspect we will want some sort of management process (airtable? etc) to track these, but I'm fine w/ moving to GH secrets to keep the 3 keys we have more secure right now, since I suspect the number of API keys we need to mange will grow. But we can research etc.

machow:
Makes sense to me, if @evansiroky is up for owning the keys / key entering process. I think the main challenge with this setup, is that he will need to store the keys somewhere else, or pull the API keys from the injected agencies.yml when it needs to be updated.

We have a jarvus bitwarden account we can store them in, and there are useful options like bitcrypt (cc @themightychris ) if we want to commit them to the repo. If non-technical users will be handling the api keys, then a sheet or airtable is likely a bit more their speed.

themightychris:
GitHub secrets seems like the best place for now. Humans can't easily pull secrets back out of GitHub Secrets, so we like to have secrets backed up in Jarvus' BitWarden/VaultWarden server as well. We have a CalITP team set up already that everyone should have access to. If anyone interested doesn't have an account yet, please DM me on Slack and we'll get you access.

On a related note, for anything that will run under Kubernetes we have [sealed secrets](https://github.com/bitnami-labs/sealed-secrets) implemented on the cluster which allows us to just commit encrypted secret files directly to the public GitHub repository and they get magically decrypted into regular Kubernetes Secret objects after they're pushed to the cluster

machow (approved):
See my above comments!

evansiroky:
I am fine with maintaining these secrets. I have gone ahead and deleted the public Google Sheet and have stored the data elsewhere (see internal chat), so I am going to merge this now.

PR #654, opened 2021-11-04
Ensure no duplicate transaction ids in cleaned micropayment_device_transactions

### LINKED ISSUE #647: Resolve invalid multiplied transaction IDs in micropayment_device_transactions

mjumbewu:
This PR contains the following code. The first item is the meat of the PR, and the later two are new functionality allowing us to validate that the data was produced correctly:

* **Ensure no duplicate transaction ids in cleaned `micropayment_device_transactions`**
  
  This resolves #647

* **Add test for table emptiness**
  
  Creates a new table test. This test is useful for validation of a general query for invalid data. The way to use it is to create a view that selects some invalid records. If that view materializes to an empty table (i.e., there are no invalid records) then the test passes. If the view materializes some non-empty records (i.e., there are invalid records), then the test fails. This is based off of similar concepts like DBT's [bespoke tests](https://docs.getdbt.com/docs/building-a-dbt-project/tests#bespoke-tests)
  
  Since the testing framework expects a list of fields, use the `check_empty` test like:

  ```yaml
  tests:
    check_empty:
      - "*"
  ```

* **Add a check for invalid micropayment_device_transaction rows**
  
  Currently this view materializes to `payments.invalid_cleaned_micropayment_device_transaction`s. We should create a new schema in BigQuery (something like `qa`) so that these validation views don't pollute the `payments` and other production schemas.

### DIFF
--- a/airflow/dags/payments_views_staging/stg_cleaned_micropayment_device_transactions.sql
+++ b/airflow/dags/payments_views_staging/stg_cleaned_micropayment_device_transactions.sql
@@ -9,7 +9,59 @@ tests:
   check_composite_unique:
     - micropayment_id
     - littlepay_transaction_id
+  check_unique:
+    - littlepay_transaction_id
 ---
+
+with
+
+deduped_micropayment_device_transaction_ids as (
+
+    select distinct
+        littlepay_transaction_id,
+        micropayment_id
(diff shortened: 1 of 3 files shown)


machow (approved):
(pairing w/ Mjumbe ) LGTM, thanks

Issue #312, opened 2021-08-25
transaction golden view / refactor mst_ridership


hunterowens:
**Is your feature request related to a problem? Please describe.**
To easily visualize data for the payments dataset, we need a golden / tidy view. 

**Describe the solution you'd like**
Would like a dataset that represents each full "ride" on the system, created by joining the tables

This should have

1) Tap on / location data 
2) Tap off  data (if applicable - some riders don't tap off and there
3) customer funding source data (card scheme, form factor, etc) 
4) fare type (penalty fare, full fare, sub fare) 
5) capping info (30 day cap, 7 day cap, etc)

mjumbewu:
I've started putting together a query for this (it's still in the exploration phase): https://console.cloud.google.com/bigquery?sq=1005246706141:879b8f7e5de944918d04f9f38764d4a8

There are a number of questions:
* There's only one duplicate transaction/micropayment pair in the `micropayment_device_transactions` table. It would be best if there were _no_ duplicates, but are we ok to just `select distinct`?
* In `device_transactions`, `type` is always `'single'`. Why is that? No multi-zone trip data here? There do appear to be multi-zone trips, according to the trip prices.
* In `device_transactions`, `transaction_outcome` is always `'allow'`. Maybe that's good. Is it normal?
* `'Route Z'` is an unknown route? I don't understand the cases where this will be used, but for now I'm filtering it out.
* We can have micropayments with the same id and different transaction times, and I don't understand why. For now I'm excluding, so that `select distinct` works.

mjumbewu:
There are a few relevant updates that have come via email. Copying here for tracking:

* The replying engineer (Terry) did not see the `micropayment_device_transactions` duplicate in the source data so it may be an issue with the extraction process. They've pulled in the data team to help on this.
* Regarding `'single'` trip types: "As MST are using our Back Office to send the amount for Littlepay to charge for each trip it is considered a single."
* Regarding duplicate `transaction_ids` across partitions: "Route Z is the default value that is being sent by the device. Initially the device sends the default value and we are expecting this to be updated via the BackOffice API. (This need to be enriched by the fare engine via the Back Office). We can see that there are instances where this data has failed to be sent to us."
[...]

PR #655, opened 2021-11-04
Add a payments_rides view

mjumbewu:
This view is in service of #312. It consolidates information about each ride taken. Variable fare rides have information about the first tap as well as the second tap in separate fields prefixed with `off_`, whereas rides with a single tap simple have `null` values in all of the `off_` fields.


machow (approved):
LGTM--I think while we tune making sure we have staging datasets in metabase let's merge and then go over some of the key metabase questions / visualizations that might come from this!

PR #665, opened 2021-11-05
feat: create user story template

holly-g:


machow (approved):
This is great--thanks! let's definitely showcase in the monday all hands!

Issue #666, opened 2021-11-05
Bug: in gtfs_views DAG, gtfs_schedule_fact_route_id_changes and gtfs_schedule_fact_stop_id_changes tasks are ran prematurely


evansiroky:
**Describe the bug**

In the `gtfs_views` DAG, the `gtfs_schedule_fact_route_id_changes` and `gtfs_schedule_fact_stop_id_changes` tasks should have added dependencies. Both of them seem to read data from the `gtfs_schedule_dim_feeds` and then the route one reads data from `gtfs_schedule_dim_routes` and the stop one reads data from `gtfs_schedule_dim_stops`. 

The only dependency of these tasks is currently on `dim_metric_date` which only depends on `dim_date` which doesn't depend on an external DAG. Therefore, these get ran without any dependency on an external DAG. Note the start datetime last night of 00:23. The start datetime of gtfs_schedule_dim_feeds is at 01:11. So it's definitely running before the gtfs_schedule_dim_feeds task/table is ran/updated.

**To Reproduce**

See screenshot:

<img width="678" alt="Screen Shot 2021-11-05 at 1 19 52 PM" src="https://user-images.githubusercontent.com/3112493/140574392-85b0922e-944e-488b-82d1-adcc0d823c15.png">

**Expected behavior**

These tasks should have added dependencies so that they wait until the daily data is available.

PR #667, opened 2021-11-05
Add additional dependencys for tasks in gtfs_views

evansiroky:
Fixes https://github.com/cal-itp/data-infra/issues/666 by adding appropriate task dependencies so that daily updates are included in these tables.


machow (approved):
Thanks!

PR #672, opened 2021-11-05
fix(docs): added better placeholder text for the docs landing page

charlie-costanzo:


charlie-costanzo:
quick change to documentation landing page content

PR #673, opened 2021-11-07
Add adjustment and product information to payments_rides

mjumbewu:
This would allow us to add to the [payments dashboard](https://metabase.k8s.calitp.jarv.us/dashboard/29) metrics such as number of riders hitting fare caps (#372).

Includes the following validations which should be added to the [payments validation dashboard](https://metabase.k8s.calitp.jarv.us/dashboard/28):
* **Ride adjustments:** Make sure that all rides where the amount paid is not equal to the nominal amount have adjustments, and rides where the amount paid is equal to the nominal amount do not have adjustments.
* **Ride counts:** Make sure that we did not lose any transactions while joining for the rides table.
* **Paired device transaction (tap on/off) common fields:** Ensure that fields that are expected to be consistent across tap on/off transactions are actually consistent. Invalid records are any that differ on the `device_id`, `device_id_issuer`, `route_id`, `mode`, `direction`, or `vehicle_id` fields.
  
  _NOTE: I'm not sure that `device_id` is appropriate in this set, as you could have a bus with a device in the front and back, or a train system where you tap on/off at the station rather than on the vehicle. Would those be different devices?_
* **Applied micropayment adjustments:** Ensure that there is only one micropayment_adjustments record with `applied` set to `True` for each micropayment.

### DIFF
--- a/airflow/dags/payments_views/payments_rides.sql
+++ b/airflow/dags/payments_views/payments_rides.sql
@@ -22,6 +22,12 @@ gtfs_routes_with_participant AS (
     JOIN `views.payments_feeds` USING (calitp_itp_id, calitp_url_number)
 ),
 
+applied_adjustments AS (
+    SELECT participant_id, micropayment_id, product_id, adjustment_id, type, time_period_type, description, amount
+    FROM `payments.stg_cleaned_micropayment_adjustments`
+    WHERE applied IS True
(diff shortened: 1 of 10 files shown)


machow (approved):
From pairing this morning--LGTM, thanks!

Issue #677, opened 2021-11-08
Document new agencies.yml API_KEYS workflow


machow:
See this PR https://github.com/cal-itp/data-infra/pull/650. I thought we might have accidentally nuked the instructions when restructuring docs, but it looks like we just need to update them quickly. @evansiroky do you mind adding to https://github.com/cal-itp/data-infra/blob/main/docs/warehouse/02_agencies.md?

machow:
cc @charlie-costanzo

PR #697, opened 2021-11-10
Update agency secret workflow docs

evansiroky:
Resolves https://github.com/cal-itp/data-infra/issues/677.

This PR depends on #701.

This previously failed on the publish-docs workflow action, but is now fixed. [See updated publish-docs workflow run here](https://github.com/cal-itp/data-infra/runs/4172811587?check_suite_focus=true).


charlie-costanzo (approved):
Thanks again for taking care of this sphinx fix, and updating the docs!

PR #701, opened 2021-11-11
Pin pydata-sphinx-theme in publish-docs GitHub Action Workflow

evansiroky:
The previous version was having some problems with Sphinx 4.3.0 (see CI run here). The package depending on pydata-sphinx-theme hasn't updated their requirement of pydata-sphinx-theme to be above 0.6.x, so this pins the version to 0.7.2 (one that is compatible with Sphinx 4.3.0).


charlie-costanzo:
Thanks for slotting this in @evansiroky!

PR #710, opened 2021-11-12
fix(docs): pulled explainer sentance to below admonition, experimente…

charlie-costanzo:
…d with vertical links between urls and docs


charlie-costanzo:
@machow this PR is just quick tweaks to the technical onboarding page:
* Pulled out a sentence from the top admonition box and placed below to make more concise
* Added vertical bars between the url and the internal docs link (let me know if you think this adds value)

machow (approved):

Issue #695, opened 2021-11-10
Load GTFS Schedule parse failures, and do not attempt to merge corresponding feeds


machow:
See #693 

* do not error on unparseable results in task: `gtfs_loader.gtfs_schedule_history_load`
* instead keep a table with a row per table and whether it succeeded (e.g. upload a file with failures to /schedule/processed)
* only attempt to merge feeds that were successfully extracted and where are files are parseable

machow:
From pairing w/ @evansiroky -- it looks like the merge should be robust. However, if we have preprocessed other pieces of a feed when it fails, we'll need to delete them (otherwise, we could half update the data for a feed).

Parsing / pre-processing happens here:

https://github.com/cal-itp/data-infra/blob/main/airflow/dags/gtfs_loader/gtfs_schedule_history_load.py#L90

machow:
I think the above PR completed this issue!

Issue #694, opened 2021-11-10
Load each new (changed) validation report, even if the feed hasn't changed


machow:
Currently, we only load validation reports when something about the feed has changed. This makes sense if the validation report is deterministic, but has 2 limitations:

* If we upgrade the validator, the new results won't necessarily load right away (see #685)
* If we need to bail out from loading a feed, it would be nice to still have validator results (see #693)

We should be able to resolve the above 2 limitations by the following: remove the processing of validation results from the task `gtfs_loader.gtfs_schedule_history_load`, so that it loads any time a validation report changes (even if gtfs_schedule_history_load fails)

machow:
See this code in gtfs_schedule_history_load, where the validator is copied into the processed folder. We should pull this into its own task...

https://github.com/cal-itp/data-infra/blob/main/airflow/dags/gtfs_loader/gtfs_schedule_history_load.py#L66-L71

machow:
should be resolved by the creation of this task: https://github.com/cal-itp/data-infra/blob/main/airflow/dags/gtfs_loader/gtfs_validation_history_load.py

PR #716, opened 2021-11-16
Tolerate corrupted feeds

evansiroky:
This PR works on the items mentioned in #694 and #695 by doing the following:

- Extracting the validation loading (copying to bucket acting as external table) out of gtfs_schedule_history_load
- Tolerating and tracking CSV parse errors that occur when loading a CSV file.

This PR hasn't yet percolated the parse error information into dim and fact tables (as recommended in #695) as it seeks to first resolve the matter of a DAG task failing completely due to a CSV parse error.

### DIFF
--- a/airflow/dags/gtfs_loader/gtfs_schedule_history_load.py
+++ b/airflow/dags/gtfs_loader/gtfs_schedule_history_load.py
@@ -7,7 +7,7 @@
 
 import pandas as pd
 
-from calitp import get_table
+from calitp import get_table, save_to_gcfs
 from calitp.config import get_bucket
 from calitp.storage import get_fs
 
@@ -48,7 +48,14 @@ def main(execution_date, ti, **kwargs):
     print(f"Number of feeds being loaded: {df_latest_updates.shape[0]}")
 
     ttl_feeds_copied = 0
+    feed_tables_process_results = []
+    feed_process_resuls = []
     for k, row in df_latest_updates.iterrows():
+        # initialize variable to track whether a parsing error occurred and which tables
+        # were loaded so far
+        parse_error_encountered_in_this_feed = False
+        id_and_url = f"{row['itp_id']}_{row['url_number']}"
+
         # process and copy over tables into external table folder ----
         for table_file, is_required, colnames in table_details:
             # validation report handled in a separate task, since it is in a subfolder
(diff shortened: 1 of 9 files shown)


airflow/dags/gtfs_loader/gtfs_schedule_history_load.py
```diff
@@ -45,57 +40,88 @@ def main(execution_date, ti, **kwargs):
...
 
         print("total feeds copied:", ttl_feeds_copied)
+
+    # save feed and feed table process results
+    save_to_gcfs(
+        pd.DataFrame(feed_process_resuls).to_csv(index=False).encode(),
+        f"schedule/{execution_date}/processed/feed_parse_result.csv",
+        use_pipe=True,
+    )
+    save_to_gcfs(
```
machow, line 123: Can we load this data as an external table? Should only require adding this to the gtfs_schedule_history DAG:

```yaml
operator: operators.ExternalTable
source_objects:
  - 'schedule/*/processed/feed_parse_result.csv'
destination_project_dataset_table: "gtfs_schedule_history.calitp_feed_parse_result"
skip_leading_rows: 1
schema_fields:
  - name: calitp_itp_id
    type: INTEGER
  - name: calitp_url_number
    type: INTEGER
  - name: calitp_extracted_at
    type: DATE
  - name: parse_error_encountered
    type: BOOLEAN
```

(We can definitely just chuck this into a issue if you want to merge)

machow (changes requested):
This looks great, thanks! The refactoring repeated code into constants and utils looks super useful. The one piece I might suggest changing is the alternate behavior in _keep_columns that returns a boolean. It may be helpful to have _keep_columns fail explicitly, so that outer code (which is already changing the behavior based on it's boolean return value) uses a try / except instead. This way, if someone fails to account for parse errors, it won't quietly succeed but create an empty file (which could be surprising for a function that is supposed to keep the original data, and enhance it)

airflow/dags/gtfs_loader/gtfs_schedule_history_load.py
evansiroky: Done in 4ae1221

airflow/dags/gtfs_loader/gtfs_schedule_history_load.py
```diff
@@ -45,57 +40,88 @@ def main(execution_date, ti, **kwargs):
...
+                ["schedule", "processed", f"{date_string}_{id_and_url}", table_file]
+            )
 
             print(f"Copying from {src_path} to {dst_path}")
 
             if not is_required and not fs.exists(f"{bucket}/{src_path}"):
                 print(f"Skipping missing optional file: {src_path}")
             else:
-                _keep_columns(
+                parse_error_encountered = _keep_columns(
```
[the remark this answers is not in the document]
evansiroky: Done in d3cec96

machow (approved):
LGTM

PR #717, opened 2021-11-16
docs: GitHub, storing data / GCS, data catalogs

machow:


charlie-costanzo:
@tiffanychu90 cool! and then if it's going to live on the page that already exists we won't need to adjust the table of contents

charlie-costanzo:
hey @tiffanychu90 are you still having issues with the markdown linter?

tiffanychu90:
@charlie-costanzo: looks like it. This time, the linter ran through, and I was able to re-stage the docs and commit, but looks like it still broke within the workflow run

charlie-costanzo:
@machow would you be able to take a look at the [errors in this log](https://github.com/cal-itp/data-infra/runs/4240869573?check_suite_focus=true#step:4:157)?

charlie-costanzo:
okay @tiffanychu90 I took a deeper look at the errors that were coming up in the workflows and tested some things out and I have some thoughts. It looks like they're not related to the linter anymore, but MYsT errors in the markdown.

### In `/08_data_catalogs.md`
It looks like there's formatting issues with the markdown for the code blocks.

* **Question** How many of the code examples should actually execute?

Right now all of them are formatted to execute, which is breaking things when they can't complete.

For those that don't need to execute, we can format them like this to add the code block and highlight syntax:

````md
```yaml
example code here
```
````
(the 'yaml' after the first backticks adds yml syntax highlighting to the block, but this could also be python etc..., leaving out a language just doesn't add syntax)

  * Here's some extra info on the differences
    * [syntax highlighting (non-executing) code block](https://jupyterbook.org/reference/cheatsheet.html#code-and-syntax-highlighting)
    * [executing code block](https://jupyterbook.org/reference/cheatsheet.html#executable-code)

### In `sample-catalog.yml`
Thinking through how to present this doc - the best way might be to make two documents out of it based on markdown and the requirements to build the docs:
  * `sample-catalog.yml`
  * `sample-catalog-viewer.md`
 
We can keep `sample-catalog.yml` as-is

In `sample-catalog-viewer.md` , we just add:
````md
(sample-catalog)=
# Sample Catalog
```{literalinclude} sample-catalog.yml
```
````
* This pulls in everything from that document and adds yaml syntax highlighting. It also adds a header to the page to meet the toctree requirements. 
* (Note also that `(sample-catalog)=` creates an internal link. I would change [the link here](https://github.com/cal-itp/data-infra/blob/ff658ee22667c45bfc32733435ab08d53deb4ee6/docs/analytics_tools/08_data_catalogs.md?plain=1#L37) to just be `(sample-catalog)`
* Then we just need to change the table of contents to point like this:

`- file: analytics_tools/sample-catalog-viewer`

### One more thing In `05_python_libraries.md`
I thin the executing Python code under 'shared utils' section is also causing the build to break. If we just instead put that in:

````md
```python
code
```
````

I think it should fix things.

Anyways, I just wanted to jot this down - if we do everything here the docs should build. Happy to get on a call to walk through it!

**Comment by tiffanychu90** [comment]
@charlie-costanzo: I'm running into issues with the internal link (sample-catalog). 

Within the Markdown file:

This hyperlink [ sample-catalog ] (sample-catalog) generates this url: https://github.com/cal-itp/data-infra/blob/analyst-docs/docs/analytics_tools/sample-catalog. 

Shouldn't the url be https://github.com/cal-itp/data-infra/blob/analyst-docs/docs/analytics_tools/sample-catalog.yml?

**Comment by charlie-costanzo** [comment]
Hmm @tiffanychu90 that syntax should work as long as the anchor tag is in `sample-catalog-viewer.md` like the picture below. It’s set up that way in other areas of the docs.
3A24939E-A7BE-4F74-AE9D-60474CF11418

**Comment by tiffanychu90** [comment]
@charlie-costanzo: so the [sample-catalog-viewer.md](https://github.com/cal-itp/data-infra/blob/analyst-docs/docs/analytics_tools/sample-catalog-viewer.md) does pull that, but the [docs page](https://github.com/cal-itp/data-infra/blob/analyst-docs/docs/analytics_tools/08_data_catalogs.md) pulls the wrong link and goes to a broken URL

**Comment by charlie-costanzo** [comment]
> Now that I'm thinking about it, I wonder if it just doesn't render in GitHub, but in the Jupyter Book, it's actually linking correctly?

Ah yes - I  think that's it. Once we get it to build it should link correctly.

**Comment by tiffanychu90** [comment]
@charlie-costanzo: I think this can be merged in. I think the toctree will resolve itself once merged? It hasn't recognized `analytics_tools/08_data_catalogs` yet.

**Comment by charlie-costanzo** [comment]
@tiffanychu90 nice! I think the only thing that's currently breaking your build might be [this executing python code block here](https://github.com/cal-itp/data-infra/blob/566f4b3a6425e122480918bf134de452f0853ea4/docs/analytics_tools/08_data_catalogs.md?plain=1#L75). If we just open it with
````md
```python
````
I think the workflow will build the book and we can take a look at it quick before merging if that works for you!

**Comment by charlie-costanzo** [comment]
Hey @tiffanychu90 - this is so great!

I just thought of a couple potential tweaks related to structure - feel free to push back on any of them as I really value your experience working on docs!

1.  Now that I think about it, in the interest of limiting the number of pages in the docs, how would you feel about removing the additional `sample-catalog-viewer.md` file and just appending this code inside to the bottom of `08_data_catalogs`?
````md
```{literalinclude} sample-catalog.yml
```
````
Then we could just add a link to it from the table of contents at the top of the page, and internally link to that section where needed. We'd then just need to adjust the table of contents for the book.

2. Changing the [h1/page title](https://github.com/cal-itp/data-infra/blob/be37c21bd1f24baaa07a4cc703140f75fc740bb8/docs/analytics_tools/08_data_catalogs.md?plain=1#L16) of `08_data_catalogs` to `# Using Data Catalogs`
  - It just might fit with the naming of the other section better

Let me know what you think, otherwise after this we should be good to merge!

tiffanychu90:
@charlie-costanzo: I removed the `sample-catalog-viewer.md`, I think it is simpler without it. But, instead of the `{literalinclude}`, I just link to the `sample-catalog.yml` file as the reference. Since there are snippets of the relevant catalog items in open data portal, GCS sections, it seems duplicative to include the entirety of  `sample-catalog.yml` in as another section within that page. 

The benefit of having `sample-catalog.yml` checked in would be so other analysts can cut & paste and set up their own catalog correctly. The docs would then walk them through relevant catalog snippets they're interested in. 

Does that work / logically make sense for an analyst who's reading the docs with fresh eyes?

charlie-costanzo:
> **follow-up**: was the original reason for the `sample-catalog-viewer.md` because the `_toc.yml` can only take markdown files and not yaml files?

Ah yes - it was a workaround because the toctree breaks when we try to include a yaml file, it appears to only allow .md or .py. I've been trying to find docs that provide info on including other files but I haven't been able to find anything concrete yet. [This actually might be a potential solution...](https://jupyterbook.org/structure/configure.html?highlight=table%20contents#exclude-pages-from-your-build)

tiffanychu90:
@charlie-costanzo: Excluding one of the files doesn't seem to be working either: https://github.com/cal-itp/data-infra/blob/analyst-docs/docs/_config.yml#L13-L15

charlie-costanzo:
> Excluding one of the files doesn't seem to be working either: https://github.com/cal-itp/data-infra/blob/analyst-docs/docs/_config.yml#L13-L15

Dang okay, thank you for trying that @tiffanychu90! 

What do you think about re-adding `sample-catalog-viewer.md` as another page, as things were before, and getting it to build for now?

Then I'll merge, but also look into:
* Potentially storing `sample-catalog.yml` in a 'references' type section
* I can also maybe ask Natalya if she came across anything related to this as she was setting things up

tiffanychu90:
Cool. I left the `exclude_patterns` in `_config.yml`, as working examples of what we might want to explore how to get that successfully excluding. Got the coveted green check mark!

charlie-costanzo (approved):
heck yeah @tiffanychu90! We got the coveted green check mark and it looks great. Thanks for working through this with me! Merging now

PR #722, opened 2021-11-16
feat:documentation updated jupyterbook documentation page of sql snippets to use the sql magic 

Nkdiaz:
worth mentioning that the script executes and the result is displayed in jupyterbook but the query is no longer displayed. Possibly another bug with how sql magic is returning the input.


charlie-costanzo (approved):
Awesome, thank you so much for getting this in there!

PR #747, opened 2021-11-17
feat:documentation added tag for sql output to be quieted

Nkdiaz:
added requested '%%sql -m -o * --quiet' tag so the output is not displayed


charlie-costanzo (approved):
Heck yeah, thanks for taking care of this!!

PR #752, opened 2021-11-18
Fix SLORTA GTFS URL

evansiroky:
The previous URL has been having download errors for multiple days. This new URL should be the correct one.


machow (approved):
thanks!

hunterowens:
:+1:

Issue #617, opened 2021-10-28
Investigate customer_funding_source cleaning options


mjumbewu:
We have ways for cleaning `micropayments` and `device_transactions`. Based on information we've recently received from Littlepay, we may be able to clean `customer_funding_sources` as well.

A few things to note:
* Each `customer_funding_source` file is a full extract of the `customer_funding_source`,  not an increment, so there is duplication across files.
* `customer_id`s may become inactive and be replaced. In these cases, both `customer_id` values will be present in the data export, but the active one will be set in the `principal_customer_id`.

Maybe the best approach is to use the most recent file only and match with other tables on either the `principal_customer_id` or `funding_source_vault_id`.

mjumbewu:
We have an outstanding question for Littlepay here, about what we see in the data. The gist of the question is:

> Hi Terry,
>
> We've been looking at this issue on-and-off for a while. I think what you describe with the vault IDs covers most of the records in the micropayments table, however there are still quite a few (around 2%) of the micropayment records where no vault ID/customer ID pair exists in the customer funding source table. For example, for the vault_id `7053c3bd-ebd4-42d7-aa1a-ab55836839df` is paired with the customer_id `0de38f1b-d82a-4c6b-9a2e-ddba50dbfbc6` in the micropayments table, but with a completely different ID in the customer_funding_source table:
>
> 
> ```sql
> select distinct funding_source_vault_id, customer_id
> from payments.micropayments
> where funding_source_vault_id = '7053c3bd-ebd4-42d7-aa1a-ab55836839df'
> ```
> ![image](https://user-images.githubusercontent.com/146749/145098801-1779578e-13d1-4317-b947-c7d45acb2a23.png)
> 
> ```sql
> select distinct funding_source_vault_id, customer_id, principal_customer_id
> from payments.customer_funding_source 
> where funding_source_vault_id = '7053c3bd-ebd4-42d7-aa1a-ab55836839df'
> ```
> ![image](https://user-images.githubusercontent.com/146749/145098855-e112f5ea-3ccf-4ff6-a0d6-73e69c553d45.png)
> 
> **Should we just disregard the customer ID in the micropayments table, and rely solely on the vault ID?** Is that a safe thing to do?
> 
> In case it's relevant information, the micropayment customer ID above (`0de38f1b-d82a-4c6b-9a2e-ddba50dbfbc6`) does show up in the customer_funding_source table, just for a completely different vault ID:
> 
> ```sql
[...]

PR #755, opened 2021-11-18
fix(payments): Escape regex parameters in formatted string

mjumbewu:
It took me a while to figure out why the `calitp_export_account` and `calitp_export_datetime` fields were empty. Then I realized we were using curly braces inside of the regex, and the regex string was inside of a format string.

This issue is relevant to #617, as we need to know the export file for resolving the funding source IDs.


machow (approved):
Ah, nice catch! It took a second to realize at which level of the template sandwich things needed to be escaped 😬

PR #757, opened 2021-11-19
feat: load GTFS RT validations for a single agency (126)

machow:
This PR adds validation results--from a single agency (126)--for trip updates, service alerts, and vehicle positions. It's meant to get some form of daily validation out into the open, so we can work on tasks related to efficient storage, transformation, and early debugging.

It uses a kubernetes pod operator. The pod's docker image comes from https://github.com/cal-itp/gtfs-rt-api/pull/3, which publishes images every release using gh actions.

TODO:

- [x] : review and merge the gtfs-rt-api PR
- [x] : add realtime DAG documentation as new page in https://docs.calitp.org/data-infra/datasets_and_tables/01_overview.html
  - descriptions of tables
  - dag maintenance
- [x] : stub out issues for next steps on validation (e.g. what happens if a gtfs schedule fails to download? data storage, etc..)


docs/datasets_and_tables/gtfs_rt.md
```diff
@@ -0,0 +1,62 @@
+# GTFS Realtime
+
+Currently, Payments data is hosted by Littlepay, who exposes the "Littlepay Data Model" as a set of files stored in an S3 bucket. To get a copy of the data docs, email hunter.
```
evansiroky, line 3: Please update from copied payments docs.

docs/datasets_and_tables/gtfs_rt.md
```diff
@@ -0,0 +1,66 @@
...
+
+| Tablename | Description | Notes |
+| --------- | ----------- | ----- |
+| `calitp_files` | Metadata on each RT feed file we sample every 20 seconds (e.g. vehicle positions) | |
+| `vehicle_positions` | One row per feed individual vehicle position reported every 20 seconds | Sampling times occur in 20 second intervals, but not on specific points in time. |
+| `validation_service_alerts` | Each row is the contents of an individual results file from the GTFS RT validator | |
+| `validation_trip_updates` | Similar to above, but for trip updates. | |
+| `validation_vehicle_positions` | Similar to above, but for vehicle positions. | |
+
+## Maintenance
```
evansiroky, line 33: Everything from this header to the end seems to be more about airflow stuff. I recently made a new `airflow` folder in the docs that I would like to advocate to be the place for all airflow-related items. The information about datasets is still helpful to include here, but the stuff about airflow maintenance is a little less relevant to people not doing stuff with airflow.

So, maybe this section could be in a new file at `airflow/realtime-data-pipeline.md`.

machow: Ah--I think the payments dataset docs also have maintenance sections that discuss airflow. Is it okay if we merge these in, and create an issue for moving all of this info out (for all datasets; the main other one is payments)?

I can go through and 

* copy out / remove maintenance sections across all files in datasets into the airflow section.
* replace them in each dataset with links to the airflow maintenance docs

machow: Added an issue here, since it stacks well with some other docs changes requested: https://github.com/cal-itp/data-infra/issues/783

airflow/dags/rt_loader/external_validation_service_alerts.sql
```diff
@@ -0,0 +1,38 @@
+---
```
evansiroky, line 1: This seems similar to the DAG tasks in `gtfs_schedule_history` in that it is creating an external table. Should we create a new DAG for creating these external RT tables where the DAG schedule run just once?

airflow/dags/rt_loader/load_rt_validations.yml
```diff
@@ -0,0 +1,23 @@
...
+cmds:
+  - gtfs-rt-validator-api
+
+arguments:
+  - validate-gcs-bucket
+  - cal-itp-data-infra
+  - cloud
+  - "gs://calitp-py-ci/gtfs-rt-validator-api/gtfs_schedule_126"
+  - "--gtfs-rt-glob-path"
+  - "gs://calitp-py-ci/gtfs-rt-validator-api/gtfs_rt_126/2021*/126/0/*"
```
evansiroky, line 13: Can you add a comment or some kind of description noting that this is for just one agency?

docs/datasets_and_tables/gtfs_rt.md
evansiroky: Sounds good to do as part of another PR.

evansiroky (changes requested):
See comment with a recommendation to create a new DAG that has one-time executions of tasks. Also, is the plan with these RT PRs to just do the work for one agency as a proof-of-concept? If not, then we shouldn't hardcode the analysis of just one agency.

docs/datasets_and_tables/gtfs_rt.md
machow: ah sorry--should be updated now

airflow/dags/rt_loader/external_validation_service_alerts.sql
machow: I regret making `gtfs_schedule_history` its own DAG (or at the very least having it run once a day) for 2 reasons:

* when changes are made to the external table definition, they are not automatically picked up the next day (because the DAG only runs once). This has been a common source of errors after merging PRs :/. In general, what happens is the tasks depending on the table fail and then you have to go in and clear the once-off DAG tasks
* it seems helpful to keep the table creation close to the loading script. It's quick to see everything together. 

The payments_loader DAG follows a similar setup, where external tables are included in the loader DAG, just to keep everything together: 

image

An alternate way to do it is to have a separate DAG for external tables (e.g. a rt_setup DAG) and have it set to `latest_only`. This it will run daily, but will skip if you clear it for days in the past.

machow:
I can tee up the issues that describe how to implement robustly for all agencies before merging. A nice thing of this PR is it will let people explore the data, let the process run a bit to see if anything needs tweaking / hardening

airflow/dags/rt_loader/external_validation_service_alerts.sql
evansiroky: ok, yeah, I agree with what has been in the PR.

airflow/dags/rt_loader/load_rt_validations.yml
machow: done!

machow:
Can resolve doc conflict prior to merging (should be quick)

machow:
Issues stubbed out:

* https://github.com/cal-itp/data-infra/issues/788
* https://github.com/cal-itp/data-infra/issues/787
* https://github.com/cal-itp/data-infra/issues/786

evansiroky (approved):

Issue #728, opened 2021-11-17
Docs page: Analytics Tools / Tools Quick Links


charlie-costanzo:
**Issue:** 
‘Tools Quick Links’ seeks to anticipate the user’s needs in their first 2-3 weeks, when they are still committing to memory the tools we use and how to access them. 

When starting on a new team the amount of information you receive can be overwhelming and it can be hard to keep track of it all. This section is meant to be a place that is easy to access and read to refer back to when needing a reminder, as simply as possible.

This section should also include a ‘still need access to a tool?’ admonition in case a user is still experiencing access trouble. 

This section is probably in a good place now, we can slot it in for review and approval.

**Format:** 
‘Lost a link? Find quick access to our tools below.’

Then a short, concise list of links to our major analytics tools **(SHOULD THIS BE A TABLE?),** with an ‘still need access to a tool on this page’ admonition at the bottom. 

**_Misc:_**

PR #760, opened 2021-11-19
fix(docs): easy doc fixes 

charlie-costanzo:
This PR accomplishes some easy fixes in the docs:
- Removes `analytics_tools/02_querying` as well as references to bigquery in favor of adding querying instructions to `analytics_tools/04_notebooks`
- Puts the body of `tools_quick_links` into a table (also removed the bigquery reference), which makes issue #728 able to close out
- Reformats `technical_onboarding/overview` to make more sense with bigquery gone
- fixes TOC to reflect new changes
- moves `sample-catalog-viewer` to `analytics_examples` section


charlie-costanzo:
cc @tiffanychu90 in this PR I'm moving the `sample-catalog-viewer` to live in the `analytics_examples` section, I think that makes sense

machow (approved):
Thanks!

PR #772, opened 2021-11-22
fix(reports): reports index was filtering those without feed info

machow:
cc @holly-g . This PR fixes a minor issue in reports_gtfs_schedule_index. This table should record one row per agency gtfs feed that *could* be reported on in reports.calitp.org.

In addition it has the following columns:

* has_feed_info - does the feed have a feed_info.txt for the last day of the month being reported?
* use_for_report - should it be used to generate a report? This is true when a feed has url number = 0 and has feed info. (note that reports currently break if an agency does not have feed_info.txt, https://github.com/cal-itp/reports/issues/26#issuecomment-883630244)

Note that use_for_reports is referenced in this line of the reports repo: https://github.com/cal-itp/reports/blob/main/reports/1_generate_ids.py#L7

The issue is that this table was accidentally filtering out rows where not has_feed_info (by accidentally using a JOIN, rather than a LEFT JOIN). This means that reports were being generated as intended, but internal users could not tell why a potential report was marked as "should not generate".

### DIFF
--- a/airflow/dags/gtfs_views/reports_gtfs_schedule_index.sql
+++ b/airflow/dags/gtfs_views/reports_gtfs_schedule_index.sql
@@ -62,9 +62,13 @@ agency_feeds_on_end_date AS (
         , S.itp_id AS calitp_itp_id
         , S.url_number AS calitp_url_number
         , S.agency_name
+        , COALESCE(FI.has_feed_info, FALSE)
+            AS has_feed_info
     FROM `gtfs_schedule_history.calitp_status` S
     JOIN publish_dates_crnt PD ON
         S.calitp_extracted_at = PD.date_end
+    LEFT JOIN has_feed_info_end_date FI
+      USING(calitp_itp_id, calitp_url_number, publish_date)      
 )
 
 SELECT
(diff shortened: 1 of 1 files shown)


machow:
Additionally, let's delete this download dataset button from reports: https://github.com/cal-itp/reports/blob/main/templates/report.html.jinja#L42

airflow/dags/gtfs_views/reports_gtfs_schedule_index.sql
```diff
@@ -62,9 +62,15 @@ agency_feeds_on_end_date AS (
         , S.itp_id AS calitp_itp_id
         , S.url_number AS calitp_url_number
         , S.agency_name
+        , COALESCE(FI.has_feed_info, FALSE)
```
evansiroky, line 65: Maybe I'm not reading this right, but it seems like even if there isn't any feed info file, this still coalesces null values into false. So, I'm not sure how this would activate reports for those feeds that don't have a feed info file. Let me know if I'm missing something.

evansiroky: @machow clarified that this is expected behavior.

evansiroky (approved):

PR #777, opened 2021-11-23
Add parse error stats to feed and files fact tables

evansiroky:
This PR adds info about feed parse errors to daily fact tables about feeds and feed files respectively.

### DIFF
--- a/airflow/dags/gtfs_views/gtfs_schedule_fact_daily_feed_files.sql
+++ b/airflow/dags/gtfs_views/gtfs_schedule_fact_daily_feed_files.sql
@@ -45,9 +45,7 @@ raw_daily_files AS (
 
     FROM `gtfs_schedule_history.calitp_files_updates` T1
     JOIN `views.gtfs_schedule_dim_feeds` T2
-        ON
-            T1.calitp_itp_id = T2.calitp_itp_id
-            AND T1.calitp_url_number = T2.calitp_url_number
+        USING(calitp_itp_id, calitp_url_number)
     LEFT JOIN `gtfs_schedule_history.calitp_feed_tables_parse_result` T3
         ON
             T1.calitp_itp_id = T3.calitp_itp_id
--- a/airflow/dags/gtfs_views/gtfs_schedule_fact_daily_feed_files.sql
+++ b/airflow/dags/gtfs_views/gtfs_schedule_fact_daily_feed_files.sql
@@ -35,16 +35,25 @@ raw_daily_files AS (
         , T1.is_agency_changed
         , T1.full_path
         , FALSE AS is_interpolated
+        , T3.parse_error_encountered
 
         -- calculate the leading date, so we can fill in missing rows, where
         -- extraction failed to run.
         , LEAD(T1.calitp_extracted_at)
-            OVER (PARTITION BY calitp_itp_id, calitp_url_number ORDER BY T1.calitp_extracted_at)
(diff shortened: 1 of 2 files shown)


airflow/dags/gtfs_views/gtfs_schedule_fact_daily_feeds.sql
```diff
@@ -50,17 +50,23 @@ daily_feeds AS (
...
+        daily_feeds.*,
+        CASE WHEN download_status.status = "success" THEN "success" ELSE "error"
             END
-            AS extraction_status
+            AS extraction_status,
+        parse_result.parse_error_encountered
 
     FROM daily_feeds
-    LEFT JOIN `gtfs_schedule_history.calitp_status` Status
+    LEFT JOIN `gtfs_schedule_history.calitp_status` download_status
```
machow, line 60: Hey, can we discuss style changes that aren't necessary for features being added, before you add them? Conventions on the placement of commas, and table naming choices / camel casing are useful to set across the project, but when they're dropped into code changes without discussion, it takes longer to identify code changes made.

machow: I think your changes are pretty in line with DBT's style guide, so this isn't an opinion on the style changes themselves, but the rolling out of style changes..!

https://github.com/dbt-labs/corp/blob/master/dbt_style_guide.md

airflow/dags/gtfs_views/gtfs_schedule_fact_daily_feed_files.sql
```diff
@@ -35,16 +35,25 @@ raw_daily_files AS (
...
         -- calculate the leading date, so we can fill in missing rows, where
         -- extraction failed to run.
         , LEAD(T1.calitp_extracted_at)
-            OVER (PARTITION BY calitp_itp_id, calitp_url_number ORDER BY T1.calitp_extracted_at)
+            OVER (PARTITION BY T1.calitp_itp_id, T1.calitp_url_number ORDER BY T1.calitp_extracted_at)
             AS tmp_next_date
 
     FROM `gtfs_schedule_history.calitp_files_updates` T1
     JOIN `views.gtfs_schedule_dim_feeds` T2
-        USING (calitp_itp_id, calitp_url_number)
```
machow, line 47: Can we put this using clause back in? This seems like the construct it was designed for (joining on keys of the same name)

machow (approved):
Thanks, this LGTM--I left one comment suggesting putting the USING clause back in, but am onboard with whatever seems best..!

airflow/dags/gtfs_views/gtfs_schedule_fact_daily_feeds.sql
evansiroky: I think it'd be great to follow style conventions. I don't even know there was a convention for SQL, so happy to follow whatever was being followed if that was happening. I'm also a big fan of using linters to enforce style conventions.

airflow/dags/gtfs_views/gtfs_schedule_fact_daily_feed_files.sql
evansiroky: I thought that would lead to an ambiguous error due to the addition of the `T3` join, but looks like it's ok on the first join. I added it back in 4b37dc1.

machow:
thanks--this is super handy

PR #781, opened 2021-11-23
Hide certain feeds in reports

evansiroky:
Hardcodes the hiding of two private feeds:

- Amtrak
- Tulare County Regional Transit Agency


evansiroky:
Receive verbal approval from @machow.

PR #784, opened 2021-11-24
push_to_gcloud.yml: add airflow2 environment as push destination

### LINKED ISSUE #746: Start pushing airflow content into v2 bucket in push_to_gcloud gh action
Let's create a second set of these lines while we upgrade, that also push to the newly provisioned v2 airflow (#743)

https://github.com/cal-itp/data-infra/blob/main/.github/workflows/push_to_gcloud.yml#L65-L70

Note that we'll need to ensure DAGs are not running before / after we do this. 

image

lottspot:


machow (commented):
Thanks, this looks great! I set it to pause dags at creation, so this should be good to go. Do you mind if I try making a branch for the v2 compatible code (should be minor changes to our airflow folder), and then merge this? (I might change it to run on pushes to a refactor to v2 branch / tweak a bit).

lottspot:
Do with this merge as your heart desires

machow:
plucked this code into #851!

PR #785, opened 2021-11-24
Add diagram of static schedule DAGs

evansiroky:
I frequently check a diagram that I made of these DAGs and their dependencies, so I figured I'd make a PR to formally include it in the docs.


machow (approved):
this is great--thanks!

PR #789, opened 2021-11-24
docs: refactor out numbers from file names

machow:
removes number from filenames (sorry @themightychris for baking in this name anti-pattern 😬😬😬)


machow:
Seems to be building. Going to merge so we can get this changes in cc @charlie-costanzo

charlie-costanzo:
> Seems to be building. Going to merge so we can get this changes in cc @charlie-costanzo 

heck yeah thanks for taking care of this @machow !

Issue #566, opened 2021-10-23
GTFS Guidelines: Presence of GTFS-RT feeds


evansiroky:
## Question

Does the provider publish all three versions of standard GTFS Realtime consistent with the GTFS Realtime Best Practices (Trip Updates, Vehicle Positions, and Alerts)?

See also #554.

## Metrics

* A provider is considered to have a complete set of associated GTFS-Realtime feeds if the existence of all three types of GTFS-Realtime feeds (Trip Updates, Vehicle Positions, Service Alerts) can be confirmed.

## Data sources

* Airtable via data warehouse?

## (Data Servicess Team to Copy and Fill Out Below)

The QuVR MD template below will be filled out by a member of the data services team.
This allows us to describe the request, in a way that is easy to hand-off for analysis.
After the research phase, we will sync with the asker to figure out if the metric and dashboard pieces are needed.

Before starting research:

* **Question**: Question written as a single sentence.
* **View**: E.g. `views.gtfs_schedule_fact_daily_feed_files`.
* **Research**:
  - How should the results be presented?
  - When are they needed by?

After reviewing research with the asker:

* **Metric**: what specific calculations are needed?
* **Dashboard**: where should we put the result?

machow:
Hey--looking at this in more detail on the airtable. It appears the calculation for this metric exists as `services.Realtime Status`.

[I've added a quick example analysis to this notebook](https://github.com/cal-itp/data-analyses/blob/main/airtable_explore/0_gtfs_rt_org_service.ipynb)

### Presence Service level GTFS RT feeds

Digging deeper, there are three rollup columns on the `services` table, based on `services data`:

* TripUpdates Status
* VehiclePostitions Status
* Alerts Status

It seems like this is sufficient to count the presence of GTFS RT feeds **per service**? Here is a count of services in the airtable marked as having complete GTFS RT feeds:

![image](https://user-images.githubusercontent.com/2574498/142494376-54e5ac98-bf10-47de-822a-49e377b478aa.png)


### Presence of provider level GTFS RT feeds 

AFAICT the provider level question just involves aggregating services grouped by provider organization.

Here's the number of services per organization

![image](https://user-images.githubusercontent.com/2574498/142494193-ca5f14ec-f2f0-446f-b136-7960f4e887b1.png)

And here is a count of the proportion of services with GTFS RT coverage within organizations

[...]

Issue #557, opened 2021-10-22
GTFS Guidelines: Stable URL


evansiroky:
## Question

Does a GTFS exist at a stable URL exist for a transit provider? 

## Metrics

For a particular provider, the answer to this question is whether or not the provider has a GTFS available at a stable URL.

## Data sources

From Airtable?

## (Data Servicess Team to Copy and Fill Out Below)

The QuVR MD template below will be filled out by a member of the data services team.
This allows us to describe the request, in a way that is easy to hand-off for analysis.
After the research phase, we will sync with the asker to figure out if the metric and dashboard pieces are needed.

Before starting research:

* **Question**: Question written as a single sentence.
* **View**: E.g. `views.gtfs_schedule_fact_daily_feed_files`.
* **Research**:
  - How should the results be presented?
  - When are they needed by?

After reviewing research with the asker:

* **Metric**: what specific calculations are needed?
* **Dashboard**: where should we put the result?

evansiroky:
The latest assessment instructions say to check the GTFS Guidelines dashboard on metabase, so closing as complete for this iteration.

PR #799, opened 2021-12-01
Load raw RT urls into dim feeds

evansiroky:
This is a work in progress until it can be figured out how to match the raw feed extraction date with the rest of the data in the dim feeds table.

### DIFF
--- a/airflow/dags/gtfs_views/gtfs_schedule_dim_feeds.sql
+++ b/airflow/dags/gtfs_views/gtfs_schedule_dim_feeds.sql
@@ -35,6 +35,36 @@ feed_feed_info AS (
             AND T1.calitp_extracted_at < T2.calitp_deleted_at
             AND T2.calitp_extracted_at < T1.calitp_deleted_at
             AND T2.calitp_itp_id != 200
+),
+
+raw_feed_with_extract_date AS (
+  SELECT
+    *,
+    PARSE_DATE(
+      '%Y-%m-%d',
+      REGEXP_EXTRACT(_FILE_NAME, ".*/([0-9]+-[0-9]+-[0-9]+)")
+    ) AS extract_date
+  FROM gtfs_schedule_history.calitp_feeds_raw
+),
+
+final_feed_info AS (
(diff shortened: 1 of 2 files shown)


machow:
Hey--do you mind linking to the issue this PR is for? I think it might be for reports, but am having trouble remembering 😬

evansiroky:
> Hey--do you mind linking to the issue this PR is for? I think it might be for reports, but am having trouble remembering 😬
> 
> edit: seems like #557, is that right?

Yeah, this would support answering the questions posed in #557 and #566.

holly-g:
@machow and @evansiroky to pair on creating a new table.

machow:
DM'd @evansiroky to see if we can pair on this--here are some of my thoughts for posterity. A lot of this is due to how the gtfs_schedule views need to be tuned to store some big fact tables.

There are some subtle issues with the SQL join, and I'm wondering whether with MSD transitdatabase on the horizon, whether it's worth creating tables in a msd_views dag (with a msd_ or similar prefix) so we can answer high-level feed / org questions without the challenges of plugging into the gtfs_schedule tables (which had to be optimized for storing / joining to a lot of stop times across time)

machow:
@evansiroky following up on our pairing--here's a graph of two ways to get this data in...

image

The big challenge with option 1 (adding to dimension) is that it requires a lot of SQL / careful checking to make sure it was correctly added to dim_feeds, because it is a [slowly changing dimension](https://en.wikipedia.org/wiki/Slowly_changing_dimension).

There are ways to make this easier down the road. For example, tools like DBT let you define queries that get the relevant data for the current day (rather than having to worry about time), and then make the SCD by running merges for every day across time.

For now, a nice option for getting the data in might be joining it directly into the fact table. This hopefully should work pretty well for the RT urls, since the metrics that need it should just need it from fact_daily_feeds?!

machow (approved):
Thanks for being so patient on this! (I sanity checked the SQL by running the original and new code, and double checking that they returned the same number of rows)

PR #816, opened 2021-12-03
clarify tools for writing geospatial data to GCS buckets by linking t…

edasmalchi:
A small docs change to Introduction to Analytics Tools --> Notebooks to clarify that directly saving geoparquets to GCS buckets doesn't work for now, linking to the Storing New Data page which already has accurate and complete info for working with geospatial data.


charlie-costanzo (approved):
Great catch, thank you so much for adding this @edasmalchi! it looks great!

PR #817, opened 2021-12-03
Include nice message upon successful feed downloads

evansiroky:
This updates the contents of the email that gets sent off each night when all feeds download successfully so that a nice message is generated instead of an empty box.


machow (approved):
Love that we have this problem!

PR #821, opened 2021-12-03
feature(payments): Add ID dupe order for each ID in funding sources

mjumbewu:
This adds **6** fields to the enriched `customer_funding_source` table:
* calitp_funding_source_id_ranking_size
* calitp_funding_source_id_rank
* calitp_funding_source_vault_id_ranking_size
* calitp_funding_source_vault_id_rank
* calitp_customer_id_ranking_size
* calitp_customer_id_rank

For each of the `..._rank` fields, if the value is `1` then it is the latest occurrence of the corresponding ID. Instances are arranged in desc order by the export datetime.


machow (approved):
LGTM--thanks for getting to the bottom of this.

PR #823, opened 2021-12-03
Remove failing device transaction data test in payments pipeline

### LINKED ISSUE #796: Fix failing device transaction data test in payments pipeline
The `validate_cleaned_device_transaction_pairs_common_fields` check is consistently failing because of 'Route Z' issues. The validation has been failing because of differences between tap-on/-off transaction data for a few (5) MST trips. Either the tap-on or -off was coded with a Route Z route in each case:

```sql
SELECT micropayment_id, route_id_1, route_id_2
FROM `cal-itp-data-infra.payments.invalid_cleaned_device_transaction_common_fields`

-- Row  micropayment_id                       route_id_1  route_id_2	
--  1   6e2b2387-2c6f-492e-ad45-80324317fb4c  001         Route Z
--  2   136d8ddd-eff3-4bcc-830b-c22097886392  001         Route Z
--  3   98e4d3a7-526a-436b-befc-27080b9792d6  Route Z     901
--  4   6fec38b5-6cb2-4650-8c96-3cb13f2f22fb  Route Z     001
--  5   baa426dd-34dc-418e-97ed-0f02696952f0  Route Z     024
```

**Proposed solution for now:**
* [x] Update the payments view to choose `route_id` that is not `'Route Z'` in these cases where the IDs don't match
* [x] Remove the validation from the Airflow pipeline, but keep it in Metabase
* [x] Set up an alert in Metabase for when the validation table (`invalid_cleaned_device_transaction_common_fields`) is not empty

mjumbewu:
* **Update payments view to use non-'Route Z' route_id** - In cases where the route IDs don't match between tap on and off 
transactions, use the route ID that is not 'Route Z'
* **Remove test from validate transaction pairs task** - It is helpful to have the table with potentially invalid rows still generated, and we can set this table up in another platform to alert, but the data is currently a little unreliable to have failures hold up the rest of the pipeline. In particular, [this question](https://metabase.k8s.calitp.jarv.us/question/180-which-correlated-tap-on-and-off-transactions-have-differing-information) in Metabase can be set up with alerts for when there are any results in the table.

This PR resolves #796

### DIFF
--- a/airflow/dags/payments_views/payments_rides.sql
+++ b/airflow/dags/payments_views/payments_rides.sql
@@ -102,7 +102,7 @@ SELECT
     p.product_type,
 
     -- Common transaction info
-    t1.route_id,
+    CASE WHEN t1.route_id <> 'Route Z' THEN t1.route_id ELSE COALESCE(t2.route_id, 'Route Z') END AS route_id,
     r.route_long_name,
     r.route_short_name,
     t1.direction,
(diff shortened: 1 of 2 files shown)


machow (approved):
Is this change of `payments_rides` supposed to be in there? The one thing I was wondering for that SQL is are we using Route Z as an internal token for missing data, or making sure missing values are set so they can be joined on data littlepay marked as Route Z?

My barbarian guess is that we're leaning into Route Z being the token for "explicitly unknown", but want to double check I've got the intent right!

In any event, LGTM :).

mjumbewu:
@machow 

As I understand it thus far, the `Route Z` values communicate that no route information is available (yet) for the given record.

I did intend to include the `payments_rides` changes in here. Since all of the discrepancies we've seen so far between tap-on and tap-off transactions are cases where one of those taps has `Route Z` for the route (see #796), I figured we could just account for that case here. It's not the most conservative choice, but given that this is the last step in a chain of views, I think it's ok.

Issue #807, opened 2021-12-01
fix: wrong URL (404) in docs


holly-g:
**Describe the bug**
Currently, user gets a 404 when accessing the link to create a PAT. 

**Expected behavior**
[Correct link](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token)

holly-g:
In light of https://github.com/cal-itp/data-infra/issues/792 (the switch to SSH) -- 
fix: wrong URL (404) in docs. --> Remove PAT in docs

Issue #792, opened 2021-11-29
[docs] Switch to SSH cloning instead of HTTPS + PAT 


hunterowens:
### Current Behavior

[currently](https://docs.calitp.org/data-infra/analytics_tools/saving_code.html), the saving code docs use a github PAT + https:// based cloning. as a consequence, users have to type the PAT into the command line every time they push / pull from GH, I think. 

### Expected Behavior
I believe we should use SSH based cloning as our preffered method, allowing users to not have to enter passwords on the hub to push / pull. The GH guide is [here](https://docs.github.com/en/authentication/connecting-to-github-with-ssh)

machow:
re-opening since the instructions as given do not allow users to push code via ssh. Let's make sure to run through new workflows when documenting

https://cal-itp.slack.com/archives/C02KH3DGZL7/p1639172356045100

edasmalchi:
I'm working on a quick docs update to address this -- any reason to suggest analysts use an ssh passphrase?

It doesn't seem necessary to me for our setup, the linked GitHub docs seem to nudge you towards making one

(I just switched to SSH from PAT authentication following those to see if they were adequate, didn't make an ssh passphrase and working fine) Oddly didn't seem to hit the same permissions issue as Ben, but I'll add that phrase to our docs since it seemed to fix for him

edasmalchi:
^merged PR should fix this

PR #832, opened 2021-12-07
docs: add SSH cloning

holly-g:
* Removes reference to [PAT](https://github.com/cal-itp/data-infra/issues/807)
* Removes HTTPS cloning in favor of[ SSH](https://github.com/cal-itp/data-infra/issues/792)


charlie-costanzo:
This should now be captured in the merged PR, #839

PR #840, opened 2021-12-08
Genericize parse vehicle_positions to other rt types

evansiroky:
DRAFT PR, Not yet tested at all.

This PR refactors the parsing of RT data such that it is genericzied to all types of RT feeds. The existing parse_rt_vehicle_positions DAG task has been mostly made into an operator that is able to handle other RT feed types.


machow:
Hey @evansiroky, I'm working on patching in additional engineering support from the Jarvus side for tasks like this, so hope you don't mind that I took a look. 

It seems like a key value with airflow Operators is separating the "what" (content) from the "how" (execution).

For example, the SqlToWarehouseOperator lets you define "what"...

* specify a sql query to run
* the resulting table name should be
* descriptions for columns

and takes care of the "how" (by running the SQL and creating the table in our warehouse). The reason is that the "what" varies a lot across tasks, but the how stays relatively the same. If we swapped databases, we could modify or use a different operator to change the "how" (e.g. to support postgres).

One potential challenge with these operators is that they handle both the what and how through methods. They don't take any "what" as arguments, except for the operator that should run (via an operator factory). It seems like for parsing a protobuff RT file, there are three big pieces of "what":

* what header fields to extract
* what body fields to extract
* what type conversions to make (or a schema, etc..)

And some smaller whats:

* prefixes
* file type names

What do you think of us separating the underlying methods (the how) from the what (configuration via arguments)? For example, this would mean changing methods like these to class init arguments:

```python
    def get_google_cloud_filename_prefix(self):
        return "tu"

    def get_rt_file_substr(self):
        return "trip_updates"
```

and the 

```python
    def get_entity_details(self, x):
        """
        Returns a list of dicts containing entity values to be added to a dataframe.
        See https://gtfs.org/reference/realtime/v2/#message-tripupdate for list of
        possible entity values.
        """
        entity = x.get("entity")
        details = []
        if entity is not None:
            for e in entity:
                d = {
                    "entity_id": self.pull_value(e, "id"),
                    "trip_id": self.pull_value(e, "trip_update.trip.trip_id"),
            ...
```

To be more general, and allow "what" arguments like,

```python
operator: BaseRealtimeProcessor
# note the fields extracted above are now specified in the task
entity_details:
  entity_id: "id"
  trip_id: "trip_update.trip.trip_id"
filename_prefix: "tu"
rt_file_substring: "trip_updates"
```

machow:
Notes from pairing w/ @lauriemerrell :

* cd to `airflow` folder, use `import debug`
* can now import from dags (e.g. `from dags.rt_loader.some_task_file import main`)
* need to install apache-airflow[google] and apache-airflow-providers-cncf-kubernetes, and python-dotenv

lauriemerrell:
I am now able to run this successfully locally and create the parquet files. 🎉 

However:
🚫 = must address before ready for review
❓  = may be able to proceed without addressing

* 🚫 Trip updates is failing, and I think it's running out of space. Need to check if that's my local Docker needing more resources or if there's something that needs to be changed about the processing itself.
* 🚫 There are a lot of null fields in `trip_updates` and `service_alerts` with the current setup. I think that the way we are referencing the fields is not working correctly for these... I need to investigate more but it seems like we are not actually pulling out the field values (i.e., field naming is not correct or the entity response structure is different than we expect). 
* 🚫  We are currently pulling a somewhat odd subset of fields. I need to clarify what fields we actually want to pull. 
* ❓ Seems like we can maybe improve performance by leveraging the `gtfs_rt.calitp_files` table to figure out what files to load rather than globbing everything for the entire day? Not sure whether we want to do that here or split into a subsequent PR.

lauriemerrell:
* 🚫 I will need assistance/pairing to figure out the memory leak issue. I have confirmed that memory usage is spiking and causing `trip_updates` to fail. It does not seem to be an issue for `service_alerts` or `vehicle_positions`. 

* 🔜 I recommend creating a new issue to add the ability to parse list-type data because that requires refactoring how the data is unpacked from the JSON in `get_entity_details` / `pull_details`.
  *  For service alerts, the `alert` level actually contains a *list* of `activePeriod.start` and `activePeriod.end` values (i.e., there can be multiple start and end values for the same alert.) **So splitting this out means that until it's implemented we can literally only pull the list of alert entity IDs with no actual information for `service_alerts`**-- literally all the other fields I think we'd care about contain a list structure at some point (even header text is a translated string that contains a list of of text, language pairs). I think it's worth it to keep this moving without that.  
  * List parsing would allow us to load `stopTimeUpdates` in `trip_updates` too. 
* ❓ Do we want to fix NA handling via Pandas? Right now for fields that are empty in the JSON, they are loaded as `<NA>` (not as actual missing values). Not sure if that's what we want.

* 🔜 Also would recommend refactoring to look at the `gtfs_rt.calitp_files` table (instead of globbing everything) in a separate issue.

lauriemerrell:
I spoke with Michael and @atvaccaro about this ticket yesterday and Michael sent me the updates here: https://github.com/cal-itp/data-infra/blob/main/airflow/dags/rt_loader/parse_rt_vehicle_positions.py#L163-L169 which makes the `trip_updates` loop more efficient, but based on testing yesterday it was still timing out. I will need to investigate more.

Also discussed this with @mjumbewu this morning, kind of wondering if we'll need to change the overall approach on how the protobuf files are loaded and written to parquet (or maybe do something else entirely)... Not sure the current structure is efficient enough for the size of the trip updates data.

airflow/plugins/operators/rectangle_realtime_data_operator.py
```diff
@@ -0,0 +1,214 @@
...
+from pathlib import Path
+
+import pandas as pd
+import tempfile
+
+# Note that all RT extraction is stored in the prod bucket, since it is very large,
+# but we can still output processed results to the staging bucket
+
+
+class RectangleRealtimeDataOperator(BaseOperator):
```
atvaccaro, line 18: what is a `Rectangle` here?

atvaccaro (commented):
Also, is there a guide to executing this locally or in some staging environment?

airflow/plugins/operators/rectangle_realtime_data_operator.py
```diff
@@ -0,0 +1,214 @@
...
+
+        # get execution_date from context:
+        # https://stackoverflow.com/questions/59982700/airflow-how-can-i-access-execution-date-on-my-custom-operator
+        iso_date = str(context["execution_date"]).split("T")[0]
+
+        # fetch files ----
+        feed_files = self.fetch_bucket_file_names(iso_date)
+
+        # parse feeds ----
+        for feed, files in feed_files.items():
```
atvaccaro, line 184: Unsure how you're tracking progress; https://github.com/tqdm/tqdm is my favorite progress bar library

airflow/plugins/operators/rectangle_realtime_data_operator.py
```diff
@@ -0,0 +1,214 @@
...
+
+                # convert protobuff objects to DataFrames
+                entities_dfs = [*map(self.rectangle_entities, parsed_entities)]
+                entities_dfs = [df for df in entities_dfs if df is not None]
+
+                print("  %s entities sub dataframes created" % len(entities_dfs))
+                if len(entities_dfs) > 0:
+                    entities_rectangle = pd.concat(entities_dfs)
+                    entities_rectangle.insert(0, "calitp_itp_id", int(feed[0]))
+                    entities_rectangle.insert(1, "calitp_url_number", int(feed[1]))
```
atvaccaro, line 204: Even with the changes I remember from our call (specifically using generators instead of lists), there's lots of in-memory operations to ultimately produce a dataframe that is written to parquet.

pandas uses either https://github.com/dask/fastparquet or https://github.com/apache/arrow/tree/master/python to write to parquet.

It seems like https://arrow.apache.org/docs/python/ipc.html#efficiently-writing-and-reading-arrow-data would be a good method for this. Something roughly like:

1.  For each feed, define the output stream and schema
2. Iterate over the files in the feed
3. Write each file in batches; if it's still too large, sub-divide into a sane number of records

Parallelization (across feeds) and asynchronous programming could help here.

Related question: is there a reason everything ends up in a single parquet file? Do you have an example of multiple file names within the same feed?

airflow/plugins/operators/rectangle_realtime_data_operator.py
lauriemerrell: Right now we aren't actually tracking progress, tqdm makes sense, I can look into adding

airflow/plugins/operators/rectangle_realtime_data_operator.py
lauriemerrell: Thanks Andrew, I do think we're going to need to move more in the streaming direction (and hoping we can work through that together) -- but I wanted to share more context, sorry in advance for length but I wanted to lay it all out in case it's helpful.

The terminology and the naming in the code here is a bit confusing. We're actually looping over an entire day's worth of files for each feed, so that's why there are multiple files. 

We start here:

**A**
[in`gtfs-data/rt`]
* [timestamp T on date D] 
   * [feed 1]
      * service_alerts protobuf file
      * trip_updates protobuf file
      * vehicle _positions protobuf file 
    * [feed 2]
      * service_alerts protobuf file
      * trip_updates protobuf file
      * vehicle _positions protobuf file 
 * [timestamp T+20s on date D] 
   * [feed 1]
      * service_alerts protobuf file
      * trip_updates protobuf file
      * vehicle _positions protobuf file 
    * [feed 2]
      * service_alerts protobuf file
      * trip_updates protobuf file
      * vehicle _positions protobuf file 
 
And we're converting to this:

**B**
[in `gtfs-data/rt-processed`]
* vehicle_positions 
   * [feed_1_vehicle_positions_date_D].parquet
   * [feed_2_vehicle_positions_date_D].parquet
* service_alerts
   * [feed_1_service_alerts_date_D].parquet
   * [feed_2_service_alerts_date_D].parquet 
* trip_updates
   * [feed_1_trip_updates_date_D].parquet
   * [feed_2_trip_updates_date_D].parquet 

So the current approach to the conversion to parquet is combining two goals:
1. Rectangling the data (converting from the JSON / key-value format to a nice rectangular dataframe)
2. Combining all 3k timestamp data points into just one file per day per feed

The raw protobuf files vary significantly in size and complexity from feed to feed. I think each input file itself is generally not too big, but when combined they start to get big in the output files.

Stepping back, I think that the **actual** goal from the pipeline perspective is just to get the data into BigQuery as an external table (I think that's the reason we need to convert file types, [BigQuery only supports certain file types in external tables from GCS](https://cloud.google.com/bigquery/external-data-cloud-storage)). 

I think that with BigQuery's [built-in JSON support](https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions) and [pyarrow's JSON support](https://arrow.apache.org/docs/python/json.html) it's possible we don't actually need to do the rectangling in Python here at all and can do all the UNNESTing stuff directly in BQ, but maybe it's possible to make the process more efficient via the approaches you mentioned. I definitely think parallelizing across feeds would help. 

I'm not sure how much we care about these intermediate parquet files, though I do think that Eric has used them for analysis recently and has built a significant amount of tooling to ingest them; we should notify him if we're going to change things substantially and move away from producing rectangled parquet files. 

A further piece of context: this pipeline is already running and working for the `vehicle_positions` data, so this current PR was basically just an attempt to apply that existing pipeline with minimal modifications to the other two data types. However, it seems that both in size and structure the `trip_updates` data in particular is probably not going to work. (I wanted to just to explain a bit why we're starting from this point; the vehicle positions data is smaller and already "more rectangular" than the other two data types, which might be why we started with it, but it also makes it harder to apply the same approach.)

lauriemerrell:
@atvaccaro 

> Also, is there a guide to executing this locally or in some staging environment?

There's some local setup info in the [README here](https://github.com/cal-itp/data-infra/tree/main/airflow). I can also definitely help you walk through it anytime, I just had to do the setup myself 😄

atvaccaro:
I recommend we close this as #1011 replaced this.

lauriemerrell:
Superseded by #1011 / #1039

PR #851, opened 2021-12-10
refactor(airflow): upgrade to airflow v2

machow:
This PR makes some small tweaks to make our pipeline airflow v2 compatible. Note that because we are on airflow v1.10.14 (and not v1.10.15) AFAIK this code is not compatible with our current setup. We will need to push this to the new cluster, and once it's working, disable the original one.

TODO before ready:

- [x] : latest only operators failing (getting on call with Chris C to understand / resolve)
  * appears resolved. Chris has never seen this, and they started working. (may have been startup quirks?)
- [x] : it looks like in some cases tasks ran even when their upstream dependencies were waiting or failed? (couldn't reproduce)

TODO for release:

- [x] : update github action to push this code to the new cluster. get from https://github.com/cal-itp/data-infra/pull/784
- [x] : turn on sandbox and verify it works
- [x] : turn on loaders and verify they work
- [x] : before 8pm, pause all DAGs on old cluster, turn on gtfs_download (our only live data extractor)
- [x] : verify feeds downloaded
  * inspect download logs
  * glance at gtfs-data/schedule/<execution date>
  * query for agency.txt from gtfs_schedule_history.calitp_files_updates

(Note that gtfs_download is the only DAG to use xcom)

### DIFF
--- a/airflow/plugins/operators/__init__.py
+++ b/airflow/plugins/operators/__init__.py
@@ -1,6 +1,4 @@
 # flake8: noqa
-# TODO: after upgrading airflow, using a relative import raises the error
-# "attempted relative import with no known parent package"
 from operators.external_table import ExternalTable
 from operators.once_off_external_task_sensor import OnceOffExternalTaskSensor
 from operators.pod_operator import PodOperator
--- a/airflow/plugins/operators/__init__.py
+++ b/airflow/plugins/operators/__init__.py
@@ -1,9 +1,11 @@
 # flake8: noqa
-from .external_table import ExternalTable
-from .once_off_external_task_sensor import OnceOffExternalTaskSensor
-from .pod_operator import PodOperator
-from .python_taskflow_operator import PythonTaskflowOperator
-from .sql_to_warehouse_operator import SqlToWarehouseOperator
-from .csv_to_warehouse_operator import CsvToWarehouseOperator
-from .python_to_warehouse_operator import PythonToWarehouseOperator
-from .sql_query_operator import SqlQueryOperator
+# TODO: after upgrading airflow, using a relative import raises the error
+# "attempted relative import with no known parent package"
+from operators.external_table import ExternalTable
+from operators.once_off_external_task_sensor import OnceOffExternalTaskSensor
+from operators.pod_operator import PodOperator
+from operators.python_taskflow_operator import PythonTaskflowOperator
+from operators.sql_to_warehouse_operator import SqlToWarehouseOperator
+from operators.csv_to_warehouse_operator import CsvToWarehouseOperator
+from operators.python_to_warehouse_operator import PythonToWarehouseOperator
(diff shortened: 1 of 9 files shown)


machow:
Okay--testing the export from v1 to v2. The one big snag I've hit is that the export script does not handle commas in table descriptions well.... :/. The cryptic errors did not help.......

image

machow:
Snag 2: the xcom database in v2 uses a compound primary key, but this key is duplicated in our v1 database for some task runs. E.g. [generate_provider_list on Aug 21](https://wffd3b3bb3b7b7ee5p-tp.appspot.com/admin/basexcom/?flt3_14=2021-08-21+00%3A00%3A00). I don't know why this happens (e.g. they have different timestamps).

Crazily, I downgraded our airflow v2 environment to airflow 2.0.1 when I hit the last error. I'm now going to try upgrading, to see if later versions of airflow fixed this...

machow:
rebased to `fixup` a bunch of tweaks to the gh action

machow:
**This code is running right now on v2, and will block the merging of new PRs. We can rebase to this branch and push up to v2 if needed. If possible, let's queue up any substantial feedback as github issues.**

Notes on key changes in behavior:

* merging to main now pushes DAGs up to `calitp-airflow2-prod` composer environment
* the new UI link is here!: https://o1d2fa0877cf3fb10p-tp.appspot.com/home
* the changes to support airflow v2 are not backwards compatible (but are fairly lightweight). We can't push this content to v1 (but v2 appears to be working well). In a sort of nightmare scenario we could port DAG code to work on v1, but I don't anticipate needing it.

image

airflow/plugins/operators/__init__.py
```diff
@@ -1,9 +1,11 @@
...
-from .external_table import ExternalTable
-from .once_off_external_task_sensor import OnceOffExternalTaskSensor
-from .pod_operator import PodOperator
-from .python_taskflow_operator import PythonTaskflowOperator
-from .sql_to_warehouse_operator import SqlToWarehouseOperator
-from .csv_to_warehouse_operator import CsvToWarehouseOperator
-from .python_to_warehouse_operator import PythonToWarehouseOperator
-from .sql_query_operator import SqlQueryOperator
+# TODO: after upgrading airflow, using a relative import raises the error
+# "attempted relative import with no known parent package"
```
evansiroky, line 3: This seems more explicit anyways. Can this comment be removed now?

evansiroky (approved):
LGTM, aside from a comment on a comment.

airflow/plugins/operators/__init__.py
machow: Will remove the comment! It's probably won't be an issue, but the challenge with this format is that it's ambiguous where they're being imported from. It could be either an installed (or the built in) operators package, or this operators folder. The dots are unambiguous that it's from our internal folder. (but I'm not sure why this error is occurring :/)

PR #852, opened 2021-12-10
fix(gtfs-rt): parse internal ids to ints to match schema

machow:
Pairing with @Nkdiaz -- the table schema has ids like calitp id as integers. however, the loader is loading strings. going to merge, since it's a pretty small change and will let us re-run / check everything


machow:
Ah, actually I can't merge w/o adding @Nkdiaz and having her press accept, since I think I'm set as a maintainer on this repo now. @evansiroky do you want to review / accept?

evansiroky (approved):

PR #855, opened 2021-12-10
fix(gtfs-rt): fix typecasting for rt files

machow:
Should fix errors parsing RT files introduced by #852. I have docker running airflow v2 right now, so need to spin things up to test, before switching this out of being a draft.

### DIFF
--- a/airflow/dags/rt_loader_files/calitp_files_process.py
+++ b/airflow/dags/rt_loader_files/calitp_files_process.py
@@ -46,8 +46,8 @@ def main(execution_date, **kwargs):
 
     ser_id = raw_res["name"].str.split("/")
 
-    raw_res["calitp_itp_id"] = int(ser_id.str.get(-3))
-    raw_res["calitp_url_number"] = int(ser_id.str.get(-2))
+    raw_res["calitp_itp_id"] = ser_id.str.get(-3).astype(int)
+    raw_res["calitp_url_number"] = ser_id.str.get(-2).astype(int)
     raw_res["calitp_extracted_at"] = ser_id.str.get(-4)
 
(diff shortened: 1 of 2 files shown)


machow:
@evansiroky sorry for messing up with #855. I setup my local dev against airflow v1 again and tested these changes by...

* deleting all outputs of this task on the gtfs-data-test bucket
* confirming in the bigquery console that the result is query-able

The big mistake was that I needed to set `skip_leading_rows = 1` in the table schema, so it knows to skip the CSV headers.

airflow/dags/rt_loader_files/calitp_files_process.py
```diff
@@ -46,8 +46,8 @@ def main(execution_date, **kwargs):
 
     ser_id = raw_res["name"].str.split("/")
 
-    raw_res["calitp_itp_id"] = int(ser_id.str.get(-3))
-    raw_res["calitp_url_number"] = int(ser_id.str.get(-2))
+    raw_res["calitp_itp_id"] = ser_id.str.get(-3).astype(int)
+    raw_res["calitp_url_number"] = ser_id.str.get(-2).astype(int)
```
evansiroky, line 50: Just curious what led to making this change.

evansiroky (approved):
Seems reasonable to me, although I have one question that I'm curious about that is non-blocking.

airflow/dags/rt_loader_files/calitp_files_process.py
machow: Initially I thought that it was a schema issue, since the itp ids w/o this change are AFAICT strings. Now that I know it was needing to drop the header, I'm not as sure, but I included when testing just to have things bolted down / matching the schema

PR #856, opened 2021-12-10
Update TCTA URL

evansiroky:
Previous URL was pointing to a zip file with a subfolder. This one has no subfolder.


evansiroky:
I went ahead and merged this to meet a 4pm deadline for when the daily DAGs run.

PR #861, opened 2021-12-13
Update agencies.yml

o-ram:
Update the feed for Yolobus

Add in the following agencies/feeds:
- Antelope Valley Transit
- Mountain Transit
- commute.org shuttles

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -46,6 +46,14 @@ anaheim-resort-transportation:
       gtfs_rt_service_alerts_url: null
       gtfs_rt_trip_updates_url: null
   itp_id: 14
+antelope-valley-transit-authority:
+  agency_name: Antelope Valley Transit Authority
+  feeds:
+    - gtfs_schedule_url: https://www.avta.com/userfiles/files/AVTA%20GTFS.zip
+      gtfs_rt_vehicle_positions_url: null
+      gtfs_rt_service_alerts_url: null
+      gtfs_rt_trip_updates_url: null
+  itp_id: 16
 arcadia-transit:
   agency_name: Arcadia Transit
   feeds:
@@ -1409,7 +1417,7 @@ westcat:
 yolobus:
   agency_name: Yolobus
   feeds:
-    - gtfs_schedule_url: http://iportal.sacrt.com/GTFS/Yolobus/google_transit.zip
+    - gtfs_schedule_url: http://www.yolobus.com/GTFS/google_transit.zip
       gtfs_rt_vehicle_positions_url: null
       gtfs_rt_service_alerts_url: null
       gtfs_rt_trip_updates_url: null
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -302,6 +302,14 @@ commerce-municipal-bus-lines:
       gtfs_rt_service_alerts_url: https://citycommbus.com/gtfs-rt/alerts
       gtfs_rt_trip_updates_url: https://citycommbus.com/gtfs-rt/tripupdates
   itp_id: 75
+commute-org-shuttles:
+  agency_name: Commute.org Shuttles
+  feeds:
+    - gtfs_schedule_url: http://data.trilliumtransit.com/gtfs/commute-ca-us/commute-ca-us.zip
+      gtfs_rt_vehicle_positions_url: null
+      gtfs_rt_service_alerts_url: null
+      gtfs_rt_trip_updates_url: null
+  itp_id: 76
 compton-renaissance-transit-service:
   agency_name: Compton Renaissance Transit Service
   feeds:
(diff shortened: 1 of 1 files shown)


airflow/data/agencies.yml
```diff
@@ -1409,7 +1433,7 @@ westcat:
 yolobus:
   agency_name: Yolobus
   feeds:
-    - gtfs_schedule_url: http://iportal.sacrt.com/GTFS/Yolobus/google_transit.zip
+    - gtfs_schedule_url: http://www.yolobus.com/GTFS/google_transit.zip
```
e-lo, line 1436: this looks like the same data as is in the iportal.sacrt.  Is SAC-RT mirroring yolobus.com? Or visa versa? (aka who is the publisher?)

airflow/data/agencies.yml
```diff
@@ -294,6 +302,14 @@ commerce-municipal-bus-lines:
       gtfs_rt_service_alerts_url: https://citycommbus.com/gtfs-rt/alerts
       gtfs_rt_trip_updates_url: https://citycommbus.com/gtfs-rt/tripupdates
   itp_id: 75
+commute-org-shuttles:
```
e-lo, line 305: This is a precursor feed to the bay area 511 feed, so duplicative of datasets.

airflow/data/agencies.yml
o-ram: Yolobus is the publisher and SacRT is mirroring. Yolobus made it clear that they consider their link to be the ultimate source of truth.

airflow/data/agencies.yml
o-ram: This one looks to be an active Trillium feed, and it's not completely identical to what I downloaded for commute shuttles from 511. Either way, shouldn't they be in the .yml file so we can easily get validation results for them and make sure they end up on reports.calitp? Right now, I have them marked as missing from the reports site. We have close to all the other bay area services in the .yml file/reports site even though they're technically part of the 511 feed too. 

I guess what I'm trying to say, for the purpose of clarity, is why wouldn't we include this feed, when we have other bay area operator feeds in our database outside of 511? (I.e. Westcat has a Trillium static link, Samtrans has their own static link, etc.)

e-lo: > This one looks to be an active Trillium feed, and it's not completely identical to what I downloaded for commute shuttles from 511

airflow/data/agencies.yml
```diff
@@ -794,6 +810,14 @@ morro-bay-transit:
...
       gtfs_rt_trip_updates_url: null
   itp_id: 213
+mountain-transit:
+  agency_name: Mountain Transit
+  feeds:
+    - gtfs_schedule_url: https://api.transloc.com/gtfs/mountaintransit.zip
+      gtfs_rt_vehicle_positions_url: null
+      gtfs_rt_service_alerts_url: null
+      gtfs_rt_trip_updates_url: null
+  itp_id: 214
```
evansiroky, line 820: I am confused about where did this itp_id of 214 come from. I'm not seeing a match in the airtable organization list.

airflow/data/agencies.yml
evansiroky: There isn't much cost to ingest this information and it is a somewhat separate issue from which feed the assessment process selects. I say we might as well be including these feeds and if there is ever a case that an agency goes rogue and decides to not be aggregated, then we'll have the history.

evansiroky (changes requested):
Just to publicly acknowledge it, I know that @e-lo is pretty adamant about the `itp_id` not being supposed to be relevant anymore, but enforcing a unique constraint on this ID is critical for this repository and the data pipeline until an eventual migration occurs. And that migration of having an explicit link to the transit database's GTFS Dataset ID is something that is probably a few months out at best. When said migration occurs it may be of some value to have ITP IDs provide some helpful clues about how to match information in this agencies.yml file to information in the transit database.

With all that being said, I checked to make sure that each of the ITP IDs were not duplicated from some value in the past and made sure a good match was found in the transit database. One of the items noted below did not have a match. Once that is resolved, this should be good to go.

airflow/data/agencies.yml
o-ram: I got it from the [Teamwork CRM](https://calitp.teamwork.com/crm/companies/28815), which originally pulled its itp_id info from the [primary providers spreadsheet](https://docs.google.com/spreadsheets/d/105oar4q_Z3yihDeUlP-VnYpJ0N9Mfs7-q4TnribqYLU/edit?usp=sharing) that was since deprecated into Airtable. I see what you're saying, the itp_id column for Mountain Transit is empty in airtable. 

@e-lo would there be a reason for that? Or did the itp_id just not port over for Mountain for some reason?

evansiroky: Thanks @e-lo for adding the ITP ID in the database!

evansiroky (approved):

PR #871, opened 2021-12-15
clarify ssh instructions

edasmalchi:
Clarifies instructions for creating an ssh key and adds command to set permissions for private key if needed.

Addresses issue [792](https://github.com/cal-itp/data-infra/issues/792)


charlie-costanzo (approved):
This looks great Eric, thank you so much for fixing this!

PR #874, opened 2021-12-17
Add customer and funding source info to payments_rides view

### LINKED ISSUE #617: Investigate customer_funding_source cleaning options

### LINKED ISSUE #704: Merge customer funding source in views.payments_rides
* This is the last piece of https://github.com/cal-itp/data-infra/issues/312
* Opening here, since everything is done, so it will be easier to track
* Customer funding source is needed to identify unique customers
  - https://github.com/cal-itp/data-infra/issues/372

### LINKED ISSUE #753: Compile payments assumptions to verify with Littlepay
Create a document (Notebook or Markdown) to send to Littlepay about the things we're doing to clean the data, to validate that all of those transformations are sensible.

mjumbewu:
Adds fields describing customers to the `payments_rides` view.

- `funding_source_vault_id`, `customer_id`: Can be used as a key to join to cleaned customer funding source.
- `principal_customer_id`: Closest thing we have that can be used like a unique rider.
- Other funding source fields, `bin`, `masked_pan`, `card_scheme`, `issuer`, `issuer_country`, `form_factor`

Also adds cleaned tables `stg_cleaned_customer_funding_sources` and `stg_cleaned_customers` to the `payments` dataset.

- `stg_cleaned_customers`: Mainly a mapping from `customer_id` to `principal_customer_id`
- `stg_cleaned_customer_funding_source_vaults`: Distinct records from `customer_funding_source` along with timestamp ranges for when the records are valid for a given vault/customer combination. This table can get used like a slowly changing dimension, where the key is (`customer_id`, `funding_source_vault_id`).

Further exploration of the dataset(s) is done in https://github.com/cal-itp/data-analyses/blob/main/littlepay_data_issues/customer_funding_source_issues.ipynb#MVP.

This issue closes #704 and closes #617.


machow (changes requested):
@mjumbewu this is kind of a nuts thing, but do you mind if we formalize setting NULLs to very early / late dates to this?:

* early NULL: 1899-01-01 00:00:00
* late NULL: 2099-01-01 00:00:00

The kind of weird reason is that pandas / numpy are not great at parsing extremely far in the past / future dates.

image

AFAIK there is recent work to get around this, and it may not be an issue in more recent if people use the `convert_dtypes()` method. (Also, this will still happen if there are nulls in the column, so in practice people should still be vigilant about type conversions!)

I'll add a section to the docs to codify this practice (which will also let us change the decision in step!)

mjumbewu:
@machow ah, makes sense. Thanks for catching and calling it out.

machow:
Ah, thanks--we can go through in a refactor and sling far future further into the future! I can't remember the range on numpy's datetime64[ns] type, but I think there are a few hundred years into the future available

machow (approved):

mjumbewu:
For posterity, the max date is [some time in 2262](https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#timestamp-limitations).

Issue #550, opened 2021-10-22
Post airflow task failures to a slack channel


machow:
Currently we email airflow task failures to Hunter and I. However, as DAGs grow / domains are added, it will be useful to have task success / failures out and convenient to access for everyone.

Chris C walked me through the process, and it seems fairly straightforward to set up. His advice was to not use the airflow provided tooling, since making the request to the slack API is straightforward (and airflow tools can get surprisingly weird).

machow:
Checking on this, but AFAICT we can use the on_failure_callback in the task default should do it. 

This should enable messages:

*  use the requests with an environment variable for the slack token or URL we'll need for posting.
* set on_failure_callback to be our message posting function
  * need to add in default_args when calling create_dag here: https://github.com/cal-itp/data-infra/blob/main/airflow/dags/dags.py#L119

Relevant docs:

* [Dag class docs](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/dag/index.html?highlight=dag#airflow.models.dag.DAG)
* [Baseoperator docs](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/index.html?highlight=baseoperator#airflow.models.BaseOperator)

holly-g:
See #data-infra-alerts

PR #875, opened 2021-12-17
Feat email slack

machow:
From pairing with @holly-g -- this PR addresses #550 by posting task failures to a slack channel.

Note that..

* calitp slack is maxed out on app integrations (unpaid plans have 10; cc @thekaveman know of any unused apps? ;)
* I tested by adding to jarvus slack and posting to a #calitp-pipeline-errors channel there

image

I wanted to at least kick-off exploration around this, since as more people join it will be useful to flip the approach...

* away from adding people to dag error emails
* toward allowing people to subscribe to them (e.g. by joining a channel!).
* 
* The code as implemented sends to a single channel, but it should be straightforward to split out if useful!


machow:
Keeping feedback from slack here so I can find it later!:

image

airflow/dags/dags.py
```diff
@@ -128,4 +147,8 @@ def sql_enrich_duplicates(schema_tbl, key_columns, order_by_columns):
             "sql_enrich_duplicates": sql_enrich_duplicates,
         },
         user_defined_filters=user_defined_filters,
+        default_args={
+            "on_failure_callback": email_failure,
+            "on_retry_callback": email_failure,
```
evansiroky, line 152: I kind of worry about emailing on retry since I think some external DAG dependency tasks may "retry" multiple times while waiting for the external dependency to finish.

evansiroky (approved):
Seems like a good idea, but if the retry emails become too noisy and not enough signally, we may want to change that.

airflow/dags/dags.py
machow: Ah, shoot--good point. Let me double check / tweak to ensure external dependencies won't fire retry messages.

themightychris:
I set up a slack integration for this: https://cal-itp.slack.com/services/B0308PJLLAC

That should be good to use, assuming our slack manages to stay pro past feb 2

themightychris:
the `CALITP_AIRTABLE_API_KEY` env var is in place on `calitp-airflow2-prod`

airflow/dags/dags.py
evansiroky: > Ah, shoot--good point. Let me double check / tweak to ensure external dependencies won't fire retry messages.

Seems like @machow never got around to checking this, and that's OK. However, let's make sure we're not getting a lot of noise from retries before merging this.

evansiroky: Also, do these callbacks override the default method of sending off error emails?

airflow/dags/dags.py
```diff
@@ -114,6 +115,24 @@ def sql_enrich_duplicates(schema_tbl, key_columns, order_by_columns):
...
+    message = f"""
+Task Failed: {ti.dag_id}.{ti.task_id}
+Execution Date: {ti.execution_date}
+Try {ti.try_number} of {ti.max_tries}
+
+<{ti.log_url}| Check Log >
+"""
+    requests.post(
+        os.environ["CALITP_SLACK_URL"], json={"text": message},
+    )
```
evansiroky, line 133: Could this result in an error if it fails? Should we surround it with try except?

evansiroky (changes requested):
Looking at this again, I have a few more questions. I'd like for a new hire who will be working regularly with DAGs to investigate.

airflow/dags/dags.py
themightychris: @atvaccaro

airflow/dags/dags.py
themightychris: @atvaccaro

airflow/dags/dags.py
```diff
@@ -114,6 +115,24 @@ def sql_enrich_duplicates(schema_tbl, key_columns, order_by_columns):
     """
 
 
+def email_failure(context):
+    slack_url = os.environ.get("CALITP_SLACK_URL")
+    if slack_url is None:
+        print("Skipping email to slack channel. No CALITP_SLACK_URL in environment")
```
atvaccaro, line 121: is this supposed to return here?

airflow/dags/dags.py
atvaccaro: In the past, I have sometimes found issues with accessing the task instance information from the context. I was never really sure _why_; I would recommend an overall structure similar to the following.

```python
try:
    msg = "..."
    requests.post(...)
except:
    requests.post("failed to log to slack")
```

With the `except` being as dumbed-down as possible as a last-ditch attempt to log to Slack.

airflow/dags/dags.py
atvaccaro: I believe they are separate but I'm not 100% sure; I can test when I have time.

For what it's worth, I have never used email-based alerting with Airflow. I've always used Slack for errors and something like Prometheus to check for positive state.

airflow/README.md
```diff
@@ -83,3 +83,5 @@ Note that the following variables were set manually in cloud composer:
 * `SENDGRID_MAIL_FROM`
 * `POD_CLUSTER_NAME` - name of the kubernetes cluster
 * `POD_LOCATION` - location of cluster (e.g. us-west-2a)
+* `CALITP_SLACK_URL` - slack webhook api url, for posting errors to a channel.
```
atvaccaro, line 86: what channel does this point at?

atvaccaro (dismissed):

airflow/README.md
evansiroky: there's an environment variable in the composer prod environment with a slack webhook URL that points to the #data-infra-notify channel.

airflow/dags/dags.py
evansiroky: I was wondering that too, not sure why I forgot to comment. It probably should return at this point.

evansiroky:
@atvaccaro can you pick this up where @machow left off?

evansiroky (approved):

PR #877, opened 2021-12-20
fix(docs): correct gtfs ckan readme link

machow:
Quick fix for a broken link in our docs (that I added a while ago 😬)


holly-g (approved):
Thanks!

PR #878, opened 2021-12-20
Update agencies.yml with latest RT URL info

evansiroky:
This PR manually adds a bunch of RT URLs that were present in the transit database that didn't make their way into the agencies.yml file yet.

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -138,9 +138,9 @@ beach-cities-transit:
   agency_name: Beach Cities Transit
   feeds:
     - gtfs_schedule_url: https://www.redondo.org/civicax/filebank/blobdload.aspx?BlobID=33555
-      gtfs_rt_vehicle_positions_url: null
-      gtfs_rt_service_alerts_url: null
-      gtfs_rt_trip_updates_url: null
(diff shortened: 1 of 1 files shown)


airflow/data/agencies.yml
```diff
@@ -1394,17 +1394,17 @@ ventura-county-transportation-commission:
   agency_name: Ventura County Transportation Commission
   feeds:
     - gtfs_schedule_url: http://data.trilliumtransit.com/gtfs/vctc-ca-us/vctc-ca-us.zip
-      gtfs_rt_vehicle_positions_url: null
-      gtfs_rt_service_alerts_url: null
-      gtfs_rt_trip_updates_url: null
+      gtfs_rt_vehicle_positions_url: https://govcbus.com/gtfs-rt/vehiclepositions
+      gtfs_rt_service_alerts_url: https://govcbus.com/gtfs-rt/alerts
+      gtfs_rt_trip_updates_url: https://govcbus.com/gtfs-rt/tripupdates
```
hunterowens, line 1399: AFAIK, here might be worthwhile to see if we should use theGMV vs Trillium feeds (and / or diff em)

Nkdiaz: Was verified by Evan that the VCTC RT feeds were marked as the Trillium on the transit database

Nkdiaz (approved):

PR #879, opened 2021-12-22
feat(gtfs-rt) SQL script to have airflow view for metabase

Nkdiaz:
Created a new table in BigQuery Views to visualize gtfs_extraction_errors from the cloud gtfs_rt_logs.std_out 
Gtfs_rt_Extraction_errors was added to cal-itp-data-infra sandbox, Then linked up Metabase to show the GTFS RT Extraction Errors by Day in metabase for easy visualization.  Rather than counts we decided to keep the raw rows so people could drilldown from summaries in metabase. We tested out two tables in metabase, Count of all errors by day (GTFS RT || Feed Extraction Errors by Day), and Count of distinct errors by day (GTFS RT || Distinct Feed Extraction Errors by Day)
@evansiroky it is worth mentioning that the urls might include the api keys which are in the URL. Not sure how you want to handle that.

### DIFF
--- a/docs/datasets_and_tables/gtfs_rt.md
+++ b/docs/datasets_and_tables/gtfs_rt.md
@@ -47,11 +47,11 @@ broadly useful across the org should live in `views`.
 
 ### Extraction
 
-Extraction of GTFS RT feeds is handled by the [gtfs-rt-archive service](../services/gtfs-rt-archive.md).
+Extraction of GTFS RT feeds is handled by the [gtfs-rt-archive service](../services/gtfs-rt-archive.md).Within this service is some basic logging functionality that exports the logs to an external table 'std.out' that reads the data in this bucket.
 
 ### Logging
 
-All GTFS RT feed logs are loaded to Bigquery via a [Google Logger Router called `rt-extract-to-bigquery`](https://console.cloud.google.com/logs/router?project=cal-itp-data-infra). Note that this data is saved to its own dataset (`gtfs_rt_logs`), since routers do not allow table names to be customized.
+All GTFS RT feed logs are loaded directly to Bigquery via a [Google Logger Router called `rt-extract-to-bigquery`](https://console.cloud.google.com/logs/router?project=cal-itp-data-infra). Note that this data is saved to its own dataset (`gtfs_rt_logs`), since routers do not allow table names to be customized. Within `gtfs_rt_logs`, the data specific to URL feeds failing to download is stored in `stdout`.
 
 See [Google Cloud Log Router docs](https://cloud.google.com/logging/docs/routing/overview) for more information.
 The raw logs may be browsed in the [Logs Explorer](https://console.cloud.google.com/logs/query?project=cal-itp-data-infra).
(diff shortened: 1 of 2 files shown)


machow:
Tested out SQL in metabase w/ Natalya -- LGTM. @evansiroky the one piece to double check which she mentions is that the API keys will be in this data (not sure the level of exposure people are targeting; I know they're also in the airtable, so want to double check)

evansiroky (changes requested):
The underlying code looks fine, but it's unclear to me how the data pipeline is working here. Can this PR include a comment either in this task or somewhere else in the docs about how the data gets into that BigQuery `stdout` table? I don't see any other documentation or other DAG tasks that load data into here. Also, if the table is outdated, please update to the up-to-date table name.

machow:
@Nkdiaz do you mind adding an excerpt to the extraction section here, explaining that we are using google cloud logger to export to bigquery. Namely, we are using a sink to output to bigquery as a destination. The table is names e.g. stdout / stderr are not changeable, sort put on the gtfs_rt_logs dataset (double check that I got the name right). If anything is unclear be sure to consult the docs, in case I didn't quite get the story right ;).

docs/datasets_and_tables/gtfs_rt.md
```diff
@@ -46,7 +46,7 @@ broadly useful across the org should live in `views`.
 
 ### Extraction
 
-Extraction of GTFS RT feeds is handled by the [gtfs-rt-archive service](../services/gtfs-rt-archive.md).
+Extraction of GTFS RT feeds is handled by the [gtfs-rt-archive service](../services/gtfs-rt-archive.md). For Logs, we are using google cloud logger with a user-definited metric as a sink to output to BigQuery. The metric used is 'gtfs-rt-url-errors' and it filters for logName='stdout', namespace_name="gtfs-rt", and severity="INFO". Currently all the url-errors are treated as severity="INFO".The destination is bigquery and because the table names are not changable (gtfs_rt.stdout), they were put in their own dataset (gtfs_rt_logs).
```
evansiroky, line 49: I feel like there's a lot being explained here, but the context is missing. This might be a good place to make a screen share of how this is setup in the google cloud console. Also, I feel like this could be in a section separate from the `Extraction` section.

evansiroky (changes requested):
Still unclear from the docs how things are setup in the Google Cloud Console. Things may be better communicated in this context with a screen share.

machow:
@evansiroky, reviewed and updated with @Nkdiaz.

Let's have @Nkdiaz pair with @mjumbewu to review the documentation / logging process, before flagging you to re-review.

docs/datasets_and_tables/gtfs_rt.md
```diff
@@ -46,7 +47,14 @@ broadly useful across the org should live in `views`.
 
 ### Extraction
 
-Extraction of GTFS RT feeds is handled by the [gtfs-rt-archive service](../services/gtfs-rt-archive.md).
+Extraction of GTFS RT feeds is handled by the [gtfs-rt-archive service](../services/gtfs-rt-archive.md).Within this service is some basic logging functionality that exports the logs to an external table 'std.out' that reads the data in this bucket.
```
evansiroky, line 50: I don't believe this sufficiently explains how the the logs from the archiver service are automagically saved to Google Cloud and analyzeable within the logging dashboard. However, that seems out of scope for this PR.

evansiroky (approved):
The code looks good, so I'm going to approve and merge. However, there is one documentation issue that ought to be fixed in a separate effort.

PR #885, opened 2021-12-23
Update Clean Air express feed, add in RT feeds

o-ram:
Clean Air Express' link is now at https instead of http. I also added in some RT feeds for agencies that were missing, and matched up various 511 RT feeds with their 511 static feed.


evansiroky (approved):
Thanks for the updates. In the interest of finally merging this in so the pipeline begins working, I'm going to go ahead and approve. However, I did notice that SolanoExpress and Soltrans now have the same URL, so I'm kind of wondering if we still need to keep SolanoExpress.

Issue #661, opened 2021-11-05
add task level documentation in gusty


machow:
I know you can add DAG level documentation in METADATA.yml, but unsure how to add in the task. I know you add docstrings when creating tasks, but an argument is what gusty would need (or we'd to add it as a feature / argument operators)

Talking to @chriscardillo -- can we add a doc_md field to our task yml files?

machow:
If you click "Task Instance Details" , it will show a list of all attributes available (but I can't see a docs field, so we might need to roll support into gusty, or into our custom operators)

![image](https://user-images.githubusercontent.com/2574498/140533593-d351a547-5966-4896-bb96-7773e7ccd890.png)

machow:
I'm still so confused by the doc_md options, and wonder if it'd have to be handled in a custom way? Opened an issue on gusty here: https://github.com/chriscardillo/gusty/issues/36

machow:
Should be resolved once we upgrade to composer v2

PR #890, opened 2021-12-27
feat(airflow): add examples of doc_md to sandbox

machow:
This PR adds examples to two sandbox tasks, which show the use of airflow v2's doc_md argument. Resolves https://github.com/cal-itp/data-infra/issues/661.

This argument allows you to add documentation that is visible from the airflow UI. It is not the most prominent, but on the upside can be styled with lists, headers, etc..!

image

### DIFF
--- a/airflow/dags/sandbox/op_python_to_warehouse.py
+++ b/airflow/dags/sandbox/op_python_to_warehouse.py
@@ -4,6 +4,9 @@
 # fields:
 #   g: The g field python
 #   x: The x field python
+# doc_md: |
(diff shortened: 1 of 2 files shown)


machow:
Note that I've opened an issue on gusty, to explore another useful option: including all of the code from a .py task into the task instance UI. This would let a person quickly glance at the code for a task, without having to switch to github (similar to how we can see the SQL for .sql tasks :)

https://github.com/chriscardillo/gusty/issues/39

evansiroky (approved):

Issue #684, opened 2021-11-09
HTTP 413 error in jupyterhub


machow:
* payload too large
* it sounds like this is happening when an output is too large (e.g. a graph, or a bunch of graphs?)
* I think there's probably a piece of config to raise this limit

edit: it sounds like we may need to (also?) tweak something with our cluster setup: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1230#issuecomment-482318568

holly-g:
May relate to #634

lottspot:
@machow we have pushed the change we believe will help with this-- please have folks keep an eye out and report if this continues to be an issue!

edasmalchi:
Thanks, this seems better!

I just tried stress testing it with a notebook that generates some big ipyleaflet maps -- saving was still possible with the first 2 map outputs displayed but with 3+ maps the HTTP 413 error reappears.

Not sure if it's possible to further increase the limit?

holly-g:
@lottspot could we increase the limit?

lottspot:
I think we're perfectly ok with bumping this limit up to 1GB. #894 stands as a reference for how to accomplish this.

lottspot:
What we believe to be a related report has come in from @amandaha8 regarding this issue:

> Hi, just wanted to flag this in jupyterhub. My notebook is failing to save even though I only have 2 bar charts in my output. I have a pop up that says "file save error for notebook valid response 4313".
<img width="539" alt="filesave" src="https://user-images.githubusercontent.com/2199417/153407448-bac5999a-784d-495d-bbc2-fa9c46472b55.PNG">

Nkdiaz:
This was solved by linked pull request

Nkdiaz:
On 03/04/22 some of the analysists reported that they still see the 413 error.  @natam1. @edasmalchi, and @amandaha8 have all reported 413 errors when trying to save in jupyterhub in the past week. According to @atvaccaro , when the analysts download the WIP notebook, refresh the page, and upload the notebook again it is able to save again. This involved a token that is given on page load so reloading the page usually forces the re-auth and a new token. One potential way forward would be to enable auto-save so that you lose less work if you are unable to save manually.

atvaccaro:
[...]

Issue #634, opened 2021-11-01
Unreliable jupyterhub autosave


machow:
cc @tiffanychu90 

![image](https://user-images.githubusercontent.com/2574498/139699592-a51eaeff-98f5-4f33-b98e-a973f8d067b6.png)

hunterowens:
FWIW, @machow / @lottspot , given that is occurring only on @tiffanychu90 / @benjaminbressette  caltrans laptop and not on personal machines, we're pretty sure this is an issue w/ Chrome settings on those machines. Will have @tiffanychu90 confirm

tiffanychu90:
I'm not really sure anymore. The same issue popped up on my personal laptop too. But today, this issue disappeared and Jupyter Hub is functioning normally again.

hunterowens:
hrmm.

benjaminbressette:
Mine still isn't autosaving, on DOT laptop and my personal computer.

edasmalchi:
I also still run into saving trouble from time to time. 

My working hypothesis is that it has something to do with the (file)size of cell outputs in the browser, since it tends to happen when I've ran a notebook that includes cells with complex outputs, such as a detailed map or even a graphic of a detailed shapely geometry

This [notebook](https://github.com/cal-itp/data-analyses/blob/a326ab0ac9e0af3fb29e01a93aa9062d3343dd27/high_quality_transit_areas/rail_ferry_brt.ipynb) might be a good example: after running `ca_shape['geometry'].iloc[0]` in the 3rd cell from the bottom the notebook will no longer save, giving `File Save Error for rail_ferry_brt.ipynb Invalid response: 413 Request Entity Too Large`

Comment out and re-run that cell so there's no output and once again the notebook will save as normal...

themightychris:
@lottspot see https://github.com/nds-org/esiphub/issues/18#issuecomment-410916251

lottspot:
The feature referenced in part two of nds-org/esiphub#18 was removed in [94a86cd](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/commit/94a86cdf09e661e16543d6eb1cc24d51a8190c00) ([PR](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1539)). Will try to apply this change directly to the ingress annotation instead (as outlined in part 1 of the comment)

lottspot:
[...]

PR #894, opened 2021-12-28
jupyterhub/values: add proxy-body-size to ingress annotations

lottspot:
Resolves (so we believe):

- #634 
- #684


themightychris (approved):
64mb feels a tad unambitious :-) but it's what the community recommends so I'm fine with deploying and seeing what happens

Issue #900, opened 2021-12-30
Fix high memory usage in validation_fact_notice_fields task view


machow:
https://github.com/cal-itp/data-infra/blob/main/airflow/dags/gtfs_views/validation_fact_notice_fields.py

This task pulls all the fields out corresponding to a code that the validator could throw. Recently this task failed, and looking at metric explorer, it appears to be maxing out memory (> 12 GB).

It looks like it's opening all validation reports into memory before looping over them. If we move the opening into the loop it should resolve. Note also that recently the GTFS validator added an option to dump its code schema, which is a simpler way to get this information (e.g. we couldn't do it before, but it was fast to infer it from the data).

evansiroky:
Resolved with #901

PR #901, opened 2021-12-30
fix(gtfs-schedule): validation notice fields

machow:
addresses #900 

* fixes a high memory brought on by validation_fact_notice_fields loading everything into memory up front
* TODO: test locally


evansiroky:
Verified locally that a significant amount of garbage collection occurs that keeps the memory usage low.

evansiroky (approved):

Issue #787, opened 2021-11-24
Load daily GTFS RT validation feed error code counts


machow:
This will allow us to calculate https://github.com/cal-itp/data-infra/issues/570. Let's add code to gtfs-rt-validator-api (https://github.com/cal-itp/gtfs-rt-validator-api) that rolls up counts per feed per error. This will let us get metrics out eTl style, while we sort out compressing the data to parquet.

machow:
Here's one possible data model. Note that the boxes at the top are my attempt to hack comments into mermaid js 😬. Also dim_feeds data could either be included in dim_files, or added as a dimension to the fact table (would hold off for now, the issue of including with files is potentially size)

[...]

PR #902, opened 2021-12-30
Validate an hour of data across all feeds

machow:
Addresses #787

(Andrew)

Realizing I don't have the Airflow template here.

The pertinent tasks
image

I've tested and confirmed that this runs successfully on an hour of data in approximately 30 minutes across all feeds. I'd like to merge this in so we get at least some data going across the board. We can follow-up with the following.

1. Make validation data queryable via bigquery
2. Change how we sample, and/or validate more data
3. Point at the new validator image one https://github.com/cal-itp/gtfs-rt-validator-api/pull/5 is merged
4. Maybe most importantly, be aware of the validity periods of GTFS schedules; this shouldn't really be a problem right now, but does have implications for backfilling, especially

### DIFF
--- a/airflow/dags/rt_loader/create_validation_params.py
+++ b/airflow/dags/rt_loader/create_validation_params.py
@@ -6,7 +6,7 @@
 # external_dependencies:
 #   - gtfs_loader: calitp_feed_status
 # ---
-
+import pandas as pd
 from calitp import query_sql, save_to_gcfs
 from calitp.config import get_bucket
 
@@ -36,18 +36,24 @@ def main(execution_date, **kwargs):
     # This prefix limits the validation to only 1 hour of data currently
     prefix_path_rt = f"gtfs-data/rt/{date_string}T00:*"
 
-    params = raw_params.assign(
-        gtfs_schedule_path=lambda d: prefix_path_schedule
-        + "/"
-        + d.calitp_itp_id.astype(str)
-        + "_"
-        + d.calitp_url_number.astype(str),
-        gtfs_rt_glob_path=lambda d: prefix_path_rt
-        + "/"
-        + d.calitp_itp_id.astype(str)
-        + "/"
-        + d.calitp_url_number.astype(str)
-        + "/*",
+    raw_params["entity"] = [
+        ("service_alerts", "trip_updates", "vehicle_positions")
+    ] * len(raw_params)
+    params = raw_params.explode("entity").reset_index(drop=True)
+
+    params = pd.concat(
+        [
+            params,
+            params.apply(
+                lambda row: {
+                    "gtfs_schedule_path": f"{prefix_path_schedule}/{row.calitp_itp_id}_{row.calitp_url_number}",
+                    "gtfs_rt_glob_path": f"{prefix_path_rt}/{row.calitp_itp_id}/{row.calitp_url_number}/*{row.entity}*",
+                },
+                axis="columns",
+                result_type="expand",
+            ),
+        ],
+        axis="columns",
     )
 
     path = f"rt-processed/calitp_validation_params/{date_string}.csv"
--- a/airflow/dags/rt_loader/create_validation_params.py
+++ b/airflow/dags/rt_loader/create_validation_params.py
@@ -32,7 +32,9 @@ def main(execution_date, **kwargs):
     # Note that raw RT data is currently stored in the production bucket,
     # and not copied to the staging bucket
     prefix_path_schedule = f"{get_bucket()}/schedule/{execution_date}"
-    prefix_path_rt = f"gtfs-data/rt/{date_string}T00:00:*"
+
+    # This prefix limits the validation to only 1 hour of data currently
+    prefix_path_rt = f"gtfs-data/rt/{date_string}T00:*"
 
     params = raw_params.assign(
         gtfs_schedule_path=lambda d: prefix_path_schedule
@@ -48,8 +50,8 @@ def main(execution_date, **kwargs):
         + "/*",
     )
 
+    path = f"rt-processed/calitp_validation_params/{date_string}.csv"
+    print(f"saving {params.shape[0]} validation params to {path}")
     save_to_gcfs(
-        params.to_csv(index=False).encode(),
-        f"rt-processed/calitp_validation_params/{date_string}.csv",
-        use_pipe=True,
+        params.to_csv(index=False).encode(), path, use_pipe=True,
     )
--- a/airflow/dags/rt_loader/create_validation_params.py
+++ b/airflow/dags/rt_loader/create_validation_params.py
@@ -1,0 +1,55 @@
+# ---
+# python_callable: main
+# provide_context: true
+# dependencies:
+#   - external_calitp_validation_params
+# external_dependencies:
+#   - gtfs_loader: calitp_feed_status
+# ---
+
+from calitp import query_sql, save_to_gcfs
+from calitp.config import get_bucket
+
+
+def main(execution_date, **kwargs):
+    date_string = execution_date.to_date_string()
+
+    raw_params = query_sql(
+        f"""
+        SELECT
+            calitp_itp_id,
+            calitp_url_number,
+            calitp_extracted_at
+        FROM gtfs_schedule_history.calitp_feed_status
+        WHERE
+            is_extract_success
+            AND NOT is_parse_error
+            AND calitp_extracted_at = "{date_string}"
+        """,
+        as_df=True,
+    )
+
+    # Note that raw RT data is currently stored in the production bucket,
+    # and not copied to the staging bucket
+    prefix_path_schedule = f"{get_bucket()}/schedule/{execution_date}"
+    prefix_path_rt = f"gtfs-data/rt/{date_string}T00:00:*"
+
+    params = raw_params.assign(
+        gtfs_schedule_path=lambda d: prefix_path_schedule
+        + "/"
+        + d.calitp_itp_id.astype(str)
+        + "_"
+        + d.calitp_url_number.astype(str),
+        gtfs_rt_glob_path=lambda d: prefix_path_rt
+        + "/"
+        + d.calitp_itp_id.astype(str)
+        + "/"
+        + d.calitp_url_number.astype(str)
+        + "/*",
+    )
+
+    save_to_gcfs(
+        params.to_csv(index=False).encode(),
+        f"rt-processed/calitp_validation_params/{date_string}.csv",
+        use_pipe=True,
+    )
--- a/airflow/dags/rt_loader/external_calitp_validation_params.sql
+++ b/airflow/dags/rt_loader/external_calitp_validation_params.sql
@@ -1,0 +1,24 @@
+---
+operator: operators.SqlQueryOperator
+description: |
+  Parameters passed to GTFS RT validation.
+fields:
+  calitp_itp_id: Internal ITP ID.
+  calitp_url_number: Internal URL number.
+  gtfs_schedule_path: Path to GTFS Schedule data used in validation.
+  gtfs_rt_glob_path: |
+    Path to GTFS RT data used in validation. Wildcards (*) allow the path to grab
+    every individual RT file we store.
+---
+
+CREATE OR REPLACE EXTERNAL TABLE gtfs_rt.validation_service_alerts (
+    calitp_itp_id INT64,
+    calitp_url_number INT64,
+    calitp_extracted_at DATE,
+    gtfs_schedule_path STRING,
+    gtfs_rt_glob_path STRING
+)
+OPTIONS (
+    uris=["{{get_bucket()}}/rt-processed/calitp_validation_params/*"],
+    format="JSON"
+)
--- a/airflow/dags/rt_loader/external_calitp_validation_params.sql
+++ b/airflow/dags/rt_loader/external_calitp_validation_params.sql
@@ -11,7 +11,7 @@ fields:
     every individual RT file we store.
 ---
 
-CREATE OR REPLACE EXTERNAL TABLE gtfs_rt.validation_service_alerts (
+CREATE OR REPLACE EXTERNAL TABLE gtfs_rt.calitp_validation_params (
     calitp_itp_id INT64,
     calitp_url_number INT64,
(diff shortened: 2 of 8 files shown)


machow:
Note, I'm unable to validate MTC 511's feed (itp id 200), due to this validator system error

image

Looks like it's areas.txt is missing a required field?

machow:
To test on dev:

* copy any missing schedule data to staging (e.g. re-run this [copy-to-test-bucket data transfer job](https://console.cloud.google.com/transfer/cloud/jobs/transferJobs%2F14958109996250338090/runs?authuser=0&project=cal-itp-data-infra))
* Run gtfs_loader DAG, ensure calitp_feed_status runs (or copy the table over from prod)
* Turn on rt_loader, mark any wait_for tasks as successful, clear tasks if needed.

lauriemerrell:
Some immediate next steps, per meeting this morning with @machow and @mjumbewu, in addition to the items that Michael already noted:
* Investigate areas.txt errors -- during our call we experienced them with ITP ID 4 / URL 1 (in addition to MTC 511 errors noted above)
* Experiment with performance when validating 1 hour of data vs. 1 minute (right now 10 minutes to validate 1 minute of data for all agencies; see how that scales with data volume)
* Keep inspecting actual validation results to better understand what types of results we are seeing.

atvaccaro:
> * Experiment with performance when validating 1 hour of data vs. 1 minute (right now 10 minutes to validate 1 minute of data for all agencies; see how that scales with data volume)

Does the Java validator end up being the biggest performance issue?

lauriemerrell:
> Does the Java validator end up being the biggest performance issue?

I think that we are not entirely sure yet how long the validation takes if we try to make the rest of the pipeline as efficient as possible -- for example, as Michael notes above, right now we're initiating validation even if there's no data, so there's room to make the whole thing smarter and see what that gets us in efficiency gains. And validating RT data requires identifying the correct static (schedule) feed to use (the RT data exists within the context of the schedule data, which contains the actual information about the trips, stops, etc.); I think we only have to do that identification once, and if that's the part that's taking a while maybe the subsequent validation won't be too bad. 

That all being said, running the Java validator is kind of the core requirement of this side of the pipeline and we don't own the validator itself. So if the validator's performance is a real problem even after we try to make improvements, then we might need to start asking about whether we should change the scope of the data that we're validating (can we validate a subsample instead of literally all the data).

themightychris:
One of the ancillary commits in this PR, 7483f1a7af16f3145ce10fcbe7932e66fc010342, just got merged in on its own via #1014

I've deleted that commit from this PR and went ahead and rebased it on top of the latest `origin/main` along the way

lauriemerrell:
Currently this just grabs the most recently-extracted GTFS schedule feed to validate against. I'm wondering whether there are going to be potential issues with that -- these may just be overly-speculative edge cases but I want to note them just in case:
- We want the feed that was "in effect" on the date that the RT data was occurring. So for today Feb 3's RT data, we want the most recent GTFS schedule feed that covers Feb 3 service.
- Is there a risk that if we use the schedule feed extracted on Feb 3 that it represents service starting the following day? I.e., do we want to pull the feed from T-1 to validate data that happened on day T? (How immediately do we think service changes take effect?) 
- Do we have any instances where a feed is published that doesn't take effect for several days in the future? I don't think that's a best practice at all, but I am wondering if it's possible.

These questions are motivated by the fact that sometimes agencies might make "breaking" changes to their feeds -- for example I believe that VTA changed all their route IDs at some point in like 2020. It is not going to be common but I wonder if we need to handle this. (Maybe we just wait and see whether it becomes an issue?)

evansiroky:
> Currently this just grabs the most recently-extracted GTFS schedule feed to validate against. I'm wondering whether there are going to be potential issues with that -- these may just be overly-speculative edge cases but I want to note them just in case:
> 
> * We want the feed that was "in effect" on the date that the RT data was occurring. So for today Feb 3's RT data, we want the most recent GTFS schedule feed that covers Feb 3 service.
> * Is there a risk that if we use the schedule feed extracted on Feb 3 that it represents service starting the following day? I.e., do we want to pull the feed from T-1 to validate data that happened on day T? (How immediately do we think service changes take effect?)
> * Do we have any instances where a feed is published that doesn't take effect for several days in the future? I don't think that's a best practice at all, but I am wondering if it's possible.
> 
> These questions are motivated by the fact that sometimes agencies might make "breaking" changes to their feeds -- for example I believe that VTA changed all their route IDs at some point in like 2020. It is not going to be common but I wonder if we need to handle this. (Maybe we just wait and see whether it becomes an issue?)

These are good points. I think we should introduce a concept of using timestamped data for when a static GTFS feed was downloaded in the daily pipeline. Since we are downloading the RT data more frequently, we should do a switch-over of validating with whatever static feed we have available in our downloaded archive at that certain point in time. The daily downloads occur at midnight UTC each night. So for example, if we're validating Feb 3 RT data we would use the downloaded static data from Feb 2 up until midnight UTC at which point, we would switch to validating with the Feb 3 static data.

This strategy should set us up well for any future plans we may make to ingest the static GTFS data at even more frequent intervals and begins to follow more of a streaming pattern than a batch processing pattern.

atvaccaro:
So I just ran the validator on an hour of data, and it took 26 minutes instead of 8 minutes for 1 minute of data. So there's clearly a less-than-linear scaling here, which makes sense; though the validator itself does still seem to process the actual proto files one by one. I'm attaching a log for reading. Some gentle extrapolation still indicates we're probably looking at several hours to process 24 hours of data at minimum, which isn't the _worst_ but also not ideal. We may also start hitting issues trying to parallelize across feeds with that amount of data per feed.

I'd like to propose deploying this to run 1 hour of data across all feeds; as Michael's message above states, we can't successfully validate feed 200 due to an error so it's difficult to know the runtime of validating our largest feed over 24 hours.

Once we're comfortable doing an hour of validation per day, we can have a couple follow-ups around changing how we sample, increasing our data, and also deploying the parallelized validator. Does that sound like a reasonable approach?

evansiroky:
> So I just ran the validator on an hour of data, and it took 26 minutes instead of 8 minutes for 1 minute of data. So there's clearly a less-than-linear scaling here, which makes sense; 

In looking at the logs, I'm seeing most feeds taking about 1 minute to validate completely with the longest one of Long Beach being about 8 minutes. I am quite optimistic about extending out the time period of processing.

> though the validator itself does still seem to process the actual proto files one by one. 

I looked through the realtime validator code, and it seems that this is necessary in order to do certain types of validation.

> I'm attaching a log for reading. 

Thanks, this was super helpful.

> Some gentle extrapolation still indicates we're probably looking at several hours to process 24 hours of data at minimum, which isn't the _worst_ but also not ideal. We may also start hitting issues trying to parallelize across feeds with that amount of data per feed.

I think these validation results are important enough that if they take a really long time to process we should still try to do it. However, I think there are lots of opportunities for parallelization after looking at the logs and underlying validator code.

> I'd like to propose deploying this to run 1 hour of data across all feeds; as Michael's message above states, we can't successfully validate feed 200 due to an error so it's difficult to know the runtime of validating our largest feed over 24 hours.

What error is being seen with ITP ID 200? Can you let us know?
 
> [rt_loader_load_rt_validations_2022-02-08T00_00_00+00_00_4.log](https://github.com/cal-itp/data-infra/files/8044662/rt_loader_load_rt_validations_2022-02-08T00_00_00%2B00_00_4.log)
> 
> Once we're comfortable doing an hour of validation per day, we can have a couple follow-ups around changing how we sample, increasing our data, and also deploying the parallelized validator. Does that sound like a reasonable approach?

I am really hesitant about going the sampling path as the validation results at all times are so critical for realtime data quality analysis. I want to try to exhaust other options before deciding to go down that route. I think there are lots of opportunities for parallelization after looking at the logs and underlying validator code.

evansiroky (approved):
Just one comment about changing up a table name, but this is probably a good progress-check opportunity to merge this in.

atvaccaro:
> What error is being seen with ITP ID 200? Can you let us know?

I encountered the same error as Michael in https://github.com/cal-itp/data-infra/pull/902#issuecomment-1014772371

A missing `areas.txt`; am I correct in understanding that ID 200 == MTC 511?

airflow/dags/rt_loader/external_calitp_validation_params.sql
```diff
@@ -0,0 +1,24 @@
...
+fields:
+  calitp_itp_id: Internal ITP ID.
+  calitp_url_number: Internal URL number.
+  gtfs_schedule_path: Path to GTFS Schedule data used in validation.
+  gtfs_rt_glob_path: |
+    Path to GTFS RT data used in validation. Wildcards (*) allow the path to grab
+    every individual RT file we store.
+---
+
+CREATE OR REPLACE EXTERNAL TABLE gtfs_rt.validation_service_alerts (
```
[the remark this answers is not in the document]
atvaccaro: Oh thanks, totally missed this. I'll double-check the names for all the tasks.

atvaccaro:
Also realizing it's still pointed at CI buckets, will change.

airflow/dags/rt_loader/create_validation_params.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+    # Note that raw RT data is currently stored in the production bucket,
+    # and not copied to the staging bucket
+    prefix_path_schedule = f"{get_bucket()}/schedule/{execution_date}"
+
+    # This prefix limits the validation to only 1 hour of data currently
+    prefix_path_rt = f"gtfs-data/rt/{date_string}T00:*"
+
+    params = raw_params.assign(
```
evansiroky, line 39: These params need to be expanded to be on a per-feed, per-rt-feed-type basis, otherwise, the rt validator won't work properly.

The RT validator notes that it automatically sorts files based on file timestamp ([see docs on `-sort`](https://github.com/CUTR-at-USF/gtfs-realtime-validator/tree/master/gtfs-realtime-validator-lib#command-line-config-parameters)). This is important because some validation rules are dependent on analyzing the previous chronological file. In order to accomplish this though, there seems to be an assumption that the files being analyzed are all of the same type, ie they're all trip updates, all vehicle positions or all service alerts. So, this code currently jumbles all these file types together which will result in validations of different file types which may incorrectly trigger (or not trigger) a validation rule. From the logs posted earlier it seems like the sorting is already occurring correctly, so that's good.

So, please update these params to grab only one specific rt file type at a time in the glob path.

This should actually provide more opportunities for parallelization than before.

atvaccaro: Nice catch, thanks! I will rework these as well as the output formats. I think it may also be worth splitting up the validations into separate tasks like the loading, so  the DAG structure is a bit cleaner, too.

lauriemerrell:
> A missing areas.txt; am I correct in understanding that ID 200 == MTC 511?

Yes, ITP ID 200 = MTC 511. But as I [noted above](https://github.com/cal-itp/data-infra/pull/902#issuecomment-1015624911), we saw the same error with ITP ID = 4, URL = 1 as well (at the time I posted that, it was Capitol Corridor data being incorrectly tagged as AC Transit, so we should actually check ITP ID 56, URL 1 to see if that's still occurring.)

I almost wonder if those errors are going to be related to Evan's point about jumbled files, because `areas.txt` is an [optional, experimental file from flex v1](https://trilliumtransit.com/gtfs/reference/#areas) so I can't understand why a missing `areas.txt` file is a problem. AFAIK no flex information is included in GTFS RT.

Basically I'm wondering if the error itself is like a red herring that's covering for something else, because on its face it doesn't really make sense.

evansiroky:
> > A missing areas.txt; am I correct in understanding that ID 200 == MTC 511?
> 
> Yes, ITP ID 200 = MTC 511. But as I [noted above](https://github.com/cal-itp/data-infra/pull/902#issuecomment-1015624911), we saw the same error with ITP ID = 4, URL = 1 as well (at the time I posted that, it was Capitol Corridor data being incorrectly tagged as AC Transit, so we should actually check ITP ID 56, URL 1 to see if that's still occurring.)
> 
> I almost wonder if those errors are going to be related to Evan's point about jumbled files, because `areas.txt` is an [optional, experimental file from flex v1](https://trilliumtransit.com/gtfs/reference/#areas) so I can't understand why a missing `areas.txt` file is a problem. AFAIK no flex information is included in GTFS RT.
> 
> Basically I'm wondering if the error itself is like a red herring that's covering for something else, because on its face it doesn't really make sense.

That's unfortunate that the whole RT validator bails on a field missing within areas.txt. That file shouldn't even be needed to analyze the GTFS-RT with its current spec. Maybe we should remove that file from the static feed before validating. Or maybe better yet, @lauriemerrell, can you try to run [the latest docker image of the mobility-data/gtfs-realtime-validator project](https://github.com/mobilitydata/gtfs-realtime-validator/#setup) and see if you can reproduce the MTC 500 error? If you can, can you please submit an issue to that project?

atvaccaro:
I was able to at least work around this by deleting the areas.txt entirely; so I assume the validator won't allow missing fields but will allow the entire file missing.

Side note I cloned the validator repo and started poking around. The schemas for the individual entity files are hard-coded in Java classes so we have limited ability to control it, unfortunately, unless we fork and build our own jars.

evansiroky:
> I was able to at least work around this by deleting the areas.txt entirely; so I assume the validator won't allow missing fields but will allow the entire file missing.

👍 

> Side note I cloned the actual validator repo (i.e. the Java app) and started poking around. The schemas for the individual entity files are hard-coded in Java classes so we have limited ability to control it, unfortunately, unless we fork and build our own jars.

The maintainers of the realtime validator should be responsive enough that we should be able to collaborate and not have to go down the path of creating a fork of our own. We can get by with some features missing if the core of it works until a fixed release is published from them.

atvaccaro:
I just published a release https://github.com/cal-itp/gtfs-rt-validator-api/releases/tag/v0.0.3 that will allow the multiprocessing we want for this; I'm currently testing that new image with the changes suggested by @evansiroky.

airflow/dags/rt_loader/create_validation_params.py
atvaccaro: This is done

PR #905, opened 2021-12-30
Fix Dumbarton Express URLs

evansiroky:
The Dumbarton Express RT URLs had some erroneous characters that resulted in an invalid API key being created.


Nkdiaz (approved):
Verified that removing the g3 allows the api key to be generated correctly

PR #906, opened 2021-12-30
Update ridemvgo RT URLs

Nkdiaz:
Ridemvgo was one of the RT feeds giving an "error fetching url https://ridemvgo.net/gtfs-rt/alerts". After some research, I realized that ridemvgo is now using .org rather than .net domain name. I updated the yml file accordingly.


evansiroky (approved):
This matches what is in the transit database.

PR #907, opened 2021-12-30
Update OCTA URLs

evansiroky:
These won't work unless they're https.


machow (approved):
LGTM, I pasted one of the urls into my browser to make sure it returned something back.

PR #911, opened 2022-01-03
Update gtfs feeds now hosted by LACMTA

e-lo:
LACMTA is now in charge of maintaining and publishing feeds in LA County which are not currently hosted by the agencies themselves. They maintain these feeds at https://github.com/LACMTA/los-angeles-regional-gtfs.

This PR:
1.  Updates `agencies.yml` to point to the LACMTA hosted feed zip files (rather than Trilliums, which are deprecated) 
2. Adds two feeds not currently in `agencies.yml`: Cities of South Gate and Sierra Madre

- [x] Airtable Transit Database has been updated accordingly. 
- [x] ITP IDs for additional agencies are from original source and are not conflicting.

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -173,7 +173,7 @@ beaumont-pass-transit:
 bell-gardens:
   agency_name: Bell Gardens
   feeds:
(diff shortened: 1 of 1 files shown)


evansiroky (approved):
Looks great, thanks!

Issue #908, opened 2021-12-30
Need a way to test out downloading all RT files in agencies.yml


machow:
Currently, when we update links to RT feeds in agencies.yml, we need to do the following to evaluate the URLs:

* review a PR
* merge into main
* wait for the GTFS RT service to restart and pick it up

However, the simple act of downloading RT feeds from agencies.yml (or rather [agencies.filled.yml](https://github.com/cal-itp/data-infra/blob/main/.github/workflows/push_to_gcloud.yml#L39)) is pretty straightforward to implement in a few lines of python. 

It would be useful to have a way for users to test the links they are adding, or debug existing links (in addition to looking at logs).

## Proposal

* we add a workflow or job that can be initiated using github [workflow dispatch](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#workflow_dispatch)
  * by default the dispatch tries to download each URL in agencies.yml (after inserting API keys)
  * could also add an input to the dispatch to let people try only a specific agency, etc..
* we add a simple script in the `script` folder that takes the filled in agencies.yml and optional extra input as command line args, and attempts the downloads
  * it should be robust to failure, and report / summarize all failures at the end
* a natural place for this might be in push_to_gcloud.yml
  * let's be careful not to accidentally trigger other parts of `push_to_gcloud.yml`

lottspot:
The existing collector could easily be run locally to perform URL validation; it already logs URL retrieval errors to stdout and has the capability to write data to /dev/null (which is actually the default location when none other is provided).

The collector could actually be wired up as-is to a docker-compose project in order to satisfy this need

PR #912, opened 2022-01-03
Add PR and Issue templates for updating GTFS data

e-lo:
We may only want/need one of these, but I think they could be useful, especially for articulating a checklist.

### DIFF
--- a/.github/ISSUE_TEMPLATE/update_gtfs.md
+++ b/.github/ISSUE_TEMPLATE/update_gtfs.md
@@ -1,0 +1,25 @@
+---
+name: GTFS Data
+about: Update GTFS data in agencies.yml
+title: 'GTFS Data: '
+labels: ''
+assignees: '@evansiroky'
+---
+
+Update `agencies.yml` in order to....
+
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/update_gtfs.md
```diff
@@ -0,0 +1,25 @@
...
+
+Update the following gtfs datsets:  
+
+- ...
+- ...
+
+## Checklist
+
+- [ ] Made sure any new feeds have `itp_id`s that are not duplicative
+- [ ] Confirmed URIs are valid
```
evansiroky, line 24: Can you add a note that this should be done manually and that ideally it will also be part of an automated check as described in https://github.com/cal-itp/data-infra/issues/908? Same goes for the PR template.

evansiroky (changes requested):
This is a great idea and will be very useful to make sure URLs get into the pipeline. I just have one comment that references how some of the checking of the URLs could be automated in the future.

evansiroky (approved):

Issue #882, opened 2021-12-22
gtfs-rt-archive service needs to be able to make requests with headers


evansiroky:
**Is your feature request related to a problem? Please describe.**

Some RT feeds require authorization via adding header information. Currently there does not appear to be a way to add header data into requests for RT feeds.

**Describe the solution you'd like**

- The agencies.yml file needs to be modified to be able to include any request headers and their values.
- The gtfs-rt-archive script needs to read in the new agencies.yml data and prepare a request with the appropriate headers.

machow:
Hey--can you document the feeds you are seeing that require headers, and what they need to be successfully downloaded here? We can hand this off to the person implementing to validate it's working

evansiroky:
All Swiftly feeds require a header. See the [Swiftly API Docs here](https://swiftly-inc.stoplight.io/docs/realtime-standalone/b3A6Mjg0MzYyMTk-gtfs-rt-trip-updates). We haven't yet added the feed URLs and keys to the transit database or to agencies.yml.

machow:
Great--thanks. Where are we in the process of downloading swiftly feeds? Do we have swiftly API keys and urls? @holly-g can we lay out an issue / enhance this one to take this end-to-end, so we can make sure it's self-contained. Seems like we need...

* a list of feeds urls that will use the swiftly API / headers (or a pointer to a list)
* a working API key, url for verification (or key to test fetching the feed)
* to specify how to add to agencies.yml (e.g. could use a nested dictionary of `<header_type>: <header_value>` on the feeds entry)
* a written strategy for how we will verify it's working (e.g. check the bucket, etc..)

chriscauley:
Just a heads up, I have a fix for this but I don't have the ability to assign it to myself (I think I need to be added on the project). PR inbound.

We'll still need all the things from @machow's comment above to verify that it's working. I can confirm that the headers are being added to the request, but it would be nice to verify with a live data source.

lottspot:
It looks like this is merged and not yet deployed-- myself, Evan, and Chris C are meeting tomorrow to review deployment patterns so there should be an update after that discussion

evansiroky:
This has been resolved.

PR #913, opened 2022-01-03
Headers can now be optionally added to a feed

chriscauley:
Re: #882 

The PR had no specification for how to change the yml file so I decided to wing it. :upside_down_face: 

## YAML changes

* Any existing yaml file should work fine as is.
* Any headers specified will be the same for every url in the feeds section. If we want to make per request headers, I have ideas on how to make that, but I prefer to start with as few moving parts as possible. 
* To add headers, put a `headers` key at the same level as `feeds`.
* In the following example, the headers `Foo=asdf1234` and `Bar=asdf1234` will be appended to every request in the ac-transit feeds and the beach-city-transit feeds will have no headers.

``` yaml

ac-transit:
  agency_name: AC Transit
  feeds:
    - gtfs_schedule_url: https://api.actransit.org/transit/gtfs/download?token=[REDACTED]
      gtfs_rt_vehicle_positions_url: https://api.actransit.org/gtfsrt/vehicles?token=[REDACTED]
      gtfs_rt_service_alerts_url: https://api.actransit.org/gtfsrt/alerts?token=[REDACTED]
      gtfs_rt_trip_updates_url: https://api.actransit.org/gtfsrt/tripupdates?token=[REDACTED]
  headers:
    foo: asdf1234
    bar: asdf1234
  itp_id: 1
beach-cities-transit:
  agency_name: Beach Cities Transit
  feeds:
    - gtfs_schedule_url: https://www.redondo.org/civicax/filebank/blobdload.aspx?BlobID=[REDACTED]
      gtfs_rt_vehicle_positions_url: https://redondobeachbct.com/gtfs-rt/vehiclepositions
      gtfs_rt_service_alerts_url: https://redondobeachbct.com/gtfs-rt/alerts
      gtfs_rt_trip_updates_url: https://redondobeachbct.com/gtfs-rt/tripupdates
  itp_id: 2
```


lottspot (approved):
This LGTM, thanks for this nice addition! As the stake holder for this change, let's see what @evansiroky has to say before we merge & deploy it.

evansiroky (changes requested):
Thanks for the quick turnaround. Now that I'm looking at all this, I'm realizing it may be more complex than originally thought. I want to minimize how any changes to the `agencies.yml` affect the code within the [gtfs_downloader.generate_provider_list DAG task](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/gtfs_downloader/generate_provider_list.py).

Can you please take a look at that DAG task code and determine whether any outputs from that task would significantly change? If whatever changes we make here would result in significant changes to the DAG task, another option would be making another file altogether with the RT header information. This could have the benefit of being able to define the header information once for multiple RT feeds. This could be advantageous since our current use case will involve using the same header for multiple realtime feeds of multiple agencies. However, that would also have a cost of having information be stored in multiple files.

Perhaps some kind of work session is needed to figure out how to design this.

chriscauley:
@evansiroky and I talked over how to move the header information to a separate yml file. This last commit does that as well as validating the header yml file and applying secrets.csv via `script/agencies_convert.py` (although we may want to rename this file since it can work with any yml file, not just agencies.yml)

evansiroky:
> @evansiroky and I talked over how to move the header information to a separate yml file. This last commit does that as well as validating the header yml file and applying secrets.csv via `script/agencies_convert.py` (although we may want to rename this file since it can work with any yml file, not just agencies.yml)

This looks pretty great. However, now that I look at the headers.yml file I think we could make it be a list at the root. I went ahead and made this change and added the initial data in https://github.com/chriscauley/data-infra/pull/1. Please go ahead and make any changes as you see fit to that PR to your repo and then once you merge it in there, this PR will get updated.

Also, I'm on board with renaming `script/agencies_convert.py` to something else.

chriscauley:
I merged in @evansiroky's PR and renamed `agencies_convert.py` to `yml_convert.py`. It looks like there are some merge conflicts in the agencies list but otherwise I think this is good to go.

docs/services/gtfs-rt-archive.md
```diff
@@ -30,6 +31,30 @@ filesystem location of the downloaded key.
...
+Additional HTTP Headers can be provided in the yaml file specified in the `CALITP_HEADERS_YML` env variable. This file should be a list of
+objects specifying the `headers-data` (headers to be applied to each request) and the URLs (a map indicating which urls the headers apply to).
+If duplicate headers are specified for a url (matched via `f'{itp_id}/{url_number}/{rt_url}`), a value error will be thrown when parsing the
+header data file.
+
+Also, secret values can be substituted at build time using `{{ DOUBLE_BRACES }}` as in the example below. These secret values are built from
+airtable and are shared with the `CALITP_AGENCIES_YML` file.
+
+```yml
+headers:
```
evansiroky, line 45: This yml syntax is now outdated. Sorry I missed this.

.github/workflows/push_to_gcloud.yml
```diff
@@ -43,6 +43,17 @@ jobs:
...
+          './airflow/data/headers.filled.yml' \
+          './airflow/data/secrets.csv'
+    - name: Validate headers.yml
+      run: |
+        python script/headers_validate.py \
+          './airflow/data/headers.filled.yml' \
+          './airflow/data/schema_headers_entry.yml'
 
     - if: ${{ github.repository == 'cal-itp/data-infra' }}
       uses: google-github-actions/setup-gcloud@master
```
evansiroky, line 59: I can't comment on the particular line directly, but in the later step `Push agencies data to google cloud and prod services`, there is some logic that seems to set an environment variable in a kubernetes configuration. I believe we need to do the same thing for `headers.filled.yml` as was done for `agencies.filled.yml`.

evansiroky (changes requested):
This is looking great and it's really close to being able to be merged. I went ahead and fixed the merge conflict. However, I see two items that should be taken care of before merging.

1. I forgot to update the docs with the updated header.yml syntax.
2. The headers environment variable should be updated in a kubernetes config.

evansiroky (approved):
Awesome work. I've gone ahead and changed the base branch to a staging branch so that I can add in the LA Metro feeds prior to merging to main.

PR #939, opened 2022-01-11
chore: gitignore Mac .DS_Store files

lauriemerrell:
Adding [.DS_Store](https://en.wikipedia.org/wiki/.DS_Store) to the gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -136,3 +136,6 @@ dmypy.json
 # Credentials & Secrets
 /kubernetes/secrets
 /kubernetes/gke/kube
+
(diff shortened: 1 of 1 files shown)


charlie-costanzo (approved):
heck yeah LGTM *mac users rejoice*

lauriemerrell:
@evansiroky in future, would you be ok to skip review on very small items like this? I'd love to merge as soon as possible since it's an immediate quality of life boost for Mac users (who otherwise have to manually remove these files from commits)

PR #954, opened 2022-01-13
Update Code Owners- remove former code owners

Nkdiaz:
Updating the Code Owners now that Michael has moved on to a different position. It might be worth further discussion if we want to change the owners we currently have and/or add @lauriemerrell @chriscauley @charlie-costanzo as reviewers


lauriemerrell:
Would it be ok to add me for `/airflow/dags/gtfs*/*` and `/airflow/dags/rt*/*` ?

evansiroky (approved):

PR #967, opened 2022-01-20
Add San Joaquin Swiftly RT URLs

evansiroky:
This PR does 2 things:

- Adds the Swiftly RT URLs for San Joaquin Regional Transit District
- Refactor the headers file.
  - Add San Joaquin Regional Transit District
  - Use a single super Swiftly auth key. It appears that all the other transit agencies associated with our Caltrans Swiftly key have been given access via the Cal-ITP key. 
  - Order agencies by ITP ID

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -1111,7 +1111,7 @@ san-joaquin-regional-transit-district:
   feeds:
     - gtfs_schedule_url: http://sanjoaquinrtd.com/RTD-GTFS/RTD-GTFS.zip
       gtfs_rt_vehicle_positions_url: https://api.goswift.ly/real-time/san-joaquin/gtfs-rt-vehicle-positions
-      gtfs_rt_service_alerts_url: https://api.goswift.ly/real-time/san-joaquin/gtfs-rt-alerts
+      gtfs_rt_service_alerts_url: null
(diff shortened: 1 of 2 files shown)


airflow/data/agencies.yml
```diff
@@ -1110,9 +1110,9 @@ san-joaquin-regional-transit-district:
   agency_name: San Joaquin Regional Transit District
   feeds:
     - gtfs_schedule_url: http://sanjoaquinrtd.com/RTD-GTFS/RTD-GTFS.zip
-      gtfs_rt_vehicle_positions_url: null
-      gtfs_rt_service_alerts_url: null
-      gtfs_rt_trip_updates_url: null
+      gtfs_rt_vehicle_positions_url: https://api.goswift.ly/real-time/san-joaquin/gtfs-rt-vehicle-positions
+      gtfs_rt_service_alerts_url: https://api.goswift.ly/real-time/san-joaquin/gtfs-rt-alerts
```
o-ram, line 1114: SJRTD only uses Swiftly for VP and TU. According to their [website](https://sanjoaquinrtd.com/developer-portal/), this is their Alerts URL we should be validating https://sanjoaquinrtd.com/rideralerts/RTDrideralerts.xml

o-ram (changes requested):
The Alerts URL is different in this case (see my comment), but other than that change this looks good

o-ram (approved):

PR #969, opened 2022-01-21
Make update airflow requirements point to airflow v2 instance

evansiroky:
Something missed in https://github.com/cal-itp/data-infra/pull/851 was making sure the reinstallation of requirements was done in the airflow v2 instance instead of in the v1 instance. This fixes that.


evansiroky:
Used admin privileges to merge since our airflow v2 instance is dead in the water without updated requirements.

PR #970, opened 2022-01-21
Update requirements to trigger GitHub Action

evansiroky:
In order to properly install all the dependencies for #518, the requirements needed to be updated in the prod environment. However, #518 was merged before #969, so the v1 prod environment had its dependencies updated instead of v2. This PR contains a dummy commit that'll now trigger an update in the v2 prod environment.

A linting error is expected here which can be fixed later.


evansiroky:
Admin-merged because our prod airflow is broken without updated requirements.

Issue #973, opened 2022-01-21
Airtable view:  California Transit


lauriemerrell:
California Transit table from #774 

Update -- breaking into multiple PRs / subtasks:
- [x] #990 imports raw data to internal tables in warehouse (`airtable` dataset; initially imported data is **not** user-friendly but is technically accessible)
- [x] Confirm subsequent approach -- prioritization of table joins to support and normalization schema
- [x] Create at least a few initial views with enough normalization to support analysis

lauriemerrell:
Per conversation with @evansiroky -- 
* Will start by focusing on some of the "core" tables in the California Transit base:
   * Organizations
   * Services
   * GTFS service data
   * GTFS datasets 
* Will start by just running the Airtable operator and seeing how the "raw" outputs look in Metabase to scope out how much processing is needed. Specifically, wondering how much unpacking we'll need to do for many-to-one relationships. 
* Per #983, will be using a new Airtable dataset in BQ for this data, and will prefix each table with its base of origin.

lauriemerrell:
Notes: 
* A bunch of the raw column names contain characters (`#` or `=` specifically) that BigQuery does not allow, and that our current `to_snakecase` function does not strip out. I am addressing them manually using the `rename_fields` parameter for now, wanted to note since it's a bit of a gotcha -- the # signs in particular are interpreted as comments in YAMLs so have to escape with double quotes. 
* Before loading to the final warehouse location, do we want to strip out the foreign key columns from the current schemas and move those to mapping tables or something? (For example, drop the `mobility_services_managed` list-type column from `organizations` and instead create an `organization_service` mapping table that captures those relationships.)
  * [BQ docs](https://cloud.google.com/bigquery/docs/nested-repeated) say *not* to do this, but this data is not particularly huge and I wonder if having those kind of explicit mapping tables would be useful to people who will be consuming the data? Need to see what the nested/repeated columns look like in Metabase. 
* Some columns are metrics that are being calculated in Airtable (for example, number of complete RT feeds). Do we want to drop those columns and instead calculate them in BigQuery? (I assume that might be a future direction/longer term question, not sure what the long term vision is for the relationship between Airtable and the warehouse.)

evansiroky:
See my feedback inline:

[...]

PR #975, opened 2022-01-21
fix(airflow): rename airtable related items

lauriemerrell:
Ran into this while working on #973 (noticed that a bunch of places said "airflow" instead of "airtable"), figured we might as well address now, even though cleaning up the DAG task names in the metadata database will be a bit annoying. 

cc @evansiroky


evansiroky (changes requested):
Would you mind also fixing the linting error in this PR? I had to create this problem as a part of #970.

lauriemerrell:
@evansiroky done ✅

lauriemerrell:
@evansiroky per discussion - I successfully re-ran the `sandbox` DAG.

image

evansiroky (approved):

Issue #592, opened 2021-10-25
Set resource limits on jupyterlab helm chart config


machow:
https://zero-to-jupyterhub.readthedocs.io/en/latest/jupyterhub/customizing/user-resources.html?highlight=memory#set-user-memory-and-cpu-guarantees-limits

Let's set it to 5Gb for now? (cc @hunterowens what kind of memory usage will analysts need?)

Currently, the cluster has 12 Gb of memory, but it can always be upsized to meet need, etc..

TODO:

* @machow to investigate how much memory analysts need
* will try to implement the change with help from Chris C, since he's been prodding me to go through kubernetes stuff anyway
* let's also install https://github.com/jupyter-server/jupyter-resource-usage if it's easy to do along the way

lottspot:
Now that we have an operation prometheus & grafana stack, we should complete #1284 in order to gain visibility into the insights we need to advance this task in an informed way.

atvaccaro:
grabbing this to do alongside https://github.com/cal-itp/data-infra/issues/1323

PR #982, opened 2022-01-24
feat: adjust apps resource requests and node pool assignments

themightychris:
Adjust both JupyterHub and RT resource requests based on [Grafana charts](https://monitoring.k8s.calitp.jarv.us/d/967c0gU7k/jupyterhub-monitoring) and [discussion](https://github.com/cal-itp/data-infra/issues/1323#issuecomment-1090764303).
Require JupyterHub user pods to schedule on the proper node pool. We want to enable this so that as we add resource requests that may spread out pods, we don't accidentally spread to other node pools.
Bumps hub to use v10 of the jupyter image to get the resource usage library.

closes https://github.com/cal-itp/data-infra/issues/592
related to https://github.com/cal-itp/data-infra/issues/1323

### DIFF
--- a/kubernetes/apps/charts/jupyterhub/values.yaml
+++ b/kubernetes/apps/charts/jupyterhub/values.yaml
@@ -10,11 +10,13 @@ jupyterhub:
       name: ghcr.io/cal-itp/calitp-py
       tag: hub-v9
     memory:
+      # Much more than 10 and we risk bumping up against the actual capacity of e2-highmem-2
       limit: 10G
-      guarantee: 4G
+      # Very roughly I have seen most usage in the 2-3GB range
+      guarantee: 3G
     cpu:
       limit: 1
-      guarantee: 1
+      guarantee: 0.5
     storage:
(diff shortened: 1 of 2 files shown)


themightychris:
With resource requests and autoscaling configured on the jupyterhub-users nodepool, we probably want to turn this on. The potential downside would be that if jupyterhub couldn't schedule a user pod for some reason after the jupyterhub nodepool was maxed or became unavailable, users attempting to access jupyterhub would get a hard error. The alternative though is that they get scheduled to nodes that are responsible for other things and silently induce chaos. I'd rather we get the hard error and then have to go figure out what's blocking dedicated capacity then we not know when this is happening until things start crashing from OOM on other node pools

lottspot:
@atvaccaro no, we don't have any alerting infrastructure yet. Once we do, we will have a lot of questions we'll need to answer around the types of things we ought to be alerting on.

atvaccaro:
jk opening a new PR so chris can maybe approve

PR #991, opened 2022-01-26
fix(airflow): use v2 cluster in local dev envs

themightychris:
The original value is pointing at our airflow-v1 cluster which is due for retirement soon

I'm not positive this is the only place we need to make any change, as I'm not sure where people have been getting the service account auth file they use for this and whether that service account has the same access to the v2 cluster as it did to the v1 cluster

Before this get merged, could someone who already had a working local setup for running airflow jobs that run on k8s switch to this branch and restart your docker containers (e.g. `docker-compose down && docker-compose up`) and verify everything still works as expected

### DIFF
--- a/airflow/docker-compose.yaml
+++ b/airflow/docker-compose.yaml
@@ -65,7 +65,7 @@ x-airflow-common:
     GOOGLE_CLOUD_PROJECT: cal-itp-data-infra
 
     # Composer variables for kubernetes
-    POD_CLUSTER_NAME: "us-west2-calitp-airflow-pro-332827a9-gke"
+    POD_CLUSTER_NAME: "us-west2-calitp-airflow2-pr-171e4e47-gke"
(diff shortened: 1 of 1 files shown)


lauriemerrell:
@themightychris I tried running the `rt_loader` DAG (which contains the `load_rt_validations` PodOperator DAG task) but I got this error: `airflow.exceptions.AirflowNotFoundException: The conn_id `google_cloud_default` isn't defined`. I actually think I *have* run this DAG locally before (not realizing that was a pod operator) and did not get that error

lauriemerrell:
After manually pulling in the change from #1012, I am able to run `rt_loader.load_rt_validations` successfully locally. 
image

lauriemerrell (approved):

themightychris:
Verified that Laurie ran a pod in the new airflow v2 cluster:

Screen Shot 2022-01-31 at 12 03 58 PM

PR #1008, opened 2022-01-28
Create one PR template for whole project

evansiroky:
# Overall Description

This PR creates one big PR template that should show up as a template for all future PRs made to this repo. This closes #980.

## Checklist for all PRs

- [x] Run `pre-commit run --all-files` to make sure markdown/lint passes
- [x] Link this pull request to all issues that it will close using keywords (see GitHub docs about [Linking a pull request to an issue using a keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)). Also mention any issues that are partially addressed or are related.

PR #1011, opened 2022-01-30
MVP of just producing JSON from RT protos

atvaccaro:

This PR updates the `insert DAG name` DAG in order to....

Adds the following DAG tasks:
- parse_rt_service_alerts
- parse_rt_trip_updates
- external_trip_updates
- external_service_alerts

- parse_rt_vehicle_positions
- external_vehicle_positions

Pre-template:
- [x] Handle all three types of RT data
- [x] Write back compressed JSON instead of parquet
- [x] Create external tables on top of the resulting files
- [x] Ensure this runs performantly on the larger data feeds

This is a proposed alternative to #840 which just parses realtime data, serializes to compressed JSON, and writes back to GCS. Rather than transform the data into a relational ("rectangular") format, the key-value nature is preserved; this is important for the trip updates feed especially. The data _is_ flattened, so an input file produces one output row per entity.

An example output row:
```json
{
  "id": "731078",
  "tripUpdate": {
    "trip": {
      "tripId": "731078",
      "routeId": "10R"
    },
    "stopTimeUpdate": [
      {
        "stopSequence": 14,
        "departure": {
          "time": "1643309761"
        },
        "stopId": "881822"
      },
      {
        "stopSequence": 15,
        "departure": {
          "time": "1643309820"
        },
        "stopId": "883125"
      }
    ],
    "vehicle": {
      "id": "2131",
      "label": "0903",
      "licensePlate": ""
    },
    "timestamp": "1643309821"
  },
  "header": {
    "gtfsRealtimeVersion": "1.0",
    "incrementality": "FULL_DATASET",
    "timestamp": "1643309837"
  },
  "calitp_filepath": "/var/folders/ls/cdr1pq6n7yn9m0cb41zkflk80000gn/T/tmpdkccuh7b/2022-01-27T18:57:08/167/1/gtfs_rt_trip_updates_url",
  "calitp_itp_id": 167,
  "calitp_url_number": 1
}
```

Advantages:
1. Delaying transforms means we make fewer assumptions about how this data will be consumed
1. This is more generic across the different filetypes; we aren't enforcing a tabular structure
1. JSON is nice

Major downsides/caveats:
1. This will likely not be as performant to load as parquet, in both storage and loading
1. I haven't tested how quickly this loads into GCP

This PR also adds some additional logging, multithreading, streamed writes, etc.

One thing that may warrant investigation: I found is that the realtime data feeds use the full ISO timestamps as keys in the GCS bucket; I think this causes unnecessarily slow globbing due to the high cardinality. I would recommend splitting the key into date and time separately, so looking up all files from a particular day is trivial.


airflow/plugins/operators/proto_to_json_operator.py
```diff
@@ -0,0 +1,209 @@
...
+                # convert protobuff objects to DataFrames
+                parsed = parse_pb(feed_fname)
+
+                if parsed and "entity" in parsed:
+                    for record in parsed["entity"]:
+                        record["header"] = parsed["header"]
+                        record["calitp_filepath"] = str(parsed["calitp_filepath"])
+                        record["calitp_itp_id"] = int(feed[0])
+                        record["calitp_url_number"] = int(feed[1])
+                        gzipfile.write((json.dumps(record) + "\n").encode("utf-8"))
```
lauriemerrell, line 123: question: Did you run into issues with the encoding here at all? (Like, did you add this because you ran into an issue?) Just curious what we know about this.

lauriemerrell (changes requested):
A few mostly nitpicky inline comments, otherwise looks good to me. I agree with all your points and I think we probably want to switch to something more like this.

A few more general notes about how this PR fits in to the big picture:

IMO post-merge we should compare performance of this approach vs. the existing approach on a datatype where the existing approach does already work (vehicle positions or service alerts). I think we need to test separately:
* The copy operation (pb to parquet vs. JSON)
* The time to query the resulting external table (external table based on JSON vs. parquet)

Also just to record some notes from an ad hoc offline conversation between me, Andrew, and @machow about this:
* Michael agrees that this ELT approach might make more sense that the original #840 approach; can partition the external tables on extraction date
* Michael wonders if the IDs are going to be a problem, since they are arbitrary strings and may therefore take a lot of space
* Michael also shared [this page](https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-json) which says that loading compressed JSONs into BQ is slower than uncompressed. However, I am not sure if/how this applies to external tables? If this becomes an issue (which I think would be identified in my proposed experiment above), we have to consider the tradeoffs of storing uncompressed JSON (i.e., enormous storage in GCS) vs. whatever the load time delay is in BQ. In that case we may also want to revisit using parquet rather than compressed JSON.

.gitignore
```diff
@@ -1,6 +1,10 @@
 # Vim
 .*.sw[po]
 
+# PyCharm, etc.
+.idea/
+.DS_Store
```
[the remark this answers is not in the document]
atvaccaro: Oh nice, we probably just need to merge/rebase

airflow/plugins/operators/proto_to_json_operator.py
atvaccaro: No, the gzip writer just expects bytes

airflow/dags/rt_loader/parse_rt_trip_updates.yml
```diff
@@ -1,4 +1,4 @@
-operator: operators.RectangleRealtimeDataOperator
+operator: operators.RealtimeToFlattenedJSONOperator
```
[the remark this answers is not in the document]
atvaccaro: Yep, I will do that

airflow/plugins/operators/proto_to_json_operator.py
```diff
@@ -0,0 +1,209 @@
...
+    # https://github.com/fsspec/gcsfs/issues/379
+    with ThreadPoolExecutor(max_workers=4) as pool:
+        args = [
+            (i, feed, files, filename_prefix, iso_date, dst_path)
+            for i, (feed, files) in enumerate(feed_files.items())
+        ]
+        list(pool.map(handle_one_feed, *zip(*args)))
+
+
+class RealtimeToFlattenedJSONOperator(BaseOperator):
```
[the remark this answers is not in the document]
atvaccaro: Definitely, I change the name to reflect the flattening that I did, and never changed the `.py` file

airflow/plugins/operators/realtime_to_flattened_json_operator.py
```diff
@@ -0,0 +1,218 @@
...
+        ]
+        list(pool.map(handle_one_feed, *zip(*args)))
+
+
+class RealtimeToFlattenedJSONOperator(BaseOperator):
+    @apply_defaults
+    def __init__(
+        self,
+        header_details,
+        entity_details,
```
lauriemerrell, line 170: Should we remove these (`header_details` and `entity_details`) since we're no longer using them? Or do you anticipate adding something like that back in in the future?

airflow/plugins/operators/realtime_to_flattened_json_operator.py
```diff
@@ -0,0 +1,218 @@
...
+if __name__ == "__main__":
+    filename_prefix = "tu"
+    rt_file_substring = "trip_updates"
+    execute(
+        context={"execution_date": "2022-01-27T00:00:00Z"},
+        filename_prefix=filename_prefix,
+        rt_file_substring=rt_file_substring,
+        src_path=f"{get_bucket()}/rt/",
+        dst_path=(
+            f"{get_bucket()}/rt-processed_test_2022-01-27/" + rt_file_substring + "/"
```
lauriemerrell, line 216: We should remove all these constants before merging, right? Maybe ideally `__main__` would allow passing command-line arguments for testing...?

airflow/plugins/operators/realtime_to_flattened_json_operator.py
```diff
@@ -0,0 +1,218 @@
...
+class RealtimeToFlattenedJSONOperator(BaseOperator):
+    @apply_defaults
+    def __init__(
+        self,
+        header_details,
+        entity_details,
+        filename_prefix,
+        rt_file_substring,
+        cast={},
+        is_timestamp=[],
```
lauriemerrell, line 174: I think `cast` and `is_timestamp` are also unused in this version, can we remove them too?

airflow/plugins/operators/realtime_to_flattened_json_operator.py
```diff
@@ -0,0 +1,218 @@
...
+        self.header_details = header_details
+        self.entity_details = entity_details
+        self.filename_prefix = filename_prefix
+        self.rt_file_substring = rt_file_substring
+        self.cast = cast
+        self.time_float_cast = {col: "float" for col in is_timestamp}
+        self.src_path = f"{get_bucket()}/rt/"
+        self.dst_path = "".join(
+            [
+                f"{get_bucket()}/rt-processed_test_2022-01-24/",
```
lauriemerrell, line 191: We should remove the testing bucket path here

lauriemerrell (changes requested):
Overall looks good, thanks for jumping in on this so quickly, just a few outstanding things:

- [x] A few inline comments, mostly related to cleaning up testing constants. 
- [ ] Just to get in the habit, would probably be good to pull in the [new Airflow PR checklist](https://github.com/cal-itp/data-infra/pull/1008) and make sure everything there is covered (checklist will be added by default on all new PRs, but I think this snuck in under the wire before that merged.)
- [x] Specifically the new process asks that you upload a screenshot showing that you can successfully run any affected DAGs locally. Ideally would do that with all the final paths filled in so we can see the tables and files for all three data types in `staging` (both in GCS and in BQ), though I know we're still working out your GCP permissions for that. But basically would like to confirm that we can successfully run this on one full day's worth of all three data types. 
- [ ] We need to add a [sandbox](https://github.com/cal-itp/data-infra/tree/main/airflow/dags/sandbox) test for the new operator. See for example the [test for the Airtable operator](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/sandbox/op_airtable_to_warehouse.yml) that was recently added -- basically just want to have a small stub task that ensures that this operator is working as expected.

airflow/plugins/operators/realtime_to_flattened_json_operator.py
atvaccaro: yeah I think we can remove; it might at some point be nice to allow a dynamic key to flatten on (i.e. not `entity`) but should be fine for now

airflow/plugins/operators/realtime_to_flattened_json_operator.py
atvaccaro: yeah I'll make them arguments with defaults

airflow/plugins/operators/realtime_to_flattened_json_operator.py
atvaccaro: I think we need a conversation around this in general; IMO paths/keys should largely be fixed between production and testing/staging, but then the bucket should differ.

lauriemerrell:
@atvaccaro I also just realized -- do we need to run a backfill to get the existing vehicle positions parquet files into JSON if we change the external table definitions?

airflow/dags/rt_loader/external_service_alerts.sql
```diff
@@ -0,0 +1,9 @@
+---
+operator: operators.SqlQueryOperator
+---
+
+CREATE OR REPLACE EXTERNAL TABLE `gtfs_rt.service_alerts`
+OPTIONS (
+    format = "PARQUET",
+    uris = ["{{get_bucket()}}/rt-processed/service_alerts/*.parquet"]
```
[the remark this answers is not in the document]
atvaccaro: Yes, I think it will.

atvaccaro:
> @atvaccaro I also just realized -- do we need to run a backfill to get the existing vehicle positions parquet files into JSON if we change the external table definitions?

Does the raw data still exist? We should theoretically just be able to backfill via Airflow if we can re-process the historical raw data.

atvaccaro:
To answer a consideration brought up in the weekly data meeting, I think we need to test proto files vs compressed JSON vs parquet in terms of storage.

atvaccaro:
I've confirmed these run locally, however it's tough for me to execute the full tasks on the larger data sets since it's relying on my internet vs intra-GCP.

```
[2022-02-02 14:41:28,714] {logging_mixin.py:109} INFO - 2022-02-02 14:41.28 [info     ] writing 85970 lines (2.5 MB) from /tmp/tmp4qibr8uu/temporary.gz to gs://gtfs-data-test/rt-processed/vehicle_positions/rt_2022-02-01_99_0.gz dst_path=gs://gtfs-data-test/rt-processed/vehicle_positions/ feed=('99', '0') filename_prefix=rt i=71 iso_date=2022-02-01 len_files=3273 total_feeds=72
[2022-02-02 14:41:31,558] {logging_mixin.py:109} INFO - 2022-02-02 14:41.31 [info     ] took 22.788161 seconds to process 3273 files dst_path=gs://gtfs-data-test/rt-processed/vehicle_positions/ feed=('99', '0') filename_prefix=rt i=71 iso_date=2022-02-01 len_files=3273 total_feeds=72
```
image

I would like to merge once I get the sandbox testing figured out!

.pre-commit-config.yaml
```diff
@@ -3,6 +3,7 @@ repos:
     rev: v2.4.0
     hooks:
       - id: flake8
+        args: ["--ignore=E501,W503"]
```
lauriemerrell, line 6: Can we comment what these are / why we're ignoring?

lauriemerrell (approved):
I think we're ok to merge, though I am still a bit curious about the performance aspect. 

One nitpicky inline comment.

Re: backfilling vehicle positions
> Does the raw data still exist? We should theoretically just be able to backfill via Airflow if we can re-process the historical raw data.

Yes, the raw data is all still there....

Just to be totally explicit (maybe this is obvious) I think we should:
1.  Merge
2. Confirm that this runs ok on new, live data for a few days
3. Assess the storage sizes 
4. Only run a backfill (for all 3 data types) after we confirm that we like this approach and we're comfortable with storing everything as compressed JSONs for the medium term

lauriemerrell (approved):
Also I just realized that *technically* it would probably be best practice to include a screenshot showing that the sandbox DAG task runs locally as well.

atvaccaro:
> Also I just realized that _technically_ it would probably be best practice to include a screenshot showing that the sandbox DAG task runs locally as well.

Can do! I think I'm missing the Airtable API key so it failed, but the sandbox test succeeded.
image

.pre-commit-config.yaml
atvaccaro: Yes I will; I would like to centralize these things soon-ish in a pyproject.toml file.

lauriemerrell:
Thanks @atvaccaro  -- re: Airtable, I think current practice is just to focus on screenshot of specific tasks that are affected by the PR, so it's fine if the overall DAG fails for an unrelated reason.

evansiroky:
> Can do! I think I'm missing the Airtable API key so it failed, but the sandbox test succeeded.

I have privately communicated with @atvaccaro on how to get setup with the sandbox airtable.

atvaccaro:
I just executed the following query to create a legacy table referencing the parquet files. cc @edasmalchi 

```sql
CREATE OR REPLACE EXTERNAL TABLE `cal-itp-data-infra.gtfs_rt.vehicle_positions_legacy`
OPTIONS (
    format = "PARQUET",
    uris = ["gs://gtfs-data/rt-processed/vehicle_positions/*.parquet"]
);
```

PR #1012, opened 2022-01-31
chore(airflow): restore google cloud connection ID

lauriemerrell:
Re-add the Airflow Google Cloud Default connection ID to allow local running of PodOperator DAG tasks (specifically, to allow testing of #991).


lauriemerrell:
This was deleted here: https://github.com/cal-itp/data-infra/pull/518/files#diff-88902fdd045281cb61b751ed478492dca758abb6d7ccc0dd28f4793bbd93ac31L59; unclear why, but it broke local testing.

themightychris (approved):

PR #1014, opened 2022-01-31
fix(airflow/local): disable catchup by default

themightychris:
I'm brand new to Airflow development, and have no sense of best practices, but while I was trying to get up and running locally I found it really annoying in my venture to run DAGs that activating any DAG immediately starting spawning months worth of backfilling task instances.

Setting this as part of the local dev environment defaults seemed to make sense to me. This won't affect deployed environments at all since the `docker-compose.yaml` only defines the local dev environments. It seems to successfully prevent activated DAGs from immediately spawning tons of catchup tasks so you can, as a local dev, start with just spawning one task


lauriemerrell:
FWIW, Michael was planning to do this too :) https://github.com/cal-itp/data-infra/pull/902/files

evansiroky (approved):
Since @machow also did this in #902, that's good enough for me to give a thumbs up. It's be nice to have those extra comments in #902 though.

themightychris:
Alright I cherry-picked Michael's commit in favor of mine, and then I'll merge this and go rebase that other PR and remove that commit

PR #1016, opened 2022-01-31
Automate metabase backups

lottspot:
This PR packages a helm cron job to dump a postgresql database to a restic repository and adds values files and deployments to run the job within the metabase namespace.


evansiroky:
I'm delegating any review responsibility I have on this to @themightychris or @atvaccaro.

atvaccaro (approved):
I have no real opinions on this.

I am curious, why are we using a k8s-hosted postgres instead of GCP managed? Cost?

kubernetes/apps/charts/postgresql-backup/values.yaml
```diff
@@ -0,0 +1,9 @@
+cronjob:
+  schedule: "15 * * * *"
```
atvaccaro, line 2: Just for my own understanding: this will backup every hour and will keep the last 36 hourly, the last 7 daily, and the last 52 weekly?

lottspot: To tell you the truth I'm not entirely certain of how the retention settings intersect. @themightychris would know better (I c/p'd those values from him)

lottspot:
> I am curious, why are we using a k8s-hosted postgres instead of GCP managed? Cost?

We actually quite commonly run postgres databases in k8s clusters. It's not at all difficult to do for use cases which don't require a full primary/replica scale architecture, and also don't have a great deal of need for vertical scaling.

This metabase DB, for example, represents a lightly changed dataset about 5 MB total in size. Running a DB like this in an IaaS service, beyond being absurdly cost ineffective, is actually more complex than simply deploying and managing the database in the same packaging as the app itself, without being large enough in scale (both data size and txn load) to enjoy any mutually exclusive benefit.

kubernetes/apps/charts/postgresql-backup/values.yaml
themightychris: @atvaccaro correct, this `restic forget` command will keep the last 36 snapshots, whatever their timing (hourly + any manual runs we might do) and then at least 1 per day for the last 7 days and at least 1 per week for the last 52 weeks

themightychris (approved):

PR #1021, opened 2022-02-01
adding script for testing urls

### LINKED ISSUE #908: Need a way to test out downloading all RT files in agencies.yml

chriscauley:
# Overall Description

Fixes #908. This adds an action to the push_to_gcloud workflow that runs `git diff airflow/data/agencies.yml OLD_HASH NEW_HASH` to get a list of all urls that have been added to the agencies.yml file. It then loads the actual agencies.yml file and downloads any urls matching the domains found in the diff.

Update: As you can see from the lovely column of red X's in the commit history, I did eventually get the workflow to run successfully. I didn't modify any actual urls so it passed with 0/0 urls checked.

## Instructions

* Checkout this branch and run `python script/agencies_verify.py /path/to/agencies.yml /path/to/headers.yml` to verify all urls. Note that on my current list (taken from the google bucket yesterday) the api.actransit.org urls cause 3xHTTP404 errors.
* Now do `git diff agencies.yml f18ac09 HEAD > DIFF` to emulate what github does in the workflow changes.
* Run `python script/agencies_verify.py /path/to/agencies.yml /path/to/headers.yml DIFF` to only check the domains listed in the DIFF file.
* modify agencies.yml by changing the path of some url (note, not the domain, but the path) and run the previous command again. This time that URL should 404.
## Checklist for all PRs

- [X] Run `pre-commit run --all-files` to make sure markdown/lint passes

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -1,5 +1,5 @@
 ac-transit:
-  agency_name: AC Transit
+  agency_name: AC Transit (testing)
   feeds:
     - gtfs_schedule_url: https://api.actransit.org/transit/gtfs/download?token={{ AC_TRANSIT_API_KEY }}
       gtfs_rt_vehicle_positions_url: https://api.actransit.org/gtfsrt/vehicles?token={{ AC_TRANSIT_API_KEY }}
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -1,5 +1,5 @@
 ac-transit:
-  agency_name: AC Transit (testing)
+  agency_name: AC Transit
   feeds:
     - gtfs_schedule_url: https://api.actransit.org/transit/gtfs/download?token={{ AC_TRANSIT_API_KEY }}
       gtfs_rt_vehicle_positions_url: https://api.actransit.org/gtfsrt/vehicles?token={{ AC_TRANSIT_API_KEY }}
--- a/script/agencies_verify.py
+++ b/script/agencies_verify.py
@@ -1,0 +1,63 @@
+import logging
+import requests
+import sys
+import urllib
+import yaml
+
+sys.path.append("services/gtfs-rt-archive/src/gtfs_rt_archive/")
+
+from mapperfns import map_agencies_urls, map_headers  # noqa
+
+USAGE = """
+Usage:
+python script/agencies_verify.py agencies.yml headers.yml [DIFF]
+
+Arguments:
+agencies.yml - Yaml file containing agencies to download
+headers.yml - Yaml file mapping headers onto urls in agencies.yml
+DIFF - optional git diff between two hashes (will check all urls if not provided)
+"""
+
+
+def main():
+    logger = logging.getLogger("gtfs-rt-archive")
+    successes = []
+    fails = []
+    changed_domains = None
+
+    with open(sys.argv[1], "r") as f:
+        agencies = map_agencies_urls(logger, yaml.load(f, Loader=yaml.SafeLoader))
+    with open(sys.argv[2], "r") as f:
+        headers = dict(list(map_headers(logger, yaml.load(f, Loader=yaml.SafeLoader))))
+    if len(sys.argv) > 3:
+        with open(sys.argv[3], "r") as f:
+            lines = f.readlines()
+            lines = [line for line in lines if line.startswith("+") and "http" in line]
+            changed_urls = ["http" + line.strip().split("http")[-1] for line in lines]
+            changed_domains = [
+                urllib.parse.urlparse(url).netloc for url in changed_urls
+            ]
+
+    for key, url in list(agencies):
+        domain = urllib.parse.urlparse(url).netloc
+        if changed_domains is not None and domain not in changed_domains:
+            continue
+        try:
+            result = requests.get(url, headers=headers.get(key, {}))
+            result.raise_for_status()
+        except requests.HTTPError:
+            print(f"HTTP{result.status_code}: {url}")
+            fails.append(url)
+            continue
+        successes.append(url)
+    print(f"{len(successes)}/{len(successes+fails)} urls successfully downloaded")
+    if fails:
+        print("Exiting with error because some urls failed to download")
+        exit(1)
+
+
+if __name__ == "__main__":
+    if len(sys.argv) < 3:
+        print(USAGE)
+        exit()
+    main()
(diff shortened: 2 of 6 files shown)


script/agencies_verify.py
```diff
@@ -0,0 +1,63 @@
...
+        agencies = map_agencies_urls(logger, yaml.load(f, Loader=yaml.SafeLoader))
+    with open(sys.argv[2], "r") as f:
+        headers = dict(list(map_headers(logger, yaml.load(f, Loader=yaml.SafeLoader))))
+    if len(sys.argv) > 3:
+        with open(sys.argv[3], "r") as f:
+            lines = f.readlines()
+            lines = [line for line in lines if line.startswith("+") and "http" in line]
+            changed_urls = ["http" + line.strip().split("http")[-1] for line in lines]
+            changed_domains = [
+                urllib.parse.urlparse(url).netloc for url in changed_urls
```
evansiroky, line 38: This checking of domains is problematic. If there is ever a change to a Swiftly URL or a 511 URL, that will result in all URLs on that domain being matched and it will trigger the checking of all the Swiftly and 511 URLs and not just those that changed. I think it would be best to not compile a list of domains and instead focus on just the exact URLs that get stored into the `changed_urls` variable.

script/agencies_verify.py
```diff
@@ -0,0 +1,63 @@
...
+    print(f"{len(successes)}/{len(successes+fails)} urls successfully downloaded")
+    if fails:
+        print("Exiting with error because some urls failed to download")
+        exit(1)
+
+
+if __name__ == "__main__":
+    if len(sys.argv) < 3:
+        print(USAGE)
+        exit()
```
evansiroky, line 62: Can this exit with a non-zero status to indicate a failure?

evansiroky (changes requested):
This looks very promising! I have a few comments that should be addressed before this is ready to be merged.

script/agencies_verify.py
chriscauley: Resolved in 2914c16

script/agencies_verify.py
```diff
@@ -0,0 +1,63 @@
...
+            ]
+
+    for key, url in list(agencies):
+        domain = urllib.parse.urlparse(url).netloc
+        if changed_domains is not None and domain not in changed_domains:
+            continue
+        try:
+            result = requests.get(url, headers=headers.get(key, {}))
+            result.raise_for_status()
+        except requests.HTTPError:
```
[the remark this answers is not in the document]
chriscauley: Resolved in 47e091b

script/agencies_verify.py
```diff
@@ -0,0 +1,63 @@
+import logging
+import requests
```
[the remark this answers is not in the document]
chriscauley: Resolved in 1b04e7e (the workflow wasn't using the requirements file so I fixed that as well)

script/agencies_verify.py
chriscauley: Resolved in 34b6287.

I used changed_domains the first time because the AGENCIES_DIFF file is between versions of `airflow/data/agencies.yml` which doesn't have secret keys in them. The script that replaces the secret keys (script/yml_convert.py) wasn't set up to work with non-yaml files. In my most recent I gave the yml_convert script an `--unsafe` option, so using changed_urls should work now.

script/agencies_verify.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def main():
+    logger = logging.getLogger("gtfs-rt-archive")
+    successes = []
+    fails = []
+    changed_urls = None
+
+    with open(sys.argv[1], "r") as f:
+        agencies = map_agencies_urls(logger, yaml.load(f, Loader=yaml.SafeLoader))
```
evansiroky, line 28: This method will only result in realtime URLs being returned. We also need to check the static URLs.

evansiroky (changes requested):
Sorry I didn't catch this the first time around, but this check needs to work for both the static and realtime URLs.

evansiroky:
I think this is really close. It seems to be just a matter of making a quick edit to somehow change this line of code so that the string "gtfs_rt" can be configurable: https://github.com/cal-itp/data-infra/blob/772fbedbb332251d5ca500d3038b4a472f4a51bc/services/gtfs-rt-archive/src/gtfs_rt_archive/mapperfns.py#L46

script/agencies_verify.py
chriscauley: Good catch. I fixed this in a677fd3 by adding a "key_prefix" kwarg to the mapper function which defaults to the rt value (since it's part of the rt-archiver code). I verified it locally that it now catches broken schedule urls.

evansiroky (approved):

airflow/data/agencies.yml
```diff
@@ -1,5 +1,5 @@
 ac-transit:
-  agency_name: AC Transit
+  agency_name: AC Transit (testing)
```
evansiroky, line 2: Can this be reverted back to the original?

evansiroky (changes requested):
Whoops, sorry, missed one thing.

airflow/data/agencies.yml
chriscauley: Thank you for catching that and sorry for leaving it in. I had to change the file somehow to test the git hook.

evansiroky (approved):

Issue #653, opened 2021-11-03
Move gtfs_loader.gtfs_schedule_tables_load into gtfs_schedule_history2


machow:
This task requires gtfs_schedule_history_load to have run across all time. This is a natural fit for gtfs_schedule_history2, since it also requires all past successes for that task. The one difference is that gtfs_schedule_tables_load can be `latest_only`, since it loads the most up-to-date gtfs schedule data for the day from the external tables (for convenience).

Setting latest_only: true in the yaml should work. You can check this by opening instance details.

lauriemerrell:
This was made obsolete by #1157

Issue #914, opened 2022-01-04
Bug: SacRT agency shows no service for several months despite being active


lauriemerrell:
**Describe the bug**
The [October](https://reports.calitp.org/gtfs_schedule/2021/10/273/index.html) and [November](https://reports.calitp.org/gtfs_schedule/2021/11/273/index.html) published reports and the [draft December report](https://development-build--cal-itp-reports.netlify.app/gtfs_schedule/2021/12/273/index.html) for SacRT show 0 service activity. Service hours for all days for all three months are 0. 

However this agency was active during these months. 

Not sure if this is an error with the data infra side (is the data being pulled) or the report.... Submitting on infra side for now?

**To Reproduce**
1. Go to any of the monthly reports linked above
2. Scroll down to the service hours chart
3. No active hours

**Expected behavior**
Should show some positive active hours for this agency.

Nkdiaz:
It was previously reported that Yolo County Transportation District (itp_id 372) is also not showing up in gtfs_schedule_fact_daily_service. I can verify that Yolo County Transportation is also not included in the list of agencies that we generate reports for October and November.

Nkdiaz:
Writing down my research journey, 
- Verified that Sacrometo RT does not have service hours within the monthly metrics, not only for the past several months but it appears that we have never generated that table with service hours. This table is calculated from `tbl_views.gtfs_schedule_fact_daily_service`
- in BigQuery, I verified that calitp-id 273 (Sacramento RT) does not appear in `tbl_daily_service_hours` nor in the `tbl.views.gtfs.schedule_fact_daily_service` but it does appear in the `gtfs_schedule_service_daily` with the values is in service marked as true
- further diving into `gtfs_schedule_fact_daily_service, which gets its data involving service hours from gtfs_schedule_fact_daily_trips. calitp_id 273 is not in gtfs_schedule_fact_daily_trips which is dependent on gtfs_schedule_stg_daily_service and gtfs_schedule_dim_feeds. I verified that Sacramento was in both the stg_daily_service and dim_feeds.
- Now that I had the problem mapped to the gtfs_schedule_fact_daily_trips, I dug deeper into the sql query. stg_daily_service subtable is able to be generated with calitp 273 but daily_service_trips is not able to be generated with calitp 273,
- it appears that the daily_service_trips is unable to join with the stg_daily_service_keyed due to its service_id. when you remove it you are able to calculate all the other variables. 
When you remove the  service id from the following join

[...]

PR #1022, opened 2022-02-01
fix(gtfs-schedule): trim whitespace before view creation

### LINKED ISSUE #946: Trim whitespace in GTFS schedule data
Spinning out from #914 after discussion in #gtfs-quality Slack channel. 

We decided we *should* trim whitespace from GTFS schedule data in our ingestion pipeline. 

Should be trimmed after validation is run to preserve whitespace validation notices (for posterity?)

At time of writing, not sure of best point in ingestion at which to do this.

lauriemerrell:
# Overall Description

This PR resolves #946 by trimming whitespace at the `gtfs_views_staging` step in the GTFS pipeline, thereby addressing #914. 🚨 **Reviewers please do not merge this without talking to me, I would like to make sure that we receive multiple reviews to confirm this approach since it has side effects.** 

We need to do *some* trimming this week so that SacRT will have service information in the January reports. If people do not agree with the approach outlined here, we can make a separate PR to just make a targeted fix for `service_id` for SacRT and postpone the larger decision. 

Please see extensive notes about considerations for how to handle the whitespace issue in #946. In the end, I chose the approach I did (trimming in SQL) based on the arguments put forth by @machow:

>   If it seems like a legit loading issue (e.g. "we are parsing csvs wrong"), then I could see changing the pre-processing task. But in the case of even slight ambiguity I would say fixing belongs in SQL transformation, because whether it's an error is a business decision.
    In other words, the validator can change how it handles this, and guidelines could change. But if pre-processed out..
        it would be annoying to migrate raw data load to reflect these business realities, when we have a staging layer for that.
        it would be difficult to understand the business impact of the decision (e.g. answering how often does this occur?)

However, I want to make sure we are all in alignment on this based on the considerations outlined below.

## Additional notes

Numbered for ease of reference.

1. ✅ `views.gtfs_schedule_fact_daily_trips` does now show trips for SacRT (the original issue we wanted to address); this is visible in `staging` now.
2. ✅ I confirmed that I didn't accidentally drop any columns in my enumeration, using a query on `INFORMATION_SCHEMA.COLUMNS`. 
3. ❓ I checked how many tables/columns have leading/trailing whitespace, to see whether we should just trim ID columns or whether we should trim all string columns. I confirmed that there are leading / trailing whitespaces in dozens of columns and across most of the tables that we load (it's not isolated to just one or two locations). It seemed better to trim across the board based on Chris and Elizabeth's feedback that preserving these spaces isn't appropriate anyway. However, see subsequent warnings. 
4. ⚠️ This trimming only occurs right before we create the `views`. This means that we are not doing any trimming on files (like `shapes`) that we do not yet load into `views`. **This means that joins between `views` and other data, like `shapes` would not be fully supported; `shape_id` in the `trips` views would be trimmed, but `shape_id` in `gtfs_schedule_type2.shapes` would not be trimmed.** This would be a new, stricter semantic boundary between `views` and earlier data. 
    - ⚠️ In particular, this would mean that the data in `GTFS Schedule Feeds Latest` in Metabase (which corresponds to the `gtfs_schedule` dataset in BigQuery) would no longer align with the `views` data and joins between the two would not be supported. **The `GTFS Schedule Feeds Latest` data would be untrimmed and subject to joining problems like the SacRT issue.**
5. ⚠️ Trimming requires that we explicitly enumerate columns. This means that we now have to keep `gtfs_views_staging` in sync with `gtfs_schedule_history` where we define what fields we load from the raw files. I don't think that this is a huge problem (I don't think we anticipate that these fields will change often, and any failure to keep them in sync will be obvious because the new field won't appear in `views` unless it's added here). 
6. ⛔  We effectively cannot handle quoted vs. unquoted fields differently. I really do not think it's practical in our pipeline, no matter what stage we trim at. So this step will trim the whitespace from `"<space>7"` and `<space>7` alike. 

## Alternatives considered

The raw GTFS data is handled as follows:
1. Downloaded in [`gtfs_downloader`](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/gtfs_downloader/download_data.py) -- we clearly do not want to do any trimming on the actual raw data
2. ⭐ Copied to `gtfs-data/schedule/processed` in [`gtfs_loader/gtfs_schedule_history_load.py`](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/gtfs_loader/gtfs_schedule_history_load.py) -- **trimming whitespace as part of this operation is the best/main alternative to the current approach**
   - Trimming here would trim ALL data before it is ingested into BigQuery at all, but not before the totally raw data is validated
   - This would trim tables including `shapes` etc. that are not (yet) loaded as views -- this would remove consideration 4 above
   - We already explicitly handle columns here, so this would mitigate consideration 5 above
   - Trimming here would mean that we do not have access to the fully raw data in BQ, so it would be more difficult to assess how prevalent leading/trailing whitespace is if we ever revisit this question / if there's a situation where analysts need the raw data, it would essentially not be accessible to them
   - Not sure of potential performance impacts, because this would add an extra Pandas operation on every table being loaded (doubt it's a big deal, but wanted to note just in case)
   - I think (?) this would require explicitly re-running Airflow history to pick up the changes, current approach does it automatically
3. Loaded as external tables in [`gtfs_schedule_history`](https://github.com/cal-itp/data-infra/tree/main/airflow/dags/gtfs_schedule_history) - from what I can tell, there is not a way to trim on external table creation
4. Updates are merged into existing data in [`gtfs_schedule_history2/merge_updates.py`](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/gtfs_schedule_history2/merge_updates.py) -- this step is already pretty complicated and right now it's table-agnostic; trimming requires us to enumerate columns so don't think this is appropriate
5. `gtfs_views_staging` creates the `_clean` tables -- this is the approach in this PR
6. `views` creates the final views; trimming at that point is basically just a worse, more complicated version of the approach in this PR

## Remaining TODOs

- [ ] Perform spot checks on trimmed data to confirm no ill effects
- [x] Double check all references from `views` to the `schedule_type2` data to ensure that we don't reference any untrimmed data
- [x] Fill out reviews requested to request targeted reviews from different stakeholders

## Reviews requested
@themightychris / @e-lo / @mjumbewu / @machow / @evansiroky -- The folks who weighed in on the general CSV interpretation debate. Please let me know what you think.
@edasmalchi -- Hoping you can weigh in on consideration 4 above. Would those limitations be blocking for you? If you think current approach is viable, what documentation is needed to make these caveats clear to analysts? (cc @charlie-costanzo on the docs question)
@Nkdiaz / @atvaccaro -- General code review.

## Checklist for all PRs

**gtfs_views_staging**: 
image

**gtfs_views**: (stop times fails because of query limit, I believe that's unrelated)
image

This PR updates the `gtfs_views_staging` DAG in order to trim whitespace from `*_clean` tables.

- `<table>_clean.sql`

### DIFF
--- a/airflow/dags/gtfs_views_staging/validation_notices_clean.sql
+++ b/airflow/dags/gtfs_views_staging/validation_notices_clean.sql
@@ -5,7 +5,55 @@ dependencies:
   - type2_loaded
 ---
 
+-- Must trim string fields that come from raw GTFS tables that we clean & load into views
+-- (To allow joining with the cleaned data after this is run)
+-- Don't trim stopId because we don't load stops into views
+-- This table has over 70 columns, so even though EXCEPT is a bit messy it still seems cleaner
+
 SELECT
-    * EXCEPT(calitp_deleted_at)
+    * EXCEPT(
+      calitp_deleted_at
+      , fareId
+      , previousFareId
+      , routeId
+      , currentDate
+      , feedEndDate
+      , routeColor
+      , routeTextColor
+      , tripId
+      , tripIdA
+      , tripIdB
+      , routeShortName
+      , routeLongName
+      , routeDesc
+      , stopId
+      , stopName
+      , serviceIdA
+      , serviceIdB
+      , departureTime
+      , arrivalTime
+      , parentStation
+      , parentStopName)
     , COALESCE(calitp_deleted_at, "2099-01-01") AS calitp_deleted_at
+    , TRIM(fareId) as fareId
+    , TRIM(previousFareId) as previousFareId
+    , TRIM(routeId) as routeId
+    , TRIM(currentDate) as currentDate
+    , TRIM(feedEndDate) as feedEndDate
+    , TRIM(routeColor) as routeColor
+    , TRIM(routeTextColor) as routeTextColor
+    , TRIM(tripId) as tripId
+    , TRIM(tripIdA) as tripIdA
+    , TRIM(tripIdB) as tripIdB
+    , TRIM(routeShortName) as routeShortName
+    , TRIM(routeLongName) as routeLongName
+    , TRIM(routeDesc) as routeDesc
+    , TRIM(stopId) as stopId
+    , TRIM(stopName) as stopName
+    , TRIM(serviceIdA) as serviceIdA
+    , TRIM(serviceIdB) as serviceIdB
+    , TRIM(departureTime) as departureTime
+    , TRIM(arrivalTime) as arrivalTime
+    , TRIM(parentStation) as parentStation
+    , TRIM(parentStopName) as parentStopName
 FROM `gtfs_schedule_type2.validation_notices`
--- a/airflow/dags/gtfs_views_staging/agency_clean.sql
+++ b/airflow/dags/gtfs_views_staging/agency_clean.sql
@@ -5,8 +5,22 @@ dependencies:
   - type2_loaded
 ---
 
+-- Trim all string fields
+-- Incoming schema explicitly defined in gtfs_schedule_history external table definition
+
 SELECT
-    * EXCEPT(calitp_deleted_at)
+    calitp_itp_id
+    , calitp_url_number
+    , TRIM(agency_id) as agency_id
+    , TRIM(agency_name) as agency_name
+    , TRIM(agency_url) as agency_url
+    , TRIM(agency_timezone) as agency_timezone
+    , TRIM(agency_lang) as agency_lang
+    , TRIM(agency_phone) as agency_phone
+    , TRIM(agency_fare_url) as agency_fare_url
+    , TRIM(agency_email) as agency_email
+    , calitp_extracted_at
+    , calitp_hash
     , FARM_FINGERPRINT(CONCAT(CAST(calitp_hash AS STRING), "___", CAST(calitp_extracted_at AS STRING))) AS agency_key
     , COALESCE(calitp_deleted_at, "2099-01-01") AS calitp_deleted_at
 FROM `gtfs_schedule_type2.agency`
--- a/airflow/dags/gtfs_views_staging/calendar_clean.sql
+++ b/airflow/dags/gtfs_views_staging/calendar_clean.sql
@@ -5,11 +5,25 @@ dependencies:
   - type2_loaded
 ---
 
+-- Trim all string fields
+-- Incoming schema explicitly defined in gtfs_schedule_history external table definition
+
 SELECT
-    * EXCEPT(start_date, end_date, calitp_deleted_at),
-    PARSE_DATE("%Y%m%d",start_date) AS start_date,
-    PARSE_DATE("%Y%m%d",end_date) AS end_date,
-    FARM_FINGERPRINT(CONCAT(CAST(calitp_hash AS STRING), "___", CAST(calitp_extracted_at AS STRING)))
-        AS calendar_key,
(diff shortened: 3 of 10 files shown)


airflow/dags/gtfs_views_staging/validation_notices_clean.sql
```diff
@@ -5,7 +5,55 @@ dependencies:
...
 ---
 
+-- Must trim string fields that come from raw GTFS tables that we clean & load into views
+-- (To allow joining with the cleaned data after this is run)
+-- Don't trim stopId because we don't load stops into views
+-- This table has over 70 columns, so even though EXCEPT is a bit messy it still seems cleaner
+
 SELECT
-    * EXCEPT(calitp_deleted_at)
+    * EXCEPT(
```
evansiroky, line 14: I don't believe we do any joins of the validation notices data back to any of the daily fact tables, but I guess this doesn't hurt.

evansiroky (approved):
Seems reasonable to me for now. Doing more stuff further upstream may be needed later, but if this helps unblock the generation of the January reports, I'm all for merging in ASAP.

lauriemerrell:
Thanks @edasmalchi -- with confirmation that analysts do in fact rely on `GTFS Schedule Feeds Latest` and joining that with `views`, I agree that I think we need to move the trimming to pre-processing (step 2 under the alternatives listed). 

Additionally, in reports right now we mix and match a bit across `views` and `schedule_type2` which just reinforces that we don't really have a firewall between `views` and earlier stages (there's [an outstanding ticket ](https://github.com/cal-itp/reports/issues/36) for me to address this, but still.)

@machow, can you take a look and let me know if you see some other alternative that I'm not thinking of? I kind of think that, short of substantially refactoring the whole pipeline, we basically have to move this to pre-processing.

In the meantime, I made a separate PR #1027 to address the narrow SacRT issue so that we can unblock report generation for January.

machow:
Hey, I can follow up a bit more tomorrow, but RE 

> with confirmation that analysts do in fact rely on GTFS Schedule Feeds Latest and joining that with views

What about changing feeds latest to be derived from the cleaned data? Or a helper function in calitp-py to do the temporal filter for latest? There is a bit more cleaning done than just this trimming, so moving trimming to pre-processing afaik would not resolve "make the latest data equiv to filtering the view data"

(That's def a limit I created by just dumping the latest raw data to feeds latest 😬)

evansiroky:
I think this could work. Definitely be sure to check with analysts during this update and maybe keep things running in parellel if possible for a certain time. Also, it seems like this work could resolve #653.

machow:
Hey, sorry for the wait. Moving staging tables into their own dataset makes sense to me (vs using the suffix _clean).

I would be careful with increasing the scope of this PR, which now seems like multiple issues:

* Trimming whitespace from modeled data
* Aligning schedule latest with modeled data
* Cleaning up internal structure

I would suggest fixing the first bullet, and then going down the list. Once that's done analysts can query the latest data from views, but doing all 3 in one move will involved a lot of moving parts to check.

lauriemerrell:
Ok, sounds good, thanks Michael! As noted in my comment I would split this into 3 PRs, one for each of the primary steps I enumerated.

Also -- based on #1079, I think as part of this cleaning step we probably want to swap the `clean` tables to use `SELECT DISTINCT` for all the columns, under all the same arguments as this whitespace discussion. 

I removed all the reviewers until this is actually ready for review again.

lauriemerrell:
Closing this PR; refactoring and will make new PRs. Subsequent work can be followed in #1137

PR #1028, opened 2022-02-03
fix: only keep one netlify link per PR

atvaccaro:
The docs publishing github workflow posts a netlify link back to the PR with the PR's example doc site; currently, a new comment is made by the workflow bot for each new commit pushed to the PR.

This PR adds a flag that will delete older comments with duplicate text.


atvaccaro:
Actually this may not be the right workflow for this flag.

PR #1031, opened 2022-02-04
chore: silence noisy retry logs and fix a function name

atvaccaro:
tiny change

### DIFF
--- a/airflow/dags/dags.py
+++ b/airflow/dags/dags.py
@@ -117,7 +117,7 @@ def sql_enrich_duplicates(schema_tbl, key_columns, order_by_columns):
     """
 
 
-def email_failure(context):
+def log_failure_to_slack(context):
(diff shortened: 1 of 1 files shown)


evansiroky:
Have we verified whether this change will silence the retries? In the attached screenshot, the first error was a legitimate error that required attention. However, the next one was an "error" from one of the waiting tasks. The slack notifications seem to have the same signature for tasks that failed legitimately and the "waiting"-style tasks that "fail" upon a timeout, but can fail up to 25 times. Also, some tasks can fail and retry up to 2 times, so it may not be necessary to receive a notification about the first failure if the retry succeeds. This is how the email notifications work: I only get an email after all retry attempts failed.

Screen Shot 2022-02-04 at 9 23 58 AM

themightychris (approved):

PR #1032, opened 2022-02-04
fix: pin click 8.0.0

atvaccaro:
I broke the Airflow requirements.txt; for some reason, the composer install failed even though typer should be compatible with click 8.0.0; this should? fix things.

I think we need a CI/CD task to test installing the requirements.

### DIFF
--- a/.github/workflows/check-requirements.yml
+++ b/.github/workflows/check-requirements.yml
@@ -1,0 +1,16 @@
+name: Check requirements
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  install:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.7'
+    - run: pip install -r airflow/requirements.txt
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 # Vim
 .*.sw[po]
 
+# JetBrains
+.idea/
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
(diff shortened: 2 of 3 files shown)


evansiroky (dismissed):
Did this end up reproducing the error received in the composer airflow? If not, it may not be best to merge something that provides false positives. Also, this doesn't seem to pin click 8.0.0.

atvaccaro:
> Did this end up reproducing the error received in the composer airflow? If not, it may not be best to merge something that provides false positives.
No, because Composer installs things incrementally; so without 8.0.0 being pinned, `pip` just keeps the 7.X installed but then fails to install something later.

> Also, this doesn't seem to pin click 8.0.0.
Yeah I've been attempting to reproduce via the CI/CD task (hence my commits adding/removing click) but unfortunately the Composer install doesn't work the same way.

I just re-worked these commits some; I'd like to merge this so I can proceed with re-merging the RT parser.

atvaccaro:
> I may be a bit out of my depth here, but just to confirm, do we expect this to interact with the existing`push_to_gcloud` installation of `scripts/requirements.txt` ([ref](https://github.com/cal-itp/data-infra/blob/0c649a85c5cb44762ca7bf18391a62ea1f714365/.github/workflows/push_to_gcloud.yml#L26))?

Nope, this is solely Airflow reqs. Ideally we'd have better dependency management in both places, but that can wait.

themightychris (approved):

atvaccaro:
Thanks! If this fails again, I'm going to go the poetry/lockfile route cause it's silly to keep poking at Composer update.

https://github.com/cal-itp/data-infra/runs/5096366649?check_suite_focus=true

PR #1056, opened 2022-02-08
fix(gtfs-rt): define json schema for vehicle positions external table

### LINKED ISSUE #1051: Bug: Querying gtfs_rt.vehicle_positions fails for most queries
**Describe the bug**
Queries on the new gtfs_rt.vehicle_positions table seem to fail almost all of the time. With a tiny limit value ~100, the query may run after several tries so perhaps something to do with timing/continuous loading+updating of the table? Larger queries consistently fail.

**To Reproduce**
1. Try a query on gtfs_rt.vehicle_positions, for example 
```
SELECT * FROM `cal-itp-data-infra.gtfs_rt.vehicle_positions`
WHERE calitp_itp_id = 300 LIMIT 1000
```
2. Observe various BigQuery JSON parse/process errors

**Expected behavior**
Query returning a valid table without errors.

**Additional context**
BigQuery errors:
```

Error while reading table: cal-itp-data-infra.gtfs_rt.vehicle_positions, error message: JSON parsing error in row starting at position 0: Could not convert value 'string_value: "vehicle_193"' to integer. Field: id; Value: vehicle_193
Error while reading table: cal-itp-data-infra.gtfs_rt.vehicle_positions, error message: JSON processing encountered too many errors, giving up. Rows: 1; errors: 1; max bad: 0; error percent: 0
Error while reading table: cal-itp-data-infra.gtfs_rt.vehicle_positions, error message: JSON parsing error in row starting at position 0: Could not convert value 'string_value: "t_2062901_b_33847_tn_8"' to integer. Field: tripId; Value: t_2062901_b_33847_tn_8
Error while reading table: cal-itp-data-infra.gtfs_rt.vehicle_positions, error message: JSON processing encountered too many errors, giving up. Rows: 2; errors: 2; max bad: 0; error percent: 0
Error while reading table: cal-itp-data-infra.gtfs_rt.vehicle_positions, error message: JSON parsing error in row starting at position 0: No such field: vehicle.currentStopSequence.
Error while reading table: cal-itp-data-infra.gtfs_rt.vehicle_positions, error message: JSON processing encountered too many errors, giving up. Rows: 3; errors: 3; max bad: 0; error percent: 0 
```

Another BigQuery error via calitp.py `query_sql`
```
BadRequest: 400 Error while reading table: cal-itp-data-infra.gtfs_rt.vehicle_positions, error message: JSON parsing error in row starting at position 0: No such field: vehicle.currentStopSequence.
```

lauriemerrell:
# Overall Description
Resolves #1051 by:
* Explicitly defining the JSON schema in `external_vehicle_positions.sql`
* Adding the [`ignore_unknown_values` flag](https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-json#json_options) to all the external table tasks here; this will allow rarely-used fields to be ignored instead of preventing access to the table (I am not sure if we want this to be enabled long-term, but it seemed like it would help smooth things out here in the short term)

**Note that this PR _narrowly_ resolves the issue by focusing only on vehicle positions data.** I wanted to unblock @edasmalchi and vehicle positions is significantly simpler than trip updates. If this approach works out well, we can apply it to service alerts and trip updates as well in a subsequent PR. 

### Resulting fields

Heads up for @edasmalchi -- the field names now follow the JSON hierarchy, so `latitude` and `longitude` are now accessed via `vehicle.position.latitude` and `vehicle.position.longitude` respectively. Right now things follow the [actual GTFS docs](https://gtfs.org/reference/realtime/v2/#message-vehicleposition) pretty closely, with the exception that they show everything in snake case but our processing (which is actually Google's default processing for GTFS pb messages) turns that into camel case, so `current_stop_sequence` in the docs becomes `currentStopSequence` in our table. 

Below are screenshots of the schema, I am hesitant to add these to our actual docs until things have stabilized more but putting here for at least some reference. 

Top-level fields & `header`:
image

`vehicle.vehicle` (🙄), `vehicle.trip`, and `vehicle.position`:
image

And finally other fields immediately under `vehicle` (`vehicle.<field>`):
image

## Checklist for all PRs

image

This PR updates the `rt_loader` DAG in order to....

Update the following DAG tasks:

- `external_vehicle_positions.sql`: explicitly define JSON schema for the external table; add ignore unknown values flag
- `external_service_alerts.sql` & `external_trip_updates.sql`: add ignore unknown values flag only

### DIFF
--- a/airflow/dags/rt_loader/external_service_alerts.sql
+++ b/airflow/dags/rt_loader/external_service_alerts.sql
@@ -8,5 +8,6 @@ dependencies:
 CREATE OR REPLACE EXTERNAL TABLE `gtfs_rt.service_alerts`
 OPTIONS (
     format = "JSON",
-    uris = ["{{get_bucket()}}/rt-processed/service_alerts/*.jsonl.gz"]
+    uris = ["{{get_bucket()}}/rt-processed/service_alerts/*.jsonl.gz"],
+    ignore_unknown_values = True
 )
--- a/airflow/dags/rt_loader/external_trip_updates.sql
+++ b/airflow/dags/rt_loader/external_trip_updates.sql
@@ -8,5 +8,6 @@ dependencies:
 CREATE OR REPLACE EXTERNAL TABLE `gtfs_rt.trip_updates`
 OPTIONS (
     format = "JSON",
-    uris = ["{{get_bucket()}}/rt-processed/trip_updates/*.jsonl.gz"]
+    uris = ["{{get_bucket()}}/rt-processed/trip_updates/*.jsonl.gz"],
+    ignore_unknown_values = True
 )
--- a/airflow/dags/rt_loader/external_vehicle_positions.sql
+++ b/airflow/dags/rt_loader/external_vehicle_positions.sql
@@ -5,8 +5,48 @@ dependencies:
   - parse_rt_vehicle_positions
 ---
 
-CREATE OR REPLACE EXTERNAL TABLE `gtfs_rt.vehicle_positions`
+CREATE OR REPLACE EXTERNAL TABLE `gtfs_rt.vehicle_positions` (
+    header STRUCT<
+      timestamp INT64,
+      incrementality STRING,
+      gtfsRealtimeVersion STRING
+    >,
+    id STRING,
+    vehicle STRUCT<
+      vehicle STRUCT <
+        licensePlate STRING,
+        label STRING,
+        id STRING
+      >,
+      trip STRUCT <
+        tripId STRING,
+        routeId STRING,
+        directionId INT64,
+        startTime STRING,
+        startDate STRING,
+        scheduleRelationship STRING
+      >,
+      position STRUCT <
+        latitude NUMERIC,
+        longitude NUMERIC,
+        bearing NUMERIC,
+        odometer INT64,
+        speed NUMERIC
+      >,
+      currentStopSequence INT64,
+      stopId STRING,
+      currentStatus STRING,
+      timestamp INT64,
+      congestionLevel STRING,
+      occupancyStatus STRING,
+      occupancyPercentage INT64
+    >,
+    calitp_itp_id INT64,
+    calitp_url_number INT64,
+    calitp_filepath STRING
+)
 OPTIONS (
     format = "JSON",
-    uris = ["{{get_bucket()}}/rt-processed/vehicle_positions/*.jsonl.gz"]
+    uris = ["{{get_bucket()}}/rt-processed/vehicle_positions/*.jsonl.gz"],
+    ignore_unknown_values = True
 )
--- a/airflow/dags/rt_loader/external_vehicle_positions.sql
+++ b/airflow/dags/rt_loader/external_vehicle_positions.sql
@@ -6,12 +6,15 @@ dependencies:
 ---
 
 CREATE OR REPLACE EXTERNAL TABLE `gtfs_rt.vehicle_positions` (
+    calitp_itp_id INT64,
+    calitp_url_number INT64,
+    calitp_filepath STRING,
+    id STRING,
     header STRUCT<
       timestamp INT64,
       incrementality STRING,
       gtfsRealtimeVersion STRING
     >,
-    id STRING,
     vehicle STRUCT<
       vehicle STRUCT <
         licensePlate STRING,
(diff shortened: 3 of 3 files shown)


edasmalchi (approved):
Looks great to me, thanks!

e-lo:
❓ Is there a reason why we would define this in a static `.sql` file rather than using something dynamically generated from the canonical `.proto` definitions? (or a [`json-schema`](https://json-schema.org/) middle-person?) 

I'm sure there is - but it pains me to see data models not written in....data.

lauriemerrell:
Thanks @e-lo for that point... I'll investigate those as alternatives, but perhaps we can merge this as a stopgap for the immediate term. 

I think the main challenge is just that any individual feed isn't usually actually using the full schema, so that's what we ran into originally - when we extracted the schema using the automatic BigQuery functionality, it had problems because there were some fields that occur only rarely. 

So if we extract a schema from any given file, it's probably not sufficient to define the entire schema for the whole table across all the feeds. 

We don't have all the files concatenated into one (and therefore, knowledge of all possible fields) until we load them into this BigQuery table. 

I'll look into the options you mentioned, but I am also wondering about extracting schema metadata across **so many files** as what we have in the realtime data. I am also wondering about imputing a schema because as we've seen in the schedule data sometimes people have random unsupported fields (we do also enumerate fields in the schedule pipeline, I think for that reason); I assume that's less likely in RT data because it is a heavier lift to produce and seems less susceptible to totally random errors like schedule data, but I would still wonder about that.

e-lo:
>  I am also wondering about extracting schema metadata across so many files as what we have in the realtime data. I am also wondering about imputing a schema because as we've seen in the schedule data sometimes people have random unsupported fields (we do also enumerate fields in the schedule pipeline, I think for that reason); I assume that's less likely in RT data because it is a heavier lift to produce and seems less susceptible to totally random errors like schedule data, but I would still wonder about that.

There are lots of feeds that will have (and we want to have) experimental fields - which we want to support.

lauriemerrell:
> There are lots of feeds that will have (and we want to have) experimental fields - which we want to support.

To clarify, defined "experimental" fields like occupancyStatus? Or ad-hoc fields that are not defined in the main spec at all? And specifically in realtime or in schedule also? 

(This is probably a bigger picture conversation than this immediate PR...)

e-lo:
> Main challenge is just that any individual feed isn't usually actually using the full schema

[not being an expert in anything related to data warehouses...] it would seem that it should be possible to include all possible fields in the db schema and have null values for data structures or values that aren't provided?

lauriemerrell:
> [not being an expert in anything related to data warehouses...] it would seem that it should be possible to include all possible fields in the db schema and have null values for data structures or values that aren't provided?

Yes, this is definitely possible, but we will need to look into it more... I think that it is harder with JSON data than with tabular data. 

That's what the original approach was (just have BigQuery load everything), but BigQuery complained about fields that were too rare. I think there are settings we can toggle that can play with that to basically tell it "no field is too rare". 

But I also thought that for these early days, it might be preferable to know what we're getting -- ideally there would be a staging layer where we load everything, and then there would be the `views` that analysts could consume. But since the RT pipeline is still so early, Eric is just working with the first layer of ingestion. So right now this PR is trying to make the data usable for him while also being the first stage of ingestion.

lauriemerrell:
> We don't want to 🗑️ data that isn't part of a "minimum viable dataset".

I'm not totally sure what this means.

I guess to clarify -- this definitely doesn't trash any raw data. This is just an incremental step on trying to load it into BigQuery. This current approach still loads significantly **more** fields than were being loaded under the prior approach (the proof of concept vehicle positions pipeline that was in place until last week, which basically only loaded lat/long and timestamp).

lauriemerrell:
Spinning that bigger picture conversation into #1058 if that's ok @e-lo -- I think the priority for this PR is getting access to this data this week, and we can revisit the overall approach for sure.... That's part of why I only implemented for `vehicle_positions` and not the other two

atvaccaro (approved):
lgtm!

evansiroky (approved):
While it certainly would be nice to generate SQL commands from a JSON schema or the files themselves, I believe this is mergeable as-is. It adds benefit as mentioned since this adds more data to the vehicle positions feed than was available previously with the rectangling which purposely omitted certain fields. Furthermore, we urgently need to resolve #1051.

PR #1061, opened 2022-02-09
fix(gtfs-rt): fix data types for vehicle positions external table schema

lauriemerrell:
# Overall Description

Some of the data types I picked in #1056 were wrong (I misunderstood how the data types in the [GTFS documentation](https://gtfs.org/reference/realtime/v2/) mapped to the [BigQuery data types](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#numeric_types)), and I didn't catch them because of the limited testing data I used in the `gtfs-data-test` bucket. 

This PR fixes `NUMERIC` (and one `INT64`) to be the correct `FLOAT64` type. 

To test this, I ensured that queries that were failing with the current definition succeed with the new definitions on the actual prod data by making an external table in `staging` that still reads data from the prod RT bucket (table is: `cal-itp-data-infra-staging.gtfs_rt.vehicle_positions_test_prod_source`)

## Checklist for all PRs

image

This PR updates the `rt_loader` DAG in order to update the following DAG tasks:

- `external_vehicle_positions.sql`: Corrects data types in external table schema definition

### DIFF
--- a/airflow/dags/rt_loader/external_vehicle_positions.sql
+++ b/airflow/dags/rt_loader/external_vehicle_positions.sql
@@ -30,11 +30,11 @@ CREATE OR REPLACE EXTERNAL TABLE `gtfs_rt.vehicle_positions` (
         scheduleRelationship STRING
       >,
       position STRUCT <
-        latitude NUMERIC,
-        longitude NUMERIC,
-        bearing NUMERIC,
-        odometer INT64,
-        speed NUMERIC
+        latitude FLOAT64,
+        longitude FLOAT64,
+        bearing FLOAT64,
+        odometer FLOAT64,
+        speed FLOAT64
       >,
       currentStopSequence INT64,
       stopId STRING,


atvaccaro (approved):
Is a float best here, or would a numeric with precision of maybe precision 9 and scale 6 work? Just glancing at the RT data, it looks like we get truncated decimals with no more than 6 places after the decimal.

atvaccaro:
> Thanks @atvaccaro -- I _think_ we need float? The proto data is actually a `float32` (see [here](https://developers.google.com/protocol-buffers/docs/proto#scalar), as linked from the RT docs). The value that was breaking it is: image
> 
> I think to be safe we need to stick with `FLOAT64` (because the underlying data is actually a float and not a fixed decimal), but let me know if that's a misunderstanding?
> 
> ETA: And BigQuery's only float type is FLOAT64.

I'm wondering if the error is due to the default parameters for `numeric`? Not a big deal either way. Also I think I was looking at the lat/lon more than anything.

lauriemerrell:
> I'm wondering if the error is due to the default parameters for numeric? Not a big deal either way. Also I think I was looking at the lat/lon more than anything.

It definitely is, but I'm still nervous about converting to `numeric` because the proto datatype really is float... I guess we could have different numeric parameters by specific field, but it would still always be a conversion and risk these problems.

Issue #1005, opened 2022-01-28
Divide views dataset by domain


lauriemerrell:
Currently, all views across domain areas (Payments, GTFS Schedule, GTFS RT, Airtable) are included in the single `views` BigQuery dataset. This was done because Metabase did not support joins across BigQuery datasets and we wanted to ensure that data could be joined across domains. 

As of version 0.42, Metabase will support joins across datasets. As a result we can reconsider whether we want to make dedicated `views_<domain>` datasets rather than having everything combined in one -- as we create more views, it seems likely that the current combined `views` dataset will become unwieldy (if it hasn't already).

j-meelah:
Obsolete by dbt

Issue #985, opened 2022-01-24
Research: How many operators is Cal-ITP assessing?


evansiroky:
## Question

How many operators is Cal-ITP assessing?

Answering this question should help inform future query design that where the unit of analysis is on the subset of operators that are categorized as assessed operators.

## Metrics

* An assessed operator is an organization in airtable that meets the following conditions:

<img width="685" alt="Screen Shot 2022-01-24 at 2 16 29 PM" src="https://user-images.githubusercontent.com/3112493/150875059-68e87b3e-0f0a-45ab-8eb6-d7573e75f644.png">

## Data sources

* Airtable via data warehouse

## Depends on

- [ ] #979

## (Data Servicess Team to Copy and Fill Out Below)

The QuVR MD template below will be filled out by a member of the data services team.
This allows us to describe the request, in a way that is easy to hand-off for analysis.
After the research phase, we will sync with the asker to figure out if the metric and dashboard pieces are needed.

Before starting research:

* **Question**: Question written as a single sentence.
* **View**: E.g. `views.gtfs_schedule_fact_daily_feed_files`.
* **Research**:
  - How should the results be presented?
  - When are they needed by?

After reviewing research with the asker:

* **Metric**: what specific calculations are needed?
* **Dashboard**: where should we put the result?

lauriemerrell:
I created a draft attempt to answer this in Metabase: https://dashboards.calitp.org/question/255-testing-answer-issue-985

Note -- I had to assume that the Service Type filter shown was service type of services **operated** since there is also a service type of services **managed**. 

Also note that I think right now this has to be answered in SQL in Metabase because they don't seem to allow you to filter on OR conditions in the question builder UI (so you can't select multiple options for things like reporting category as shown in the screenshot).

evansiroky:
Status update: @lauriemerrell is working on creating some mapping tables that are derived from the raw airtable data to be able to do standard database joins via metabase.

Issue #984, opened 2022-01-24
Research: percent of providers with a static GTFS dataset.


evansiroky:
## User Story

**As a** Cal-ITP program manager or more senior Caltrans executive, 
**I want to** know what percent of transit agencies have a GTFS Schedule dataset 
**so that** I can quantify the statewide quality of transit data across all agencies and track improvement to this quality over time.

This question should build on build on top of the research performed in cal-itp/data-analyses#379 and calculate a percent of relevant transit agencies that have GTFS Schedule data.

If needed, research could be performed with various stakeholders to determine how to calculate the answer to this question since some transit agencies could have a GTFS Schedule for some of their services but not others. Therefore, there is a decision that needs to be made whether this high level ask is asking for whether some agencies are at least partially accounted for or fully accounted for with GTFS Schedule data. If none of the stakeholders can give a clear answer about how to calculate this baseline, a deliverable of this report should propse at least one recommended option for calculating this baseline.

## Acceptance Criteria

**Given** the data Cal-ITP has collected about transit agencies with respect to how they are funded, what kind of service they operate, and the presence of GTFS Schedule data
**When** applying all relevant criteria about what qualifies as a transit agency for reporting purposes and analyzing for the presence of GTFS Schedule data
**Then** a percentage should be calculated.

The deliverable of this should be as a metabase question that simply shows a number as a percent.

## Sprint Ready Checklist

- [ ]  Acceptance criteria defined
- [ ]  Team understands acceptance criteria
- [ ]  Team has defined solution / steps to satisfy acceptance criteria
- [ ]  Acceptance criteria is verifiable / testable
- [ ] Dependencies identified

evansiroky:
Answered via work on GTFS Guideliens Dashboard.

PR #1063, opened 2022-02-09
Feat initial airtable views

lauriemerrell:
# Overall Description

This PR adds the initial Airtable views that should be sufficient to answer the questions in #984 and #985. 🚨 Reviewers please don't merge without confirming that the questions below have been addressed.

Specifically, this PR:
1. Creates `views` versions of `organizations`, `services`, `gtfs service data`, and `gtfs datasets` -- these tables contain only a limited subset of columns that seemed necessary as a first pass
   * ❓ Do these columns look sufficient for a first attempt? Are any really core, stable facts about these entities missing? (Columns are enumerated in each `california_transit_<table name>.sql` file) -- would appreciate @edasmalchi or @e-lo check here
   * ❓ How do we feel about the naming for the main views / tables (`airtable_california_transit_<table name>`)? The names are long, but until #1005 I think this might be our best bet? -- hope @evansiroky can weigh in 
2. Creates a `generate_airtable_mapping_table.py` helper that can be used to generate mapping DAG tasks programmatically 
   * ❓ Would appreciate feedback on the right place store this -- maybe @evansiroky, @atvaccaro, or @mjumbewu can advise
   * ❓ Do we want to store the initial CSV that I used to generate this first batch in the repo? If so, where? (File attached for reference) -- also a Q for @evansiroky, @atvaccaro, or @mjumbewu 
3. Using the helper, creates a bunch of mapping tables between the listed tables
   * ❓ Do these capture the specific relationships we care about to answer #984 / #985 ? -- hoping @evansiroky, @edasmalchi, or @e-lo can advise
   * ❓ How do we feel about the naming convention? -- appreciate thoughts from anyone 

## Checklist for all PRs

- [ ] Run `pre-commit run --all-files` to make sure markdown/lint passes
- [ ] Link this pull request to all issues that it will close using keywords (see GitHub docs about [Linking a pull request to an issue using a keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)). Also mention any issues that are partially addressed or are related.

- [ ] Include this section whenever any change to a DAG in the `airflow/dags` folder occurs, otherwise please omit this section.
- [ ] Verify that all affected DAG tasks were able to run in a local environment
- [ ] Take a screenshot of the graph view of the affected DAG in the local environment showing that all affected DAG tasks completed successfully
image
   - ❓ Do we want to add docs about these to the `Transit Database` section of the docs, the `Airflow` section of the docs, both, neither? I don't have any docs info in here yet because I wanted to confirm that we feel good about this approach and ask what we think is most important to document -- cc @charlie-costanzo here
   - ❓ Is it ok for me to add docs in a follow-up PR?
- [ ] Fill out the following section describing what DAG tasks were added/updated

This PR **creates** the `airtable_views` DAG in order to....

Add the following DAG tasks:

**Primary views**:
- `california_transit_gtfs_datasets.sql`
- `california_transit_gtfs_service_data.sql`
- `california_transit_organizations.sql`
- `california_transit_services.sql`

**Mapping tables**: 
- `california_transit_map_gtfs_datasets_aggregated_to_x_self.sql`
- `california_transit_map_gtfs_datasets_dataset_producers_x_organizations_gtfs_datasets_produced.sql`
- `california_transit_map_gtfs_datasets_dataset_publisher_x_organizations_gtfs_datasets.sql`
- `california_transit_map_gtfs_datasets_gtfs_service_mapping_x_gtfs_service_data_gtfs_dataset.sql`
- `california_transit_map_gtfs_service_data_reference_static_gtfs_service_x_self.sql`
- `california_transit_map_organizations_mobility_services_managed_x_services_provider.sql`
- `california_transit_map_organizations_mobility_services_operated_x_services_operator.sql`
- `california_transit_map_organizations_parent_organization_x_self.sql`
- `california_transit_map_services_gtfs_services_association_x_gtfs_service_data_services.sql`
- `california_transit_map_services_paratransit_for_x_self.sql`

Here's a screenshot of what one of these mapping tables looks like (`airtable_california_transit_map_organizations_mobility_services_managed_x_services_provider.sql`), just to illustrate:
image

**Here's the CSV file I fed into my helper to generate the mapping DAG tasks**:
[mapping_tables.csv](https://github.com/cal-itp/data-infra/files/8036379/mapping_tables.csv)

### DIFF
--- a/airflow/dags/airtable_views/california_transit_gtfs_service_data.sql
+++ b/airflow/dags/airtable_views/california_transit_gtfs_service_data.sql
@@ -1,0 +1,34 @@
+---
+operator: operators.SqlToWarehouseOperator
+dst_table_name: "views.airtable_california_transit_gtfs_service_data"
+
+description: Data from the gtfs service data table in the California Transit Airtable base. See Airtable data documentation for more information about its contents. This table represents only the most recent extract and does not track changes in data over time.
+
+fields:
+  service_id: Internal Airtable ID for this GTFS service data record, used for joining with other Airtable views tables
+  name: Name for the GTFS dataset <> service relationship
+  services: Internal Airtable ID for the service record in this relationship
+  gtfs_dataset: Internal Airtable ID for the gtfs_dataset record in this relationship
+  category: Indicates whether this is a primary, precursor, or unknown-type GTFS dataset/service relationship
+
+tests:
+  check_null:
+    - gtfs_service_data_id
+    - name
+  check_unique:
+    - gtfs_service_data_id
+
+external_dependencies:
+  - airtable_loader: california_transit_gtfs_service_data
+---
+
+  SELECT
+    gtfs_service_data_id,
+    name,
+    -- service and gtfs_dataset are 1:1 foreign key fields
+    -- but they export as an array from airtable
+    -- turn them into a string for joining
+    REPLACE(REPLACE(REPLACE(services, "'",""), "[", ""), "]", "") services,
+    REPLACE(REPLACE(REPLACE(gtfs_dataset, "'",""), "[", ""), "]", "") gtfs_dataset,
+    category
+  FROM `airtable.california_transit_gtfs_service_data`
--- a/airflow/dags/airtable_views/california_transit_map_gtfs_service_data_reference_static_gtfs_service_x_self.sql
+++ b/airflow/dags/airtable_views/california_transit_map_gtfs_service_data_reference_static_gtfs_service_x_self.sql
@@ -3,9 +3,7 @@ operator: operators.SqlToWarehouseOperator
 
 dst_table_name: views.airtable_california_transit_map_gtfs_service_data_reference_static_gtfs_service_x_self
 
-description: | Self-join mapping table for the GTFS gtfs_service_data table,
-reference_static_gtfs_service column in the California Transit Airtable base.
-Each row represents a relationship between two gtfs_service_data records.
+description: Self-join mapping table for the GTFS gtfs_service_data table, reference_static_gtfs_service column in the California Transit Airtable base. Each row represents a relationship between two gtfs_service_data records.
 
 fields:
   gtfs_service_data_id: Internal Airtable ID for a gtfs_service_data record
@@ -24,7 +22,7 @@ tests:
     - reference_static_gtfs_service_id
 
 external_dependencies:
-- airtable_loader: california_transit_gtfs_service_data
+  - airtable_loader: california_transit_gtfs_service_data
 ---
 
 -- follow the sandbox example for unnesting airtable data
--- a/airflow/dags/airtable_views/california_transit_map_gtfs_service_data_reference_static_gtfs_service_x_self.sql
+++ b/airflow/dags/airtable_views/california_transit_map_gtfs_service_data_reference_static_gtfs_service_x_self.sql
@@ -1,0 +1,53 @@
+---
+operator: operators.SqlToWarehouseOperator
+
+dst_table_name: views.airtable_california_transit_map_gtfs_service_data_reference_static_gtfs_service_x_self
+
+description: | Self-join mapping table for the GTFS gtfs_service_data table,
+reference_static_gtfs_service column in the California Transit Airtable base.
+Each row represents a relationship between two gtfs_service_data records.
+
+fields:
+  gtfs_service_data_id: Internal Airtable ID for a gtfs_service_data record
+  reference_static_gtfs_service_id: Internal Airtable ID for a gtfs_service_data record
+  gtfs_service_data_name: gtfs_service_data record name
+  reference_static_gtfs_service_name: gtfs_service_data record name
+
+tests:
+  check_null:
+    - gtfs_service_data_id
+    - reference_static_gtfs_service_id
+    - gtfs_service_data_name
+    - reference_static_gtfs_service_name
+  check_composite_unique:
+    - gtfs_service_data_id
+    - reference_static_gtfs_service_id
+
+external_dependencies:
+- airtable_loader: california_transit_gtfs_service_data
+---
+
+-- follow the sandbox example for unnesting airtable data
+
+WITH
+
+unnested_t1 AS (
+    SELECT
+        T1.gtfs_service_data_id as gtfs_service_data_id
+        , T1.name as gtfs_service_data_name
+        , CAST(reference_static_gtfs_service AS STRING) AS reference_static_gtfs_service_id
+    FROM
+        `airtable.california_transit_gtfs_service_data` T1
+        , UNNEST(JSON_VALUE_ARRAY(reference_static_gtfs_service)) reference_static_gtfs_service
+),
+t2 AS (
+    SELECT
+        T2.gtfs_service_data_id as reference_static_gtfs_service_id
+        , T2.name as reference_static_gtfs_service_name
+    FROM
+        `airtable.california_transit_gtfs_service_data` T2
+)
+
+SELECT *
+FROM unnested_t1
+LEFT JOIN t2 USING(reference_static_gtfs_service_id)
--- a/airflow/dags/airtable_views/california_transit_organizations.sql
+++ b/airflow/dags/airtable_views/california_transit_organizations.sql
@@ -1,0 +1,34 @@
+---
+operator: operators.SqlToWarehouseOperator
+dst_table_name: "views.airtable_california_transit_organizations"
+
+description: Data from the organizations table in the California Transit Airtable base. See Airtable data documentation for more information about its contents. This table represents only the most recent extract and does not track changes in data over time.
+
+fields:
+  organization_id: Internal Airtable ID for this organization record, used for joining with other Airtable views tables
+  name: Organization name
+  organization_type: Categorical data; example values "City/Town", "Joint Powers Agency"
+  roles: Categorical choices imported as a string where a comma separates different selections; example values "Regional Transportation Planning Agency, Metropolitan Planning Organization"; "Metropolitan Planning Organization"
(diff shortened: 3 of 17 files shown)


airflow/plugins/generate_airtable_mapping_table.py
```diff
@@ -0,0 +1,202 @@
...
+# and col1 and col2 are the corresponding (warehouse) column names
+# that define the linked record relationship
+# in cases of self-joins, leave table2 and col2 blank
+# and just define the table in table1 and the foreign key column in col1
+# (for example, services.paratransit_for)
+
+import pandas as pd
+
+
+def generate_sql(table1, table2, col1, col2):
```
evansiroky, line 23: In the times I have seen code used to generate code that is then copied and pasted elsewhere to be executed it is a red flag of code that probably shouldn't exist in that form. This is an anti-pattern for the following reason in my opinion:

- It creates two codebases that need to be maintained (one for the code that generates the code and another for the other code that is generated)
- The generated code is not always easily testable in various fashions (in execution, static typing or via automated testing)
- It requires extra build steps that are not needed had there been just one codebase

This appears to be trying to make the text for `SqlToWarehouseOperator` DAG tasks. A better approach for this is to refactor this file into two new operators:

1. AirtableTwoTableJoinOperator
2. AirtableSelfJoinOperator

In doing this, you lose some of the automagical creation of some documentation of the DAG task, but also lose the burdens that come with the "executing code to generate executable code" anti-pattern. Furthermore, any additional DAG tasks will then be able to use one of these operators instead of having to manually run some code to generate another Gusty configuration.

evansiroky (changes requested):
In the times I have seen code used to generate code that is then copied and pasted elsewhere to be executed it is a red flag of code that probably shouldn't exist in that form. This is an anti-pattern for the following reasons in my opinion:

- It creates two codebases that need to be maintained (one for the code that generates the code and another for the other code that is generated)
- The generated code is not always easily testable in various fashions (in execution, static typing or via automated testing)
- It requires extra build steps that are not needed had there been just one codebase
- If the generated code has a bug that needs to be fixed or a new feature should be introduced, this could result in a need to update all of the generated code files (of which there are already about a dozen)

The `generate_airtable_mapping_table.py` file appears to be trying to make the text for `SqlToWarehouseOperator` DAG tasks. A better approach for this is to refactor this file into two new custom airflow operators:

1. AirtableTwoTableJoinOperator
2. AirtableSelfJoinOperator

In doing this, you lose some of the automagical creation of some documentation of the DAG task, but also lose the burdens that come with the "executing code to generate executable code" anti-pattern. Furthermore, any additional DAG tasks will then be able to use one of these operators instead of having to manually run some code to generate another Gusty configuration.

lauriemerrell:
Thanks @atvaccaro and @evansiroky -- I understand that the current approach is not ideal. @mjumbewu made me aware of this existing example of SQL templating in the Payments pipeline -- [stg_enriched_micropayments](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/payments_views_staging/stg_enriched_micropayments.sql#L10), which is [added as a defined macro in `dags.py`](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/dags.py#L153).

This seems like a pretty solid option to me -- I think it addresses Evan's primary concerns about having a two-step code generation process (I understand/agree with the concerns there), while balancing Andrew's concerns about proliferation of very limited-scope Airflow operators. I'm planning to go that route for now.

evansiroky:
Thanks, @atvaccaro for engaging in healthy dialogue around development patterns and methodologies. I really appreciate the points you bring up and think this adds knowledge for the whole team. I myself am still fairly new to the data science world and am not entirely sure how dbt is structured and how this PR relates to it, but am interested to learn more.

Regarding the path forward here, that templating option via the macros does seem like it would involve less custom code due to it not having all the overhead that custom operators have. However, that option introduces a potential problem of proliferation of macros which may not be that big of a deal actually. It seems like the macros we currently have are limited in scope. So with all that being said, the macros option sounds good to me.

lauriemerrell:
Screenshot showing that DAG tasks can all run locally after refactoring to use macro:
image

airflow/dags/airtable_views/california_transit_gtfs_service_data.sql
```diff
@@ -0,0 +1,34 @@
...
+---
+
+  SELECT
+    gtfs_service_data_id,
+    name,
+    -- service and gtfs_dataset are 1:1 foreign key fields
+    -- but they export as an array from airtable
+    -- turn them into a string for joining
+    REPLACE(REPLACE(REPLACE(services, "'",""), "[", ""), "]", "") services,
+    REPLACE(REPLACE(REPLACE(gtfs_dataset, "'",""), "[", ""), "]", "") gtfs_dataset,
```
evansiroky, line 32: It seems a bit fragile to be doing string replacement. Is there a way to use built-in JSON-parsing functions instead?

airflow/dags/airtable_views/california_transit_map_gtfs_service_data_reference_static_gtfs_service_x_self.sql
```diff
@@ -0,0 +1,33 @@
...
+    - gtfs_service_data_id
+    - reference_static_gtfs_service_id
+
+external_dependencies:
+  - airtable_loader: california_transit_gtfs_service_data
+---
+{{
+
+  sql_airtable_mapping(
+    table1 = "gtfs_service_data",table2 = "", col1 = "reference_static_gtfs_service", col2 = ""
```
evansiroky, line 30: oi, this brings back [memories](https://cal-itp.slack.com/archives/C02JFS4LAMU/p1638909693026300?thread_ts=1638810002.018000&cid=C02JFS4LAMU)

Screen Shot 2022-02-10 at 1 42 34 PM

airflow/dags/airtable_views/california_transit_organizations.sql
```diff
@@ -0,0 +1,34 @@
...
+external_dependencies:
+  - airtable_loader: california_transit_organizations
+---
+-- roles is a multi-select field in airtable
+-- turn it into a string that's comma-delimited
+  SELECT
+    organization_id,
+    name,
+    organization_type,
+    REPLACE(REPLACE(REPLACE(roles, "'",""), "[", ""), "]", "") roles,
```
evansiroky, line 31: Maybe this is fine for now, but another thought is that instead of turning this into something comma-delimited, perhaps there can be another table consisting of the organizations and their roles. Also, maybe this translation doesn't even need to occur if the unnesting is possible using SQL in BigQuery.

evansiroky (approved):
Seems good enough to go, but I have a few comments.

airflow/dags/airtable_views/california_transit_services.sql
```diff
@@ -0,0 +1,33 @@
...
+  - airtable_loader: california_transit_services
+---
+
+  SELECT
+    service_id,
+    name,
+    -- service type and mode are multi-select fields in airtable
+    -- turn them into a string that's comma-delimited
+    REPLACE(REPLACE(REPLACE(service_type, "'",""), "[", ""), "]", "") service_type,
+    REPLACE(REPLACE(REPLACE(mode, "'",""), "[", ""), "]", "") mode,
```
evansiroky, line 31: Same comment as https://github.com/cal-itp/data-infra/pull/1063#discussion_r804133702

airflow/dags/dags.py
```diff
@@ -151,6 +152,7 @@ def log_failure_to_slack(context):
             **user_defined_macros,
             "scd_join": scd_join,
             "sql_enrich_duplicates": sql_enrich_duplicates,
+            "sql_airtable_mapping": __macros__.generate_airtable_mapping_table.generate_sql,
```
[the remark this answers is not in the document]
lauriemerrell: Done, and verified that they still run ok: 
image

image

airflow/dags/airtable_views/california_transit_gtfs_service_data.sql
lauriemerrell: I went ahead and changed these since they are single response / foreign keys

airflow/dags/airtable_views/california_transit_organizations.sql
lauriemerrell: These multi-response fields are why I did this gross string manipulation and the reason I did it is that I find unnested data very difficult to work with in BQ. I thought that for analysts it would be easier to do a query on `roles like "%MPO%"` than to deal with the unnested data. Making a separate table is ok too, that's just going pretty far down the denormalized route and I wasn't sure how much we wanted to commit to that?

airflow/dags/dags.py
```diff
@@ -150,9 +60,9 @@ def log_failure_to_slack(context):
         latest_only=False,
         user_defined_macros={
             **user_defined_macros,
-            "scd_join": scd_join,
-            "sql_enrich_duplicates": sql_enrich_duplicates,
-            "sql_airtable_mapping": __macros__.generate_airtable_mapping_table.generate_sql,
+            "scd_join": __macros__.sql_gtfs_schedule_scd_join.scd_join,
+            "sql_enrich_duplicates": __macros__.sql_payments_enrich_duplicates.sql_enrich_duplicates,
+            "sql_airtable_mapping": __macros__.sql_airtable_generate_mapping_table.generate_sql,
```
evansiroky, line 65: Maybe these can also be added via the `**` command:

```suggestion
            **data_infra_macros
```

lauriemerrell: Refactored again, now you don't have to make any changes to `dags.py` to add a macro (which seems good?)

PR #1082, opened 2022-02-14
fix(airtable): add dummy airtable loader for external dependencies

### LINKED ISSUE #1080: Bug: airtable_views.california_transit_map_services_gtfs_services_association_x_gtfs_service_data_services external dependencies not complete
**Describe the bug**

It appears that the `airtable_views.california_transit_map_services_gtfs_services_association_x_gtfs_service_data_services` external dependency DAGs are not all being registered properly in airflow.

**To Reproduce**

The DAG defines the external dependencies as follows:

```yml
external_dependencies:
  - airtable_loader: california_transit_services
  - airtable_loader: california_transit_gtfs_service_data
```

However, this is what shows up in Airflow (note the single external dependency):

Screen Shot 2022-02-13 at 11 18 33 PM

**Expected behavior**

There should be 2 external dependencies instead of 1.

**Additional context**

This may be happening in other DAG tasks as well.

lauriemerrell:
# Overall Description

Restructure dependencies for the `airtable_views` DAG to resolve #1080. Basically, it seems likely that there's a gusty bug such that specifying multiple `external_dependencies` is not working correctly. 

Also, it seems that elsewhere in our pipeline when we have a "staging" and  "views", we usually have the views depend on a `DummyOperator` that waits for the entire staging DAG to complete. 

So, I am making the Airtable flow align with that convention, which also addresses the original problem. 

## Checklist for all PRs

image

This PR updates the `airtable_views` DAG in order to standardize the external dependency handling.

Adds the following DAG tasks:

- `dummy_airtable_loader.yml` - waits for `airtable_loader` to complete, following pattern from elsewhere in our codebase

- all existing `airtable_views` tasks were updated to have their dependencies point to the dummy rather than the external dependencies

### DIFF
--- a/airflow/dags/airtable_views/california_transit_gtfs_datasets.sql
+++ b/airflow/dags/airtable_views/california_transit_gtfs_datasets.sql
@@ -15,8 +15,8 @@ tests:
   check_unique:
     - gtfs_dataset_id
 
-external_dependencies:
-  - airtable_loader: california_transit_gtfs_datasets
+dependencies:
+  - dummy_airtable_loader
 ---
 
 SELECT
--- a/airflow/dags/airtable_views/california_transit_gtfs_service_data.sql
+++ b/airflow/dags/airtable_views/california_transit_gtfs_service_data.sql
@@ -18,8 +18,8 @@ tests:
   check_unique:
     - gtfs_service_data_id
 
-external_dependencies:
-  - airtable_loader: california_transit_gtfs_service_data
+dependencies:
+  - dummy_airtable_loader
 ---
 
   SELECT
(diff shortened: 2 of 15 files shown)


lauriemerrell:
@evansiroky @atvaccaro I fully acknowledge that this solution is not necessarily great since it essentially obscures the actual dependencies, but it does seem consistent with our normal approach and I'm not sure of another workaround (since Gusty seemingly didn't work with the original approach, which I think was "correct").

evansiroky:
I'll let @atvaccaro handle this one.

lauriemerrell:
Idea from @evansiroky: have more specific dummy tasks that wait for the actual individual external tasks rather than waiting for the entire previous DAG, so we can preserve the more specific task-level dependencies.

I think that could be good because that way if an unrelated upstream task fails it won't block all these.

atvaccaro (approved):
Seems fine in the current architecture.

I do think that this exposes an anti-pattern of always tying load intervals to transform/warehouse intervals. i.e. we could easily sync airflow hourly and separately run the warehouse transforms daily or whatever (which is a common pattern in the "modern data stack" with fivetran/stitch/some other SaaS data loader with less frequent transform jobs via dbt or airflow)

lauriemerrell:
Merged with general dependencies because it seems more consistent with rest of pipeline.

PR #1100, opened 2022-02-16
Jupyterhub memory increase

Nkdiaz:
As part of a pairing session with @lottspot, removed the memory limitation in the jupyterhub values.yml and created a production deployment environment


lottspot (approved):
:+1:

PR #1102, opened 2022-02-17
fix: avoid double prefixing gcs bucket url

atvaccaro:
See https://cal-itp.slack.com/archives/C0332307HKM/p1645073980149109 for discussion.

Currently, the RT validator Airflow job fails after all multiprocessing is finished, as the final status JSON file is uploaded with a double-prefixed GCS URL, i.e. `gs://gs://gtfs-data-test...`. This removes the prefix since `get_bucket()` already returns a protocol prefix.

### DIFF
--- a/airflow/dags/rt_loader/load_rt_validations.yml
+++ b/airflow/dags/rt_loader/load_rt_validations.yml
@@ -11,11 +11,11 @@ arguments:
     validate_gcs_bucket_many(
         project_id="cal-itp-data-infra",
         token="cloud",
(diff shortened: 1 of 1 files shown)


lauriemerrell (approved):
LGTM 😅 Although I do wonder how we missed this in the local testing?

atvaccaro:
> LGTM 😅 Although I do wonder how we missed this in the local testing?

That's on me, and honestly I wonder if the underlying client somewhat handles it? Cause it's clearly running and uploading files despite this.

PR #1105, opened 2022-02-18
Gtfs logger errors

Nkdiaz:
WIP to create a big query view table that analysts can query that will give detailed information of gtfs-archiver fetching/writing errors as logged in big query.  The table parses the stdout/stderr table to give information on the archiver process and the resulting error. 
It is worth mentioning that this is a stop-gap measure as eventually we will Prometheus/grafana/loki to give more detailed metrics as the presence of the exception takes less priority over the presence of the file in gcp storage.

### DIFF
--- a/airflow/dags/rt_views/gtfs_rt_thread_errors.sql
+++ b/airflow/dags/rt_views/gtfs_rt_thread_errors.sql
@@ -93,4 +93,4 @@ t2.error_time,
 t2.textpayload AS error_message
 FROM log_table t1
 JOIN stderr_table t2 ON t1.feed_file = t2.feed_file
-AND t2.error_time >= t1.start_fetch_time
+AND t2.error_time >= t1.start_fetch_time AND t2.error_time < t1.completed_write_time
--- a/airflow/dags/rt_views/gtfs_rt_thread_errors.sql
+++ b/airflow/dags/rt_views/gtfs_rt_thread_errors.sql
@@ -2,21 +2,25 @@
 operator: operators.SqlToWarehouseOperator
 dst_table_name: "views.gtfs_rt_thread_errors"
 
-description: |
+description: | Each row is a unique file_hash of a feed that generated an exception within GCP logger
 
 
 fields:
-  file_hash:
-  calitp_itp_id:
-  calitp_url_number:
-  feed_file
-  start_fetch_time:
-  completed_fetch_time:
-  start_write_time:
-  completed_fetch_time:
-  error_time:
-  error_message:
+  file_hash: unique identifier for each feed download attempt.
+  calitp_itp_id: Feed ITP ID.
+  calitp_url_number: Feed URL number.
+  feed_file: Combination calitp_itp_id, URL number, and feed URL name.
+  feed_file_name: feed url name.
+  start_fetch_time: timestamp of start_fetch thread as logged by archiver.
+  completed_fetch_time: timestamp of completed_fetch thread as logged by archiver.
+  start_write_time: timestamp of start_write thread as logged by archiver.
+  completed_write_time: timestamp of completed_write thread as logged by archiver.
+  error_time: timestamp when thread exception occured.
+  error_message: error message contained within the exception.
 ---
+
+-- note that archiver will log a completed write time regardless if fetch was successful due to them being on separate threads
+-- this will be changed in future iterations
 
 With
 start_fetch AS (
--- a/airflow/dags/rt_views/gtfs_rt_thread_errors.sql
+++ b/airflow/dags/rt_views/gtfs_rt_thread_errors.sql
@@ -1,0 +1,96 @@
+---
+operator: operators.SqlToWarehouseOperator
+dst_table_name: "views.gtfs_rt_thread_errors"
+
+description: |
+
+
+fields:
+  file_hash:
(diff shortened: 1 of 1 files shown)


airflow/dags/rt_views/gtfs_rt_thread_errors.sql
```diff
@@ -0,0 +1,100 @@
...
+log_table AS (
+    SELECT
+t1. file_hash,
+t1.calitp_itp_id,
+t1.calitp_url_number,
+t1.feed_file,
+t1. start_fetch_time,
+t1.completed_fetch_time,
+t2.start_write_time,
+t2.completed_write_time,
```
evansiroky, line 78: Please have consistent indenting.

airflow/dags/rt_views/gtfs_rt_thread_errors.sql
```diff
@@ -0,0 +1,100 @@
...
+    FROM `cal-itp-data-infra.gtfs_rt_logs.stdout`
+    WHERE textPayload like "%completed write%"
+),
+write_join_table AS (
+    SELECT * FROM start_write t3
+    JOIN complete_write t4  ON t3.file_hash = t4.c_file_hash
+),
+log_table AS (
+    SELECT
+t1. file_hash,
```
evansiroky, line 71: Is this space needed? Can it be changed to:

```suggestion
t1.file_hash,
```

airflow/dags/rt_views/gtfs_rt_thread_errors.sql
```diff
@@ -0,0 +1,100 @@
...
+t1.calitp_itp_id,
+t1.calitp_url_number,
+t1.feed_file,
+t1. start_fetch_time,
+t1.completed_fetch_time,
+t1. start_write_time,
+t1.completed_write_time,
+t2.error_time,
+t2.textpayload AS error_message
```
evansiroky, line 97: Needs indentation.

evansiroky (commented):
Just a few formatting comments.

airflow/dags/rt_views/gtfs_rt_thread_errors.sql
```diff
@@ -0,0 +1,100 @@
+---
+operator: operators.SqlToWarehouseOperator
+dst_table_name: "views.gtfs_rt_thread_errors"
```
lauriemerrell, line 3: cc @atvaccaro here - since he's been doing some investigation with the logs too, just wondering if he has thoughts on naming here

atvaccaro: Yeah I think you could leave out `thread` and probably specify that it's the archiver log errors; otherwise we may confuse it with GTFS RT _data_ errors.

airflow/dags/rt_views/gtfs_rt_thread_errors.sql
```diff
@@ -0,0 +1,100 @@
...
+
+-- note that archiver will log a completed write time regardless if fetch was successful due to them being on separate threads
+-- this will be changed in future iterations
+
+With
+start_fetch AS (
+    SELECT
+        textpayload,
+        timestamp AS start_fetch_time,
+        REGEXP_EXTRACT(textPayload, r'\[txn (.*?)\]') AS file_hash,
```
atvaccaro, line 30: should this column be called `transaction_id`? I think it's a generated uuid, not a hash of anything.

Also do you need a lazy evaluation here? `)` The UUID won't contain a bracket character.

airflow/dags/rt_views/gtfs_rt_thread_errors.sql
```diff
@@ -0,0 +1,100 @@
...
+t1.completed_fetch_time,
+t2.start_write_time,
+t2.completed_write_time,
+ FROM fetch_join_table t1 JOIN write_join_table t2 ON t1.file_hash = t2.file_hash
+),
+stderr_table AS (
+    SELECT textPayload, timestamp AS error_time,
+    REGEXP_EXTRACT(textpayload,"fetcher ([0-9]+/[0-9]+/?[a-zA-Z0-9\\.\\_\\-]+)?") AS feed_file,
+    FROM `cal-itp-data-infra.gtfs_rt_logs.stderr`
+    WHERE textPayload LIKE "%fetcher%"
```
atvaccaro, line 85: do you have an example log line that this pulls successfully?

atvaccaro (changes requested):

airflow/dags/rt_views/gtfs_rt_thread_errors.sql
Nkdiaz: good point, I had heard it referred to as a hash but it isn't actually a hash. Will rename it across both tables

airflow/dags/rt_views/gtfs_rt_thread_errors.sql
Nkdiaz: Could you clarify this statement.. pulls what? the feed file?

atvaccaro: had a quick slack convo

holly-g:
Closing this ticket due to ongoing work with RT archiver v3.

PR #1106, opened 2022-02-18
feat(gtfs-rt): add calitp files to views

lauriemerrell:
# Overall Description

Expose the existing `gtfs_rt.calitp_files` table in `views` so we can use it in Metabase. I also reordered the dependencies so that they make sense (we identify the files, then we recreate the external table, then we make the view.)

Notes about things I've done that reviewers may find controversial:
- I left this in the `rt_loader_files` DAG rather than the `views` DAG. I did this because I think we will want this `views` table to update more often than daily. After this merges I think we should consider changing this DAG to run hourly or similar. 
- I pre-extracted date, hour, minute from `calitp_extracted_at`. This is kind of duplicative but I really think everyone is going to need these columns and working with them manually in Metabase all the time would be annoying. 

## Checklist for all PRs

- [x] ~Link this pull request to all issues that it will close using keywords (see GitHub docs about [Linking a pull request to an issue using a keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)). Also mention any issues that are partially addressed or are related.~ Not directly related to any issues.

image

This PR updates the `rt_loader_files` DAG in order to add the following DAG tasks:

- `rt_views_gtfs_rt_fact_files`: pulls `gtfs_rt.calitp_files` into `views` so it can be accessed in Metabase

### DIFF
--- a/airflow/dags/rt_loader_files/rt_views_gtfs_rt_fact_files.sql
+++ b/airflow/dags/rt_loader_files/rt_views_gtfs_rt_fact_files.sql
@@ -1,0 +1,38 @@
+---
+operator: operators.SqlToWarehouseOperator
+# write to views even though this is not in views DAG so we can run more often if needed
+dst_table_name: "views.gtfs_rt_fact_files"
+
+description: |
+  Each row is a GTFS realtime file that was downloaded.
+  Note that presence of a file is not a guarantee that the downloaded file is complete or valid.
+
+
+fields:
+  calitp_extracted_at: Timestamp when this file was extracted
+  calitp_itp_id: Feed ITP ID
+  calitp_url_number: Feed URL number
+  name: File type (service_alerts, trip_updates, or vehicle_positions)
+  size: File size in bytes
+  md5_hash: Hash of file contents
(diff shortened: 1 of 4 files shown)


airflow/dags/rt_loader_files/rt_views_gtfs_rt_fact_files.sql
```diff
@@ -0,0 +1,38 @@
...
+  - external_calitp_files
+---
+
+SELECT
+  calitp_extracted_at,
+  calitp_itp_id,
+  calitp_url_number,
+  -- turn name from a file path like gtfs_rt_<file_type>_url
+  -- to just file_type
+  REPLACE(REPLACE(name, "gtfs_rt_", ""), "_url", "") as name,
```
atvaccaro, line 32: regex might be easier to read but idc

atvaccaro (approved):

atvaccaro (approved):

PR #1110, opened 2022-02-22
feat(gtfs-archiver) : added view table specific for http error codes

Nkdiaz:
Added a table to view to aid in RT-Archiver analysis. This table selects all the log messages that have a generated HTTP code error

### DIFF
--- a/airflow/dags/rt_views/gtfs_rt_extraction_html_errors.sql
+++ b/airflow/dags/rt_views/gtfs_rt_extraction_html_errors.sql
@@ -1,6 +1,6 @@
 ---
 operator: operators.SqlToWarehouseOperator
-dst_table_name: "views.gtfs_rt_extraction_html_errors"
+dst_table_name: "views.gtfs_rt_extraction_http_errors"
 
 description: |
   Each row is a unique text payload returned for a HTTP Error for the calitp RT_archiver
--- a/airflow/dags/rt_views/gtfs_rt_extraction_html_errors.sql
+++ b/airflow/dags/rt_views/gtfs_rt_extraction_html_errors.sql
@@ -26,7 +26,7 @@ WITH
     textPayload,
     timestamp,
     REGEXP_EXTRACT(textPayload, r'\[txn (.*?)\]') AS file_hash,
-    REGEXP_EXTRACT(textpayload,r'error fetching url ([a-zA-Z].*): HTTP Error') AS url_name,
+    REGEXP_EXTRACT(textpayload,r'error fetching url ([a-zA-Z].*): HTTP Error') AS download_url,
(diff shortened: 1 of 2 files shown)


airflow/dags/rt_views/gtfs_rt_extraction_html_errors.sql
```diff
@@ -8,7 +8,7 @@ description: |
 fields:
   calitp_itp_id: Feed ITP ID.
   calitp_url_number: Feed URL number.
-  textPayload: Error message that contains feed key, header length, download url, and error response.
+  token_cleaned_textPayload: Error message that contains feed key, header length, download url, and error response. API Tokens are replaced in the string with API_TOKEN
```
evansiroky, line 11: Let's try to be consistent with what kind of case is being used.

airflow/dags/rt_views/gtfs_rt_extraction_html_errors.sql
```diff
@@ -58,7 +57,7 @@ WITH
 SELECT
   calitp_itp_id,
   calitp_url_number,
-  textPayload,
+  token_cleaned_textpayload,
```
evansiroky, line 60: Let's try to be consistent with what kind of case is being used. In this instance, I would expect something like:

```suggestion
  token_cleaned_text_payload,
```

evansiroky (commented):
Just a comment about consistent casing.

airflow/dags/rt_views/gtfs_rt_extraction_html_errors.sql
```diff
@@ -0,0 +1,78 @@
+---
+operator: operators.SqlToWarehouseOperator
+dst_table_name: "views.gtfs_rt_extraction_html_errors"
```
lauriemerrell, line 3: Do we want to call this `http_errors` rather than `html_errors`?

lauriemerrell (changes requested):

airflow/dags/rt_views/gtfs_rt_extraction_html_errors.sql
Nkdiaz: ahh good call

airflow/dags/rt_views/gtfs_rt_extraction_html_errors.sql
Nkdiaz: in the gtfs_archiver it is "textPayload".. so would you prefer "textpayload" or "text_payload"

Nkdiaz: changed it to text_payload

holly-g:
Rendered obsolete by #1463

PR #1115, opened 2022-02-23
fix(payments): Correct cardinality for micropayments:transactions

### LINKED ISSUE #1068: Bug: Payments pipeline blocked due to invalid transaction assumptions
**Describe the bug**
We're currently enforcing that a transaction (essentially a tap) is only associated with one micropayment (essentially, we thought, a trip). However, as can be seen below, transactions can be associated with multiple micropayments. As a result, our pipeline cannot proceed. If we allow it to, it will result in fan-out in the final `payments_rides` table.

image

image

Currently (as of 2022 Feb 11) there is only one transaction that is breaking, so the impact of any resulting fan-out would be very small.

I think what should happen is that we:
* Relax the invariant that a device transaction can only be associated with one micropayment by removing [the uniqueness constraint on `littlepay_transaction_id`](https://github.com/cal-itp/data-infra/blob/2ee7ed8d92ae9bc26b495812e260bac601230f25/airflow/dags/payments_views_staging/stg_cleaned_micropayment_device_transactions.sql#L12-L13) and removing [the micropayment device transaction validation table](https://github.com/cal-itp/data-infra/blob/2ee7ed8d92ae9bc26b495812e260bac601230f25/airflow/dags/payments_views_staging/validate_cleaned_micropayment_device_transactions.sql)
* Remove the expectation that each `payment_rides` record only corresponds to one micropayment.
  1. In the near term, we can just use the earliest `micropayment_id` associated with the transactions
  2. In the longer term, perhaps we should remove `micropayment_id`, `adjustment_id`, and `product_id` from the rides table since these potentially become 1:many relationships. This may be the necessary impetus to switch to a more star-like schema and rethink the rides table (though it is very nice to allow agencies to download a nice wide flat version of their ride data).
* To calculate ride charges, either:
  * Sum the `charge_amount` across all micropayments associated with device transactions to arrive at the final cost of a ride (this will result in some rides with `charge_amount` values differing from the `nominal_amount` values with no immediately apparent adjustments)
  * Or, have a `debited_amount` column, a `credited_amount` column, and a `final_amount` column that is the sum of the two.

**To Reproduce**
1.  Run the following in BQ to get all transactions associated with multiple micropayments:
    ```sql
    with

    multplied_transaction_ids as (
        select littlepay_transaction_id
        from `payments.stg_cleaned_micropayment_device_transactions`
        group by 1
        having count(*) > 1
    )

    select littlepay_transaction_id, m.*
    from `payments.stg_cleaned_micropayment_device_transactions`
    join `payments.stg_cleaned_micropayments` m using (micropayment_id)
    join multplied_transaction_ids using (littlepay_transaction_id)
    order by transaction_time desc, littlepay_transaction_id, charge_type desc
    ```
    One of those will be `0d4aae68-e4ea-4319-8fd1-6932184596e9`.
2.  Run the following to get the micropayments associated with the above transaction:
    ```sql
    select *
    from payments.stg_cleaned_micropayment_device_transactions
    join payments.stg_cleaned_micropayments using (micropayment_id)
    where littlepay_transaction_id = '0d4aae68-e4ea-4319-8fd1-6932184596e9'
    ```
    This should result in two micropayments, one with `type=DEBIT` and `charge_type=flat_fare` and `charge_amount=1.75`, and the other with `type=CREDIT` and `charge_type=refund` and `charge_amount=-1.75`.

**Expected behavior**
The `rides` table still has a single record for each complete trip, and takes all the credits, debits, and adjustments into account.

mjumbewu:
# Overall Description

We're currently enforcing that a transaction (essentially a tap) is only associated with one micropayment (essentially, we thought, a trip). However, transactions _can_ be associated with multiple micropayments. Normally we would expect only 1 micropayment per transaction except in the case of refunds, where we will see an additional micropayment with `type='CREDIT'`. This PR takes the distinction between DEBIT and CREDIT micropayments into account.

Resolves #1068

## Airflow DAG Updates

This PR updates tasks across the `payments_views_staging` and `payments_views` DAGs.

* `payments_views`
  * `payments_rides`
    * Only count DEBIT micropayments as rides
    * Add a `refund_amount` column to account for refunds
  * `sbmtd_transactions`
    * Only count DEBIT micropayments
* `payments_views_staging`
  * `stg_cleaned_micropayment_device_transactions`
    * Remove the unique `littlepay_transaction_id` check
    * Update descriptions and comments
  * `validate_cleaned_micropayment_device_transactions`
    * Check that `littlepay_transaction_id` values are unique only when joined with DEBIT micropayments

## Successful Run Screenshots

Screenshot-2022-02-23-17:12:17

Screenshot-2022-02-23-17:13:33

### DIFF
--- a/airflow/dags/payments_views/payments_rides.sql
+++ b/airflow/dags/payments_views/payments_rides.sql
@@ -15,6 +15,7 @@ fields:
   issuer_country: "From payments.customer_funding_source.issuer_country"
   form_factor: "From payments.customer_funding_source.form_factor"
   charge_amount: "From payments.micropayments.charge_amount"
+  refund_amount: "If there is a refund micropayment associated with the trip, the charge_amount on that refund"
   nominal_amount: "From payments.micropayments.nominal_amount"
   charge_type: "From payments.micropayments.charge_type"
   adjustment_id: "From payments.micropayments.adjustment_id"
@@ -72,6 +73,23 @@ gtfs_routes_with_participant AS (
     JOIN `views.payments_feeds` USING (calitp_itp_id, calitp_url_number)
 ),
 
+debited_micropayments AS (
+    SELECT *
+    FROM `payments.stg_cleaned_micropayments`
+    WHERE type = 'DEBIT'
+),
+
+refunded_micropayments AS (
+    SELECT m_debit.micropayment_id, m_credit.charge_amount AS refund_amount
+    FROM debited_micropayments m_debit
+    JOIN `payments.stg_cleaned_micropayment_device_transactions` dt_debit USING (micropayment_id)
+    JOIN `payments.stg_cleaned_micropayment_device_transactions` dt_credit USING (littlepay_transaction_id)
+    JOIN `payments.stg_cleaned_micropayments` m_credit ON
+        dt_credit.micropayment_id = m_credit.micropayment_id
+    WHERE m_credit.type = 'CREDIT'
+      AND m_credit.charge_type = 'refund'
+),
+
 applied_adjustments AS (
     SELECT participant_id, micropayment_id, product_id, adjustment_id, type, time_period_type, description, amount
     FROM `payments.stg_cleaned_micropayment_adjustments`
@@ -110,6 +128,7 @@ SELECT
     v.form_factor,
 
     m.charge_amount,
+    mr.refund_amount,
     m.nominal_amount,
     m.charge_type,
     a.adjustment_id,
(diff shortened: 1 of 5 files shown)


evansiroky (changes requested):
Can you please post a screenshot of the affected DAG tasks passing on a local airflow instance?

mjumbewu:
@evansiroky added screenshots to the PR description

evansiroky (approved):

airflow/dags/payments_views/payments_rides.sql
```diff
@@ -153,7 +172,8 @@ SELECT
     t2.latitude AS off_latitude,
     t2.longitude AS off_longitude
 
-FROM `payments.stg_cleaned_micropayments` AS m
+FROM debited_micropayments AS m
+LEFT JOIN refunded_micropayments AS mr USING (micropayment_id)
 JOIN `payments.stg_cleaned_customers` AS c USING (customer_id)
 LEFT JOIN `payments.stg_cleaned_customer_funding_source_vaults` AS v
```
atvaccaro, line 178: nit: is this intended to be an inner join? if so, could it be made explicit?

atvaccaro (approved):
lgtm!

airflow/dags/payments_views/payments_rides.sql
evansiroky: I wanted to make sure these payment fixes got in tonight so I created https://github.com/cal-itp/data-infra/issues/1120 in case we want to work on this later.

PR #1081, opened 2022-02-14
feat(amplitude-ingest): Amplitude to Airflow task

angela-tran:
# Overall Description
Related to #960, closes #1046 

Merges this PR into `feat-amplitude-operator`

This PR adds to the `amplitude_to_df` function some logic to call the Amplitude Export API and read the returned zipped stream into a Pandas DataFrame. 

## What this PR does and does not do
- See the first commit [`98403dd`](https://github.com/cal-itp/data-infra/pull/1081/commits/98403ddfb5f402225407f9753651b5c67f8c933a) for the basic implementation
- This PR does not add error handling.
- This PR does not attempt to transform the raw data into any specific schema.
- This PR does not have Pytests. (We could write proper pytests which I think would be done as a separate issue.)

## Pre-requisites for testing locally

1. Change `.devcontainer/Dockerfile` line 4:
```diff
-ARG VARIANT="3.9"
+ARG VARIANT="3.8"
```
2. Export `CALITP_AMPLITUDE_BENEFITS_API_KEY` and `CALITP_AMPLITUDE_BENEFITS_SECRET_KEY` key value pairs. Get key/secret from Amplitude.

## How to test

I tested these changes myself by writing some simple, throwaway code to call the function with various start/end dates and keys I got from Amplitude. 

1. Modify `airflow/debug.py`
```diff
-import os
-import sys
+# import os
+# import sys
 
-import dotenv
+# import dotenv
 
-dotenv.load_dotenv()
+# dotenv.load_dotenv()
 
-sys.path.append(os.path.abspath("plugins"))
+# sys.path.append(os.path.abspath("plugins"))
+
+import plugins.operators.amplitude_to_warehouse as atw
+
+df = atw.amplitude_to_df("20220101T05", "20220102T05")
+print(df)
```
2. Run `python -m debug.py` and expect data to be returned:
```bash
       app                             device_id  ... $schema             processed_time
0   304110  2842c0b8-a104-847e-63b7-dab84adb1196  ...      13 2022-01-01 05:07:43.196789
1   304110  9bc14448-275d-9755-9425-61aed7b38605  ...      13 2022-01-01 05:03:55.122934
2   304110  cb836441-943a-a9a4-baaf-8f28e9603d62  ...      13 2022-01-01 05:08:51.600550
3   304110  92d55c52-da91-ab9f-957e-1bf327d915e1  ...      13 2022-01-01 05:10:04.664107
```

## Checklist for all PRs

[...]

Issue #960, opened 2022-01-19
Ingest Amplitude events for Benefits metrics in Metabase


thekaveman:
**Is your feature request related to a problem? Please describe.**

Currently, an Amplitude user must log in and manually view and/or export dashboard charts and metrics for Benefits app activity. This is not such a problem now, but we're anticipating more frequent requests from a more diverse audience for this type of information. Having it centrally located with other key Cal-ITP metrics provides for more self-service opportunities and consistent reporting across the program.

**Describe the solution you'd like**

Metabase should have enough information to be able to represent the key metrics related to activity in the Benefits app:

* Total number of eligibility checks
* Number of unique users that checked eligibility
* Total number of enrollments
* Number of unique users that enrolled
* Total number of application page views

Additionally we'd like to be able to segment by:

* Language (English, Spanish)
* Transit Agency (MST, eventually SacRT, perhaps others)

**Describe alternatives you've considered**

We currently have these metrics/charts built out in Amplitude. It doesn't seem feasible to provide accounts for anyone/everyone wishing to view the metrics, nor does it make sense to have folks jumping between various dashboard tools to get a high-level picture of Cal-ITP.

**Additional context**

I can see at least three ways to make this happen (with not much preference other than 1 coming with a $$$ tag):

1. Pay for Amplitude to get access to advanced "Data Destinations" (built in event exports), which includes a [GCS connector](https://help.amplitude.com/hc/en-us/articles/360061685151-Export-Amplitude-data-to-Google-Cloud-Storage). From there, the raw event data could move into the warehouse and then on to Metabase. Likely huge overkill unless Cal-ITP wanted to start using Amplitude across various projects.
2. Use Amplitude's [Export API](https://developers.amplitude.com/docs/export-api) via Airflow or similar, writing code to do something like what the paid Amplitude plans do from Amplitude --> GCS --> Warehouse --> Metabase
3. Use Amplitude's [Dashboard API](https://developers.amplitude.com/docs/dashboard-rest-api) to export the pre-computed metrics, maybe directly into Metabase? I don't know if that even makes sense.

holly-g:
Based on backlog grooming call with Evan, Kegan will be taking a first pass at this. @thekaveman please let us know if you need any support from us!

thekaveman:
[...]

PR #1121, opened 2022-02-24
feat(amplitude-ingest): Amplitude API export

### LINKED ISSUE #1046: Implement `amplitude_to_df` function
This is a task related to #960, and completes the **"Amplitude to Airflow"** task, as detailed below.

## Amplitude to Airflow task

1. [x] Create a module in [`airflow/plugins/operators`](https://github.com/cal-itp/data-infra/tree/main/airflow/plugins/operators) like `amplitude_to_flattened_json_operator.py`; see [`realtime_to_flattened_json_operator.py`](https://github.com/cal-itp/data-infra/blob/main/airflow/plugins/operators/realtime_to_flattened_json_operator.py) as an example
2. [x] Create a function in the module like `amplitude_to_df()` to download Amplitude data and return it in a DataFrame. This function should take as parameters all API keys, querystrings, etc. needed to get at the data, perhaps reading from `os.env` if a value is not provided. See the [Amplitude Export API docs](https://developers.amplitude.com/docs/export-api)
3. [x] Create a class in the module inheriting from `airflow.models.BaseOperator` like `AmplitudeToFlattenedJSONOperator`

machikoyasuda:
## Description

Closes #1046 as a step in closing #960

This PR merges this branch into the feature branch, `feat-amplitude-operator`. @angela-tran and @machikoyasuda paired on it.

This PR adds to the `amplitude_to_df` function some logic to call the Amplitude Export API and read the returned zipped stream into a Pandas DataFrame. 

Note: I @machikoyasuda created this PR and closed this one https://github.com/cal-itp/data-infra/pull/1081 - cos I messed up the rebasing from `main`.

## What this PR does and does not do

- See the first commit [`98403dd`](https://github.com/cal-itp/data-infra/pull/1081/commits/98403ddfb5f402225407f9753651b5c67f8c933a) for the basic implementation
- This PR uses Google style docstring format for documenting methods and classes. 

- This PR does not add error handling.
- This PR does not attempt to transform the raw data into any specific schema.
- This PR does not have Pytests. (We could write proper pytests which I think would be done as a separate issue.)

## Pre-requisites for testing locally

1. Change `.devcontainer/Dockerfile` line 4:
```diff
-ARG VARIANT="3.9"
+ARG VARIANT="3.8"
```
2. Export `CALITP_AMPLITUDE_BENEFITS_API_KEY` and `CALITP_AMPLITUDE_BENEFITS_SECRET_KEY` key value pairs. Get key/secret from Amplitude.

## How to test

I tested these changes myself by writing some simple, throwaway code to call the function with various start/end dates and keys I got from Amplitude. 

1. Modify `airflow/debug.py`
```diff
-import os
-import sys
+# import os
+# import sys
 
-import dotenv
+# import dotenv
 
-dotenv.load_dotenv()
+# dotenv.load_dotenv()
 
-sys.path.append(os.path.abspath("plugins"))
+# sys.path.append(os.path.abspath("plugins"))
+
+import plugins.operators.amplitude_to_warehouse as atw
+
+df = atw.amplitude_to_df("20220101T05", "20220102T05")
+print(df)
```
2. Run `python -m debug.py` and expect data to be returned:
```bash
       app                             device_id  ... $schema             processed_time
0   304110  2842c0b8-a104-847e-63b7-dab84adb1196  ...      13 2022-01-01 05:07:43.196789
1   304110  9bc14448-275d-9755-9425-61aed7b38605  ...      13 2022-01-01 05:03:55.122934
2   304110  cb836441-943a-a9a4-baaf-8f28e9603d62  ...      13 2022-01-01 05:08:51.600550
3   304110  92d55c52-da91-ab9f-957e-1bf327d915e1  ...      13 2022-01-01 05:10:04.664107
```

## Checklist for all PRs


atvaccaro:
Based on our conversation last week, is there a compelling reason to use a custom Operator and/or pandas here? I feel like a generic PythonOperator will work, with a function that simply downloads Amplitude zipped files and then uploads to GCS?

thekaveman:
> Based on our conversation last week, is there a compelling reason to use a custom Operator and/or pandas here? I feel like a generic PythonOperator will work, with a function that simply downloads Amplitude zipped files and then uploads to GCS?

Thanks for this feedback @atvaccaro! I think we were modeling this based on the [`RealtimeToFlattenedJSONOperator`](https://github.com/cal-itp/data-infra/blob/main/airflow/plugins/operators/realtime_to_flattened_json_operator.py). Can you point us to an example of using a generic `PythonOperator`? The use of an intermediate dataframe was mostly a convenience for appending multiple JSON files together. 

Note too that this code is only one half, the Amplitude API consumption, which is done in normal Python function outside of the custom operator. If this pattern makes sense / works with a `PythonOperator` I think we'd be happy to use that instead of a custom operator. 

And finally, as we discussed we are anticipating additional use-cases in Amplitude for other Cal-ITP projects. Does that impact our decision to use a `PythonOperator` or create a custom operator as we've (started to do) here?

airflow/plugins/operators/amplitude_to_flattened_json.py
```diff
@@ -0,0 +1,51 @@
...
+from airflow.models import BaseOperator
+
+DATE_FORMAT = "%Y%m%dT%H"
+
+
+def amplitude_to_df(
+    start: str,
+    end: str,
+    api_key: str = os.environ.get("CALITP_AMPLITUDE_BENEFITS_API_KEY"),
+    secret_key: str = os.environ.get("CALITP_AMPLITUDE_BENEFITS_SECRET_KEY"),
```
thekaveman, line 19: It seems a little awkward / unnecessary to have the default values do the environment lookup?

I think it's more natural for the defaults be `None` and do the lookup inside the function body only if needed.

thekaveman (changes requested):

atvaccaro:
Yeah a generic PythonOperator (see any `.py` file with https://github.com/cal-itp/data-infra/blob/main/airflow/dags/rt_loader/create_validation_params.py as a good example, since gusty picks up those as PythonOperators) could work fine here (for the one-off), you can still import functions, etc. Gusty really helps to reduce the gap between a PythonOperator and a custom operator; there really wouldn't be too much of a difference between a PythonOperator whose arguments are the configuration, vs a custom operator where the yaml is the configuration (and the yaml just becomes Python kwargs anyways).

I had definitely forgotten about plans for future integrations. I think a dedicated operator makes sense; I would probably scope it to download an Amplitude zip and uploading to GCS. Then, use SQL to create the external bigquery tables (gusty picks up any `.sql` file as a SQLOperator).

thekaveman:
> I think a dedicated operator makes sense; I would probably scope it to download an Amplitude zip and uploading to GCS.

That is definitely the scope we had in mind! The name `AmplitudeToWarehouse` is probably a misnomer from our previous plan/scope, so we can rename to `AmplitudeToFlattenedJSON` or similar. 

Thanks for your continued assistance here!

airflow/plugins/operators/amplitude_to_flattened_json.py
```diff
@@ -0,0 +1,51 @@
...
+        for name in export.namelist():
+            with export.open(name) as compressed_file:
+                events = gzip.decompress(compressed_file.read()).decode()
+                temp_df = pd.read_json(StringIO(events), lines=True)
+                df_list.append(temp_df)
+
+    return pd.concat(df_list)
+
+
+class AmplitudeToWarehouseOperator(BaseOperator):
```
thekaveman, line 41: Since this PR renames the file (and based on our new plan), let's also rename the operator class. Something like `AmplitudeToFlattenedJSONOperator`.

thekaveman (changes requested):

machikoyasuda:
@thekaveman https://github.com/cal-itp/data-infra/pull/1121/ is ready for re-review.

Changes:
1. Rename the class
2. Move API key and secret lookup
3. Add more detailed docstrings

thekaveman (approved):
This is great! I was able to run it locally and get output from the DataFrame! 

🚢 on to GCS! 📂

atvaccaro (approved):
lgtm!

Issue #1151, opened 2022-03-01
Epic (GTFS Schedule): Implement typecasting / type-checking and schema validity checks in views_staging


lauriemerrell:
Inspired by #1148. 

Right now, we don't check data types in a structured way in the GTFS schedule pipeline. In #1148 we were causing issues by setting data types in `gtfs_schedule_history` external table definitions that weren't actually checked or handled at all, so any violations caused a total pipeline failure.

The proper place to check and enforce data types is `gtfs_views_staging` where the `_clean` tables are created. 

To do this, we should decide how to handle data type validations like what occurred in #1148 - as an interim there we just used `SAFE_CAST` so that the data made it into `views` with a null instead of the corrupt value. 

* Should rows with illegal data types be dropped? Or should we pass the nulls through?
* If we drop rows, how do we want to handle the resulting foreign key violations?
* Are there other fields we want to `CAST`? Right now it's just stop lat/long.

PR #872, opened 2021-12-16
Fix regression causing data to not be loaded

evansiroky:
This fixes a regression created in 926fa255331d440a30b0031a189e0f542fd9fd3e that is now resulting in numerous feeds not being properly loaded in subsequent DAG tasks.

Since DAGs need to reran, I am saying this refs #848 and #849 instead of fixes them.

PR #1154, opened 2022-03-02
Hotfix stops typing

### LINKED ISSUE #1148: Bug: the data pipeline should be able to parse incomplete files
**Describe the bug**

Amtrak's new (to us) GTFS file broke the data pipeline. It appears that the stops.txt file is incomplete:

Screen Shot 2022-02-28 at 5 30 25 PM

For some reason, the CSV parsing in `gtfs_loader.gtfs_schedule_history_load` did not get flagged as having an error with a row not having enough columns, so this file was marked as being OK. This resulted in the `gtfs_loader.gtfs_schedule_tables_load` and `gtfs_schedule_history2.merge_updates` failing due to there being an unfortunate value of the very last item written to the stops.txt file.

**To Reproduce**

See error logs for [gtfs_loader.gtfs_schedule_tables_load](https://o1d2fa0877cf3fb10p-tp.appspot.com/log?execution_date=2022-02-28T00%3A00%3A00%2B00%3A00&task_id=gtfs_schedule_tables_load&dag_id=gtfs_loader) and [gtfs_schedule_history2.merge_updates](https://o1d2fa0877cf3fb10p-tp.appspot.com/log?execution_date=2022-02-28T00%3A00%3A00%2B00%3A00&task_id=merge_updates&dag_id=gtfs_schedule_history2).

**Expected behavior**

The data pipeline should not have a fatal error upon any incomplete file(s) that are included in a GTFS feed, but should flag such problems accordingly.

**Additional context**

We may end up needing to modify `gtfs_loader.gtfs_schedule_history_load`. A similar effort was reported in #691, #693, and sort of resolved with #716 and finally fixed with #872. Once a fix has been made and deployed to airflow, we should clear the affected days to get data flowing again.

lauriemerrell:
# Overall Description

This PR refactors how data types are applied for `stop_lat` and `stop_lon` to resolve #1148. 

Currently we assert that they are floats in `gtfs_schedule_history`, but there's no standard way for the pipeline to handle data type violations at that point. 

This PR specifically casts them as floats in `gtfs_views_staging` instead, which is the appropriate place to handle data typing issues. This will allow the bad Amtrak data into the warehouse as a null (there will be a record with a latitude but no longitude), but we thought that was a good first step to unblock the pipeline.

See #1151 for thoughts on subsequent steps we might want to take.

 🚨 **After merging this PR, we will need to take the following manual steps _in this order_:**
1. Manually re-run the `gtfs_schedule_history` DAG to get the data type changes for `gtfs_schedule_history` (this DAG is a once-only -- it does not have a scheduled re-run to pick up the changes)
2. Manually run the following query in production with the screenshotted query settings to propagate the `gtfs_schedule_history` data type changes into `gtfs_schedule_type2`
```
-- you have to enumerate all the columns *in order* so the INSERT statement in merge_updates will work
SELECT 
    calitp_itp_id,
    calitp_url_number,
    stop_id,
    tts_stop_name,
    CAST(stop_lat AS STRING) as stop_lat,
    CAST(stop_lon AS STRING) as stop_lon,
    zone_id,
    parent_station,
    stop_code,
    stop_name,
    stop_desc,
    stop_url,
    location_type,
    stop_timezone,
    wheelchair_boarding,
    level_id,
    platform_code,
    calitp_extracted_at,
    calitp_deleted_at,
    calitp_hash
FROM `cal-itp-data-infra.gtfs_schedule_type2.stops`
```
(substitute `cal-itp-data-infra` as the project in the "Dataset" field instead of `cal-itp-data-infra-staging`)
image

## Checklist for all PRs

`gtfs_schedule_history`: re-run manually
image

`gtfs_loader.gtfs_schedule_tables_load` -- succeeds after `schedule_history` is re-run:
image

`gtfs_schedule_history2` (only runs successfully after manual interventions described above): 
image

`gtfs_views_staging` (stop times fails because of query limit):
image

`gtfs_views` -- again stop times fails because of query limits, but `stops` and its immediate dependencies succeed:
image

This PR updates the `gtfs_schedule_history` and `gtfs_views_staging` DAGs in order to move the assertion of data types later in the pipeline.

- `gtfs_schedule_history.stops` -- change `stop_lat` and `stop_lon` to strings
- `gtfs_schedule_views_staging` -- adds safe casting of `stop_lat` and `stop_lon` to floats -- this will turn invalid data into nulls

### DIFF
--- a/airflow/dags/gtfs_views_staging/stops_clean.sql
+++ b/airflow/dags/gtfs_views_staging/stops_clean.sql
@@ -6,7 +6,9 @@ dependencies:
 ---
 
 SELECT
-    * EXCEPT(calitp_deleted_at)
+    * EXCEPT(calitp_deleted_at, stop_lat, stop_lon)
+    , SAFE_CAST(stop_lat AS FLOAT64)
+    , SAFE_CAST(stop_lon AS FLOAT64)
     , FARM_FINGERPRINT(CONCAT(CAST(calitp_hash AS STRING), "___", CAST(calitp_extracted_at AS STRING))) AS stop_key
     , COALESCE(calitp_deleted_at, "2099-01-01") AS calitp_deleted_at
 FROM `gtfs_schedule_type2.stops`
--- a/airflow/dags/gtfs_views_staging/stops_clean.sql
+++ b/airflow/dags/gtfs_views_staging/stops_clean.sql
@@ -7,8 +7,8 @@ dependencies:
 
 SELECT
     * EXCEPT(calitp_deleted_at, stop_lat, stop_lon)
-    , SAFE_CAST(stop_lat AS FLOAT64)
-    , SAFE_CAST(stop_lon AS FLOAT64)
+    , SAFE_CAST(stop_lat AS FLOAT64) as stop_lat
+    , SAFE_CAST(stop_lon AS FLOAT64) as stop_lon
     , FARM_FINGERPRINT(CONCAT(CAST(calitp_hash AS STRING), "___", CAST(calitp_extracted_at AS STRING))) AS stop_key
     , COALESCE(calitp_deleted_at, "2099-01-01") AS calitp_deleted_at
 FROM `gtfs_schedule_type2.stops`
--- a/airflow/dags/gtfs_schedule_history/stops.yml
+++ b/airflow/dags/gtfs_schedule_history/stops.yml
@@ -15,9 +15,9 @@ schema_fields:
   - name: tts_stop_name
     type: STRING
   - name: stop_lat
-    type: FLOAT64
+    type: STRING
   - name: stop_lon
-    type: FLOAT64
+    type: STRING
 
   - name: zone_id
     type: STRING


airflow/dags/gtfs_views_staging/stops_clean.sql
```diff
@@ -6,7 +6,9 @@ dependencies:
 ---
 
 SELECT
-    * EXCEPT(calitp_deleted_at)
+    * EXCEPT(calitp_deleted_at, stop_lat, stop_lon)
+    , SAFE_CAST(stop_lat AS FLOAT64) as stop_lat
+    , SAFE_CAST(stop_lon AS FLOAT64) as stop_lon
```
atvaccaro, line 11: do we want to alert on nulls somehow?

atvaccaro (approved):

airflow/dags/gtfs_views_staging/stops_clean.sql
evansiroky: It seems like the validator would flag this with a fairly serious error. Not sure what kind of action we'd want to do with an alert about null values from within here that we couldn't do with a similar validator error output.

lauriemerrell:
> I'm a little confused by this step. In the efforts associated with https://github.com/cal-itp/data-infra/pull/716 and https://github.com/cal-itp/data-infra/pull/872 I didn't manually run queries and was able to clear previous failed DAG task runs and then the updated DAG tasks would complete successfully. Also, the screenshot seems to be doing a self-reference where the destination table is the same as the table being selected from in the listed query.

@evansiroky -- The `merge_updates` task fails because it's inserting data of new types into existing columns. You need this query to manually change the data types of the existing table. That's why it extracts and re-inserts in the same table. It follows the example here: https://cloud.google.com/bigquery/docs/manually-changing-schemas#option_1_use_queries_2. 

I assume this wasn't encountered before because you weren't changing the data types of existing fields in prior fixes. 

The issue here was that `stop_lat` and `stop_lon` are treated differently than basically every other field; they are two of the ONLY fields where we defined a non-string data type in the `gtfs_schedule_history` external table. That shouldn't have been done. And undoing it requires these manual steps because the `merge_updates` task merges old and new data (unlike most other DAG tasks that re-create the new table from scratch each time.)

lauriemerrell:
We do the same thing for a few fields in `attributions`; see #1155. We should ideally fix those too but didn't want to hold this up to add them.

airflow/dags/gtfs_views_staging/stops_clean.sql
lauriemerrell: A few points here:
* `stop_lon` is only conditionally required (if `location_type` = 3 or 4 it is not required; those are uncommon but still a naive check for nulls would have false positives.)
* This feed did throw a bunch of `invalid_row_length` notices (severity `ERROR`) in its validation results.

evansiroky (approved):
Thanks for responding, I understand the reasoning now.

Issue #1186, opened 2022-03-07
Write the gusty SQL task to create/append to BigQuery table from .jsonl file


angela-tran:
This work corresponds to the [#960 bullet point](https://github.com/cal-itp/data-infra/issues/960#issuecomment-1029348771) which reads:
> In `airflow/dags/amplitude_loader` create... another that creates the external table in BigQuery. 

The relevant Google documentation is 
https://cloud.google.com/bigquery/external-data-cloud-storage#ddl,
and there are examples in [`airflow/dags/rt_loader`](https://github.com/cal-itp/data-infra/tree/main/airflow/dags/rt_loader), e.g. external_vehicle_positions.sql.

This DAG task will need to depend on the one defined in #1126.

An open question is if there are any examples in `data-infra` of appending to a BigQuery table. If not, we will need to figure out how to do that.

PR #1209, opened 2022-03-10
Amplitude: Create gusty sql table loader for amplitude

machikoyasuda:
# Overall Description
Closes https://github.com/cal-itp/data-infra/issues/1186, a part of #960 

This PR adds a gusty SQL task to create a BigQuery external table that queries from the Amplitude `.jsonl` files at the specified GCS bucket URI.

## Notes
- Data types: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types
- Removed some fields that don't apply to us (see cdba57e for more notes):
   - `global_user_properties`
   - `data`
   - `groups` and `group_properties`
   - `plan`
- Data schema from Amplitude:

```json
{
    "app": 304110,
    "device_id": "0ae92304-37f1-486c-3229-c063dbf269ad",
    "user_id": "f083aa6e-502f-402a-aaf6-d6e142caae8a",
    "client_event_time": "2021-12-09 00:45:29.430000",
    "event_id": 2,
    "session_id": 1639010649366,
    "event_type": "started eligibility",
    "amplitude_event_type": null,
    "version_name": "0.0.1",
    "platform": null,
    "os_name": null,
    "os_version": null,
    "device_brand": null,
    "device_manufacturer": null,
    "device_model": null,
    "device_family": null,
    "device_type": null,
    "device_carrier": null,
    "location_lat": null,
    "location_lng": null,
    "ip_address": null,
    "country": null,
    "language": "English",
    "library": "http\/2.0",
    "city": null,
    "region": null,
    "dma": null,
    "event_properties": {
        "provider_name": "Monterey-Salinas Transit",
        "path": "\/eligibility\/confirm"
    },
    "user_properties": {
        "referring_domain": "benefits.calitp.org",
        "user_agent": "Mozilla\/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit\/537.36 (KHTML, like Gecko) Chrome\/96.0.4664.55 Safari\/537.36",
        "referrer": "https:\/\/benefits.calitp.org\/eligibility\/confirm"
    },
    "global_user_properties": {},
    "group_properties": {},
    "event_time": "2021-12-09 00:45:29.430000",
    "client_upload_time": "2021-12-09 00:45:29.458000",
    "server_upload_time": "2021-12-09 00:45:29.465000",
    "server_received_time": "2021-12-09 00:45:29.458000",
    "amplitude_id": 332931753288,
    "idfa": null,
    "adid": null,
    "data": {},
    "paying": null,
    "start_version": "0.0.1",
    "user_creation_time": "2021-12-09 00:44:09.374000",
    "uuid": "4f652b07-5889-11ec-885b-02932cde97dd",
    "groups": {},
    "sample_rate": null,
    "$insert_id": "8421157f-9f73-4fa8-b145-5ca480c5d2cf",
    "$insert_key": "01b1faf9a29618ce2c14d9dd0d645cfeeb#108",
    "is_attribution_event": false,
    "amplitude_attribution_ids": null,
    "plan": {},
    "partner_id": null,
    "$schema": 13,
    "processed_time": "2021-12-09 00:45:30.804696"
}
```
## Checklist for all PRs


airflow/dags/amplitude_benefits_loader/jsonl_bigquery.sql
```diff
@@ -0,0 +1,74 @@
...
+      provider_name STRING,
+      path STRING
+      >,
+  user_properties STRUCT<
+    referring_domain STRING,
+    user_agent STRING,
+    referrer STRING
+    >,
+  global_user_properties {},
+  group_properties {},
```
machikoyasuda, line 46: Not sure how to format these empty arrays

airflow/dags/amplitude_benefits_loader/jsonl_bigquery.sql
```diff
@@ -0,0 +1,74 @@
...
+  event_time STRING,
+  client_upload_time DATETIME,
+  server_upload_time DATETIME,
+  server_received_time DATETIME,
+  amplitude_id INT64,
+  idfa STRING,
+  adid STRING,
+  data {},
```
machikoyasuda, line 54: Ditto here - Not sure how to format these empty arrays. Just a STRUCT?

machikoyasuda:
Screen Shot 2022-03-09 at 2 38 59 PM

machikoyasuda:
Current status:

- DAG runs properly
- BigQuery table created
- BigQuery table creation running

Current error/issue:

```
Error running query
Error while reading table: cal-itp-data-infra-staging.amplitude.benefits_events, error message: JSON parsing error in row starting at position 0: Could not convert non-string JSON value to DATETIME type. Field: client_event_time; Value: 1646873103124
```

Need to figure out how to get the `client_event_time` and other `event_time` field data types correct.
D

thekaveman:
One line of our earlier conversation with @atvaccaro and @lauriemerrell regarding this initial load was to allow BigQuery to interpret the schema, not trying to specify the whole thing until we get into the warehouse in another stage (or if BQ complained when trying to load this data). 

Would that help deal with some of the issues here with datetime, empty objects, etc.?

angela-tran:
> One line of our earlier conversation with @atvaccaro and @lauriemerrell regarding this initial load was to allow BigQuery to interpret the schema, not trying to specify the whole thing until we get into the warehouse in another stage (or if BQ complained when trying to load this data).
> 
> Would that help deal with some of the issues here with datetime, empty objects, etc.?

I tried out having BigQuery auto-detect the schema, and it ended up throwing an error. Here, it was saying the field name is invalid:
image
I think even if we wrote some code to get past that issue, there would more errors to handle (for example, `groups` is a reserved keyword in BigQuery, and that happens to be one of the keys in Amplitude's event schema.)

I am also hesitant to use schema auto-detection with Amplitude's data because I've seen the Export API return only a subset of the possible properties. (I guess it somehow decides which ones to exclude for a given data set?) From [BigQuery's docs](https://cloud.google.com/bigquery/docs/schema-detect#auto-detect), the auto-detection works by picking a random file and scanning the first 500 rows, so it seems very possible for the table to be created with only a subset of the schema we need.

**So what _should_ we do?**
Currently, I'm thinking we can utilize our [`Event`](https://github.com/cal-itp/benefits/blob/4a1336768e982f7719fe971f4411e5a6c81fecc5/benefits/core/analytics.py#L21) definitions as well as Amplitude's docs ([Export](https://developers.amplitude.com/docs/export-api#returns) and [HTTP](https://developers.amplitude.com/docs/http-api-v2#keys-for-the-event-argument)) to figure out the superset of properties, which will hopefully solve our empty-objects problem. We would still need to do something about field names that conflict with BQ reserved words. (Maybe something like the renaming in [`airtable_to_warehouse.py`](https://github.com/cal-itp/data-infra/blob/1d27b283d0e5b091ad3d2a0e0b20a65c649c2b7e/airflow/plugins/operators/airtable_to_warehouse.py#L43-L46)?)

I was able to find a resolution for the datetime error (have `pandas` write out ISO-8601 datetimes, which BQ recognizes).

lauriemerrell:
In case it's helpful, a few thoughts:
* Re: reserved keywords -- you can potentially also get around it by using backticks ([Google documentation](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#reserved_keywords)). But yes I did find the renaming option helpful in `airtable_to_warehouse` when we ran into this in the Airtable data.
* Re: data typing: We actually use two different approaches in the GTFS pipeline (just wanted to share both in case it sparks anything for the Amplitude use case). 
   * In the RT pipeline, we explicitly define the types in the external table definitions ([example](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/rt_loader/external_service_alerts.sql)). There are some options you can set ([here are the docs for JSON](https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-json#json_options)) about whether additional fields can be ignored -- we set `ignore_unknown_values` to True in the RT pipeline so that if people are using experimental fields in their RT data, we will just ignore it (I see you already have that set as well here). However this setting has also had the undesirable consequence that when we have mis-filed data (for example we are ingesting one RT feed thinking it is a different type), we just load that data into the warehouse silently. So, YMMV.
   * In the schedule pipeline, we just load everything as strings at the initial ingest (except the IDs and date metadata that we ourselves have appended and for which we are 100% confident about data type; [example](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/gtfs_schedule_history/calendar_dates.yml)). And then we handle the data typing later in the pipeline using for example `SAFE_CAST` in BigQuery.

thekaveman (approved):
This looks great!

PR #1212, opened 2022-03-11
add start fetch as an info-level log

atvaccaro:


atvaccaro:
> @atvaccaro do you still believe this change should be merged into main?

nah let's skip for now

PR #1224, opened 2022-03-15
Add Elk Grove Transit

evansiroky:
# Overall Description

This adds Elk Grove Transit to agencies.yml.

## Checklist for all PRs

## agencies.yml changes checklist

- [x] Include this section whenever any change to the `airflow/data/agencies.yml` file occurs, otherwise please omit this section.
- [x] Manually made sure any new feeds have `itp_id`s that are not duplicative
- [ ] Confirmed URIs are valid (the `move DAGs to GCS folder` GitHub action should successfully pass)
- [ ] Made sure the Airtable database has consistent information
- [x] Fill out the following section describing what feeds were added/updated

This PR updates `agencies.yml` in order to add Elk Grove Transit.

Even though this agency is now being operated by SacRT, it is not showing up in SacRT's feed. I confirmed this by running SacRT's feed and Elk Grove's feed through [gtfs-to-geojson](https://github.com/BlinkTagInc/gtfs-to-geojson) and then comparing the outputted stops and shapes in QGIS:

Screen Recording 2022-03-15 at 4 45 08 PM

@e-lo as the airtable data maintainer, the `gtfs service data` table should be updated to note that the Sacramento Regional Schedule and RT feeds do not yet contain Elk Grove Transit.

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -1604,3 +1604,11 @@ time:
     gtfs_rt_service_alerts_url: https://tularetransit.com/gtfs-rt/alerts
     gtfs_rt_trip_updates_url: https://tularetransit.com/gtfs-rt/tripupdates
   itp_id: 483
+elk-grove:
+  agency_name: Elk Grove Transit
+  feeds:
+  - gtfs_schedule_url: https://elkgrovetransit.com/gtfs
+    gtfs_rt_vehicle_positions_url: https://elkgrovetransit.com/gtfs-rt/vehiclepositions
+    gtfs_rt_service_alerts_url: https://elkgrovetransit.com/gtfs-rt/alerts
+    gtfs_rt_trip_updates_url: https://elkgrovetransit.com/gtfs-rt/tripupdates
+  itp_id: 105


evansiroky:
Well, I did some research into the history of Elk Grove and here's what I came up with:

1. Before agencies.yml in b143a4dd12020bf82b6ba787fba35465efb82a43, it appears that a Google Sheet was used to generate a list of feeds to download. At that time the Google sheet did list the City of Elk Grove having an ITP ID of 105.
2. Elk Grove was added in the first version of agencies.yml with ITP ID 5 in 7c9c9f88418715c6104fe61d2916afc24c7c8636. Side note, a whole lot of ITP IDs are null at this point.
3. Elk Grove's ITP ID was corrected along with a whole bunch of other feeds in 1f377039c5604f1b8d78f4c22e559f3cf655e69c.
4. Elk Grove was removed since they were assumed to be a part of SacRT in https://github.com/cal-itp/data-infra/pull/214
5. Airtable has an entirely different Elk Grove URL than what was originally present. I'm not sure about the origins of the data in Airtable other than the initial creation which seems to have happened 4 months ago.

So, it does seem like ITP ID 105 is actually the appropriate one to use. I have updated the City of Elk Grove Organization in Airtable accordingly.

lauriemerrell (approved):
Thanks for confirming!

Issue #1184, opened 2022-03-07
GTFS Schedule: We should check for file-level deletion in type 2 / SCD logic


lauriemerrell:
In December, we changed URL 0 for SolTrans. The old feed had a `feed_info.txt` file, while the new feed does not. However, the pipeline did not mark the old `feed_info` as deleted when that change occurred. 

This means that we still show that old `feed_info` file as active (for example it is still in `feed_info_clean` with `calitp_deleted_at = 2099-01-01`, i.e., "active"). 

This leads to information from that old feed appearing in the reports for SolTrans even though the rest of the feed is correct and updated, because some fields in the report are calculated from `feed_info`.

**We need to figure out why `feed_info` wasn't marked as deleted in this case.**

lauriemerrell:
AFAICT, we don't check in our pipeline to see if an overall *file* has been deleted. `gtfs_schedule_history2.merge_updates.py` makes row-level comparisons for files that are present on both sides, but it looks at `processed` files. So if you have a file that was deleted, it has no way to know. We do mark *feeds* as deleted (not entirely sure where but I confirmed that for example if you remove a feed from `agencies.yml` it gets a `calitp_deleted_at` date in `gtfs_schedule_dim_feeds`).

We need to make `merge_updates` handle deleted *files*.

lauriemerrell:
Affected feed is `310.0`

lauriemerrell:
@holly-g, I'm removing the `needs investigation` label -- I know what's going on here.

We need to add a second check either before or after `merge_updates` that checks the files present in a feed day over day and, if a *file* has been deleted, we need to populate its entries' `calitp_deleted_at` date. (I.e. - I don't think that investigation is needed, the cause here is known, we have a mechanism to find these.)

Affected files are saved in ` cal-itp-data-infrastaging:laurie_test_general.gtfs_schedule_type2_deleted_files_investigation`, they are the results of the following query, run on 2022-03-09: 
```
WITH check_next_file AS (
    SELECT *,
        LEAD(md5_hash)
            OVER (PARTITION BY calitp_itp_id, calitp_url_number, name ORDER BY calitp_extracted_at)
            AS next_md5_hash,
        DATE_ADD(calitp_extracted_at, INTERVAL 1 DAY) as next_day
    FROM `cal-itp-data-infra.gtfs_schedule_history.calitp_files_updates`
),
check_next_feed AS (
SELECT * EXCEPT(calitp_extracted_at),
[...]

PR #1230, opened 2022-03-16
fix(gtfs-schedule): fix get-latest logic

lauriemerrell:
# Overall Description

_🚨 Merging into `refactor_gtfs_views_staging` to merge all of #1137 fixes together._ 

Realized that my logic in #1162 might lead to unexpected outcomes because of issues like #1184 (ideally they **should** be equivalent, but right now they're not). I refactored the "get latest" logic to better match the way that it is done currently. 

Technically this doesn't touch any DAG task objects directly (just the macro that they call), but I did confirm that `gtfs_schedule` still runs (except `stop_times` which hits the query limit):

image

## Checklist for all PRs

### DIFF
--- a/airflow/dags/macros.py
+++ b/airflow/dags/macros.py
@@ -105,12 +105,11 @@ def get_latest_schedule_data(table):
 
     return f"""
         SELECT
-            * EXCEPT(calitp_deleted_at)
+            t1.* EXCEPT(calitp_deleted_at)
         FROM gtfs_views_staging.{table}_clean t1
-        LEFT JOIN gtfs_schedule_history.calitp_feed_latest t2
+        -- inner join to only get the ones that are latest load
+        INNER JOIN gtfs_schedule_history.calitp_feed_latest t2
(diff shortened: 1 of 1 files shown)


airflow/dags/macros.py
```diff
@@ -106,9 +106,11 @@ def get_latest_schedule_data(table):
     return f"""
         SELECT
             * EXCEPT(calitp_deleted_at)
-        FROM gtfs_views_staging.{table}_clean
+        FROM gtfs_views_staging.{table}_clean t1
+        LEFT JOIN gtfs_schedule_history.calitp_feed_latest t2
+            USING(calitp_itp_id, calitp_url_number, calitp_extracted_at)
         WHERE
-            calitp_deleted_at = "2099-01-01"
+            t2.is_latest_load = true
```
atvaccaro, line 113: could you maybe include a comment that you are intentionally excluding tables without any corresponding `calitp_feed_latest` row

also don't need `= true`

atvaccaro (approved):

atvaccaro (approved):

Issue #1140, opened 2022-02-25
Point Payments pipeline to cleaned data rather than raw GTFS Schedule


lauriemerrell:
See #1022 for details; right now the Payments pipeline references `gtfs_schedule` data in a few locations where it should reference `views` or at least cleaned data.

PR #1233, opened 2022-03-18
Fix payments point to gtfs views

lauriemerrell:
# Overall Description

_🚨 Merging into refactor_gtfs_views_staging to merge all #1137 fixes together._

This PR removes all references to `gtfs_schedule` data in `payments_views` and replaces them with references to more appropriate `views` data. This will address #1140 once it merges as part of #1157. 

Making this adjustment requires:

1. Changing the tables (to reference `views.gtfs_schedule_dim_routes` instead of `gtfs_schedule.routes`, for example)
2. Making the merges date-aware (since `gtfs_schedule` is latest-only, joins with it don't have to account for dates; joins with `dim` tables need to take account of the slowly-changing dimension)
3. Trimming IDs on the payments side. I did this as a precaution; I don't think that this data is affected by leading/trailing whitespace issues based on a few diagnostic queries, but it seemed better to be safe than sorry. 

This has a significant side effect (not the original intention for doing this work as part of the #1337 batch of issues): The current implementation was potentially getting inappropriate data, and was certainly getting incomplete data, by using `gtfs_schedule`. `gtfs_schedule` contains a **latest-only** cut of the GTFS schedule data. These payments tables are not latest-only. So by looking only at `gtfs_schedule`, some rows were losing their GTFS schedule annotations over time. See for example the results of the following query -- most of the results **did** have route info at the time they occurred, but the associated route is no longer in the latest dataset so the information is missing. If you look at the results of this query in `staging`, you will see 10% as many results, indicating that 90% of the transactions that were missing route info had it restored by looking at `views` (in a date-aware way) instead. 
```
select distinct route_id, participant_id, extract(date from datetime(timestamp(transaction_date_time_utc))) as dt
from `cal-itp-data-infra.views.mst_transactions_gtfs_enhanced`
where route_id is not null and route_short_name is null and route_long_name is null and route_id <> "Route Z"
order by dt desc
```

## Checklist for all PRs

image

Note: `validate_payments_ride_adjustments` fails its test in my local dev environment but I verified that it fails them on `main` too (same # of failures, so not caused by my changes)

This PR updates the `payments_views` DAG in order to update the following DAG tasks:

- `sbmtd_transactions`, `payments_rides`, and `mst_transactions_gtfs_enhanced`: use `views` instead of `gtfs_schedule` as described above

### DIFF
--- a/airflow/dags/payments_views/mst_transactions_gtfs_enhanced.sql
+++ b/airflow/dags/payments_views/mst_transactions_gtfs_enhanced.sql
@@ -26,7 +26,7 @@ device_transactions AS (
         DATETIME(TIMESTAMP(transaction_date_time_utc), "America/Los_Angeles") as transaction_date_time_pacific,
         location_id,
         location_name,
-        TRIM(route_id) as route_id,
+        route_id,
         latitude,
         longitude,
         vehicle_id,
--- a/airflow/dags/payments_views/mst_transactions_gtfs_enhanced.sql
+++ b/airflow/dags/payments_views/mst_transactions_gtfs_enhanced.sql
@@ -10,8 +10,8 @@ dependencies:
 WITH
 
 gtfs_routes_with_participant AS (
-    SELECT participant_id, route_id, route_short_name, route_long_name
-    FROM `gtfs_schedule.routes`
+    SELECT participant_id, route_id, route_short_name, route_long_name, calitp_extracted_at, calitp_deleted_at
+    FROM `views.gtfs_schedule_dim_routes`
     JOIN `views.payments_feeds` USING (calitp_itp_id, calitp_url_number)
 ),
 
@@ -26,7 +26,7 @@ device_transactions AS (
         DATETIME(TIMESTAMP(transaction_date_time_utc), "America/Los_Angeles") as transaction_date_time_pacific,
         location_id,
         location_name,
-        route_id,
+        TRIM(route_id) as route_id,
         latitude,
         longitude,
         vehicle_id,
@@ -41,6 +41,9 @@ SELECT
     t2.route_short_name
 FROM device_transactions as t1
 LEFT JOIN gtfs_routes_with_participant t2
-    USING (participant_id, route_id)
+    ON t1.participant_id = t2.participant_id
+    AND t1.route_id = t2.route_id
+    AND DATETIME(TIMESTAMP(transaction_date_time_utc)) >= t2.calitp_extracted_at
+    AND DATETIME(TIMESTAMP(transaction_date_time_utc)) < t2.calitp_deleted_at
 WHERE
     transaction_outcome = 'allow'
--- a/airflow/dags/payments_views/payments_rides.sql
+++ b/airflow/dags/payments_views/payments_rides.sql
@@ -27,7 +27,7 @@ fields:
   product_code: "From payments.product_data.product_code"
   product_description: "From payments.product_data.product_description"
   product_type: "From payments.product_data.product_type"
-  route_id: "The route_id of the first tap transaction"
+  route_id: "The route_id of the first tap transaction with a non-unknown route_id, else unknown ('Route Z')"
   route_long_name: "The route_long_name of the first tap transaction"
   route_short_name: "The route_short_name of the first tap transaction"
   direction: "The direction of the first tap transaction"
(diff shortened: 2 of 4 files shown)


airflow/dags/payments_views/mst_transactions_gtfs_enhanced.sql
```diff
@@ -26,7 +26,7 @@ device_transactions AS (
         DATETIME(TIMESTAMP(transaction_date_time_utc), "America/Los_Angeles") as transaction_date_time_pacific,
         location_id,
         location_name,
-        route_id,
+        TRIM(route_id) as route_id,
```
mjumbewu, line 29: Do you know if we _need_ to trim the `route_id`? Even if we do, I think this trimming would be better further upstream (probably in `stg_cleaned_device_transactions`) rather than here -- in the GTFS we're considering a clean `route_id` to be one that is already trimmed, correct?

airflow/dags/payments_views/payments_rides.sql
```diff
@@ -97,15 +97,17 @@ applied_adjustments AS (
 ),
 
 initial_transactions AS (
-    SELECT *
+    SELECT * EXCEPT(route_id),
+        TRIM(route_id) AS route_id
```
mjumbewu, line 101: Same note about trimmed `route_id`. I'd rather we expect to be using cleaned data at this step. I can add a `TRIM` into `stg_cleaned_device_transactions` as a follow-on.

airflow/dags/payments_views/mst_transactions_gtfs_enhanced.sql
lauriemerrell: Yeah fair, I don't think that there are any `route_id` values where it makes a difference right now but yes the data has been trimmed on the GTFS side (`dim_routes` has trimmed values). Happy to move that to `stg_cleaned_device_transactions` if that's the right place.

airflow/dags/payments_views/payments_rides.sql
lauriemerrell: If you don't mind me doing it, I'm happy to

airflow/dags/payments_views/sbmtd_transactions.sql
```diff
@@ -50,17 +50,48 @@ micropayment_device_transactions_dedupe AS (
...
+            (st.calitp_extracted_at >= r.calitp_extracted_at))
+            then st.calitp_extracted_at
+          when ((t.calitp_extracted_at >= st.calitp_extracted_at) and
+            (t.calitp_extracted_at >= r.calitp_extracted_at))
+            then t.calitp_extracted_at
+          when ((r.calitp_extracted_at >= st.calitp_extracted_at) and
+            (r.calitp_extracted_at >= t.calitp_extracted_at))
+            then r.calitp_extracted_at
+        end
+        as calitp_extracted_at,
```
mjumbewu, line 68: [`greatest`](https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#greatest) should work here, right?

airflow/dags/payments_views/sbmtd_transactions.sql
```diff
@@ -50,17 +50,48 @@ micropayment_device_transactions_dedupe AS (
...
+            (st.calitp_deleted_at <= r.calitp_deleted_at))
+            then st.calitp_deleted_at
+          when ((t.calitp_deleted_at <= st.calitp_deleted_at) and
+            (t.calitp_deleted_at <= r.calitp_deleted_at))
+            then t.calitp_deleted_at
+          when ((r.calitp_deleted_at <= st.calitp_deleted_at) and
+            (r.calitp_deleted_at <= t.calitp_deleted_at))
+            then r.calitp_deleted_at
+        end
+        as calitp_deleted_at
```
mjumbewu, line 81: And [`least`](https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#least).

airflow/dags/payments_views/mst_transactions_gtfs_enhanced.sql
mjumbewu: That would be the right place. Go for it.

lauriemerrell: Added in https://github.com/cal-itp/data-infra/pull/1233/commits/30bd391ddc996c8488aea90e88856cf4013c1479

airflow/dags/payments_views/payments_rides.sql
```diff
@@ -105,7 +105,8 @@ initial_transactions AS (
 ),
 
 second_transactions AS (
-    SELECT *
+    SELECT * EXCEPT(route_id),
+        TRIM(route_id) AS route_id
```
mjumbewu, line 109: One more unnecessary `trim`

mjumbewu (approved):
One last `TRIM` to be removed, but then looks good!

Issue #1197, opened 2022-03-08
GTFS RT: Vehicle positions mart models -- interpolate time at stops


lauriemerrell:
After we have initial vehicle positions mart models available, we want to experiment with bringing some of @edasmalchi's existing geoprocessing work into the warehouse itself.

One such item is producing a version of the data that contains a vehicle's interpolated time at each stop on the route (because vehicle position updates can occur anywhere any may not actually contain time at stops). 

This will be a delicate operation:

- We need to see if it can be implemented in SQL (we assume not and that this may require a Python microservice)
- We need to make it very clear that this data has had these assumptions imposed on it; tables or columns containing this adjusted data should be labeled as such (perhaps with a prefix like est_ or adj_ or similar indicating that it has been manipulated.
- We need to document the adjustment process very clearly and explicitly in the model documentation so that users understand the assumptions that have been imposed.

lauriemerrell:
This is still an open need in the warehouse, I believe @edasmalchi and @tiffanychu90 have done a lot of work on this on the notebook analysis side, unclear what the goals/timeline would be to get it into the warehouse. Would probably require using Python models in dbt.

PR #1261, opened 2022-03-28
Release candidate pruning workflow

lottspot:
The git flow changes merged in #1197 introduced the concept of a "release candidate" as an auto-built deployment branch. That changeset did not however include a method to cleanup these branches when no longer required. This changeset introduces said cleanup method.

### DIFF
--- a/.github/workflows/cleanup-release-candidates.yml
+++ b/.github/workflows/cleanup-release-candidates.yml
@@ -1,0 +1,32 @@
+name: Cleanup release candidate branches
+
+on:
+  delete:
+    branches:
+      - '!candidates/*'
+
+env:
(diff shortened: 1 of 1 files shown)


themightychris (approved):
I was expecting something more targeted to the deleted HEAD triggering the event, but this "Sweep all the orphaned candidates" method will probably keep things tidier in practice so, LGTM

lottspot:
@themightychris I originally attempted to implement the workflow this way, but the [delete event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#delete) does not provide any context related to the ref which is actually deleted (`GITHUB_SHA` and `GITHUB_REF` point to the head of the *default* branch in the context of this event). So this implementation is the next best thing I could think of.

PR #1263, opened 2022-03-28
Better diffing of helm deployments for the CI pipeline

lottspot:
Helm template has some weird behaviors which the development team has opted to not change or offer alternatives for which prevent its output from being useful in the context of performing a `kubectl diff`. This changeset implements a workaround which will provide idempotency for helm-based deployments.

### DIFF
--- a/ci/steps/release-helm-changes.sh
+++ b/ci/steps/release-helm-changes.sh
@@ -62,7 +62,7 @@ fi
 # helm template output cannot be passed directly to kubectl diff.
 # this workaround wraps the output of helm template in a kustomize directory to allow
 # native diffing.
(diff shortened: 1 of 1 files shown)


themightychris (approved):
Looks good.

If you're trying to make diffing better though I'd also highly recommend doing this: https://github.com/JarvusInnovations/cluster-template/blob/develop/github-workflows/k8s-prepare.yml#L32

This gets noise like managedFields changes filtered out

lottspot:
Within this context, the diffing is only used as an idempotency check, so neatness is not needed. The title of this PR is probably a bit of a misnomer since without this change the diff actually altogether fails in its role as a checker of idempotency, and this is actually a workaround fix for that.

PR #1265, opened 2022-03-28
dbt CI/CD and k8s work

atvaccaro:
This PR gets the dbt operator to run successfully on the airflow jobs node pool. It also adds some configuration options to the underlying python script and configures the operator to run against staging for now.


.github/workflows/dbt-build.yml
```diff
@@ -0,0 +1,27 @@
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          context: warehouse
+          push: true
+          tags: ghcr.io/${{github.repository}}:latest
```
themightychris, line 27:
```suggestion
          tags: ghcr.io/${{github.repository}}/warehouse:latest
```

We should append a component name here so it can be pathed as a sibling alongside any other container images we publish from this repo

atvaccaro: aha! can do

lottspot (changes requested):
Some feedback on the kubernetes content:

- The `airflow-jobs` directory should be placed under the `kubernetes/apps/manifests` path rather than the `kubernetes/apps/overlays` path. The reason for this is that the `resources:` parameter in the `kustomization.yaml` doesn't refer to any other directories, but only to yaml files within the same directory. If you look at other `kustomization.yaml` files underneath the `kubernetes/apps/overlays` path, you'll notice they all contain a reference to another directory.
- In order to facilitate CI deployment, we should add a deployment config under `ci/vars/releases`. It would be a pretty simple one in this case-- just need to specify the deploy driver (which is kustomize) and the directory the manifests live under. E.g.:

  ```bash
  # ci/vars/releases/prod-airflow-jobs.env
  RELEASE_DRIVER=kustomize
  RELEASE_KUSTOMIZE_DIR=kubernetes/apps/manifests/airflow-jobs
  ```

PR #1305, opened 2022-04-01
prevent tmpdir flakes from marking the job as failed

atvaccaro:


lauriemerrell (commented):
**question:** Looks like the [SO post you found](https://stackoverflow.com/questions/45170671/python-accessing-exception-message-of-original-exception/45170931#45170931): indicates checking `e.__cause__ or e.__context__` -- should we do that? Are we sure `e.__cause__` is sufficient?

atvaccaro:
> **question:** Looks like the [SO post you found](https://stackoverflow.com/questions/45170671/python-accessing-exception-message-of-original-exception/45170931#45170931): indicates checking `e.__cause__ or e.__context__` -- should we do that? Are we sure `e.__cause__` is sufficient?

ooh yeah good catch, thanks; we can't assume all underlying code will use `raise from`

lauriemerrell (approved):
LGTM, I am curious to see if there are underlying issues or if it is just the flake

lauriemerrell (approved):
Looks good. In case anyone is looking at this later re: `max_active_tis_per_dag`: https://airflow.apache.org/docs/apache-airflow/stable/faq.html#how-to-improve-dag-performance

PR #1307, opened 2022-04-01
rt parsing pool to limit simultaneous resource consumption

atvaccaro:


lauriemerrell (commented):
**hesitation:** I am conflicted about waiting 30 minutes for the archiver to finish. When this originally came up we said that wasn't ideal because it's trying to manage a dependency via a guess at delay and the solution for [calitp_files_process.py](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/rt_loader_files/calitp_files_process.py) was to look back the following day. 

Do we want to consider any alternatives?

1. We could wait at least an hour, just for more padding.
2. We could have the parser run for the **prior** day instead of the current day. I think that would be more robust but take more thought.

I'm wondering if we should split that part out into a separate ticket for further consideration, and just do the `pool` part in this PR.

atvaccaro:
yeah fair, I will remove that change for now

An hour certainly makes sense. I don't think we _need_ to wait a full day, and at some point it'd be nice to be able to have validation results live by the next business day

Technically we could probably be using some external file sensor or similar but that feels more fragile.

lauriemerrell (approved):

Issue #1304, opened 2022-04-01
Fix github action that is supposed to build the new gtfs-rt-parser image


atvaccaro:
I probably misinterpeted how this selector works.

```yaml
on:
  push:
    branches:
      - 'master'
    paths:
      - 'jobs/gtfs-rt-parser/**'
```

This isn't super urgent, as the image can be rebuilt manually for now. But it should be fixed next sprint IMO.

PR #1311, opened 2022-04-04
Couple small fixes from observations over the weekend

atvaccaro:


lauriemerrell (changes requested):
Everything looks good to me, but can you add a PR description that briefly summarizes the changes? (Especially because we didn't write up issues for the behavior this is addressing... I see that the comments describe some of what was happening but would be nice to surface in the PR too.)

lauriemerrell:
Also I think that #1304 can be marked as fixed by this

lauriemerrell (approved):

PR #1316, opened 2022-04-04
fix(workflows): removing old board-move-issues board automation workf…

charlie-costanzo:
# Overall Description

The board-move-issues workflow was set up to work with our previous project board configuration, and had begun to break the new project board configuration that we've implemented. The board-move-issues workflow was removed to have more predictable issue creation.

**Files effected**:
`workflows/board-move-issues.yml`
* removed board automations related to the `analytics research` and `new team member` labels in the data-infra repo


holly-g:
LGTM, thank you!

.github/workflows/board-move-issues.yml
```diff
@@ -10,18 +10,10 @@ jobs:
...
-        GITHUB_PROJECT_URL: https://github.com/cal-itp/data-infra/projects/6
-        GITHUB_PROJECT_COLUMN_NAME: Backlog
-
     - name: add-new-issues-to-repository-based-project-column
       uses: takanabe/github-actions-automate-projects@v0.0.2
       if: github.event_name == 'issues' && github.event.action == 'labeled' && github.event.label.name == 'new team member'
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        GITHUB_PROJECT_URL: https://github.com/cal-itp/data-infra/projects/6
+        GITHUB_PROJECT_URL: https://github.com/cal-itp/data-infra/projects#workspaces/data-services-current-sprint-620d721f9d810a00186ed9c9/board
```
evansiroky, line 18: This link appears to be for Zenhub, so I don't think this will work properly.

evansiroky (changes requested):
I don't believe we're using these GitHub project boards anymore, so maybe this whole `board-move-issues.yml` file can be removed.

.github/workflows/board-move-issues.yml
charlie-costanzo: yeah that's a great point @evansiroky  - I was going to try to keep the `new team member` label workflow and see if the zenhub link worked with it but it seems unlikely. I just removed the workflow and re-requested your review if you don't mind giving it an approval when you get a chance

evansiroky (approved):

PR #1272, opened 2022-03-29
implement dbt MVP, both models and infrastructure

atvaccaro:
See [the relevant TDD](https://docs.google.com/document/d/17G89089wpzGGUoud41VMbId3XubUubcvsKbAMJ289fc/edit#) for the _why_ of this PR.

## Infra changes https://github.com/cal-itp/data-infra/pull/1265
1. Set up a dbt project and poetry project to support it
2. Build and deploy a docker image that contains the project
3. Create an Airflow pod operator to execute the dbt project on k8s
4. Deletes the Airflow tasks that create the migrated tables (see below).

## Modelling changes https://github.com/cal-itp/data-infra/pull/1310
1. Creates models to replace views.gtfs_rt_fact_files and views.gtfs_rt_fact_files_wide_hourly as part of the dbt MVP (initial models). These tables were selected because they are relatively functionally isolated and simple.
2. Changes the gtfs_rt schema to views -- in the interest of being able to migrate existing content, I think we can change this later

## Data pipeline schedule considerations
Currently this operator is scheduled to run at 12 UTC, so we can more easily debug at the beginning of each day. Eventually we will want this operator to run at least 2-3 times per day, so we can exercise the code regularly and prevent engineers from having to manually re-run tasks that may flake during an otherwise once-daily execution.

## Successful testing!
```
[2022-04-04 16:39:01,951] {pod_launcher.py:149} INFO - 16:39:02  Done. PASS=2 WARN=0 ERROR=0 SKIP=0 TOTAL=2
[2022-04-04 16:39:04,155] {pod_launcher.py:198} INFO - Event: dbt-run-and-upload-artifacts.8f4783abd2384b64b377ec1883666b64 had an event of type Succeeded
[2022-04-04 16:39:04,155] {pod_launcher.py:311} INFO - Event with job id dbt-run-and-upload-artifacts.8f4783abd2384b64b377ec1883666b64 Succeeded
[2022-04-04 16:39:04,260] {pod_launcher.py:198} INFO - Event: dbt-run-and-upload-artifacts.8f4783abd2384b64b377ec1883666b64 had an event of type Succeeded
[2022-04-04 16:39:04,260] {pod_launcher.py:311} INFO - Event with job id dbt-run-and-upload-artifacts.8f4783abd2384b64b377ec1883666b64 Succeeded
[...]

PR #1317, opened 2022-04-04
also build the gtfs rt parser image on the workflow file changing

atvaccaro:
follow-up from https://github.com/cal-itp/data-infra/pull/1272#discussion_r842057004


charlie-costanzo (approved):
Approve

PR #1321, opened 2022-04-05
Point prod Airflow dbt task to cal-itp-data-infra rather than staging

lauriemerrell:
# Overall Description

This PR has the prod Airflow dbt run write to `cal-itp-data-infra` by default to maintain compatibility with existing architecture (for example, data availability in Metabase) during the transition. Previously it was writing to `cal-itp-data-infra-staging` by default, which caused the RT files views to become stale even though the dbt Airflow task ran successfully (🎉).

## Checklist for all PRs

### DIFF
--- a/warehouse/profiles.yml
+++ b/warehouse/profiles.yml
@@ -5,7 +5,7 @@ calitp_warehouse:
       &prod
       execution_project: cal-itp-data-infra
       database: cal-itp-data-infra
-      schema: views
+      schema: staging
       fixed_retries: 1
       location: us-west2
       method: oauth
--- a/warehouse/profiles.yml
+++ b/warehouse/profiles.yml
@@ -1,11 +1,11 @@
 calitp_warehouse:
-  target: staging
+  target: prod
   outputs:
     prod:
       &prod
       execution_project: cal-itp-data-infra
       database: cal-itp-data-infra
-      schema: staging
+      schema: views
       fixed_retries: 1
       location: us-west2
       method: oauth
--- a/warehouse/profiles.yml
+++ b/warehouse/profiles.yml
@@ -1,5 +1,5 @@
 calitp_warehouse:
-  target: prod
+  target: staging
   outputs:
     prod:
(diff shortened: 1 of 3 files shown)


warehouse/profiles.yml
```diff
@@ -1,11 +1,11 @@
 calitp_warehouse:
-  target: staging
+  target: prod
   outputs:
     prod:
       &prod
       execution_project: cal-itp-data-infra
       database: cal-itp-data-infra
-      schema: staging
+      schema: views
```
atvaccaro, line 8: I wouldn't change this default, I think? Helps to avoid accidentally exposing models if we change around the schemas in dbt_project.yml.

atvaccaro (changes requested):
I'm not sure if we also want to change https://github.com/cal-itp/data-infra/blob/main/warehouse/macros/generate_schema_name.sql#L9

I think we might? So that running in staging builds in the staging project but doesn't prefix the schemas.

Doesn't have to be in this PR but food for thought.

warehouse/profiles.yml
```diff
@@ -1,11 +1,11 @@
 calitp_warehouse:
-  target: staging
+  target: prod
```
atvaccaro, line 2: oh also this is just the "dbt default" target if you do `dbt run` without anything else

the target is overridden by https://github.com/cal-itp/data-infra/blob/main/airflow/dags/transform_warehouse/dbt_run_and_upload_artifacts.yml#L22 and https://github.com/cal-itp/data-infra/blob/main/warehouse/scripts/run_and_upload.py#L34

lauriemerrell: Ok, I think I fixed this?

atvaccaro: I still think the default target should be `staging` in case we ever miss passing --target somewhere, or someone accidentally runs it locally with this profiles.yml and no target

atvaccaro (approved):

PR #1332, opened 2022-04-05
docs(requirements): created docs/requirements.txt, now referenced in workflow

charlie-costanzo:
# Overall Description

Previously, the docs build process was using the `airflow/requirements.txt` file and lines of pip installations in the build docs workflow. This PR moves those packages into a `docs/requirements.txt` file which is also now referenced in the workflow file.

Files Edited:
`.github/workflows/publish-docs.yml`
* install dependencies from new docs requirements file

Files Added:
`docs/requirements.txt`
* moved packages being installed line by line in workflow to their own requirements file

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,0 +1,6 @@
+git+https://github.com/machow/siuba.git@stable
+jupyter-book==0.12.0
+# Temporary pinning of pydata-sphinx-theme to workaround Sphinx 4.3.0 error
+# https://github.com/pydata/pydata-sphinx-theme/issues/508
+pydata-sphinx-theme==0.7.2
+sphinxcontrib-mermaid==0.7.1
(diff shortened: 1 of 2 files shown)


docs/requirements.txt
```diff
@@ -0,0 +1,6 @@
+git+https://github.com/machow/siuba.git@stable
```
atvaccaro, line 1: does this work in a reqs.txt?

atvaccaro (approved):

docs/requirements.txt
charlie-costanzo: > does this work in a reqs.txt?

it did work for me, I was able to build and install siuba from `docs/requirements.txt` in my venv... do you think it's reliable enough to keep though @atvaccaro ?

Issue #1180, opened 2022-03-07
GTFS RT Validations: service alerts files not being found by external table URI definition


lauriemerrell:
Our `gtfs_rt.validation_trip_updates` table is essentially empty, and we suppressed the creation of `gtfs_rt.validation_service_alerts`, because our external table definition (trip updates example: https://github.com/cal-itp/data-infra/blob/main/airflow/dags/rt_loader/external_validation_trip_updates.sql#L25) is seemingly not finding any files (for service alerts, we thought no files existed). However, valid files exist for both feed types.

Specific example: `110.1`. All three validation files exist and are populated for 3/6:
![image](https://user-images.githubusercontent.com/55149902/157059993-eb2b86fb-dbe1-4b6d-85c9-689c75f2b2cc.png)

However, only the `vehicle_positions` table is successfully created and reflects this data.

atvaccaro:
I've started digging into this today; it's tricky to really compare gcsfs globbing against bigquery URI wildcards as I don't think they are really apples-to-apples. I'm probably going to timebox this for today so I don't prevent myself from progressing on other projects.

atvaccaro:
I am closing this as unable to reproduce. It didn't occur last night, and now service alerts are working as well. Will re-open if it occurs again.

lauriemerrell:
Issue happened again yesterday and today; seems like it's service alerts.

We *know* that there are files present for service alerts violations (for example, see the screenshot in the original issue description). We can't figure out why the external table definition isn't finding them; it should find the files every day.

lauriemerrell:
cc @atvaccaro 

We think that the underlying issues here will ultimately be best ameliorated by the longer-term work on bucket keying and the dbt transition. 

In the meantime, we have some ideas for things to try: 
1. Removing the dependency on the `rt_loader` DAG; this dependency is not necessary (since the view does not depend on `execution_date`) -- we will try this first to just get a sense for how often this issue is occurring, right now a lot of the failures are happening because the `wait_for` task is failing so it's hard to assess
2. Potentially suppressing the `service alerts` again (commenting it out), since that one seems the buggiest and the other two seem to be working 
   a. If we can get it to run for `trip updates` and `vehicle positions`, we can maybe see if it's actually finding all the files we expect which might help us understand what (if anything) is going on with the globbing that's failing for `service alerts`

atvaccaro:
Ok so I figured this out, and it wasn't a globbing issue.

[...]

PR #1349, opened 2022-04-07
run the fixed RT validator API image in airflow

atvaccaro:
This image also downloads schedule data _after_ RT, so we will avoid downloading schedule data that has no corresponding RT data.

### DIFF
--- a/airflow/dags/rt_loader/external_validation_service_alerts.sql
+++ b/airflow/dags/rt_loader/external_validation_service_alerts.sql
@@ -1,5 +1,6 @@
 ---
 operator: operators.SqlQueryOperator
+
 description: |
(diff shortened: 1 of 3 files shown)


lauriemerrell:
Just clarifying, I believe:

> This image also downloads schedule data after RT, so we will avoid downloading schedule data that has no corresponding RT data.

Refers to: https://github.com/cal-itp/gtfs-rt-validator-api/commit/422cc1dfbefa408a793ad14adba7f21b85f40d3e

And the primary purpose of this PR is to pick up the fix for https://github.com/cal-itp/data-infra/issues/1180

lauriemerrell (approved):

PR #1350, opened 2022-04-07
Replace loki-stack with current loki and promtail charts

lottspot:
The current deployment stack for loki is outdated, which causes labels to not properly parse. This change pulls in more current versions of the components of the loki stack to enable its full log parsing power.


themightychris (approved):

themightychris:
Looks good, needs a rebase though

PR #1372, opened 2022-04-13
payments: delete sbmtd_transactions airflow dag task

lauriemerrell:
## Describe your changes 

Delete the `payments_views.sbmtd_transactions` DAG task.

My date-aware updates in https://github.com/cal-itp/data-infra/pull/1233 have caused tests to start failing on this table, and @mjumbewu has indicated that since no one is actually using this table any more, it is preferable to just stop updating it than to change the logic (@mjumbewu I do just want to confirm how concerned you are about potential bad data since #1233... Do we need to create a follow-up ticket to clean up the actual table?)

## Type of change
- [ ] New feature
- [ ] Documentation

## How has this been tested?

I have not done any dedicated testing since this is just deleting a DAG task that has no dependencies.


mjumbewu:
LGTM. Regarding the broken data due to #1233, we can make a ticket to fix the data, but it should be a pretty low-priority ticket. For posterity, the rows we should throw away are these:
```sql
select
    *,
    count(*) over (partition by micropayment_id) as duplicates,
    rank() over (partition by micropayment_id order by length(all_route_ids)) as priority
from `cal-itp-data-infra.views.sbmtd_transactions`
qualify duplicates > 1 and priority > 1
```

mjumbewu (approved):

Issue #1258, opened 2022-03-28
transition GTFS Schedule (sans type 2) from gusty to dbt


atvaccaro:
AC:
Migrate the following DAGs' warehouse options (table creation) to dbt:
- [ ] `gtfs_views_staging` 
- [ ] `gtfs_schedule`
- [ ] `gtfs_views`

PR #1373, opened 2022-04-13
GTFS Schedule latest only dbt migration

lauriemerrell:
🚨 _Note that this is merging into the `gtfs-schedule-dbt-migration` branch. We are migrating this (as well as `gtfs_views_staging` and `gtfs_views`) before merging to main so we don't have a back and forth between dbt and non-dbt warehouse tasks_

## Describe your changes 

Migrates the `gtfs_schedule` DAG to dbt. Specifically:

* Modifies `warehouse/dbt_project.yml`
  *    * adds `gtfs_schedule` configuration to project
* Modifies `warehouse/macros`
  * adds macro `get_latest_schedule_data.sql` based on `airflow/dags/macros.py`: `def get_latest_schedule_data()`
* Creates the `models/gtfs_schedule_latest_only` folder, containing:
   * `_gtfs_schedule_latest_only.yml` (setup / docs / tests)
   * `*.sql` files for each existing table in the `gtfs_schedule` BQ dataset -- core SQL is identical to existing DAG tasks, except:
      * Uses dbt format of defining source/reference rather than direct reference to table as we do in gusty
      * Also uses dbt convention of adding a final `SELECT * FROM` 
      * Note: there is now a row count discrepancy found in the following tables when compared to production due to the discovery of duplicates and deduplication of rows in the dbt migration process.
        * Affected tables: `attributions`, `fare_rules`, `transfers`, `trips`
* 🚨 Maybe controversial: we're still materializing everything as tables, just in the interest of making as few changes as possible in this initial migration. I think we can revisit what is a view vs. table later?
* Deletes the `gtfs_schedule` Airflow DAG and its associated latest-only macro

Partially addresses #1258 (one of three DAGs / datasets to be migrated)

## Type of change
- [x] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) - deletes a DAG?
- [x] Documentation (dbt documentation)

## How has this been tested?
* Verified I can run `dbt run` 
* Verified `dbt test` passes
* Verified `dbt docs generate` runs


atvaccaro (approved):
lgtm!

I assume this would be easier in a file-level versioning world?

lauriemerrell:
> I assume this would be easier in a file-level versioning world?

Not sure it would be all that different... I think no matter what we'll basically have a pretty simple filter macro to grab these.

PR #1383, opened 2022-04-15
actually run dbt test, and deploy the docs!

atvaccaro:
Note this still deploys to a preview URL; will deploy to the prod site when we feel comfortable (i.e. when @lauriemerrell has merged her GTFS work I think?)


lauriemerrell (commented):
LGTM, just two questions:
1. Confirming we are fine w docs being public...? 
2. What will need to be changed to deploy to prod?

atvaccaro:
> 1. Confirming we are fine w docs being public...?

The project and our existing docs are both public, so I think so. I will remove `run_results.json` as it's theoretically possible that it logs some information in exceptions.

> 2. What will need to be changed to deploy to prod?

Just passing a `--prod` flag and then optionally getting a calitp.org subdomain CNAME'd at the actual prod netlify URL, if we want.

lauriemerrell (approved):

PR #1391, opened 2022-04-18
fix bad type usage

atvaccaro:
Pathlib is super useful, but unfortunately path objects don't implement `__iter__` so we can't do checks like `if some_str in path`. I failed to test this properly prior to deploying https://github.com/cal-itp/data-infra/pull/1383.

### DIFF
--- a/warehouse/scripts/run_and_upload.py
+++ b/warehouse/scripts/run_and_upload.py
@@ -82,7 +82,7 @@ def run(
 
             # avoid copying run_results is unnecessary for the docs site
             # so just skip to avoid any potential information leakage
-            if "run_results" not in artifact:
+            if "run_results" not in str(artifact):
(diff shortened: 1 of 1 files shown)


lauriemerrell (commented):
Can you add a description outlining the error that was encountered that led to this fix? (Might help with searching later for similar errors etc.)

lauriemerrell (approved):
Thanks, LGTM. 

For posterity, exact error was: `TypeError: argument of type 'PosixPath' is not iterable`, encountered [here](https://o1d2fa0877cf3fb10p-tp.appspot.com/log?dag_id=transform_warehouse&task_id=dbt_run_and_upload_artifacts&execution_date=2022-04-17T12%3A00%3A00%2B00%3A00) on 3rd attempt

PR #1409, opened 2022-04-19
fix(jupyter): removed service key gcloud-auth for jupyter users in values.yml

### LINKED ISSUE #1376: Switch Jupyter/calitp-py to using google oauth instead of service account
Currently, there's an env var set in the JupyterHub values.yaml `CALITP_SERVICE_KEY_PATH`. This is eventually used in calitp-py's `get_engine()` function. We are looking to add some new Jupyter users who should not have access to BigQuery by default, so we should remove this key from Jupyter, or at least have a custom spawner check by user for access.

In addition, we may want to consider re-configuring "warehouse-users" in the Cal-ITP github org, as that is the group that grants access to Jupyter right now. The name is misleading if it's no longer aligned with BigQuery access.

charlie-costanzo:
# Overall Description
`kubernetes/apps/charts/jupyterhub/values.yaml`
* removed service key gcloud-auth for jupyter users to in `values.yml` disassociate jupyter from the warehouse and facilitate use by a larger group of caltrans users

closes #1376 

## Checklist for all PRs

### DIFF
--- a/kubernetes/apps/charts/jupyterhub/values.yaml
+++ b/kubernetes/apps/charts/jupyterhub/values.yaml
@@ -17,18 +17,6 @@ jupyterhub:
     cpu:
       limit: 1
       guarantee: 0.5
-    storage:
-      extraVolumes:
-      - name: gcloud-auth
-        projected:
-          sources:
-            - secret:
-                name: jupyterhub-gcloud-service-key
-      extraVolumeMounts:
-        - name: gcloud-auth
-          mountPath: /usr/local/secrets
-    extraEnv:
-      CALITP_SERVICE_KEY_PATH: /usr/local/secrets/service-key.json
     lifecycleHooks:
       postStart:
         exec:


lottspot:
Will this break required functionality for existing users?

charlie-costanzo:
> Will this break required functionality for existing users?

Andrew and I did some testing this morning and it shouldn't. We just deployed the helm chart upgrades and on server restart the gcloud config is populating
Screen Shot 2022-04-19 at 3 22 25 PM

atvaccaro:
In this case, the relevant check would be testing that SQL queries in Jupyter notebooks work.

There is a warning about a quota project, which should be resolved with an app default login.
image

After `./google-cloud-sdk/bin/gcloud auth application-default login`:
image

charlie-costanzo:
> In this case, the relevant check would be testing that SQL queries in Jupyter notebooks work.
thanks for the clarification Andrew

atvaccaro (approved):

lottspot (approved):

lottspot:
Thanks for the clarifications y'all!

PR #1413, opened 2022-04-20
if using the prod account, deploy to prod netlify

atvaccaro:


atvaccaro:
> Pedant all day today -- can you add a PR description?

Yes! Sorry being lazy.

> Also.... Just want to confirm why we are doing `target.startswith()` -- that way it will match `prod` or `prod-service-account` (or whatever it's called), is that right?

Yep exactly, I'm planning on having that prefix be the "environment" long-term.

lauriemerrell (approved):

PR #1415, opened 2022-04-21
also sync metadata to the schedule metabase database

atvaccaro:
Quick follow-up now that https://github.com/cal-itp/data-infra/pull/1369 is merged.

We actually have two Metabase databases right now that contain GTFS-related tables in the warehouse. This PR simply tries to sync dbt documentation to the second database.

### DIFF
--- a/warehouse/scripts/run_and_upload.py
+++ b/warehouse/scripts/run_and_upload.py
@@ -98,24 +98,25 @@ def run(
             subprocess.run(args)
 
     if sync_metabase:
-        subprocess.run(
-            [
-                "dbt-metabase",
-                "models",
-                "--dbt_manifest_path",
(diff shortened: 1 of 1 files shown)


lauriemerrell (approved):
I don't suppose there's a way to sync specific datasets in these calls? Just because their logging is so verbose, this is probably going to be a little annoying but not a big deal.

lauriemerrell:
Ope sorry, didn't realize this was still in draft.

atvaccaro:
> I don't suppose there's a way to sync specific datasets in these calls? Just because their logging is so verbose, this is probably going to be a little annoying but not a big deal.

beat you to it 😂

lauriemerrell (approved):
LGTM 🙏

PR #1418, opened 2022-04-21
Payments dbt migration

### LINKED ISSUE #1257: transition payment models from gusty to dbt
### User Story
As a Data Engineer, I want to transition views for payments data from gusty to dbt, so that extraction steps are done separately from analytics engineering/transformation steps per the dbt integration.

### Work Involved
Implement models in `data-infra/airflow/dags/payments_views/` in dbt

### Acceptance Criteria
Idempotent dbt processes that transform extracted payments data to the models currently developed by payments view DAGs

mjumbewu:
:rotating_light: This PR is ready for review, aside from:
- upon review, there is actually a discrepancy of 20 rows between the `payments_rides` table being generated by the airflow dag and the new dbt model - I'm having a hard time determining if this is anticipated or abnormal
  - Digging a little deeper - the uniqueness tests that are failing in both tables have 20 more results in the airflow dag table than the dbt model table

# Description

Transition payment models from Gusty to DBT.

- [x] Created model folders corresponding to Airflow DAGs `payments_views_staging` and `payments_views`. 
  - These will not be ideal going forward, as we'll likely want to separate the transformations for each vendor and `UNION` the results in `payments` models after cleaning.
- [x] Copied these models over verbatim, **with modifications only to get them running in DBT and passing linting**.
  - The only payments views that were migrated so far in this PR are `payments_feeds` and 'payments_rides` 
- [x] Add tests corresponding to Gusty validations -- no new tests.
  - [x] Persisted the results of these tests into bq schema `dbt_test__audit``
- [x] Annotated table/field names and descriptions in each folders' YAML schema file. Pulled table/field descriptions primarily from Littlepay's schema documentation.
- [x] Added source schema YAML file, and all enrich tables are using source instead of the source tables directly

Resolves #1257 

## Notes:
- This PR's scope sought to migrate the airflow payments dags to dbt models exactly as they were, making changes only to get the models working in dbt. However, over the course of troubleshooting I did break out the `payments_rides` view into a more modular model called `payments_rides_refactor.sql` in the [payments_rides_refactored](https://github.com/cal-itp/data-infra/tree/payments_rides_refactored) branch, along with a number of intermediate files, to potentially be used in the future.
- All files in the airflow `payments_views_staging` dags directory have been deleted except for the metadata file and the directory itself
- The only files deleted from the `payments_views` directory are `payments_rides`, `payments_feeds`, `validate_payments_ride_adjustments`, and `validate_payments_ride_counts`
- All of the staging/enriched and cleaned tables are persisted as views except for the following which are persisted as tables to prevent excessive cte errors in `payments_rides`: `stg_cleaned_device_transaction_types`, `stg_cleaned_micropayment_adjustments`, `stg_cleaned_micropayment_device_transactions`, and `stg_cleaned_micropayments`
- I'm curious if down the line the current enriched tables should live as intermediate or base tables
- Tables do not have introduced unique keys with unique/null tests, this should happen with refactoring
- Tables do not have relationship tests, something else that should likely be added in refactoring
- I wasn't able to migrate over the description metadata from the tests tables that previously lived in gusty, but they live as comments in the dbt test files

## Currently failing tests**:
**Generic test in YAML files**:
- unique_payments_rides_littlepay_transaction_id
  - Got 41 results, configured to fail if != 0
- unique_payments_rides_micropayment_id
  - Got 41 results, configured to fail if != 0

**Custom tests in `tests` directory**:
- validate_cleaned_device_transaction_pairs_common_fields
  - Got 610 results, configured to fail if != 0
- validate_payments_ride_counts
  - Got 4 results, configured to fail if != 0
  - **This is failing because the query that backs it isn't actually set up to evaluate to 0**, as custom dbt tests typically do, it's set to return information. It's currently still living in the dbt tests directory and failing, do we want to instead turn this into a view?

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
- The new models have been verified to produce the same shape and size results as the existing airflow dags
- **NOTE**: upon review, there is actually a discrepancy of 20 rows between the `payments_rides` table being generated by the airflow dag and the new dbt model - I'm having a hard time determining if this is anticipated or abnormal
  - Digging a little deeper - the uniqueness tests that are failing in both tables have 20 more results in the airflow dag table than the dbt model table


mjumbewu:
Keeping a log of follow-on things to be addressed here:
* In the latest Littlepay documentation, the field descriptions on the `authorisations` table are not for fields in the `authorisations` table. It looks like they are `micropayments` fields. For now, using descriptions from the previous schema docs.

mjumbewu:
Things to be sure to sanity-check before/during review:
* All enrich tables are using `source` instead of the source tables directly

charlie-costanzo:
@mjumbewu -- I'm not sure why I'm not able to tag you as a reviewer (maybe because you started the PR?) but I wanted to flag it for you

warehouse/dbt_project.yml
```diff
@@ -48,6 +48,14 @@ models:
       +materialized: table
       schema: views
 
+    payments_views_staging:
+      +materialized: view
+      schema: payments
+
+    payments_views:
+      +materialized: table
+      schema: views
```
atvaccaro, line 57: nit: do you mind adding TODO comments that `payments_views_staging` should probably end up in `staging` and `payments_views` should end up in `mart_payments`?

warehouse/models/docs/_docs_payments_views.md
```diff
@@ -0,0 +1,41 @@
+{% docs customer_funding_source_customer_id %}
```
atvaccaro, line 1: are these docs actually referenced in YAML? they won't show up anywhere without that

warehouse/models/payments_views/payments_feeds.sql
```diff
@@ -0,0 +1,25 @@
...
+    FROM UNNEST(
+        ARRAY<
+            STRUCT<
+                calitp_itp_id INT64,
+                calitp_url_number INT64,
+                participant_id STRING
+            >
+        > [
+
+            (208, 0, 'mst'),
```
atvaccaro, line 16: I think you can select from a list of value tuples? Idk

warehouse/models/payments_views_staging/stg_cleaned_authorisations.sql
```diff
@@ -0,0 +1,11 @@
+WITH stg_cleaned_authorisations AS (
+    SELECT DISTINCT * EXCEPT (
```
atvaccaro, line 2: I don't love `select distinct *` models; is there a primary key we can use to de-dupe instead?

warehouse/tests/validate_payments_ride_counts.sql
```diff
@@ -0,0 +1,62 @@
...
+
+--tests:
+--  check_empty:
+--    - "*"
+
+WITH payments_rides_counts AS (
+    SELECT
+        'on' AS type,
+        COUNTIF(charge_type = 'complete_variable_fare') AS val
+    FROM views.payments_rides UNION ALL
```
atvaccaro, line 15: these are the `union all` that are failing the linter

atvaccaro (approved):
lgtm, I will defer on the actual modeling :)

warehouse/models/docs/_docs_payments_views.md
charlie-costanzo: yep - I added them to the YAML, and it looks like they're persisting into the docs, unless I'm missing something
Screen Shot 2022-07-26 at 12 05 55 PM

warehouse/models/payments_views_staging/stg_cleaned_authorisations.sql
charlie-costanzo: there is not, aside from `calitp_hash`, which I believe we're trying to get away from? would it be worthwhile to use it here?

j-meelah (approved):

warehouse/models/payments_views_staging/stg_cleaned_authorisations.sql
atvaccaro: ah no that's fine then, we can leave for now

warehouse/dbt_project.yml
```diff
@@ -48,6 +48,16 @@ models:
       +materialized: table
       schema: views
 
+    # To-do: should probably end up in `staging`
+    payments_views_staging:
```
mjumbewu, line 52: **suggestion:** When there's no inherent ordering to a list (like this list of schemas) I suggest ordering things alphabetically. It removes some of the mental work necessary to scan the list.

I see that the `marts` schema already violates this, so I don't know whether an ordering has previously been discussed.

PR #1420, opened 2022-04-22
dbt docs reference python 3

lauriemerrell:
Tiny docs change for the dbt README -- I accidentally used just `python -m` in #1419 which will have people running things with Python 2.


atvaccaro (approved):
I blame apple

PR #1433, opened 2022-04-27
move external table creation to @once dag

atvaccaro:
Currently these external tables fail due to AUTO hive partition schema detection timeouts. Move them out permanently, as this unblocks the dbt DAG with minimal fuss.


lauriemerrell (approved):
TODO: move `gtfs_schedule_history` into this new DAG

PR #1457, opened 2022-04-27
add pyup.yml back

atvaccaro:


lauriemerrell (approved):
LGTM. 

For posterity: undoing change that was made here: https://github.com/cal-itp/data-infra/pull/1433. Caused Pyup to open a bunch of PRs. We do still intend to retire Pyup in favor of Dependabot but need to do more groundwork first.

Issue #1429, opened 2022-04-26
log DecodeErrors as files in the RT pipeline


atvaccaro:
We aren't doing anything with DecodeErrors other than logging them. To satisfy the RT validation report requests, we need to be able to count DecodeErrors over time. Since a DecodeError means we can't output a parsed file at all, we should just output an error file instead.

Issue #1410, opened 2022-04-19
Move RT parsing to run hourly in smaller batches, and implement hive keys


atvaccaro:
Also make any other necessary adjustments to the overall architecture to achieve below AC. One thing to note is that globbing is going to be rough here; we'll be doing it 24 times a day instead of 1.

See https://github.com/cal-itp/data-infra/issues/1112 as potentially related.

AC:
1. External tables readable by dbt; one table per RT type
2. These tables must sit on top of hive partitioned GCS JSONL files.
3. A specific RT parsing DAG exists that runs hourly.
4. One output file is written per input file, rather than the current aggregated files.
5. https://github.com/cal-itp/data-infra/issues/1309 is closeable

Optional:
1. Consider flagging DecodeErrors more loudly.

lauriemerrell:
Would this include backfilling to re-process old data?

atvaccaro:
> Would this include backfilling to re-process old data?

Backfilling parsed data, yes. I think the easiest way will be an actual backfill in Airflow vs. writing a migration script.

atvaccaro:
this is done and backfill is running! making a new ticket for decoderrors

Issue #1345, opened 2022-04-07
Change RT validation to output raw (or minimally transformed) data


atvaccaro:
somewhat related to https://github.com/cal-itp/data-infra/issues/1083

The RT validator already outputs JSON which BigQuery should be able to read. Currently, the validator pipeline transforms the JSON into pre-aggregated tabular data and serializes as parquet. This has a number of downsides, including (I think) empty parquet files breaking external tables (https://github.com/cal-itp/data-infra/issues/1180) and the loss of granularity inherent in aggregation.

lauriemerrell:
cc @holly-g - I think we should try to get to this next sprint. See https://github.com/cal-itp/gtfs-rt-validator-api/pull/7#issuecomment-1092015491

holly-g:
👍  queued up. thanks for flagging, @lauriemerrell !

Issue #1309, opened 2022-04-01
GTFS RT: Have we missed protobuf files in our `rt_loader.parse_*` jobs due to globbing too close to midnight?


lauriemerrell:
Realized today that we did not assess whether #1123 also affected the globbing in the `rt_loader.parse_*` jobs. 

The problem is that the RT archiver takes some time to write its files. If the globbing in `rt_loader` is done promptly at midnight, its list of files to parse may be missing files from the 2300 hour UTC that haven't finished writing yet. 

Unless the job is re-run later, those files will not be picked up.

I propose a few actions to resolve this:

- [ ] Investigate prevalence: we can compare all the timestamps from `views.gtfs_rt_fact_files` with timestamps in the actual external tables (would have to impute them from `calitp_filepath`) to see what files are missing
- [ ] Address the actual issue: short to medium term. A few options:
   * Delay running the `parse` jobs until 30 minutes or an hour after midnight, to give time for the archiver to finish (not necessarily preferred; see [this comment](https://github.com/cal-itp/data-infra/pull/1307#pullrequestreview-929315268))
   * Have the `parse` jobs process the prior, rather than the current, day -- this is what we did for `rt_loader_files.calitp_files_process`, but see [this comment ](https://github.com/cal-itp/data-infra/pull/1307#issuecomment-1086249859) for potential concerns
- [ ] Address the actual issue: long term -- worth considering @j-meelah's suggestions around streaming. Can we process this data **as it comes in**?

atvaccaro:
As discussed in a huddle, processing hourly can help with this.

PR #1458, opened 2022-04-27
Aggregate processed RT files to hourly per URL and validate all of RT

atvaccaro:
# Description

Changes the new RT parsing to aggregate to hourly. We're more limited by BigQuery external partitioned tables than I thought; specifically we need to keep our total files under 10 million, which we easily surpass with RT. Aggregating to the hour keeps us under that limit.

In addition, this also bundles together validation at that hourly interval, so we can validate 100% of RT data!

Both parsing and validation now use the same image and pod operator, as there is substantial overlap between the code that wasn't worth abstracting into an installable library.

**One callout:** This PR uploads errors in JSONL format to GCS, but doesn't fail the Airflow job itself. This would be better served by something like Sentry, where we can log errors to Sentry (with appropriate alerting) but avoid re-trying Airflow jobs automatically, etc.

Actually resolves https://github.com/cal-itp/data-infra/issues/1410
Resolves https://github.com/cal-itp/data-infra/issues/1309
Resolves https://github.com/cal-itp/data-infra/issues/1345
Resolves https://github.com/cal-itp/data-infra/issues/1429

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

Ran a single hour backfill locally.
```
[2022-05-03 16:52:44,092] {dagrun.py:477} INFO - Marking run <DagRun parse_and_validate_rt @ 2022-04-25T18:30:00+00:00: backfill__2022-04-25T18:30:00+00:00, externally triggered: False> successful
[2022-05-03 16:52:44,098] {backfill_job.py:388} INFO - [backfill progress] | finished run 1 of 1 | tasks waiting: 0 | succeeded: 6 | running: 0 | failed: 0 | skipped: 0 | deadlocked: 0 | not ready: 0
[2022-05-03 16:52:44,102] {backfill_job.py:831} INFO - Backfill done. Exiting.
```

## Screenshots (optional)
image


airflow/plugins/operators/external_table.py
```diff
@@ -41,9 +41,11 @@ def _bq_client_create_external_table(
             len(source_objects) == 1
         ), "cannot use hive partitioning with more than one URI"
         opt = bigquery.external_config.HivePartitioningOptions()
+        # _Strongly_ recommend using CUSTOM mode and explicitly-defined
+        # key schema for more than a trivial number of files
         opt.mode = hive_options.get("mode", "AUTO")
         opt.require_partition_filter = hive_options.get(
-            "require_partition_filter", False
+            "require_partition_filter", True
```
lauriemerrell, line 48: Ok so just to confirm I could override this, for example for small Airtable data?

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -129,51 +117,75 @@ def identify_files(glob, rt_file_type: RTFileType, progress=False) -> List[RTFil
...
-        if pbar:
-            pbar.write(msg)
+            for rt_file in hour.source_files:
+                feed = gtfs_realtime_pb2.FeedMessage()
+
+                try:
+                    with open(os.path.join(tmp_dir, rt_file.path), "rb") as f:
+                        feed.ParseFromString(f.read())
+                    parsed = json_format.MessageToDict(feed)
+                except DecodeError:
```
lauriemerrell, line 147: Do we want to just add an error file in this refactor? Re #1429

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -200,32 +212,42 @@ def main(
...
+    feed_hours = [
+        RTAggregation(
+            hive_path=hive_path,
+            source_files=files,
+        )
+        for hive_path, files in feed_hours.items()
+    ]
 
     if limit:
         structlog.get_logger().warn(f"limit of {limit} feeds was set")
```
lauriemerrell, line 231: Should this say "hours" rather than "feeds" now?

lauriemerrell (commented):
I think `yesterday` as defined on line 33 of `gtfs_rt_parser` is now unused, can maybe delete?

Just a few questions.

jobs/gtfs-rt-parser/gtfs_rt_parser.py
lauriemerrell: I know I said that I'm not sure we need all the history, but if you're going to run the backfill anyway....?

airflow/plugins/operators/external_table.py
atvaccaro: yes! for safety's sake I wanted to make it the default, but it's very over-rideable in the yml

jobs/gtfs-rt-parser/gtfs_rt_parser.py
atvaccaro: it's still feeds, it's just that N is now feeds rather than all records (the log was inaccurate before tbh)

i.e. each `feed_hours` entry is one hour of one feed

jobs/gtfs-rt-parser/gtfs_rt_parser.py
atvaccaro: Mixed feelings but yes, probably.... otherwise we'd have to backfill yet again. I'll come up with something.

lauriemerrell: Could be a good chance to establish a pattern for metadata/error files in the new paradigm. If you think it's too much of a headache, we can defer, but seems like it would be nice to just do it now if we can.

atvaccaro: I've added a generic "RT processing error" concept in https://github.com/cal-itp/data-infra/pull/1423 that gets uploaded in the same manner as other files.

lauriemerrell (commented):
I agree that the pydantic structure for file types makes the code more readable, I like it. Thanks for establishing that pattern.

A few inline comments and questions. 

Overall thing: I am still concerned/confused about the degree of intertwining between the validation and the parsing are here. They are logically separate operations and I don't think that it does us favors long term to conflate the two too closely. I understand that there is convenience to running them together but I really don't want us to lose sight of the fact that they are in fact separate things. I think there is room to adjust that balance here to get the best of both worlds.  Like, it seems to me like the main benefit of doing both together is mostly just the globbing, right? It seems possible to leverage that (glob once) without coupling the rest too tightly. 

I think that you already know my concerns but just to be explicit / for any other readers:
1. If (when) we update validator versions, if we decide to run a backfill with the new validator, it is not appropriate that we would need to re-parse that data as well as re-validating it.
2. If we discover a bug in our parsing logic, similarly, we shouldn't have to re-validate that data. 

Since these operations **are** big and expensive, it seems like it's always in our best interest to give ourselves as granular control as possible and limit the extent to which separate operations are combined (while still trying to avoid wasteful things like globbing everything twice all the time.)

A few specific things I would change based on this:

1. (Probably uncontroversial) I think it would be good for the naming of the Python script and the Airflow DAG tasks to be changed to reflect the fact that they now include validation. 
2. I think it would be better if there were explicit flags that were passed (implemented in this PR) that control whether we parse, validate, or both. I think it would make more sense to have separate functions that are called if possible. Like it seems like it's already pretty close to that with the `if_validate` checks? So is it possible to separate a bit more? 
3. I think that there should be separate error files for validation vs. parsing.

I think that this is possibly a genuine difference of opinion but I feel moderately strongly about this. 

Also -- can you update the PR description to include the stuff from #1423?

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -123,109 +216,359 @@ def identify_files(glob, rt_file_type: RTFileType, progress=False) -> List[RTFil
...
+            data_records.append(
+                {
+                    "header": parsed["header"],
+                    # back and forth so we use pydantic serialization
+                    "metadata": json.loads(rt_file.json()),
+                    **record,
+                }
+            )
+
+        if validate:
```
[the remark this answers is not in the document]
atvaccaro: Parse and validate were split out into separate functions entirely.

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -123,109 +216,359 @@ def identify_files(glob, rt_file_type: RTFileType, progress=False) -> List[RTFil
...
+            except FileNotFoundError:
+                if verbose:
+                    log(
+                        f"WARNING: no validation output found in {str(rt_file.timestamped_filename)}",
+                        fg=typer.colors.YELLOW,
+                        pbar=pbar,
+                    )
+                continue
+
+            for record in records:
```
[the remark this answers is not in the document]
atvaccaro: Yes, split out now.

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -123,109 +216,359 @@ def identify_files(glob, rt_file_type: RTFileType, progress=False) -> List[RTFil
...
+                    exception=str(e),
+                    hive_path=hour.hive_path,
+                    body=e.stderr,
+                )
+            )
+            validate = False
+
+    data_records = []
+    validation_records = []
+    for rt_file in hour.source_files:
```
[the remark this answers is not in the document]
atvaccaro: Yes, split out now.

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -123,109 +216,359 @@ def identify_files(glob, rt_file_type: RTFileType, progress=False) -> List[RTFil
...
+    # fetch and zip gtfs schedule
+    zipname = f"{dst_path[:-1]}.zip"
+    log(
+        f"Fetching gtfs schedule data from {gtfs_schedule_path} to {zipname}",
+        pbar=pbar,
+    )
+
+    files = [
+        "agency.txt",
+        "calendar_attributes.txt",
```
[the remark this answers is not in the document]
lauriemerrell: Ok I may have had a change of heart here (sorry)... I feel weird about us filtering this list at all. If you look in `cal-itp-data-infra.views.gtfs_schedule_dim_files` that should have a list of every file we've ever encountered, I kind of think we should stick with that. 

I actually think my preferred approach here would be to zip everything except what we know we put there: `validation.json` and the `processed/` subdirectory. I'm not sure how feasible it would be to do that, probably less performant but I think that that's like intellectually the correct thing.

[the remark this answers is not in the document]
atvaccaro: I think that's probably fair. I'll make the change.

atvaccaro:
These are now split out at the Airflow level, with separate flags for validation and parsing.

atvaccaro:
> 1. (Probably uncontroversial) I think it would be good for the naming of the Python script and the Airflow DAG tasks to be changed to reflect the fact that they now include validation.

I've renamed things on the Airflow side; I can rename the Docker image at some point but I want to minimize significant breaking changes like that in this PR.

atvaccaro:
> * I think it would be better if there were explicit flags that were passed (implemented in this PR) that control whether we parse, validate, or both. I think it would make more sense to have separate functions that are called if possible. Like it seems like it's already pretty close to that with the `if_validate` checks? So is it possible to separate a bit more?
> * I think that there should be separate error files for validation vs. parsing.

These are done!

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -123,109 +218,440 @@ def identify_files(glob, rt_file_type: RTFileType, progress=False) -> List[RTFil
...
+    ).check_returncode()
+
+
+def validate_and_upload(
+    fs,
+    jar_path: Path,
+    dst_path_gtfs,
+    dst_path_rt,
+    tmp_dir,
+    hour: RTAggregation,
```
lauriemerrell, line 298: **nitpick:** can we not call this "hour" but instead "aggregation"? to make it generic in case we change granularity?

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -123,109 +218,440 @@ def identify_files(glob, rt_file_type: RTFileType, progress=False) -> List[RTFil
...
 
-        gzip_fname = str(tmp_dir + "/" + "temporary" + EXTENSION)
-        written = 0
 
-        with gzip.open(gzip_fname, "w") as gzipfile:
+def parse_and_upload(
+    fs,
+    dst_path_rt,
+    tmp_dir,
+    hour: RTAggregation,
```
lauriemerrell, line 417: **nitpick**: same Q as above, can/should we rename this?

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -123,109 +218,440 @@ def identify_files(glob, rt_file_type: RTFileType, progress=False) -> List[RTFil
...
-            pbar.update(1)
+    if written:
+        log(
+            f"writing {written} lines to {hour.data_hive_path}{JSONL_GZIP_EXTENSION}",
+            pbar=pbar,
+        )
+        put_with_retry(fs, gzip_fname, f"{hour.data_hive_path}{JSONL_GZIP_EXTENSION}")
+    else:
+        log(
+            f"WARNING: no records at all for {hour.data_hive_path}",
```
lauriemerrell, line 478: **question**: This kind of thing is making me wonder.... Is there any value in writing out a more generic metadata file that lists the feeds that were validated or something, with the outcome? As you've mentioned to me, no files found can be because there were no errors or because validation didn't run. Is there a way for us to distinguish between those cases / do we need to be able to do that?

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -46,46 +88,99 @@ class RTFile(BaseModel):
...
+        )
+
+    def data_hive_path(self, bucket: str):
         return os.path.join(
             bucket,
             self.file_type,
             f"dt={self.tick.to_date_string()}",
             f"itp_id={self.itp_id}",
             f"url_number={self.url}",
             f"hour={self.tick.hour}",
```
lauriemerrell, line 116: **question**: This idea was actually mentioned to me in an unrelated conversation by @j-meelah -- is there any chance that we want to still partition on minute / second but just have those values constant, as a way of future proofing in case we ultimately end up processing more often? 

She brought this question up in the context of GTFS schedule. Right now, we might be tempted to partition on day only. But I know you're interested in scraping more often, so should I add an hourly partition level as a method of future-proofing against that eventuality?

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -123,109 +218,440 @@ def identify_files(glob, rt_file_type: RTFileType, progress=False) -> List[RTFil
...
+            f"WARNING: no records at all for {hour.validation_hive_path}",
+            fg=typer.colors.YELLOW,
+            pbar=pbar,
+        )
+
+    if errors:
+        upload_gzipped_records(
+            fs,
+            gzip_fname=str(tmp_dir + f"/errors_{hour.suffix}" + JSONL_GZIP_EXTENSION),
+            out_path=f"{hour.errors_hive_path}_validation{JSONL_GZIP_EXTENSION}",
```
lauriemerrell, line 400: **question**: Why do we affix "_validation" here if it's also present in the second level folder name? Just for extra clarity?

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -123,109 +218,440 @@ def identify_files(glob, rt_file_type: RTFileType, progress=False) -> List[RTFil
...
+def parse_and_aggregate_hour(
+    hour: RTAggregation,
+    jar_path: Path,
+    tmp_dir: str,
+    verbose: bool = False,
+    parse: bool = False,
+    validate: bool = False,
+    pbar=None,
+):
+    if not parse and not validate:
```
lauriemerrell, line 511: **nitpick:** see below, seems like we also assert this, do we need to do both?

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -123,109 +218,440 @@ def identify_files(glob, rt_file_type: RTFileType, progress=False) -> List[RTFil
...
-    try:
-        parse_file(*args, **kwargs)
-    except Exception as e:
-        typer.echo(
-            f"got exception while handling feed: {str(e)} from {str(e.__cause__ or e.__context__)} {traceback.format_exc()}",
-            err=True,
+
+# Originally this whole function was retried, but tmpdir flakiness will throw
+# exceptions in backoff's context, which ruins things
+def parse_and_aggregate_hour(
```
lauriemerrell, line 502: **question**: Does this need to have `hour` hardcoded as part of name? Version of same question above. And should this be renamed to be like `process` or something to account for both parsing and validation?

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -123,109 +218,440 @@ def identify_files(glob, rt_file_type: RTFileType, progress=False) -> List[RTFil
...
     progress: bool = typer.Option(
         False,
         help="If true, display progress bar; useful for development but not in production.",
     ),
+    verbose: bool = False,
     threads: int = 4,
+    jar_path: Path = JAR_DEFAULT,
 ):
-    typer.secho(f"Parsing {file_type} from {glob}", fg=typer.colors.MAGENTA)
+    assert parse ^ validate, "must either parse xor validate"
```
lauriemerrell, line 566: **question**: you've got that other check above for if not parse or validate -- do we need to check in both places?

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -123,109 +218,440 @@ def identify_files(glob, rt_file_type: RTFileType, progress=False) -> List[RTFil
...
+            pbar=pbar,
         )
-        # a lot of these are thrown by tmpdir, and are potentially flakes; but if they were thrown during
-        # handling of another exception, we want that exception still
-        if isinstance(e, OSError) and e.errno == 39:
-            return e.__cause__ or e.__context__
-        return e
 
 
 def main(
```
lauriemerrell, line 551: **super nitpick / very much just my opinion**: seems like kind of a lot going on for a `main` -- is there anything we can/should refactor into separate functions at all?

lauriemerrell (changes requested):
Thanks for addressing the first round of comments, I just have a few more things -- mostly pretty nitpicky which I realize is annoying at this stage but it's a very good step forward so I am holding it to high standards 😅 

Most substantial question is about how generic we want this to be, manifests in a few places -- could this run as-is at a non-hourly level if we wanted to change the granularity? If so, have some suggestions for places to remove hour-specific language. If not, I guess that's moot and maybe we should rename `RTAggregation` to `RTHourlyAggregation` and go the other direction.  

Related substantive question (credit to @j-meelah) is whether we want to pre-make further partitions to future-proof, in case we do want to adjust that granularity.

atvaccaro:
> Most substantial question is about how generic we want this to be, manifests in a few places -- could this run as-is at a non-hourly level if we wanted to change the granularity? If so, have some suggestions for places to remove hour-specific language. If not, I guess that's moot and maybe we should rename `RTAggregation` to `RTHourlyAggregation` and go the other direction.

Good call-out, I will rename.

I don't think we can really change the granularity since we're explicitly defining hive partitions. The _job_ could process a wider interval but that's just handled by the glob input. The output will always have to be hourly.

jobs/gtfs-rt-parser/gtfs_rt_parser.py
atvaccaro: This is a separate check; the check in `main()` is to assert the job is called with either parse or validate, as a safety check for Airflow.

The underlying `parse_and_aggregate_hour()` could technically do both in the future if we want.

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -123,109 +218,440 @@ def identify_files(glob, rt_file_type: RTFileType, progress=False) -> List[RTFil
...
 
-    with tempfile.TemporaryDirectory() as tmp_dir:
-        parsed = parse_pb(rt_file.path, open_func=fs.open)
+    try:
+        get_with_retry(fs, gtfs_schedule_path, dst_path, recursive=True)
+    except FileNotFoundError:
+        raise ScheduleDataNotFound(f"no schedule data found for {gtfs_schedule_path}")
+
+    try:
+        os.remove(os.path.join(dst_path, "areas.txt"))
```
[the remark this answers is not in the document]
atvaccaro: yep!

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -123,109 +218,440 @@ def identify_files(glob, rt_file_type: RTFileType, progress=False) -> List[RTFil
...
 
-        if not parsed or "entity" not in parsed:
-            msg = f"WARNING: no records found in {str(rt_file.path)}"
-            if pbar:
-                pbar.write(msg)
-                pbar.update(1)
-            else:
-                typer.secho(msg, fg=typer.colors.YELLOW)
-            return
+    if records_to_upload:
```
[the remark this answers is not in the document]
atvaccaro: I didn't necessarily want `upload_gzipped_records()` to handle the case where there are no records... but maybe it's not a bad idea. Will change.

jobs/gtfs-rt-parser/gtfs_rt_parser.py
atvaccaro: I think the situations are different.

For schedule, I think that's totally reasonable and most accurate because we actually did download the data at an instant, and we don't plan to aggregate up.

For RT, I don't think so. This is aggregated/transformed data, and we can't partition down to the minute/second without blowing up BigQuery. It's the same problem as having a TIMESTAMP with zeros to indicate a DATE. It's misleading if a data consumer sees timestamps and doesn't have the contextual knowledge that the data is aggregated to the hour.

jobs/gtfs-rt-parser/gtfs_rt_parser.py
atvaccaro: Because the parse errors and validation errors are written out separately, we'd overwrite one or the other without the extra prefix.

I agree it's unclear, I will make it another hive path attribute.

jobs/gtfs-rt-parser/gtfs_rt_parser.py
atvaccaro: yes, typer entrypoint vs function that may be called elsewhere theoretically

jobs/gtfs-rt-parser/gtfs_rt_parser.py
atvaccaro: Eh I think it's marginal, I don't necessarily like functions that are only called once for the sake of shortening a function.

jobs/gtfs-rt-parser/gtfs_rt_parser.py
lauriemerrell: Ahhh yeah I misunderstood this: https://github.com/cal-itp/data-infra/blob/aab55d004aad4a70d55215aae525c58ccf2c1db3/jobs/gtfs-rt-parser/gtfs_rt_parser.py#L134 and thought that "file type" included the validation or not distinction. I think that this should be specified at the top level underneath the bucket -- these errors should end up in separate tables (parse vs. validate errors).

jobs/gtfs-rt-parser/gtfs_rt_parser.py
lauriemerrell: I guess I'm wondering if there's value in refactoring the `error` files to just be `outcome` files so that you have like a record of what feeds were operated on. Can a single feed throw more than one error (within the context of just parsing or just validating? Like I assume a feed can throw an error on parsing AND on validation, but within parsing is it possible to have multiple errors written? If not, think we could just log an error outcome with whatever the error was.)

Basically... Refactor the `RTError` class to be like an `RTOutcome` class that gets created either way?

jobs/gtfs-rt-parser/gtfs_rt_parser.py
lauriemerrell: Ok, that distinction makes sense. I think we should maybe share that out as a decision that was made here for awareness.

lauriemerrell (commented):
ok just a few more small questions

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -46,46 +92,122 @@ class RTFile(BaseModel):
...
+
+
+class RTFileProcessingOutcome(BaseModel):
+    step: str
+    success: bool
+    file: RTFile
+    n_output_records: Optional[int]
+    exception: Optional[Exception]
+    hive_path: Optional[str]
+    body: Optional[str]
```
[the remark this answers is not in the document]
atvaccaro: No, my goal here was to have generic information; for example `processed X files` or other similar things.

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -123,109 +245,468 @@ def identify_files(glob, rt_file_type: RTFileType, progress=False) -> List[RTFil
...
+    try:
+        os.remove(os.path.join(dst_path, "validation.json"))
+    except FileNotFoundError:
+        pass
+
+    written = []
+    with ZipFile(zipname, "w") as zf:
+        for file in os.listdir(dst_path):
+            full_path = os.path.join(dst_path, file)
+            if not os.path.isfile(full_path):
```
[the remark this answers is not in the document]
atvaccaro: is that an "acceptable" set of schedule data?

I can drop just the `processed` directory.

jobs/gtfs-rt-parser/gtfs_rt_parser.py
```diff
@@ -123,109 +245,468 @@ def identify_files(glob, rt_file_type: RTFileType, progress=False) -> List[RTFil
...
+    pbar=None,
+) -> List[RTFileProcessingOutcome]:
+    try:
+        gtfs_zip, included_files = download_gtfs_schedule_zip(
+            fs,
+            hour.source_files[0].schedule_path,
+            dst_path_gtfs,
+            pbar=pbar,
+        )
+    except ScheduleDataNotFound:
```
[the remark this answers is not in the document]
atvaccaro: It gets re-raised and caught in the calling function

jobs/gtfs-rt-parser/gtfs_rt_parser.py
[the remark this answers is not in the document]
lauriemerrell: That's kind of my question, I actually think that the validator may be able to handle it even though it's not actually "officially" legit... May be safer to just drop `processed` specifically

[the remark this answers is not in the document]
atvaccaro: Oh wait sorry no, we should be OK still. In this example, is there still an actual zipfile or a subdirectory?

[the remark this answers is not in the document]
atvaccaro: Per slack discussion, we are leaving as-is for now with a TODO. Weird corner case we don't really support anyways.

lauriemerrell (approved):
🚀 🚀 🚀

Issue #675, opened 2021-11-08
Change jupyterhub url from hubtest.k8s.calitp.jarv.us to notebooks.calitp.org 


machow:
want to change to jupyterhub.calitp.org (cc @hunterowens to confirm)

themightychris:
The biggest barrier to this is just settling on a desired hostname and getting the official DNS record created. It would be great to knock this out so we can stop distributing the test URL sooner than later

evansiroky:
@thekaveman, is this something you can help with?

thekaveman:
@evansiroky yes once we have the desired hostname (assuming a subdomain of calitp.org), just let me know and I'll get the DNS rolling.

blakeflei:
@thekaveman @evansiroky how about `jupyter.calitp.org`?

evansiroky:
I don't have a strong opinion about the name, so if this sounds good to everyone else then it sounds good to me.

thekaveman:
I also don't have a strong opinion. I will just point out that I personally have had a lot of trouble spelling the word `jupyter` in the past 😉

themightychris:
well we did `dashboards.calitp.org` instead of `metabase.calitp.org` — so might we consider doing `notebooks.calitp.org` for this to hide the implementation detail from the consumers?

charlie-costanzo:
> well we did `dashboards.calitp.org` instead of `metabase.calitp.org` — so might we consider doing `notebooks.calitp.org` for this to hide the hard-to-spell-phonetically implementation detail from the consumers? (sorry for the last minute thought :-)

I actually had this same thought after initially thinking we should go with `jupyter.calitp.org`, now I think we should set it to `notebooks.calitp.org` if it's not too late @thekaveman 

cc @blakeflei

blakeflei:
Issuing a new CSR seems have resolved the certificate issuance

PR #1460, opened 2022-04-28
feat(jupyter):add notebooks.calitp.org to ingress

blakeflei:
# Description

Add notebooks.calitp.org ingress to the jupyterhub cluster.

Resolves #[675](https://github.com/cal-itp/data-infra/issues/675)

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Recommendations are very welcome

### DIFF
--- a/kubernetes/apps/charts/jupyterhub/values.yaml
+++ b/kubernetes/apps/charts/jupyterhub/values.yaml
@@ -63,7 +63,7 @@ jupyterhub:
     tls:
       - secretName: jupyterhub-tls
         hosts:
-          - notebooks.calipt.org
+          - notebooks.calitp.org
           - hubtest.k8s.calitp.jarv.us
   proxy:
     service:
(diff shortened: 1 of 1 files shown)


themightychris:
this looks good but we'll need to do something to force the cert to refresh _i think_ — @lottspot should we change the secret name here to force that or just watch and see if it gets handled automatically and manually delete the existing cert if needed?

blakeflei:
> do we want jupyter.calitp.org as well?
I'm certainly not opposed, but parsimony might be the best path to start. If there's a strong preference from end users we can always tune if needed

lottspot:
> should we change the secret name here to force that or just watch and see if it gets handled automatically and manually delete the existing cert if needed?

There's really no risk here in waiting and watching since we're only adding an additional hostname and not subtracting anything that currently exists. I would honestly advocate the wait and watch approach just for science even if no other reason

lottspot (approved):

lottspot:
Note that this change will unfortunately have to be deployed by hand because jupyterhub doesn't play nicely with the current pipeline

blakeflei:
@lottspot thanks for the input. Let's pair when you get back - I'll wait to merge until then

themightychris:
@atvaccaro @blakeflei I recommend we try to stick to a single canonical name per external service, so we can keep docs consistent and only worry about providing a good experience at a single hostname. Consider for example things like GitHub authentication needs to redirect back somewhere, and things might get wonky if the user started at a different domain and then gets redirected back to the canonical one. I don't think the potential convenience of additional hostnames will outweigh the potential confusion

Issue #1010, opened 2022-01-29
RT Pipeline: Track detailed information about GTFS-RT file downloads and attempts


evansiroky:
**Is your feature request related to a problem? Please describe.**

BigQuery should have certain tables with information describing information about the GTFS-RT extraction process including:

- Attempts to download GTFS-RT data files
- Download attempts that were successful
- Download attempts that were unsuccessful due to problems with Cal-ITP infrastructure
- Download attempts that were unsuccessful due to problems outside of Cal-ITP infrastructure

**Describe the solution you'd like**

- [x] #267
- [ ] Track whether download attempts were successful or not

**Describe alternatives you've considered**

We may log this, but need it in BigQuery to be able to provide insight on questions such as whether agencies meet GTFS Guidelines of having RT feeds available every 20 seconds.

**Additional context**

We also eventually want to track the timestamps within the feeds that get downloaded to see if the contents of the files are current too. However, the data for that is to be loaded in via #1002.

Nkdiaz:
Using the following query:
```
With
start_fetch AS (
    SELECT 
        textpayload, 
        timestamp AS start_fetch_time,
        REGEXP_EXTRACT(textPayload, r'\[txn (.*?)\]') AS file_hash,
        REGEXP_EXTRACT(textPayload, "mapper_key=([0-9]+)")AS calitp_itp_id,
        REGEXP_EXTRACT(textPayload, "mapper_key=[0-9]+/([0-9]+)")AS calitp_url_number,
        regexp_extract(textpayload,"mapper_key=[0-9]+/[0-9]+/?([a-zA-Z0-9\\.\\_\\-]+)?") AS feed_file,

    FROM `cal-itp-data-infra.gtfs_rt_logs.stdout` 
    WHERE textPayload like "%start fetch%"  
), 
complete_fetch AS (
    SELECT 
        timestamp as completed_fetch_time,
        REGEXP_EXTRACT(textPayload, r'\[txn (.*?)\]') AS c_file_hash,
    FROM `cal-itp-data-infra.gtfs_rt_logs.stdout` 
    WHERE textPayload like "%completed fetch%"
),
fetch_table AS (
    select * from start_fetch t1
    JOIN complete_fetch t2  ON t1.file_hash = t2.c_file_hash
),
start_write AS (
    SELECT 
        textpayload, 
        timestamp AS start_write_time,
        REGEXP_EXTRACT(textPayload, r'\[txn (.*?)\]') AS file_hash,
[...]

PR #1463, opened 2022-04-29
Rt views to dbt

### LINKED ISSUE #1256: implement GTFS RT models in dbt

lauriemerrell:
# Description

Migrates remaining `rt_views` DAG tasks to dbt instead of Airflow & deletes the `rt_views` DAG. 

Highlights:

* `validation_code_descriptions`: Because it's a Python operator that scrapes a Google Sheet, I just moved the DAG task and created a dbt model on top of the table that DAG task creates. 
   * We will have to figure out where to put this DAG task long term. I put it in `rt_loader` right now, but that might get deprecated in #1423. cc @atvaccaro 
   * I followed the exact same pattern as the GTFS schedule version of this data.
*  `gtfs_rt_extraction_errors`: This one I refactored more substantially, to make it a bit more useful and implement some of the work that @Nkdiaz had started in #1110 (I think that that PR should now be closed). I think that the work here represents significant progress towards #1010.
   * Accounted for new log structure: More recent logs no longer directly record the ITP ID / URL number in `textPayload`, based on some of the changes in the RT archiver logging. So I find the ITP ID / URL number by looking them up based on the actual URL from the `textPayload`. 
   * Dedupe by "tick": A challenge with this data is that, as noted above, we are no longer directly logging ITP identifiers. This means that when there's an issue with the MTC 511 service alerts feed (which we download 30 copies of), we get 30 instances of the same error with the same URL but *slightly* (<1sec) different logging timestamps. Because we can only use the URL to map back to ITP ID / URL number, this would result in 30 copies of each error **for each feed**, which is undesirable and inflated. Instead, I rounded the error timestamp seconds down to one of `:00`, `:20`, `:40` (rounding done per error text instance, to make sure we don't lose any unique errors), and then dedupe at that "tick timestamp" level. **This results in a slightly lower # of rows in the new table compared to the old one (<1% difference), because when multiple instances of the same error appear for the same URL within 20 sec of each other, we are only keeping one copy.**
   * Use safe URL: The full `textPayload` contains the URL with API keys filled in. I stop bringing in that raw text field and instead provide the "safe" URL with keys suppressed. 
   * Sometimes (~3% of the time) ITP ID / URL number are null, for example because the URL was deleted from `agencies.yml` on that day. I think that's ok because these are error logs, so inherently they involve cases where something went wrong. Sometimes that "something" leads to inadequate labeling and I think that's just the reality. I was going to try to interpolate labels and realized that logic would be fragile and it's probably safer to just leave things blank, especially because the prevalence is low. 

Resolves #1256

## Type of change

- [x] New feature
- [x] Documentation - changes documentation about Airflow maintenance to reflect deletion of `rt_views` DAG
- [ ] agencies.yml

## How has this been tested?

1. Ran `dbt run` and verified that everything ran as expected
   * For `validation_code_descriptions`, verified that new and old tables have same # of rows.
   * For `extraction_errors`, see above on the discrepancy in # of rows
2. Ran `dbt test` and verified that tests passed 
3. Verified that the moved `validation_code_descriptions` DAG task runs successfully in local Airflow.


lauriemerrell:
`raw_url` is null more often than I expected in the extraction errors table -- double checking why

lauriemerrell:
Turns out I just needed to refresh my tables in dbt :) All good!

atvaccaro (approved):

PR #1475, opened 2022-05-04
vendor jar in github and copy in dockerfile

atvaccaro:
The old S3 file location of the RT validator jar is returning 403s, so I'm just vendoring a copy and using it in the Dockerfile.


lauriemerrell (commented):
Can we add a credit README or something somewhere? Would feel better if there were something acknowledging where the JAR came from

lauriemerrell (approved):

PR #1476, opened 2022-05-04
external table operator: use prefix bucket template now that we figur…

lauriemerrell:
…ed out how it works

# Description

Resolves # [issue]

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml


lauriemerrell:
We've determined that this won't work because you can only template fields that are not keyword arguments. Closing PR, we will continue to use the existing `prefix_bucket` flags that Andrew added in #1412

Issue #1159, opened 2022-03-02
Transit Technology Stacks


vykster:
Ingest Transit Technology Stacks AirTable base into Metabase to support streamlining Data Assessment Workflow #1019. 

The contracts table is most essential to support our current needs. This table currently includes pdf uploads as attachments to records, these are not needed as part of the import but if a url to the pdf attachments are available that would be ideal.

PR #1477, opened 2022-05-04
Airtable operator refactor

lauriemerrell:
🚨 _This is merging into the `airtable-updates` branch, not `main`, to prepare for a bundle of related changes_

# Description

This PR establishes the groundwork for the "new" Airtable pattern, in line with the RT pipeline and to prepare for #1159.

It refactors the Airtable operator to:
* Stop writing simultaneously to GCS & BigQuery and instead only write to hive-partitioned GCS buckets 
* Use Pydantic typing to enforce data types

In doing so, I had to add a bunch of handling for arrays that contain nulls because BigQuery does not allow them. Specifically, if an array only has null values, this just return an empty array -- this collapses the distinction between [null] and [] which may be meaningful in cases where the array is the result of a lookup in Airtable, but the distinction should be recoverable by checking the lookup relation directly.

It also adds external table creation to the `create_external_tables` DAG. 

@atvaccaro -- Would appreciate particular review on: https://github.com/cal-itp/data-infra/blob/d2e52c06f92308867d59c95c5cf1cd4d28226e71/airflow/plugins/operators/airtable_to_warehouse.py#L18 -- this is where I do the null handling that we discussed. 

**Upcoming follow-ups to be handled in following PRs:**
* Rewrite `airtable_views` in dbt (this will include latest-only logic for the new external tables)
* Extract Transit Stacks

## Type of change

- [x] New feature
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) -- breaks Airtable
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

- Ran the updated Airflow tasks locally
- Manually spot checked data & new BQ tables

### DIFF
--- a/airflow/dags/sandbox/load_airtable_child.yml
+++ b/airflow/dags/sandbox/load_airtable_child.yml
@@ -3,10 +3,11 @@ operator: operators.AirtableToWarehouseOperator
 air_base_id: apphcwzP8cR82NeJo
 air_base_name: sandbox
 air_table_name: child
-bucket: "gs://{{ prefix_bucket('airtable') }}"
+bucket: "gs://{{ prefix_bucket('calitp-airtable') }}"
 
-id_name: "child_id"
-column_prefix: "child_"
+airtable_options:
+  id_name: "child_id"
+  column_prefix: "child_"
 
 
 dependencies:
--- a/airflow/dags/sandbox/load_airtable_child.yml
+++ b/airflow/dags/sandbox/load_airtable_child.yml
@@ -1,11 +1,13 @@
 operator: operators.AirtableToWarehouseOperator
 
 air_base_id: apphcwzP8cR82NeJo
+air_base_name: sandbox
 air_table_name: child
-table_name: sandbox.airtable_child
+bucket: "gs://{{ prefix_bucket('airtable') }}"
 
 id_name: "child_id"
 column_prefix: "child_"
+
 
 dependencies:
   - op_airtable_to_warehouse
--- a/airflow/plugins/operators/airtable_to_warehouse.py
+++ b/airflow/plugins/operators/airtable_to_warehouse.py
@@ -103,7 +103,7 @@ class AirtableExtract(BaseModel):
             raise ValueError(
                 "An extract time must be set before a hive path can be generated."
             )
-        safe_air_table_name = "_".join(self.air_table_name.split(" "))
+        safe_air_table_name = str.lower("_".join(self.air_table_name.split(" ")))
         return os.path.join(
             bucket,
             f"{self.air_base_name}__{safe_air_table_name}",
@@ -115,7 +115,7 @@ class AirtableExtract(BaseModel):
     def save_to_gcs(self, fs, bucket):
         hive_path = self.make_hive_path(bucket)
         print(f"Uploading to GCS at {hive_path}")
-        assert self.data, "data does not exist, cannot save"
+        assert self.data.any(None), "data does not exist, cannot save"
         fs.pipe(
             hive_path,
             gzip.compress(self.data.to_json(orient="records", lines=True).encode()),
--- a/airflow/plugins/operators/airtable_to_warehouse.py
+++ b/airflow/plugins/operators/airtable_to_warehouse.py
@@ -1,11 +1,13 @@
-import pandas as pd
-import re
 import os
+import pandas as pd
+import pendulum
 
 from pyairtable import Table
 from pydantic import BaseModel
+from slugify import slugify
 from typing import Optional
-from calitp import save_to_gcfs, to_snakecase, write_table
+from calitp import to_snakecase
+from calitp.storage import get_fs
 
 from airflow.models import BaseOperator
 
@@ -15,10 +17,12 @@ AIRTABLE_API_KEY = os.environ["CALITP_AIRTABLE_API_KEY"]
 class AirtableExtract(BaseModel):
     api_key: str
     air_base_id: str
+    air_base_name: str
     air_table_name: str
     id_name = "__id"
     rename_fields = {}
-    column_prefix: Optional[str] = None
+    column_prefix: Optional[str]
+    extract_time: Optional[pendulum.DateTime]
 
     def airtable_to_df(self):
         """Download an Airtable table as a DataFrame.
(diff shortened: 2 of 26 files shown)


airflow/requirements.txt
```diff
@@ -17,3 +17,5 @@ structlog==21.5.0
 typer==0.4.0
 humanize==3.14.0
 backoff==1.11.1
+pydantic==1.9.0
+slugify==0.0.1
```
atvaccaro, line 21: I'd recommend https://pypi.org/project/python-slugify/ seems to be better supported/active

airflow/dags/sandbox/load_airtable_child.yml
```diff
@@ -1,11 +1,13 @@
 operator: operators.AirtableToWarehouseOperator
 
 air_base_id: apphcwzP8cR82NeJo
+air_base_name: sandbox
 air_table_name: child
-table_name: sandbox.airtable_child
+bucket: "gs://{{ prefix_bucket('airtable') }}"
```
atvaccaro, line 6: do both test-airtable and airtable buckets exist yet?

airflow/plugins/operators/airtable_to_warehouse.py
```diff
@@ -1,126 +1,192 @@
...
+        # in the foreign table
+        # however, this distinction should be recoverable from other fields
+        new_array = []
+    return new_array
+
+
+def make_arrays_bq_safe(raw_data):
+    safe_data = {}
+    for k, v in raw_data.items():
+        if type(v) == dict:
```
atvaccaro, line 51: nit: I think `isinstance()` is supposed to be more pythonic

airflow/plugins/operators/airtable_to_warehouse.py
```diff
@@ -1,126 +1,192 @@
...
+            make_arrays_bq_safe(v)
+        elif type(v) == list:
+            if None in v:
```
atvaccaro, line 54: I'd just process all non-empty lists

airflow/plugins/operators/airtable_to_warehouse.py
```diff
@@ -1,126 +1,192 @@
...
-
-    Note that airtable records have rows structured as follows:
-        [{"id", "fields": {colname: value, ...}, ...]
 
-    This function applies renames in the following order.
-
-        1. rename id
-        2. rename fields
-        3. apply column prefix (to columns not renamed by 1 or 2)
+def fix_array_containing_null(arr):
```
atvaccaro, line 18: nit: I think this can be simplified, it's a bit hard to follow as-is

airflow/plugins/operators/airtable_to_warehouse.py
```diff
@@ -1,126 +1,192 @@
...
+            f"{slugify(self.air_table_name, separator='_')}.jsonl.gz",
         )
 
-    return final_df
+    def save_to_gcs(self, fs, bucket):
+        hive_path = self.make_hive_path(bucket)
+        print(f"Uploading to GCS at {hive_path}")
+        fs.pipe(
+            hive_path,
+            gzip.compress(self.data.to_json(orient="records", lines=True).encode()),
```
atvaccaro, line 137: maybe assert that `self.data` exists at this point

airflow/plugins/operators/airtable_to_warehouse.py
```diff
@@ -1,126 +1,192 @@
...
-            clean_gcs_path = re.sub(r"\/+$", "", self.gcs_path)
-            gcs_file = (
-                f"{clean_gcs_path}/{context['execution_date']}/{self.table_name}.csv"
-            )
-            print(f"Uploading to gcs at {gcs_file}")
-            save_to_gcfs(df.to_csv(index=False).encode(), f"{gcs_file}", use_pipe=True)
+    def execute(self, **kwargs):
+        self.extract.fetch_from_airtable(self.api_key)
+        fs = get_fs()
+        self.extract.save_to_gcs(fs, self.bucket)
```
atvaccaro, line 192: I've kinda been wondering if bucket should be part of these pydantic models... mixed feelings.

atvaccaro (changes requested):
the only hard request is to change the slugify module, everything else is optional

airflow/dags/sandbox/load_airtable_child.yml
lauriemerrell: Good call, `airtable` was taken (I forgot bucket names have to be *globally* unique 🥴). Switched both to `calitp-airtable` which I did create (test & prod).

airflow/plugins/operators/airtable_to_warehouse.py
lauriemerrell: I had it just process all lists because I didn't see why checking `len(v) > 0` was preferable to checking `None in v`. If you think that distinction is important I can add it back in.

airflow/plugins/operators/airtable_to_warehouse.py
```diff
@@ -1,126 +1,186 @@
...
+
+def make_arrays_bq_safe(raw_data):
+    safe_data = {}
+    for k, v in raw_data.items():
+        if isinstance(v, dict):
+            make_arrays_bq_safe(v)
+        elif isinstance(v, list):
+            safe_data[k] = process_arrays_for_nulls(v)
+        else:
+            safe_data[k] = v
```
atvaccaro, line 53:
```suggestion
    for k, v in raw_data.items():
        if isinstance(v, dict):
            v = make_arrays_bq_safe(v)
        elif isinstance(v, list):
            v = process_arrays_for_nulls(v)
        safe_data[k] = v
```

airflow/plugins/operators/airtable_to_warehouse.py
```diff
@@ -1,126 +1,186 @@
...
+        else:
+            new_array = [x if x is not None else "" for x in arr]
+    else:
+        # if array only has null values, just return an empty array --
+        # this collapses the distinction between [null] and []
+        # which may be meaningful in cases where the array is the
+        # result of a lookup in Airtable, but the distinction
+        # should be recoverable by checking the lookup relation directly
+        new_array = []
+    return new_array
```
atvaccaro, line 42:
```suggestion
types = set(type(entry) for entry in arr if entry is not None)

if not types:
    return []

filler = -1 if types <= {int, float} else ""

return [x if x is not None else "" for x in arr]
```

airflow/requirements.txt
lauriemerrell: We ended up having to drop this dependency entirely because it introduced too many conflicts & made the docker image un-buildable.

atvaccaro (approved):

PR #1483, opened 2022-05-06
More RT refactor fixes

atvaccaro:
# Description

This PR fixes two problems that popped up during the backfill.

1. Sometimes, there are no validation notices output for a given hour of RT data (usually service alerts). Currently we upload empty files, but BigQuery fails when parsing these. Instead, we should write a single newline character since the data is JSONL.
2. Our schedule data is keyed via Airflow logical execution date in GCS, so we have to subtract a day from the RT ticks to find it.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Locally via docker-compose.

### DIFF
--- a/jobs/gtfs-rt-parser/gtfs_rt_parser.py
+++ b/jobs/gtfs-rt-parser/gtfs_rt_parser.py
@@ -109,7 +109,8 @@ class RTFile(BaseModel):
         return os.path.join(
             get_bucket(),
             "schedule",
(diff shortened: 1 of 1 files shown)


lauriemerrell (approved):
LGTM, let's please make a follow up ticket to have GTFS schedule data written with actual execution time not Airflow execution date and revert the date manipulation part of this.

Issue #1482, opened 2022-05-06
Run fill RT backfill and fix bad filepaths


atvaccaro:
Follow-up from https://github.com/cal-itp/data-infra/pull/1479, we need to run the full backfill and also move the files written to the singular validation_outcomes filepath in GCS.

From that PR
> Important follow-up to this: migrate existing written data to the new entity paths for validation outcomes.
gsutil -m mv gs://test-rt-parsed/service_alerts_validation_outcomes gs://test-rt-parsed/service_alerts_validations_outcomes
The above, but in the prod bucket.

PR #1486, opened 2022-05-09
Full RT backfill, separate pools, and a couple tweaks

atvaccaro:
# Description

The new RT pipeline has been running for a few days now successfully! We only backfilled to the beginning of May, so we could catch up quickly and keep new data flowing in roughly realtime. Now, we are ready to backfill back to the start of RT data in July.

This PR also has a few small tweaks.
1. It splits up parsing and validation into separate pools. We can run much more parsing at once than validation, so we can get historical RT _data_ backfilled much more quickly by having a larger pool.
2. It skips uploading single-newline gzip files to represent zero validation notices; BigQuery does not like these.
3. Adds a retry on ServerDisconnectedError, which we didn't before.
4. It uses post-hooks to test that external tables are queryable after creation.

Resolves https://github.com/cal-itp/data-infra/issues/1482

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

I removed the empty-ish files from `gs://rt-parsed/` and can now run the following type of query on all 3 types of validations.

```sql
select distinct errorMessage.validationRule.errorDescription
from `cal-itp-data-infra`.external_gtfs_rt.trip_updates_validations
;
```

In addition, all external tables can be created and queried properly. https://o1d2fa0877cf3fb10p-tp.appspot.com/tree?dag_id=create_external_tables

## Screenshots (optional)
image

### DIFF
--- a/airflow/dags/create_external_tables/rt_service_alerts_outcomes.yml
+++ b/airflow/dags/create_external_tables/rt_service_alerts_outcomes.yml
@@ -1,7 +1,11 @@
 operator: operators.ExternalTable
 bucket: gs://rt-parsed
 prefix_bucket: true
-test: true
+
+post_hook: |
+  select distinct step
+  from `{{ get_project_id() }}`.external_gtfs_rt.service_alerts_outcomes;
+
 source_objects:
   - "service_alerts_outcomes/*.jsonl.gz"
 destination_project_dataset_table: "external_gtfs_rt.service_alerts_outcomes"
--- a/airflow/dags/create_external_tables/rt_service_alerts_validations.yml
+++ b/airflow/dags/create_external_tables/rt_service_alerts_validations.yml
@@ -1,7 +1,11 @@
 operator: operators.ExternalTable
 bucket: gs://rt-parsed
 prefix_bucket: true
-test: true
+
+post_hook: |
+  select distinct errorMessage.validationRule.errorDescription
+  from `{{ get_project_id() }}`.external_gtfs_rt.service_alerts_validations;
+
 source_objects:
   - "service_alerts_validations/*.jsonl.gz"
 destination_project_dataset_table: "external_gtfs_rt.service_alerts_validations"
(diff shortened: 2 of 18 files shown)


lauriemerrell (approved):

lauriemerrell:
Ughhhh sorry this keeps happening, didn't realize it was in draft

airflow/dags/create_external_tables/rt_service_alerts_outcomes.yml
```diff
@@ -1,7 +1,11 @@
 operator: operators.ExternalTable
 bucket: gs://rt-parsed
 prefix_bucket: true
-test: true
+
+post_hook: |
+  select distinct step
```
lauriemerrell, line 6: **question**: why `select distinct step`? Will this catch if there are issues in other columns? Did you find that `select * limit 1` (which would be my instinct) was bad for some reason?

airflow/dags/create_external_tables/rt_service_alerts_validations.yml
```diff
@@ -1,7 +1,11 @@
 operator: operators.ExternalTable
 bucket: gs://rt-parsed
 prefix_bucket: true
-test: true
+
+post_hook: |
+  select distinct errorMessage.validationRule.errorDescription
```
lauriemerrell, line 6: same question as above

lauriemerrell (commented):
Looks good just a few small clarifications -- 

* Inline questions see below
* For the backfill, given the newline change, are you going to re-run May? Wasn't sure how that works.... Can we easily clear all of prod history and re-run? 
* Nitpick but PR description says RT data since September but backfill date is since July

airflow/dags/create_external_tables/rt_service_alerts_outcomes.yml
atvaccaro: This is just to test that we can parse a particular column and it's not going to return a ton of results even if we query the entire data set.

atvaccaro:
> * For the backfill, given the newline change, are you going to re-run May? Wasn't sure how that works.... Can we easily clear all of prod history and re-run?

I did manually delete the bad files, but I was planning to _fully_ backfill since adding the last few days onto a several-months backfill isn't the biggest deal.

> * Nitpick but PR description says RT data since September but backfill date is since July

Fixing, thanks.

lauriemerrell (approved):

Issue #696, opened 2021-11-10
Implement freshness check table


machow:
A freshness check is a process that often includes a table (e.g. with columns data source, date of last load), and a set of processes to kick off depending on the distance from today and last load dates.

It has 2 major benefits

* allow analysts to check how up-to-date a view is
* alert org, while allowing flexibility (e.g. alert after 2 days, or escalate in various ways)

Since this implementation will likely apply to all our data domains (e.g. gtfs schedule, payments, rt), we should take some time to tease out what it might look like. See https://docs.getdbt.com/reference/resource-properties/freshness for this concept in DBT :).

holly-g:
Adding this to the dbt Adoption milestone 🎉

Issue #1468, opened 2022-05-03
Migrate dbt models and Metabase from old RT validation job to new validation job


atvaccaro:
Follow-up from https://github.com/cal-itp/data-infra/pull/1458/files which does _not_ delete the existing validator DAG. Once the new hourly job is running, we can delete the old validation job.

We also need to move the dbt models over! This may involve ensuring that Metabase is pointed at the right dbt-produced table.

PR #1491, opened 2022-05-11
Freshness checks

atvaccaro:
# Description

This PR adds freshness checks using dbt source freshness as well as custom Python-based checks for more complex use cases such as checking freshness of files stored in a GCS hive pattern.

Resolves https://github.com/cal-itp/data-infra/issues/696 (though this doesn't actually create a table, just alerts on job failures)

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml


warehouse/models/rt_views/_rt_views.yml
```diff
@@ -1,6 +1,45 @@
...
+          warn_after: { count: 1, period: day }
+      - name: vehicle_positions_validations
+      - name: vehicle_positions_validations_outcomes
+        loaded_at_field: cast(dt as timestamp)
+        freshness:
+          warn_after: { count: 1, period: day }
+
+
+
+  # Note: this source is deprecated by the above
```
lauriemerrell, line 42: Do we have a plan for how to actually deprecate these? I guess that's part of #1468?

lauriemerrell (changes requested):
Whenever a new DAG is created, we should document its rerun considerations here: https://docs.calitp.org/data-infra/airflow/production-maintenance.html. Can you add docs about whether it is safe to rerun etc?

warehouse/models/rt_views/_rt_views.yml
atvaccaro: Yes, exactly.

atvaccaro:
> Whenever a new DAG is created, we should document its rerun considerations here: https://docs.calitp.org/data-infra/airflow/production-maintenance.html. Can you add docs about whether it is safe to rerun etc?

Yep! Will add.

docs/airflow/production-maintenance.md
```diff
@@ -28,6 +28,7 @@ DAGs are listed in alphabetical order, as they appear in the Airflow UI.
 `airtable_loader` | **⛔ No*** | No | **🔂 No** | N/A | All tasks are unsafe after 24 hours |
 `airtable_views` | Yes | No | Yes* | `airtable_loader` | Latest-only data |
 `amplitude_benefits` | Yes | No | **🔂 No** | N/A | |
+`check_data_freshness` | Yes | No | N/A | N/A | |
```
lauriemerrell, line 31: The 3rd column is generally only N/A for once-only DAGs... Can you add a note clarifying what that means in this context? Like is the implication just that it's not worth re-running because it runs often enough on its own?

atvaccaro: Oh yeah sure, and yes exactly; it's like the Composer uptime DAG

lauriemerrell (approved):

Issue #427, opened 2021-10-05
Update and Enhance Open Dataset


hunterowens:
**Is your feature request related to a problem? Please describe.**
Currently, we publish [some aggregated GTFS data](https://data.ca.gov/dataset/cal-itp-gtfs-ingest-pipeline-dataset) to the CA open data portal. 

There are two notable problems with this currently 

1) we don't publish the universe of files such as `fares.txt` or `calendar.txt` in aggegrate form
2) we don't have an automated script (see issue #233) since we need to use a slightly different CKAN API. 

Additionally, even if we published the entire GTFS "dataset", we would want end users to maybe have access to different datasets that can be derived from GTFS, such as [High Quality Transit Areas](https://indicatrix.org/post/identifying-high-quality-transit-corridors/). 

**Describe the solution you'd like**

1) Do user research to meet w/ stakeholders determine dataset needs 
2) Data documentation through Caltrans GIS office process 
3) Publish data 

**Additional context**
People to meet with (for @hunterowens to intro) 

1) Helen Campbell, Gov Office of Planning and Research 
2) Samer Bartaseh, Caltrans Data  and Geospatial Mangement Branch Chief 
3) Chad Baker / walter yu, Caltrans  (for data documentation) 
4) Henry McKay, Planner, Caltrans

holly-g:
First pass: Tiffany can be lead analyst to develop documentation for datasets.

acouch:
Want to note that some of the files are missing a data dictionary and format, so they are not previewable in the browser: https://data.ca.gov/dataset/cal-itp-gtfs-ingest-pipeline-dataset/resource/e8f9d49e-2bb6-400b-b01f-28bc2e0e7df2 vs https://data.ca.gov/dataset/cal-itp-gtfs-ingest-pipeline-dataset/resource/53c05c25-e467-407a-bb29-303875215adc

Issue #284, opened 2021-08-16
Add a file with agency names, calitp ids to CKAN


machow:
Currently, we upload our GTFS Schedule data, but not metadata with the names of agencies in our data. This means that people with see columns like `calitp_itp_id`, `calitp_url_number`, but will likely not know what agencies they represent.

Nkdiaz:
Just checking the CKAN portal and an Agency file was uploaded on October 21, 2021 that contains a CSV with the following columns - Calitp_itp_id, url number, agency id, agency_name, and agency_url, calitp_extracted_at. For the purposes of this ticket, a file has been uploaded. Is the request for a more updated file?

PR #1505, opened 2022-05-13
MVP of dbt-based data publishing framework

atvaccaro:
# Description

This PR implements a minimum-viable data publishing framework based on [dbt exposures](https://docs.getdbt.com/docs/building-a-dbt-project/exposures). Initially, our goal is to primarily support uploads to CKAN; that process is documented in this PR as well.

Once deployed, https://github.com/cal-itp/data-infra/issues/284 and https://github.com/cal-itp/data-infra/issues/427 can be closed

## Type of change

- [x] New feature
- [x] Documentation
- [ ] agencies.yml

## How has this been tested?

I've tested metadata generation and GCS writing locally. However, we can't test publishing to CKAN until we actually go through the process, which we will be doing once this PR is merged.


warehouse/scripts/dbt_artifacts.py
```diff
@@ -0,0 +1,280 @@
...
+
+class Column(BaseModel):
+    name: str
+    description: Optional[str]
+    meta: Optional[Dict[str, Any]]
+    parent: "BaseNode" = None  # this is set after the fact
+
+    @property
+    def publish(self) -> bool:
+        return not self.meta.get("publish.ignore", False)
```
lauriemerrell, line 67: lol at double negative, but I guess it makes sense

warehouse/scripts/json_to_docblocks.py
```diff
@@ -0,0 +1,34 @@
+"""
```
lauriemerrell, line 1: so to clarify/confirm, this is what you used to automatically generate docblocks for the GTFS schedule data?

warehouse/scripts/publish.py
```diff
@@ -0,0 +1,398 @@
...
+
+                        # shapely errors if the array contains only one point
+                        if len(x) > 1:
+                            # each point in the array is wkt
+                            # so convert them to shapely points via list comprehension
+                            as_wkt = [shapely.wkt.loads(i) for i in x]
+                            return shapely.geometry.LineString(as_wkt)
+
+                    # apply the function
+                    df["geometry"] = df.pt_array.swifter.apply(make_linestring)
```
lauriemerrell, line 138: this stuff seems hard coded based on the dim_shapes_geo table specifically... is that right/appropriate? and also is `make_linestring` copied from somewhere? just wondering if we want to include like a reference on that for context or anything

docs/publishing/sections/7_ckan.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Document data
+California Open Data requires two documentation files for published datasets.
+1. `metadata.csv` - one row per resource (i.e. file) to be published
+2. `dictionary.csv` - one row per column across all resources
+
+If you are using dbt exposure-based data publishing, you can automatically generate
+these two files using the main `publish.py` script (specifically the `generate-exposure-documentation`
+subcommand). The documentation from the dbt models' corresponding YAML will be
+converted into appropriate CSVs and written out locally.
```
lauriemerrell, line 36: can you add a literal example call for this command? and specify where it would be run.... is it getting a local manifest or are you providing a prod manifest that it uses? (I may be misunderstanding something below but I thought it needs a manifest?)

docs/publishing/sections/7_ckan.md
```diff
@@ -0,0 +1,65 @@
...
+An example from the latest-only GTFS data exposure.
+```yaml
+meta:
+  destinations:
+    - type: ckan
+      bucket: gs://calitp-publish
+      format: csv
+      url: https://data.ca.gov/api/3/action/resource_update
+      ids:
+        agency: e8f9d49e-2bb6-400b-b01f-28bc2e0e7df2
```
lauriemerrell, line 56: just confirming, we don't think there's a concern with having these listed publicly? (are these real IDs / do they need to be kept secure?)

warehouse/models/gtfs_schedule_latest_only/agency.md
```diff
@@ -0,0 +1,33 @@
+Original definitions from https://gtfs.org/reference/static#agencytxt
```
lauriemerrell, line 1: Can `gtfs_views` access docs MD files in this directory? A lot of these fields are shared across the corresponding `dim` table, just wondering if we want to move these markdown files somewhere else where it's accessible in both places

lauriemerrell (changes requested):
1. **thank you** for writing all these docs, I know it is not your favorite 🙏 
2. this is cool and I am excited!
3. mostly nit picky comments, I just want the docs to be as detailed as possible in case this sits for a bit and then we have to come back to it later..... I think my implicit standard is -- could someone who's not Andrew and can't talk to Andrew successfully publish based on the documentation here alone? (if you don't think that level is possible yet, I guess perhaps that will need to wait)
4. ETA: can you update description? seems like linked Google doc is obsolete, so can probably remove. and info on any testing would be good. also -- I don't think this PR **does** close all those tickets? we still need to actually implement, right? this gives the framework / infra to publish but does not actually publish

warehouse/scripts/dbt_artifacts.py
atvaccaro: yeah lol I think it makes more sense to explicitly flag columns to skip, and include them by default

warehouse/scripts/dbt_artifacts.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+class Manifest(BaseModel):
+    nodes: Dict[str, Node]
+    sources: Dict[str, Source]
+    macros: Dict
+    docs: Dict
+    exposures: Dict[str, Exposure]
+
+    # @validator("nodes", "sources")
```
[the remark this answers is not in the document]
atvaccaro: Oh thanks, removed. It was from the original article I took the initial code from, but it's not useful for us.

warehouse/scripts/json_to_docblocks.py
atvaccaro: yes, I will comment

warehouse/models/gtfs_schedule_latest_only/agency.md
atvaccaro: all dbt artifacts are global, including docs, so we can pick whatever location we want

docs/publishing/sections/7_ckan.md
atvaccaro: They're public on CKAN itself (for example https://data.ca.gov/dataset/cal-itp-gtfs-ingest-pipeline-dataset/resource/e8f9d49e-2bb6-400b-b01f-28bc2e0e7df2) so I think it's fine?

warehouse/scripts/publish.py
atvaccaro: Adding a comment. I need to circle back to this (i.e. map tiles) after our HQTA discussion.

lauriemerrell:
Another request per meeting with Eric: can we update to give an option (maybe a modification of `--dry-run` where you write to somewhere in GCS (maybe a `test-` version of the bucket) but doesn't push to CKAN or the final endpoint, so people can see exactly the files that would be published?

atvaccaro:
> Another request per meeting with Eric: can we update to give an option (maybe a modification of `--dry-run` where you write to somewhere in GCS (maybe a `test-` version of the bucket) but doesn't push to CKAN or the final endpoint, so people can see exactly the files that would be published?

done! the bucket is a CLI option still, but there's separate dry-run and deploy concepts now.

lauriemerrell (commented):
ok just one q

warehouse/scripts/publish.py
```diff
@@ -0,0 +1,373 @@
...
+                            usage_notes=None,
+                        ).json()
+                    )
+                )
+
+
+@app.command()
+def publish_exposure(
+    exposure: ExistingExposure,
+    project: str = typer.Option("cal-itp-data-infra"),
```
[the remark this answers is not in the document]
lauriemerrell: yeah I guess I was wondering about GCS specifically, I guess we just use the `test-` prefix on the bucket.... So can we make the default bucket the `test` one unless `--deploy` is specified? maybe I am overthinking I just want to err on the side of  caution

lauriemerrell (commented):
ok sorry, just wondering if we can update docs based on the changing defaults

lauriemerrell (approved):
yay 🎉

Issue #1206, opened 2022-03-10
Metabase Question: Which GTFS Schedule feeds have zero service hours in the future?


evansiroky:
## User Story

**As a** research data analyst,
**I want to** know what agency GTFS Schedule URLs have downloaded successfully, but their data indicates that in the future zero service hours are scheduled to occur,
**so that** I can properly categorize which transit agencies are having problems with GTFS quality
**and so that** I can know which transit agencies I need to reach out to about expired feeds

For this user story, only the GTFS Schedule and not the GTFS Realtime URLs are considered.

## Acceptance Criteria

**Given** the data describing transit operations within successfully downloaded GTFS Schedule datasets in the nightly pipeline,
**When** the [Catastrophic GTFS Errors Dashboard](https://dashboards.calitp.org/dashboard/46-catastrophic-gtfs-errors-dashboard) is opened,
**Then** a table should be shown that lists out all GTFS Schedule URLs that have zero future service hour and when the last service date occurred according to the schedule data and sorted by descending number of days since last service

**Example:**

Something like the following table should be shown:

|Agency|URL|Number of Days of Zero Service|Last Service Date|
|--|--|--|--|
| Fresno County Rural Transit Agency | http://data.trilliumtransit.com/gtfs/fresnocounty-ca-us/fresnocounty-ca-us.zip | 831 | 2020-01-01 |
| Plumas Transit Systems | http://data.trilliumtransit.com/gtfs/plumas-ca-us/plumas-ca-us.zip | 528 | 2020-10-30 |
...

Nkdiaz:
Metabase dashboard is available at the following link: 
https://metabase.k8s.calitp.jarv.us/question/390-which-gtfs-schedule-feeds-have-zero-service-hours-in-the-future

evansiroky:
This issue should not be considered resolved. On the date of June 8, 2022, the following transit agencies were not showing up in the question when they should have:

* Plumas Transit
* Manteca Transit

The above 2 agencies represent a non-exhaustive list.

lauriemerrell:
Fixed. Looks like Plumas and Manteca (and a few other feeds) have no service at all in `gtfs_schedule_fact_daily_service`, so the relevant new columns in `fact_daily_feeds` are simply null. I updated the Metabase question to include that as a condition and they are now appearing in the dashboard. The `# days` column is null but that basically indicates that they have no service at all in the period covered by our data. I will close again. 

[...]

PR #1539, opened 2022-05-27
Max_service_date and days_from_service_date columns added to gtfs_schedule_fact_daily_feeds

Nkdiaz:
# Description
To help generate the metabase dashboard request for https://github.com/cal-itp/data-infra/issues/1206, max_service_date and days_from_service_date columns were added to `gtfs_schedule_fact_daily_feeds`

Resolves # [ihttps://github.com/cal-itp/data-infra/issues/1206]

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
tested in big query

### DIFF
--- a/warehouse/models/gtfs_views/_gtfs_views.yml
+++ b/warehouse/models/gtfs_views/_gtfs_views.yml
@@ -301,6 +301,12 @@ models:
         description: |
           Days since the feed's contents (the contents of any loadable file
           within the feed) have changed, relative to date
+      - name: max_service_date
+        description: |
+          Maximum service date by feed
+      - name: days_from_last_service
+        description: |
+          Days since the last service date by feed
   - name: gtfs_schedule_fact_daily_feed_files
     description: |
       Each row of this table is a file extracted from a feed on a given day. Note that on days where
--- a/warehouse/models/gtfs_views/gtfs_schedule_fact_daily_feeds.sql
+++ b/warehouse/models/gtfs_views/gtfs_schedule_fact_daily_feeds.sql
@@ -27,6 +27,11 @@ gtfs_schedule_dim_feeds AS (
     FROM {{ ref('gtfs_schedule_dim_feeds') }}
 ),
 
+gtfs_schedule_service AS (
+    SELECT *
+    FROM {{ ref('gtfs_schedule_fact_daily_service') }}
+)
+
 dim_date AS (
     SELECT *
     FROM {{ ref('dim_date') }}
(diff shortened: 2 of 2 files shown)


warehouse/models/gtfs_views/_gtfs_views.yml
```diff
@@ -301,6 +301,17 @@ models:
         description: |
           Days since the feed's contents (the contents of any loadable file
           within the feed) have changed, relative to date
+      - name: is_service_valid
+        description: |
+          Whether the service is valid (active) on this date (boolean). Service date is based on when feed_key is active.
+      - name: days_from_service_start_date
+        description: |
+          Number of days between date and service start date. if is_service_valid is False, this may be negative
```
lauriemerrell, line 309: Would specify: service start date is the first day on which service occurs on any service in this feed (just because "service" is also used to mean like an individual service pattern identified by a `service_id`.... may just be worth explicitly stating that this is a feed-wide summary.... That should maybe be obvious given what this table is but these are all kind of confusing so I think it's worth spelling out)

warehouse/models/gtfs_views/_gtfs_views.yml
```diff
@@ -301,6 +301,17 @@ models:
...
+          indicating that feed is not yet valid as of date
+      - name: days_until_service_end_date
+        description: |
+          Number of days between date and service end date. if is_service_valid is False, this may be negative
```
lauriemerrell, line 313: See above, would maybe even make this more explicit

warehouse/models/gtfs_views/gtfs_schedule_fact_daily_feeds.sql
```diff
@@ -63,6 +68,43 @@ daily_feeds AS (
             AND D.is_in_past_or_present
 ),
 
+daily_service AS (
```
lauriemerrell, line 71: **optional/nitpick**: I'm just wondering if we need to alias this table given that there's no join and nothing particularly fancy going on... seems like we could drop the `Services` qualifier on everything. And similarly, I personally prefer using the column name to `group by 1` but that's literally just me, so up to you on whether to change, but wanted to mention that I think in general we do not use column # and have been referencing things by name in group by statements.

lauriemerrell (changes requested):
Overall looks good, just one docs change that I specifically think we need to make (in `is_service_valid` -- I think we need to clarify because the lingo is very tricky around these concepts, want to be very explicit), two optional docs clarifications, and one optional SQL style comment.

PR #1541, opened 2022-05-27
Deploy stricter gtfs-rt-archiver resource limits into production

lottspot:
This deploys changes merged in #1540


lottspot:
Jupyterhub changes have already been deployed by hand and so should not be considered a concern as part of this deployment

blakeflei (approved):

Issue #1481, opened 2022-05-06
Delete old gtfs_rt.external_* tables and Airflow job


atvaccaro:
Follow-up from https://github.com/cal-itp/data-infra/pull/1479, we should delete the old external tables.

PR #1545, opened 2022-06-01
Build out dbt models for new RT external tables

atvaccaro:
# Description

Now that we have new fancy hive-partitioned external tables, we should use them in dbt/BigQuery. This PR starts that process without too much additional modeling. In addition, it starts to form an opinion of materialization for these large sets of data with an initial staging table for vehicle positions. Specifically, we should be defaulting to views and partitioned external tables for raw data, and persisting aggregations (or as-needed tables that may be aggregated across large periods of time) in an incremental manner.

Closes https://github.com/cal-itp/data-infra/issues/1468

Follow-up will be https://github.com/cal-itp/data-infra/issues/1481 as well as re-running the `create_external_tables` DAG.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

Locally via dbt run/test.


warehouse/models/rt_staging/_rt_staging.yml
```diff
@@ -0,0 +1,55 @@
+version: 2
+
+sources:
+  - name: gtfs_rt_external_tables
+    description: New hive-partitioned external tables reading from GCS.
```
lauriemerrell, line 5: Don't think we need to call these "new" any more do we?

warehouse/models/rt_staging/_rt_staging.yml
```diff
@@ -0,0 +1,55 @@
...
+          warn_after: { count: 1, period: day }
+
+models:
+  - name: stg_rt_validation_errors
+    description: Unioned and unnested table of all RT validator output.
+    columns:
+      - &rt_key
+        name: key
+        tests:
+          - not_null:
```
lauriemerrell, line 47: Is this to avoid doing expensive testing? Just test the last two days always? Just want to make sure I understand the pattern

warehouse/models/rt_staging/_rt_staging.yml
```diff
@@ -0,0 +1,55 @@
+version: 2
+
+sources:
```
lauriemerrell, line 3: Can we make a `_src_rt.yml` [like I did for Airtable](https://github.com/cal-itp/data-infra/blob/extract-transit-tech-stacks-airtable/warehouse/models/staging/transit_database/_src_airtable.yml)? (I am trying to move more towards the actual dbt defaults as I do new work that's not just conversion from Airflow)

warehouse/models/rt_views/_rt_views.yml
```diff
@@ -1,45 +1,7 @@
...
-      - name: vehicle_positions_validations
-      - name: vehicle_positions_validations_outcomes
-        loaded_at_field: cast(dt as timestamp)
-        freshness:
-          warn_after: { count: 1, period: day }
-
-
-
-  # Note: this source is deprecated by the above
+    # note: these tables should probably be moved at some point
```
lauriemerrell, line 4: genuine question... where should they go? 😅 don't belong in `external_*`

lauriemerrell (commented):
I have questions here about naming/project organization, both inline and overall

I thought in our last conversation about this we had landed on: 

> Staging tables live in a single staging dataset that is not qualified by domain, tables are prefixed with `stg` and we use double underscores to delimit source from entity, e.g. `staging.stg_airtable__some_table` 

So, this would have `stg_rt__validation_errors` instead of `stg_rt_validation_errors`. And I think that `stg_vehicle_positions` should be `stg_rt__vehicle_positions`. 

Does that seem ok?

Less important, for my Airtable PR I had made a staging directory with a transit database subdirectory.... So `staging/rt` instead of `rt_staging`. Thoughts on that? I would like to start having more organization in `models` and getting rid of so many subdirectories at that top level but maybe that is just a personal preference.

warehouse/models/rt_staging/_rt_staging.yml
atvaccaro: nah, will change

warehouse/models/rt_staging/_rt_staging.yml
atvaccaro: yeah sure, I have no opinion either way

atvaccaro:
> So, this would have `stg_rt__validation_errors` instead of `stg_rt_validation_errors`. And I think that `stg_vehicle_positions` should be `stg_rt__vehicle_positions`.
> 
> Does that seem ok?
> 

Yes! I will address.

> Less important, for my Airtable PR I had made a staging directory with a transit database subdirectory.... So `staging/rt` instead of `rt_staging`. Thoughts on that? I would like to start having more organization in `models` and getting rid of so many subdirectories at that top level but maybe that is just a personal preference.

Yeah I hadn't thought too much about this yet, but I will group under staging.

lauriemerrell (approved):
🎉

Issue #1386, opened 2022-04-15
bigquery | track warehouse query history


blakeflei:
### Background
To better understand usage patterns, debug and understand usage (and cost) spikes, know which tables are being used prior to deprecation, and audit data governance patterns, tracking/logging the GCP Bigquery warehouse queries is requested.

### User Story
**As a** provider of GCP BigQuery resources, **I want** to be able to track prior queries to the GCP warehouse **so that** we can diagnose issues with and help better direct BigQuery warehouse usage, and be able to delete data that is incorrect and unused.

### Acceptance Criteria
- There is a method to obtain _who_ has run _what_ queries and _when_ they were run.
- If there is a spike in BigQuery usage, we can find out what was done by whom (e.g. via a query is sufficient - no dashboards needed).
- If we want to deprecate a table, we can find out who has used that table and what queries were run against it.

PR #1548, opened 2022-06-03
start on audit models

atvaccaro:
# Description

Resolves # [issue]

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

### DIFF
--- a/warehouse/models/staging/audit/_stg_audit.yml
+++ b/warehouse/models/staging/audit/_stg_audit.yml
@@ -1,0 +1,5 @@
+version: 2
+
+models:
+  - name: stg_audit__cloudaudit_googleapis_com_activity
+  - name: stg_audit__cloudaudit_googleapis_com_data_access
--- a/warehouse/models/staging/audit/stg_audit__cloudaudit_googleapis_com_activity.sql
+++ b/warehouse/models/staging/audit/stg_audit__cloudaudit_googleapis_com_activity.sql
@@ -1,0 +1,48 @@
+{{
+    config(
+        materialized='incremental',
+        incremental_strategy='insert_overwrite',
+        partition_by={
+            'field': 'date',
+            'data_type': 'date',
+            'granularity': 'day',
+        },
+        partitions=['current_date()'],
+    )
+}}
+-- Note this uses a direct reference instead of source, because a new table is created daily
+
+WITH latest AS (
+    {% set today = modules.datetime.date.today() %}
+
+    SELECT *, current_date() as date
+    FROM cal-itp-data-infra.audit.cloudaudit_googleapis_com_data_access_{{ today.strftime('%Y%m%d') }}
+),
+
+everything AS (
+    {% set start_date = modules.datetime.date(year=2022, month=4, day=11) %}
+    {% set days = (modules.datetime.date.today() - start_date).days + 1 %}
+
+    {% for add in range(days) %}
+
+    {% set current = start_date + modules.datetime.timedelta(days=add) %}
+
+    SELECT *, EXTRACT(DATE FROM timestamp) AS date
+    FROM cal-itp-data-infra.audit.cloudaudit_googleapis_com_data_access_{{ current.strftime('%Y%m%d') }}
+    {% if not loop.last %}
+    UNION ALL
+    {% endif %}
+
+    {% endfor %}
+),
+
+stg_audit__cloudaudit_googleapis_com_activity AS (
+    SELECT *
+    {% if is_incremental() %}
+    FROM latest
+    {% else %}
+    FROM everything
+    {% endif %}
+)
+
+SELECT * FROM stg_audit__cloudaudit_googleapis_com_activity


warehouse/models/staging/audit/stg_audit__cloudaudit_googleapis_com_data_access.sql
```diff
@@ -0,0 +1,95 @@
...
+            'data_type': 'date',
+            'granularity': 'day',
+        },
+        partitions=['current_date()'],
+        cluster_by='job_type',
+    )
+}}
+-- we should use https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions#parse_json when available
+
+-- Note these two source CTEs a direct reference instead of source, because a new table is created daily
```
lauriemerrell, line 16: incredible nitpick: I think this comment is missing a word? 

```suggestion
-- Note these two source CTEs use a direct reference instead of source, because a new table is created daily
```

warehouse/models/staging/audit/stg_audit__cloudaudit_googleapis_com_data_access.sql
```diff
@@ -0,0 +1,95 @@
...
+        *,
+        EXTRACT(DATE FROM timestamp) AS date,
+        -- do some initial renaming/unnesting here since we can't refer to same-SELECT aliases
+        protopayload_auditlog AS payload,
+        protopayload_auditlog.metadataJson AS metadata,
+
+        JSON_QUERY(protopayload_auditlog.metadataJson, '$.jobChange.job') AS job
+
+    {% if is_incremental() %}
+    FROM latest
```
lauriemerrell, line 52: this is a new pattern for us (checking for incremental), not sure if it's worth flagging for a short talk through at some data services internal team meeting about cases where you'd do this / the rationale

warehouse/models/staging/audit/_stg_audit.yml
```diff
@@ -0,0 +1,57 @@
...
+# note: much of these models are derived from blog posts
+# https://towardsdatascience.com/bigquery-3-simple-ways-to-audit-and-monitor-your-dabatase-1cf09027f0de
+# https://dataform.co/blog/exporting-bigquery-usage-logs
+# https://towardsdatascience.com/monitoring-your-bigquery-costs-and-reports-usage-with-data-studio-b77819ffd9fa
+
+models:
+  - name: stg_audit__cloudaudit_googleapis_com_activity
+
+  - name: stg_audit__cloudaudit_googleapis_com_data_access
+    description: Parsed/unnested record of all data access from a log sink.
```
lauriemerrell, line 12: is there a link to anything that we can provide (like Google documentation) that describes what the fields mean at all?

warehouse/models/staging/audit/stg_audit__cloudaudit_googleapis_com_data_access.sql
```diff
@@ -0,0 +1,95 @@
...
+
+-- Note these two source CTEs a direct reference instead of source, because a new table is created daily
+WITH latest AS (
+    {% set today = modules.datetime.date.today() %}
+
+    SELECT *
+    FROM cal-itp-data-infra.audit.cloudaudit_googleapis_com_data_access_{{ today.strftime('%Y%m%d') }}
+),
+
+everything AS (
```
lauriemerrell, line 24: any chance we should add the is_incremental check higher up here to bypass any of these CTEs / make it clearer up front what path we're going to follow? I know that they don't actually get evaluated unless needed, and I like the `latest` / `everything` naming but just wondering

lauriemerrell (approved):
this seems fine as a starting point, so I'm cool to merge the code, however:

1. I'm not sure that this truly resolves #1386 as-is... like it definitely lays the groundwork but I'm not sure you can answer the example questions listed in that ticket directly based on these tables as created. specifically I am wondering if there's an immediate follow up to try to see how/if we can actually identify table/view names in a dedicated column
2. a few inline notes, nbd / up to you on whether to make changes

atvaccaro:
> this seems fine as a starting point, so I'm cool to merge the code, however:
> 
> 1. I'm not sure that this truly resolves #1386 as-is... like it definitely lays the groundwork but I'm not sure you can answer the example questions listed in that ticket directly based on these tables as created. specifically I am wondering if there's an immediate follow up to try to see how/if we can actually identify table/view names in a dedicated column

I think it's technically possible now; `referenced_tables` can be unnested, and unfortunately without Metabase OAuth we're limited in the `who` question.

I did just use this to help with https://github.com/cal-itp/data-infra/pull/1554; I think we should close 1386 when this is merged, and open a new one if someone has a future need that isn't immediately answerable.

warehouse/models/staging/audit/stg_audit__cloudaudit_googleapis_com_data_access.sql
atvaccaro: Eh technically we could but bigquery will ignore CTEs that aren't ultimately used, and I think it's useful to use the CTEs as the "switch" but I could be convinced otherwise.

PR #1551, opened 2022-06-07
rename column

atvaccaro:
_description added by @lauriemerrell_

# Description

Addresses `google.api_core.exceptions.BadRequest: 400 POST https://bigquery.googleapis.com/bigquery/v2/projects/cal-itp-data-infra/datasets/airtable/tables?prettyPrint=false: Invalid field name "#_schedule_feeds". Fields must contain only letters, numbers, and underscores, start with a letter or underscore, and be at most 300 characters long.` error being thrown by new column name in `services` table in California Transit Airtable base.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested? n/a

### DIFF
--- a/airflow/dags/airtable_loader/california_transit_services.yml
+++ b/airflow/dags/airtable_loader/california_transit_services.yml
@@ -9,8 +9,7 @@ id_name: service_id
 
 rename_fields:
   "# Static": num_static
-  # this typo exists in Airtable currently
-  "# Schedule FEeds": num_schedule_feeds
+  "# Schedule Feeds": num_schedule_feeds
   "# TripUpdates": num_trip_updates
   "# Alerts": num_service_alerts
   "# VehiclePositions": num_vehicle_positions
--- a/airflow/dags/airtable_loader/california_transit_services.yml
+++ b/airflow/dags/airtable_loader/california_transit_services.yml
@@ -9,6 +9,8 @@ id_name: service_id
 
 rename_fields:
   "# Static": num_static
+  # this typo exists in Airtable currently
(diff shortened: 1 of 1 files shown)


airflow/dags/airtable_loader/california_transit_services.yml
```diff
@@ -9,6 +9,8 @@ id_name: service_id
 
 rename_fields:
   "# Static": num_static
+  # this typo exists in Airtable currently
+  "# Schedule FEeds": num_schedule_feeds
```
lauriemerrell, line 13: I just fixed this in the Airtable to be "# Schedule Feeds", can just use the normal capitalization

lauriemerrell (commented):
Sorry, I see this is still in draft but just wanted to chime in that I fixed the column name in Airtable. Also, re: offline comment that renaming should happen in dbt: Ideally yes, but these columns (with hashes) are actually illegal in BQ so I don't think we can do that.... We can perhaps discuss on #1509

atvaccaro:
> Sorry, I see this is still in draft but just wanted to chime in that I fixed the column name in Airtable. Also, re: offline comment that renaming should happen in dbt: Ideally yes, but these columns (with hashes) are actually illegal in BQ so I don't think we can do that.... We can perhaps discuss on #1509

Right, IMO we should be replacing anything that isn't `\w` aka `[a-zA-Z0-9_]`. [Technically dashes are valid](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#identifiers) but they require special handling so just easier to just avoid.

lauriemerrell (approved):

PR #1558, opened 2022-06-09
agencies.yml: fix MST RT urls

lauriemerrell:
# Description

MST RT URLs started failing on Tuesday evening and failed all through yesterday (Wednesday). When I checked this morning, I was unable to access the URLs we were using. 

Checking [their website](https://mst.org/about-mst/developer-resources/), it seems that they have new RT URLs.

@o-ram and @evansiroky -- I just wanted to confirm that there's nothing special here because of Payments or anything else? (Like, just wanted to make sure that we weren't using some special URL before for some other reason?) 

## Type of change

- [ ] New feature
- [ ] Documentation
- [x] agencies.yml

## How has this been tested?

I hit the new URLs manually and verified that I can parse the resulting proto files. 

## Screenshots

 I updated the Airtable too: 

image

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -798,9 +798,9 @@ monterey-salinas-transit:
   agency_name: Monterey-Salinas Transit
   feeds:
     - gtfs_schedule_url: http://www.mst.org/google/google_transit.zip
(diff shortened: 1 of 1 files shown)


o-ram (approved):
Can confirm that there are no special considerations here! Looks like the URLs just changed

Issue #1107, opened 2022-02-18
Warehouse Views: Update semantic types in Metabase 


lauriemerrell:
Right now, a lot of our fields in Metabase in the `Warehouse Views` database do not have appropriate semantic types set: 
![image](https://user-images.githubusercontent.com/55149902/154732175-a7b754e9-9e7e-4237-8e5a-0b67213c6249.png)

This creates annoyance because Metabase will automatically bin numeric ID fields and such, which you then have to manually undo every single time you use the field:
![image](https://user-images.githubusercontent.com/55149902/154732384-866f6cb7-64d8-4191-96bb-e2d07b49815d.png)
 
Fixing this would be a pretty significant effort, especially for things like `calitp_itp_id` that might need to be set as foreign keys potentially across many tables (we may need to experiment with how exhaustive those designations need to be.)

atvaccaro:
We can now use https://github.com/gouline/dbt-metabase#semantic-types to define these! In addition, dbt-metabase automatically converts dbt relationship tests into FK semantics, which would help a ton.

PR #1559, opened 2022-06-09
start typing and anchoring important columns

atvaccaro:
# Description

Adds a bunch of semantic types! We should be using non-specific FKs to prevent auto-binning of numerical identifiers such as `calitp_itp_id`.

One opinion in this PR: every table should have an explicit primary key column with not-null and unique tests, rather than just a unique-combination-of-columns test. This is useful when using YAML anchors.

Resolves https://github.com/cal-itp/data-infra/issues/1107

## Type of change

- [x] New feature
- [x] Documentation
- [ ] agencies.yml

### DIFF
--- a/warehouse/models/gtfs_views_staging/agency_clean.sql
+++ b/warehouse/models/gtfs_views_staging/agency_clean.sql
@@ -22,7 +22,7 @@ agency_clean AS (
         TRIM(agency_email) AS agency_email,
         calitp_extracted_at,
         calitp_hash,
-        FARM_FINGERPRINT(CONCAT(CAST(calitp_hash AS STRING), "___", CAST(calitp_extracted_at AS STRING))) AS agency_key,
+        {{ farm_surrogate_key(['calitp_hash', 'calitp_extracted_at']) }} AS agency_key,
         COALESCE(calitp_deleted_at, "2099-01-01") AS calitp_deleted_at
     FROM agency
 )
--- a/warehouse/models/rt_views/_rt_views.yml
+++ b/warehouse/models/rt_views/_rt_views.yml
@@ -42,14 +42,20 @@ models:
         description: Timestamp when this file was extracted
         tests:
           - not_null
-      - name: calitp_itp_id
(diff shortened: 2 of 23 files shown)


warehouse/models/rt_views/_rt_views.yml
```diff
@@ -197,11 +185,15 @@ models:
         tests:
           - dbt_utils.not_null_proportion:
               at_least: 0.95
+        meta:
+          metabase.semantic_type: type/FK
```
lauriemerrell, line 189: TIL you can specify the FK type without specifying a target

warehouse/models/gtfs_views_staging/agency_clean.sql
```diff
@@ -22,7 +22,7 @@ agency_clean AS (
         TRIM(agency_email) AS agency_email,
         calitp_extracted_at,
         calitp_hash,
-        FARM_FINGERPRINT(CONCAT(CAST(calitp_hash AS STRING), "___", CAST(calitp_extracted_at AS STRING))) AS agency_key,
+        {{ farm_surrogate_key(['calitp_hash', 'calitp_extracted_at']) }} AS agency_key,
```
lauriemerrell, line 25: can we just go ahead and use the macro in all the `gtfs_views_staging` tables? I think this pattern is the same everywhere.

lauriemerrell (changes requested):
One request inline, and can we set the `PK` type on all the `dim_<thing>.<thing>_key` columns (like `dim_routes.route_key`)?

warehouse/models/gtfs_views_staging/agency_clean.sql
atvaccaro: done!

atvaccaro:
> One request inline, and can we set the `PK` type on all the `dim_<thing>.<thing>_key` columns (like `dim_routes.route_key`)?

done!

lauriemerrell (approved):
🎉

PR #1562, opened 2022-06-10
Audit and followups

atvaccaro:
# Description

This PR makes a couple fixes based on an investigation into high bigquery billing last week.

1. Removes very expensive and largely useless freshness checks on outcomes files. (BigQuery does not efficiently select the pseudo-columns in hive-partitioned external tables.)
2. Fixes the audit table to increment properly and also include more information, including the dbt node name.
3. Requires partition filters for outcomes external tables; they got bigger than expected.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Locally via dbt CLI.


lauriemerrell (approved):
lgtm, I see that the new dbt columns seem to often be empty, I guess this will be an iterative process of figuring out various attributes we can extract from the `job` and `metadata` columns and just keep adding over time

PR #1575, opened 2022-06-15
Update dbt docs

atvaccaro:
# Description

This PR updates the warehouse README with more detailed setup instructions that should work from scratch and minimize other method/workarounds.

## Type of change

- [ ] New feature
- [x] Documentation
- [ ] agencies.yml

## How has this been tested?
We went through this fresh last week on @edasmalchi's machine.

### DIFF
--- a/warehouse/README.md
+++ b/warehouse/README.md
@@ -37,16 +37,19 @@ This dbt project is intended to be the source of truth for the cal-itp-data-infr
 ### Install poetry and Python/dbt dependencies
 1. Install [poetry](https://python-poetry.org/docs/#osx--linux--bashonwindows-install-instructions) (used for package/dependency management).
    1. If `curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python -`
-      does not work, you can `curl` to a file `curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py`
+      does not work, you can `curl` to a file `curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py -o get-poetry.py`
       and then execute the file with `python get-poetry.py`.
    2. Answer `yes` to adding the tool to your path
 2. Restart your terminal and confirm `poetry --version` works.
 3. `poetry install` to create a virtual environment and install requirements
    1. Now would be the time to [clone](https://docs.github.com/en/repositories/creating-and-managing-repositories/cloning-a-repository)
       the `data-infra` repo if you haven't already. Use [SSH](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account),
-      not HTTPS.
+      not HTTPS. If you haven't made a folder/directory for your git repos yet,
+      you can create one with `mkdir git` (within your home directory, usually).
    2. This needs to be run from within the `warehouse/` directory; you may need
-      to `cd` to it via `cd <git-repos-path>/data-infra/warehouse/` or similar.
+      to `cd` to it via `cd <git-repos-path>/data-infra/warehouse/` or similar;
+      for example, if you had created your directory with `mkdir git`, you will
+      navigate to the warehouse directory with `cd git/data-infra/warehouse/`.
    3. If this doesn’t work because of an error with Python version, you may need to install Python 3.9
    4. `brew install python@3.9`
    5. `brew link python@3.9`
@@ -55,7 +58,14 @@ This dbt project is intended to be the source of truth for the cal-itp-data-infr
 
 ### Initialize your dbt profiles.yml
 1. `poetry run dbt init` inside `warehouse/` will create a `.dbt/` directory in your home directory and a `.dbt/profiles.yml` file.
-2. Fill out your `~/.dbt/profiles.yml` with the following content.
+2. You can enable [displaying hidden folders/files in macOS Finder](https://www.macworld.com/article/671158/how-to-show-hidden-files-on-a-mac.html)
+   but generally, we recommend using the terminal when possible for editing
+   these files. Generally, `nano ~/.dbt/profiles.yml` will be the easiest method
+   for editing your personal profiles file. `nano` is a simple terminal-based
+   text editor; you use the arrows keys to navigate and the hotkeys displayed
+   at the bottom to save and exit. Reading an [online tutorial](https://www.howtogeek.com/howto/42980/the-beginners-guide-to-nano-the-linux-command-line-text-editor/)
+   may be useful if you haven't used a terminal-based editor before.
+3. Fill out your `~/.dbt/profiles.yml` with the following content.
 ```yaml
 calitp_warehouse:
   target: dev
@@ -71,7 +81,7 @@ calitp_warehouse:
       timeout_seconds: 300
       type: bigquery
 ```
-3. See [the dbt docs on profiles.yml](https://docs.getdbt.com/dbt-cli/configure-your-profile) for more background on this file.
+4. See [the dbt docs on profiles.yml](https://docs.getdbt.com/dbt-cli/configure-your-profile) for more background on this file.
 
 ## Running the project locally
 
--- a/warehouse/README.md
+++ b/warehouse/README.md
@@ -4,6 +4,10 @@ This dbt project is intended to be the source of truth for the cal-itp-data-infr
 
 ##  Setting up the project on your local machine
 
+> Note: These instructions assume you are on macOS, but are largely similar for
+> other operating systems. Most *nix OSes will have a package manager that you
+> should use instead of Homebrew.
+
 > Note: if you get `Operation not permitted` when attempting to use the terminal,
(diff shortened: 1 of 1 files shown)


warehouse/README.md
```diff
@@ -3,9 +3,42 @@
...
+> you may need to [fix your terminal permissions](https://osxdaily.com/2018/10/09/fix-operation-not-permitted-terminal-error-macos/)
+
+### Install Homebrew (if you haven't)
+1. Follow the installation instructions at https://brew.sh/
+2. Then, `brew install gdal` to install a geospatial library needed later
+
+### Install the Google SDK (if you haven't)
+0. Implied: make sure that you have GCP permissions (i.e. that someone has added you to the GCP project)
+1. Follow the installation instructions at https://cloud.google.com/sdk/docs/install
+   1. Use `tar -xvf <drag file from Finder window to get file name> ~` to unzip into your home directory
```
lauriemerrell, line 17: per feedback from @edasmalchi I think we need to be a bit more explicit -- saying how to open Terminal & how to navigate to home directory within Terminal

warehouse/README.md
```diff
@@ -3,9 +3,42 @@
...
+   3. Most recent macOS versions use `zsh` as the default shell, so ensure the path modification is added to `~/.zshrc` when prompted
+   4. Restart your terminal, and run `gcloud init`
+   5. Answer `yes` to log in, and select the Google account associated with GCP
+   6. Set `cal-itp-data-infra` as the default project, and do not set a region
+2. You should also set the [set the application default](https://cloud.google.com/sdk/gcloud/reference/auth/application-default)
+3. If `bq ls` shows output, you are good to go.
+
+
+### Install poetry and Python/dbt dependencies
+1. Install [poetry](https://python-poetry.org/docs/#osx--linux--bashonwindows-install-instructions) (used for package/dependency management).
```
lauriemerrell, line 28: I think we need to specify that people should curl to `get_poetry.py` and then run the Python command separately (I had the same problem as Eric where it just printed the results rather than saving to file)

warehouse/README.md
```diff
@@ -3,9 +3,42 @@
...
+   1. Answer `yes` to adding the tool to your path
+2. Restart your terminal and confirm `poetry --version` works.
+3. `poetry install` to create a virtual environment and install requirements
```
lauriemerrell, line 31: this needs to be run within the repo warehouse directory right? should specify

warehouse/README.md
```diff
@@ -3,9 +3,42 @@
...
-1. Create a file `~/.dbt/profiles.yml` (i.e., in your home directory inside a `.dbt` subdirectory) that contains the following:
-```
+
+> Note: if you get `Operation not permitted` when attempting to use the terminal,
+> you may need to [fix your terminal permissions](https://osxdaily.com/2018/10/09/fix-operation-not-permitted-terminal-error-macos/)
+
+### Install Homebrew (if you haven't)
+1. Follow the installation instructions at https://brew.sh/
+2. Then, `brew install gdal` to install a geospatial library needed later
+
```
lauriemerrell, line 13: should add a section here saying that people need to clone the `data-infra` repo if they haven't

lauriemerrell (commented):
If these are going to be the only docs, I think we need more detail per the comments that @edasmalchi was making.... I can potentially add some stuff directly if you want based on my notes from the walkthrough. If we're going to have a supplemental set of instructions for analysts somewhere else maybe this is ok. 

Would also not want to merge until @edasmalchi approves

atvaccaro:
> If these are going to be the only docs, I think we need more detail per the comments that @edasmalchi was making.... I can potentially add some stuff directly if you want based on my notes from the walkthrough. If we're going to have a supplemental set of instructions for analysts somewhere else maybe this is ok.

Which specific comments? I tried to mirror our notes from our test install as much as possible.

warehouse/README.md
```diff
@@ -3,9 +3,60 @@
...
+   6. Answer `yes` to log in, and select the Google account associated with GCP
+   7. Set `cal-itp-data-infra` as the default project, and do not set a region
+3. You should also [set the application default](https://cloud.google.com/sdk/gcloud/reference/auth/application-default)
+4. If `bq ls` shows output, you are good to go.
+
+
+### Install poetry and Python/dbt dependencies
+1. Install [poetry](https://python-poetry.org/docs/#osx--linux--bashonwindows-install-instructions) (used for package/dependency management).
+   1. If `curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python -`
+      does not work, you can `curl` to a file `curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py`
```
edasmalchi, line 40: I think this just puts it on stdout -- should probably be `curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py -o get-poetry.py`

warehouse/README.md
```diff
@@ -3,9 +3,60 @@
...
+      and then execute the file with `python get-poetry.py`.
+   2. Answer `yes` to adding the tool to your path
+2. Restart your terminal and confirm `poetry --version` works.
+3. `poetry install` to create a virtual environment and install requirements
+   1. Now would be the time to [clone](https://docs.github.com/en/repositories/creating-and-managing-repositories/cloning-a-repository)
+      the `data-infra` repo if you haven't already. Use [SSH](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account),
+      not HTTPS.
+   2. This needs to be run from within the `warehouse/` directory; you may need
+      to `cd` to it via `cd <git-repos-path>/data-infra/warehouse/` or similar.
```
edasmalchi, line 49: I think here and in the clone instruction above we should recommend cloning repos in a ~/git/ directory, with instruction to mkdir if necessary

warehouse/README.md
```diff
@@ -3,9 +3,60 @@
...
+   3. If this doesn’t work because of an error with Python version, you may need to install Python 3.9
+   4. `brew install python@3.9`
+   5. `brew link python@3.9`
+   6. After restarting the terminal, confirm with `python3 --version` and retry `poetry install`
+4. `poetry run dbt deps` to install the dbt dependencies defined in `packages.yml` (such as `dbt_utils`)
+
+### Initialize your dbt profiles.yml
+1. `poetry run dbt init` inside `warehouse/` will create a `.dbt/` directory in your home directory and a `.dbt/profiles.yml` file.
```
edasmalchi, line 57: I still think we need specific instructions on how to find+edit a hidden file on macos.

Maybe something like:

"
Note that folders that begin with a dot, like `.dbt` are hidden folders in macOS. `profiles.yml` is just a plain text file, so once we find it we can easily edit it. First, open a new Finder window and make sure Finder displays your home directory under "Favorites" on the left side of the window. If it isn't there, select Finder -> Preferences... using the menu bar, select "Sidebar" in the preferences window, and check the box next to the home directory.

After navigating to the home directory in a finder window, use the shortcut **Shift - Command - .** to show hidden folders. You should then see the `.dbt` folder appear and be able to open `profiles.yml` with TextEdit and edit your profile. After you do, use **Shift - Command - .** again to hide the hidden folders.

If you're comfortable with a command-line text editor, like `nano` or `vim`, you can also use that to edit the file via the terminal.
"

edasmalchi (changes requested):
Thanks @atvaccaro and @lauriemerrell, commented in a couple places where I think more guidance would be helpful

warehouse/README.md
lauriemerrell: Personally I feel like the instructions to access via Finder are harder than just explaining to people how to use `vim`, especially since they can copy-paste and only edit one field (name)? But if you think Finder is worth it that's ok with me

edasmalchi: Honestly not sure, on the one hand yeah it seems simple to use `vim` but writing out those instructions in an equally verbose way also seems complex: 

> First, open your profile using `vim` by running `vim ~/.dbt/profiles.yml`, it should be blank(?).
> Copy the template below, then return to the terminal window with `vim` running. Press i to enter insert mode, then Command-V to paste the template. Press escape to exit insert mode, and move the cursor to the `schema: ` line by using the k and j keys to move up and down, and the h and l keys to move left and right.
> With the cursor on the < character of that line, replace the comment with your name by typing c then $, entering your name, and pressing escape.
> If all looks good, press :wq then enter to save the file.

Like those instructions would probably work, but then if something goes wrong I think a new person would be really stuck. I know the Finder instructions are somewhat convoluted, but I'm hoping they leverage familiar concepts.

lauriemerrell: Yeah fair.... In between could be just having them edit the YML in a text editor and then paste it in using vim?

edasmalchi: That works for me! Probably more future-proof for when we eventually support Windows/WSL, too

atvaccaro: I think I am going to include `nano ~/.dbt/profiles.yml` instructions... that prevents the messiness of GUI text editing (and possibly messing up indentation from pasting) but also is easier than using `vim`.

edasmalchi: That also sounds good

edasmalchi (approved):
lgtm, thanks!

PR #1580, opened 2022-06-17
GTFS downloader v2 - download_feeds task and Pydantic/GCS work!

atvaccaro:
# Description

Must wait on https://github.com/cal-itp/data-infra/pull/1509!

This PR creates some initial Pydantic models to faciliate easy reading/writing of hive-partitioned artifacts in GCS. It also creates a new DAG `download_gtfs_schedule_v2` with a single task to download GTFS Schedule feeds based on Airtable records.

1. Raw GTFS Schedule zip files are saved in GCS
2. Pydantic bases classes drive the GCS I/O, and the subclasses configure metadata, partitions, etc.
3. GCS files keep metadata (defined in #2) describing the configuration that led to their creation

I've also set up the [Airflow variables](https://o1d2fa0877cf3fb10p-tp.appspot.com/variable/list/) containing our existing secrets.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Locally! Right now I'm testing with an Airtable extract from 2022-06-10 as that was the last time that task was run in development. With those records, we end up with `successfully fetched 177 of 189` which is just below our 95% threshold. We will be able to test with newer data once https://github.com/cal-itp/data-infra/pull/1509 is merged.

### DIFF
--- a/airflow/dags/download_gtfs_schedule_v2/download_feeds.py
+++ b/airflow/dags/download_gtfs_schedule_v2/download_feeds.py
@@ -2,6 +2,7 @@
 # python_callable: download_all
 # provide_context: true
 # ---
+import cgi
 import logging
 import traceback
 
@@ -39,19 +40,28 @@ def download_feed(
     resp = s.send(r)
     resp.raise_for_status()
 
-    # we can try a few different locations for the filename
-    disp1 = re.search('filename="(.+)"', resp.headers.get("content-disposition", ""))
-    disp2 = re.search('filename="(.+)"', resp.headers.get("Content-Disposition", ""))
+    disposition_header = resp.headers.get(
+        "content-disposition", resp.headers.get("Content-Disposition")
+    )
+
+    if disposition_header:
+        if disposition_header.startswith("filename="):
+            # sorry; cgi won't parse unless it's prefixed with the disposition type
+            disposition_header = f"attachment; {disposition_header}"
+        _, params = cgi.parse_header(disposition_header)
+        disposition_filename = params.get("filename")
+    else:
+        disposition_filename = None
 
     filename = (
-        (disp1.group(0) if disp1 else None)
-        or (disp2.group(0) if disp1 else None)
+        disposition_filename
         or (os.path.basename(resp.url) if resp.url.endswith(".zip") else None)
         or "feed.zip"
     )
 
     extract = GTFSFeedExtractInfo(
-        filename=filename,
+        # TODO: handle this in pydantic?
+        filename=filename.strip('"'),
         config=record,
         response_code=resp.status_code,
         response_headers=resp.headers,
--- a/airflow/dags/download_gtfs_schedule_v2/download_feeds.py
+++ b/airflow/dags/download_gtfs_schedule_v2/download_feeds.py
@@ -18,7 +18,8 @@ from utils import (
     AirtableGTFSDataExtract,
     AirtableGTFSDataRecord,
     AirtableGTFSDataRecordProcessingOutcome,
-    GTFSFeedExtract,
+    GTFSFeedExtractInfo,
+    GTFSFeedType,
 )
 
 GTFS_FEED_LIST_ERROR_THRESHOLD = 0.95
@@ -48,7 +49,7 @@ def download_feed(
         or "feed.zip"
     )
 
-    extract = GTFSFeedExtract(
+    extract = GTFSFeedExtractInfo(
         filename=filename,
         config=record,
         response_code=resp.status_code,
@@ -75,6 +76,9 @@ def download_all(task_instance, execution_date, **kwargs):
     outcomes: List[AirtableGTFSDataRecordProcessingOutcome] = []
 
     for record in records:
+        if record.data != GTFSFeedType.schedule:
+            continue
+
         try:
             # this is a bit hacky but we need this until we split off auth query params from the URI itself
             jinja_pattern = r"(?P<param_name>\w+)={{\s*(?P<param_lookup_key>\w+)\s*}}"
--- a/airflow/dags/download_gtfs_schedule_v2/download_feeds.py
+++ b/airflow/dags/download_gtfs_schedule_v2/download_feeds.py
@@ -2,6 +2,7 @@
 # python_callable: download_all
 # provide_context: true
 # ---
+import logging
 import traceback
 
 import os
@@ -72,12 +73,17 @@ def download_all(task_instance, execution_date, **kwargs):
     with create_session() as session:
         auth_dict = {var.key: var.val for var in session.query(Variable)}
 
-    records = AirtableGTFSDataExtract.get_latest().records
+    records = [
+        record
+        for record in AirtableGTFSDataExtract.get_latest().records
+        if record.data == GTFSFeedType.schedule
+    ]
     outcomes: List[AirtableGTFSDataRecordProcessingOutcome] = []
 
-    for record in records:
-        if record.data != GTFSFeedType.schedule:
-            continue
+    logging.info(f"processing {len(records)} records")
+
+    for i, record in enumerate(records, start=1):
+        logging.info(f"attempting to fetch {i}/{len(records)} {record.uri}")
 
         try:
             # this is a bit hacky but we need this until we split off auth query params from the URI itself
@@ -90,9 +96,8 @@ def download_all(task_instance, execution_date, **kwargs):
                 record.uri = re.sub(jinja_pattern, "", record.uri)
             outcomes.append(download_feed(record, auth_dict=auth_dict))
         except Exception as e:
-            print(
-                f"exception occurred while attempting to download feed: {str(e)}",
-                traceback.format_exc(),
+            logging.error(
+                f"exception occurred while attempting to download feed {record.uri}: {str(e)}\n{traceback.format_exc()}"
             )
             outcomes.append(
                 AirtableGTFSDataRecordProcessingOutcome(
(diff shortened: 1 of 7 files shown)


airflow/dags/download_gtfs_schedule_v2/download_feeds.py
```diff
@@ -0,0 +1,135 @@
...
+    return AirtableGTFSDataRecordProcessingOutcome(
+        success=True,
+        input_record=record,
+        extract=extract,
+    )
+
+
+def download_all(task_instance, execution_date, **kwargs):
+    start = pendulum.now()
+    # TODO: use laurie's method for this
```
lauriemerrell, line 82: obsolete comment?

airflow/plugins/utils.py
```diff
@@ -1,8 +1,80 @@
...
+def get_auth_secret(auth_secret_key: str) -> str:
+    try:
+        secret = Variable.get(auth_secret_key, os.environ[auth_secret_key])
+        return secret
+    except KeyError as e:
+        logging.error(f"No value found for {auth_secret_key}")
+        raise e
+
+
+def make_name_bq_safe(name: str):
```
lauriemerrell, line 39: not sure why this is showing up in the diff here? 👀 assume it's ok but just odd

airflow/plugins/utils.py
```diff
@@ -81,7 +153,369 @@ def get_successfully_downloaded_feeds(execution_date):
...
+    trip_updates = "trip_updates"
+    vehicle_positions = "vehicle_positions"
+
+
+class AirtableGTFSDataRecord(BaseModel):
+    name: str
+    uri: Optional[str]
+    data: GTFSFeedType
+    schedule_to_use_for_rt_validation: Optional[List[str]]
+    auth_query_param: Dict[str, str] = {}
```
lauriemerrell, line 173: `#TODO: headers` here?

airflow/plugins/utils.py
```diff
@@ -81,7 +153,369 @@ def get_successfully_downloaded_feeds(execution_date):
...
+class PartitionedGCSArtifact(BaseModel, abc.ABC):
+    """
+    This class is designed to be subclassed to model "extracts", i.e. a particular
+    download of a given data source.
+    """
+
+    filename: str
+
+    class Config:
+        json_encoders = {
```
lauriemerrell, line 225: do you still need this here with the `PETITION_SERIALIZERS` above? seems the same thing?

airflow/plugins/utils.py
```diff
@@ -81,7 +153,369 @@ def get_successfully_downloaded_feeds(execution_date):
...
+
+    class Config:
+        arbitrary_types_allowed = True
+        json_encoders = {Exception: lambda e: str(e)}
+
+    # TODO: is this really useful?
+    @property
+    @abc.abstractmethod
+    def input_type(self) -> typing.Type[PartitionedGCSArtifact]:
+        """The input type that was processed to produce this outcome.."""
```
lauriemerrell, line 329: extra period at end

airflow/plugins/utils.py
```diff
@@ -81,7 +153,369 @@ def get_successfully_downloaded_feeds(execution_date):
...
+    directory = GCSDirectoryInfo(**fs.info(table_path))
+
+    for key in keys:
+        # TODO: this just sorts by string values, but we should parse stuff out
+        directory = sorted(
+            directory.children(fs), key=lambda o: o.partition[key], reverse=True
+        )[0]
+
+    children = directory.children(fs)
+    if len(children) != 1:
```
lauriemerrell, line 396: clarification: is this an *us* thing or a general hive partitions thing to expect only one file in the final folder? are we sure this will always be the case?

airflow/plugins/utils.py
```diff
@@ -81,7 +153,369 @@ def get_successfully_downloaded_feeds(execution_date):
...
+    def time(self) -> pendulum.Time:
+        return self.download_time.time()
+
+
+class AirtableGTFSDataRecordProcessingOutcome(ProcessingOutcome):
+    input_type: ClassVar[typing.Type[PartitionedGCSArtifact]] = GTFSFeedExtractInfo
+    extract: Optional[GTFSFeedExtractInfo]
+
+
+# class GTFSFeedExtractOutcome(PartitionedGCSArtifact):
```
lauriemerrell, line 477: just to confirm -- I should actually implement this as part of the `gtfs_loader` work, right?

airflow/plugins/utils.py
```diff
@@ -81,7 +153,369 @@ def get_successfully_downloaded_feeds(execution_date):
...
+#     def table(self) -> str:
+#         return f"{self.extract.feed.type}_outcomes"
+#
+#     @property
+#     def bucket(self) -> str:
+#         pass
+
+
+if __name__ == "__main__":
+    # records = AirtableGTFSDataExtract.get_latest().records
```
lauriemerrell, line 497: do we need this commented stuff?

airflow/plugins/utils.py
```diff
@@ -81,7 +153,369 @@ def get_successfully_downloaded_feeds(execution_date):
...
+    #     download_time=pendulum.now(),
+    # )
+    # extract.save_content(get_fs(), b"")
+    # print(
+    #     AirtableGTFSDataRecordProcessingOutcome(
+    #         success=True,
+    #         extract=extract,
+    #     ).json(indent=4)
+    # )
+    objs = fetch_all_in_partition(
```
lauriemerrell, line 512: do we want to leave this hard coded in here? maybe this should be more clearly labeled as like a test date or something?

lauriemerrell (changes requested):
A few tiny nitpicks inline, and how do we want to track substantive TODOs from here? I think that:

* Saving Airtable processing outcomes
* Typed sorting of partitions 

Are both fairly significant. Do they need dedicated follow up tickets?

airflow/dags/download_gtfs_schedule_v2/download_feeds.py
```diff
@@ -0,0 +1,135 @@
...
+GTFS_FEED_LIST_ERROR_THRESHOLD = 0.95
+
+
+def download_feed(
+    record: AirtableGTFSDataRecord, auth_dict: Dict
+) -> AirtableGTFSDataRecordProcessingOutcome:
+    if not record.uri:
+        raise ValueError("")
+
+    parse_obj_as(HttpUrl, record.uri)
```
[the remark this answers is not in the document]
atvaccaro: I think we do want HttpUrl as it's a bit stricter, especially for checking there's a good TLD... I could be convinced otherwise.

https://pydantic-docs.helpmanual.io/usage/types/#urls

airflow/dags/download_gtfs_schedule_v2/download_feeds.py
```diff
@@ -0,0 +1,135 @@
...
+                )
+            )
+
+    # TODO: save the outcomes somewhere
+
+    print(f"took {pendulum.now() - start} to process {len(records)} records")
+
+    assert len(records) == len(
+        outcomes
+    ), f"we somehow ended up with {len(outcomes)} from {len(records)}"
```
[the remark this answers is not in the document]
atvaccaro: oops yes!

airflow/plugins/utils.py
atvaccaro: that is odd... I will rebase to double checK?

airflow/plugins/utils.py
atvaccaro: ah yes we do, the partition de/serializers are specific to partitions, whereas `json_encoders` configure Pydantic-serialized JSON

airflow/plugins/utils.py
atvaccaro: This is specific to us and I could be convinced otherwise. For example, maybe we just partition down to the e.g. hour and then timestamp within the underlying files.

airflow/plugins/utils.py
atvaccaro: Yes I just left this stub in, I can remove it if you'd prefer.

airflow/plugins/utils.py
atvaccaro: It's nice for easy debugger entry in PyCharm, but that's it.

airflow/plugins/utils.py
atvaccaro: Yes I can add a comment.

atvaccaro:
> * Saving Airtable processing outcomes
> * Typed sorting of partitions

These are both done now!

airflow/plugins/utils.py
```diff
@@ -1,8 +1,78 @@
...
+import gcsfs
+import humanize
+import pandas as pd
+import pendulum
+from airflow.models import Variable
 from calitp import read_gcfs, save_to_gcfs
+from calitp.config import is_development
+from calitp.storage import get_fs
 from pandas.errors import EmptyDataError
+from pydantic import AnyUrl, validator, Field
```
lauriemerrell, line 20: should we change this (and the usage below) to HttpUrl?

airflow/plugins/utils.py
lauriemerrell: this is still weird... just want to triple check?

airflow/plugins/utils.py
lauriemerrell: I think that this is ok but I wonder if we want to at least add a comment here being explicit about this just in case we do need to change it later

lauriemerrell (approved):
optional comments 👍

airflow/plugins/utils.py
atvaccaro: ok I think this is fine and just a consequence of the rebase

PR #1582, opened 2022-06-22
Reports to warehouse

edasmalchi:
# Description

New table to consolidate and improve metric generation for reports. Replaces complex Siuba code in reports notebook and allows for visibility into service hour changes before the report generation process.

Resolves https://github.com/cal-itp/reports/issues/129

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

Ran for 2 months via dbt, that processed 1.4GB.

### DIFF
--- a/warehouse/models/gtfs_views/_gtfs_views.yml
+++ b/warehouse/models/gtfs_views/_gtfs_views.yml
@@ -843,7 +843,8 @@ models:
           - not_null
   - name: reports_daily_service_route_stops
     description: |
-      New table to consolidate and improve metric generation for reports.
+      Table consolidating daily service hours, routes, and stops by feed,
+      filtered to the specific feed used for each monthly report.
       Replaces complex Siuba code in reports notebook and allows for 
       visibility into service hour changes before the report generation process.
       Primary key is composite of service_date, calitp_itp_id, calitp_url_number.
--- a/warehouse/models/gtfs_views/_gtfs_views.yml
+++ b/warehouse/models/gtfs_views/_gtfs_views.yml
@@ -845,7 +845,7 @@ models:
     description: |
       Table consolidating daily service hours, routes, and stops by feed,
       filtered to the specific feed used for each monthly report.
-      Replaces complex Siuba code in reports notebook and allows for 
+      Replaces complex Siuba code in reports notebook and allows for
       visibility into service hour changes before the report generation process.
       Primary key is composite of service_date, calitp_itp_id, calitp_url_number.
     tests:
@@ -878,11 +878,11 @@ models:
       - name: last_departure_ts
         description: Time of last departure on date
       - name: n_stops
-        description: | 
+        description: |
          Total number of stops present in active feed
          (not necessarily served on date)
       - name: n_routes
-        description: | 
+        description: |
          Total number of routes present in active feed
          (not necessarily operating on date)
 
--- a/warehouse/models/gtfs_views/_gtfs_views.yml
+++ b/warehouse/models/gtfs_views/_gtfs_views.yml
@@ -841,6 +841,49 @@ models:
         description: Field in validation output from GTFS validator
         tests:
           - not_null
+  - name: reports_daily_service_route_stops
+    description: |
+      New table to consolidate and improve metric generation for reports.
+      Replaces complex Siuba code in reports notebook and allows for 
+      visibility into service hour changes before the report generation process.
+      Primary key is composite of service_date, calitp_itp_id, calitp_url_number.
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - service_date
+            - calitp_itp_id
+            - calitp_url_number
+    columns:
+      - name: service_date
+        description: Date on which this service level was present
+        tests:
+          - not_null
+      - name: calitp_itp_id
+        description: '{{ doc("column_calitp_itp_id") }}'
+        tests:
+          - not_null
+      - name: calitp_url_number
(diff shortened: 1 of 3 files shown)


edasmalchi:
will add yaml change and link reports repo pr before merging

lauriemerrell (changes requested):
Eric to add YAML

warehouse/models/gtfs_views/_gtfs_views.yml
```diff
@@ -841,6 +841,49 @@ models:
         description: Field in validation output from GTFS validator
         tests:
           - not_null
+  - name: reports_daily_service_route_stops
+    description: |
+      New table to consolidate and improve metric generation for reports.
```
lauriemerrell, line 846: I would refactor this a bit to be a generic description of the table content / purpose -- this description will persist in general, so table will no longer be "new".

warehouse/models/gtfs_views/reports_daily_service_routes_stops.sql
```diff
@@ -0,0 +1,146 @@
...
+    SELECT
+        T1.*,
+        T2.calitp_itp_id,
+        T2.calitp_url_number,
+    FROM stops_routes_agg AS T1
+    INNER JOIN gtfs_schedule_dim_feeds AS T2
+        ON T1.feed_key = T2.feed_key
+),
+
+date_renamed AS (
```
lauriemerrell, line 73: stylistic optional thing: you could refactor the original CTE to just be the below instead of having dim_date separately. not a big deal but don't necessarily need to have such a minimal CTE as a separate item


```
dates AS (
    SELECT full_date AS service_date
    FROM {{ ref('dim_date') }}
),
```

warehouse/models/gtfs_views/reports_daily_service_routes_stops.sql
```diff
@@ -0,0 +1,146 @@
...
+    SELECT full_date AS service_date
+    FROM dim_date
+),
+
+reports_feeds_distinct AS (
+    SELECT DISTINCT calitp_itp_id, calitp_url_number, use_for_report
```
lauriemerrell, line 79: question: do you actually need `use_for_report` here?

warehouse/models/gtfs_views/reports_daily_service_routes_stops.sql
```diff
@@ -0,0 +1,146 @@
...
+        T1.n_stops,
+        T1.service_date,
+        T2.n_routes,
+    FROM stops_agg AS T1
+    INNER JOIN routes_agg AS T2
+        ON T1.service_date = T2.service_date
+            AND T1.feed_key = T2.feed_key
+),
+
+stops_routes_metadata_joined AS (
```
lauriemerrell, line 63: optional suggestion: this could probably be combined with the above (just add one more join)

lauriemerrell (changes requested):
One specific change re: table description, otherwise looks ok to me... 

One stylistic thing is that this is kind of a lot of CTEs, if it were me I might try to combine a few just to make it a bit shorter, but I don't think it's a huge deal/requirement. [This post](https://discourse.getdbt.com/t/why-the-fishtown-sql-style-guide-uses-so-many-ctes/1091) says that they try to keep their queries to <100 lines of SQL, which might be a good ballpark goal.

warehouse/models/gtfs_views/reports_daily_service_routes_stops.sql
edasmalchi: ah, probably not since the second join should handle it

warehouse/models/gtfs_views/reports_daily_service_routes_stops.sql
```diff
@@ -0,0 +1,132 @@
...
+        T1.n_stops,
+        T1.service_date,
+        T2.n_routes,
+        T3.calitp_itp_id,
+        T3.calitp_url_number
+    FROM stops_agg AS T1
+    INNER JOIN routes_agg AS T2
+        ON T1.service_date = T2.service_date
+            AND T1.feed_key = T2.feed_key
+    INNER JOIN gtfs_schedule_dim_feeds AS T3
```
edasmalchi, line 63: Is this the right way to combine those joins?

edasmalchi:
thanks for the comments @lauriemerrell, tried to fix in the latest commit

atvaccaro (approved):
lgtm I think the distincts might be unnecessary tho

warehouse/models/gtfs_views/reports_daily_service_routes_stops.sql
lauriemerrell: Yes, looks good to me!

lauriemerrell (approved):
LGTM! I see Andrew has at least one comment so I defer to him on that. Congrats on first dbt/SQL PR in data-infra! 🎉

atvaccaro:
oh @edasmalchi have you set up pre-commit locally? it looks like they are failing in CI/CD

warehouse/models/gtfs_views/reports_daily_service_routes_stops.sql
```diff
@@ -0,0 +1,132 @@
...
+),
+
+routes_agg AS (
+    SELECT
+        feed_key,
+        date AS service_date,
+        COUNT(DISTINCT route_key) AS n_routes
+    FROM gtfs_schedule_fact_daily_feed_routes
+    GROUP BY 1, 2
+),
```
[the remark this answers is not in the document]
edasmalchi: good point, removed them here

PR #1583, opened 2022-06-22
dbt migration of remaining Airtable views

charlie-costanzo:
🚨 _Note that this is merging into the `airtable-updates` branch and not main

# Description

This is another intermediate PR that will be merged into main via #1509. 

This PR:
* Migrates the remaining airtable tables to dbt, reading from the new external airtable tables
* Creates latest-only staging tables for airtable transit database
* Unnests single value arrays in columns that include links to other records
* Updates `_stg_transit_database.yml` to include tests and descriptions
* Utilizes new naming conventions

## Type of change
- [x] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) - deletes a DAG?
- [ ] Documentation

## How has this been tested?
* Verified I can run `dbt run` 
* Verified `dbt test` passes
* Verified `dbt docs generate` runs


lauriemerrell (changes requested):
Thanks Charlie! This looks great as a start, but I think we should also make staging tables for the other Airtable tables listed here (i.e., would be good to have staging tables even for those that don't make it into the mart):

image

And for the two listed here, I would actually maybe list all the columns that are in those tables in the screenshot rather than listing just the tables that we have in `views` now (I would say that the `views` tables are more the equivalent of the `mart` tables, and for these staging tables we probably want to have basically all the columns available and then we make curation decisions at the mart level.) I hope that makes sense -- we can def discuss on our call later but wanted to get this comment in in case you have time to review before that.

warehouse/models/staging/transit_database/stg_transit_database__county_geography.sql
```diff
@@ -0,0 +1,28 @@
...
+
+stg_transit_database__county_geography AS (
+    SELECT
+        county_geography_id AS key,
+        {{ trim_make_empty_string_null(column_name = "name") }},
+        fips,
+        msa,
+        caltrans_district,
+        caltrans_district_name,
+        unnested_rtpa AS rtpa,
```
lauriemerrell, line 19: this was my bad -- can we update these to be like `rtpa_organization_key` to list what table it's a foreign key to? (would also need to update YAML with new name)

warehouse/models/staging/transit_database/stg_transit_database__eligibility_programs.sql
```diff
@@ -0,0 +1,31 @@
...
+        order_by = 'dt DESC, time DESC'
+        ) }}
+),
+
+stg_transit_database__eligibility_programs AS (
+    SELECT
+        eligibility_program_id AS key,
+        {{ trim_make_empty_string_null(column_name = "program") }},
+        unnested_administering_entity AS administering_entity,
+        unnested_eligibility_types AS eligibility_types,
```
lauriemerrell, line 16: same comment as above -- this would become administering_entity_organization_key and eligibility_type_rider_requirement_key and services below would become service_key

warehouse/models/staging/transit_database/stg_transit_database__fare_systems.sql
```diff
@@ -0,0 +1,58 @@
...
+        reg_adult_ticket_price_min,
+        reg_adult_ticket_price_max,
+        pass_price_min,
+        pass_price_max,
+        fare_capping,
+        paratransit_fare_url,
+        demand_response_fare_url,
+        mobility_services_managed__from_transit_provider_,
+        gtfs_dataset__from_mobility_services_managed___from_transit_provider_,
+        unnested_transit_provider AS transit_provider,
```
lauriemerrell, line 49: transit_provider_organization_key

warehouse/models/staging/transit_database/stg_transit_database__gtfs_datasets.sql
```diff
@@ -0,0 +1,46 @@
...
+        {{ trim_make_empty_string_null(column_name = "name") }},
+        data,
+        fares_v2_status,
+        fares_notes,
+        pathways_status,
+        schedule_comments,
+        uri,
+        future_uri,
+        api_key,
+        unnested_aggregated_to AS aggregated_to,
```
lauriemerrell, line 23: aggregated_to_gtfs_dataset_key

warehouse/models/staging/transit_database/stg_transit_database__gtfs_service_data.sql
```diff
@@ -0,0 +1,47 @@
...
+        external_table = source('airtable', 'california_transit__gtfs_service_data'),
+        order_by = 'dt DESC, time DESC'
+        ) }}
+),
+
+stg_transit_database__gtfs_service_data AS (
+    SELECT
+        gtfs_service_data_id AS key,
+        {{ trim_make_empty_string_null(column_name = "name") }},
+        unnested_services AS services,
```
lauriemerrell, line 15: service_key and below gtfs_dataset_key and reference_static_gtfs_service_data_key

warehouse/models/staging/transit_database/stg_transit_database__ntd_agency_info.sql
```diff
@@ -0,0 +1,59 @@
...
+        tribal_area_name,
+        population,
+        density,
+        sq_miles,
+        voms_do,
+        voms_pt,
+        total_voms,
+        volunteer_drivers,
+        personal_vehicles,
+        unnested_organizations AS organizations,
```
lauriemerrell, line 52: organization_key

warehouse/models/staging/transit_database/stg_transit_database__place_geography.sql
```diff
@@ -0,0 +1,35 @@
...
+    SELECT
+        place_geography_id AS key,
+        {{ trim_make_empty_string_null(column_name = "name") }},
+        place_fips,
+        placename,
+        type,
+        place_base,
+        place_formal,
+        county,
+        unnested_county_base AS county_base,
```
lauriemerrell, line 21: county_geography_key

lauriemerrell (changes requested):
mostly looks good, just one more naming thing -- I think all foreign keys should be like `<optional relationship summary>_table_name_key` to make it explicit that they are a foreign key

lauriemerrell (approved):
LGTM! 🎉

PR #1584, opened 2022-06-23
Update Montebello Schedule URL

evansiroky:
# Description

Montebello Bus Lines was able to give us a publicly-available URL to download data from.

## Type of change

- [ ] New feature
- [ ] Documentation
- [x] agencies.yml

## How has this been tested?

Downloaded locally.


lauriemerrell (approved):
Thanks Evan -- confirmed can download & that downloaded file appears to be an active feed; confirmed Airtable also updated with new URL

PR #1588, opened 2022-06-24
fix: transit tech stacks -- fix unnesting

lauriemerrell:
_🚨 merging into feature branch not main_

# Description

Fixes unnesting-- switches from `CROSS JOIN` to `LEFT JOIN` where applicable to stop dropping rows where the unnested array is null. 

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml


atvaccaro (approved):
legit

charlie-costanzo (approved):
nice, LGTM!

PR #1591, opened 2022-06-27
Add all current MTC 511 agencies

evansiroky:
# Description

After inspecting all current agencies that 511 claims to support, I'm adding the ones that were missing in agencies.yml.

## Type of change

- [ ] New feature
- [ ] Documentation
- [x] agencies.yml

## How has this been tested?

Locally download missing agencies.

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -313,6 +313,10 @@ commute-org-shuttles:
       gtfs_rt_vehicle_positions_url: null
       gtfs_rt_service_alerts_url: null
       gtfs_rt_trip_updates_url: null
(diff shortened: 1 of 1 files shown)


evansiroky:
> At least some of these don't actually return substantive realtime data

I think we ought to keep the URLs in the event that they do end up producing realtime data in the future. Eventually those will be the URLs.

> Treasure Island Ferry has not actually been present in the combined MTC feed

That is very interesting. I didn't check the regional feed against all of this.

> For Golden Gate, the list of routes is actually fairly different between URL 0 and the new MTC URL.

Yes, I believe that for some reason the MTC feed separates the bus and ferry routes in the precursor Golden Gate feed.

PR #1592, opened 2022-06-27
Revert "Add all current MTC 511 agencies"

atvaccaro:
Reverts cal-itp/data-infra#1591

We think we've hit a throughput limit; [seeing a step function in Grafana](https://monitoring.k8s.calitp.jarv.us/d/xPWY4jL7z/gtfs-rt-achiver?orgId=1&from=1656328443522&to=1656347186379) that started when this was deployed.


lauriemerrell (approved):
approving for troubleshooting, we will add back as we figure out what's going on

PR #1603, opened 2022-06-28
add missing columnName column to schedule validation notices schema

atvaccaro:
# Description

We had a failure last night in `validation_notices_load` due to bad JSON parsing. A new schedule feed is resulting in a new validation notice (specifically, a non-ASCII character) which has a field called `columnName` which we haven't used before. I'm unsure how this differs from `fieldName`.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
tbd

### DIFF
--- a/airflow/dags/gtfs_schedule_history/validation_report.yml
+++ b/airflow/dags/gtfs_schedule_history/validation_report.yml
@@ -29,6 +29,8 @@ schema_fields:
       fields:
         - name: csvRowNumber
           type: INTEGER
+        - name: columnName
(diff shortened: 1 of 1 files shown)


lauriemerrell:
As we've discussed offline I think we should submit an issue to the MD validator for this -- this seems like it should be `fieldName` like it is elsewhere

lauriemerrell (approved):
I submitted https://github.com/MobilityData/gtfs-validator/issues/1205. Good to merge

PR #1604, opened 2022-06-29
Update links for Angel Island-Tiburon Ferry Company

evansiroky:
# Description

Angel Island-Tiburon Ferry Company published data for their ferry company on MTC 511. This PR adds the URLs for this respective feed.

## Type of change

- [ ] New feature
- [ ] Documentation
- [x] agencies.yml

## How has this been tested?

Confirmed with agency, downloaded locally.

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -1575,10 +1575,10 @@ santa-rosa-citybus:
 angel-island-tiburon-ferry-company:
   agency_name: Angel Island-Tiburon Ferry Company
   feeds:
-    - gtfs_schedule_url: http://data.trilliumtransit.com/gtfs/bayareaferries-ca-us/bayareaferries-ca-us.zip
-      gtfs_rt_vehicle_positions_url: null
-      gtfs_rt_service_alerts_url: null
-      gtfs_rt_trip_updates_url: null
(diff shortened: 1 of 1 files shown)


lauriemerrell (approved):
Thanks @evansiroky -- before merging, I just wanted to confirm that it looks like we are keeping the Trillium Bay Area Ferries link in `agencies.yml` for the `Blue and Gold Fleet` (ITP ID `41`). I just want to confirm that that is correct?

I also think that this would be related to some potentially complicated Airtable updates -- right now there is just one entry that is combined between a bunch of ferries. I can add this to the Airtable TODO doc if appropriate.

evansiroky:
> Thanks @evansiroky -- before merging, I just wanted to confirm that it looks like we are keeping the Trillium Bay Area Ferries link in `agencies.yml` for the `Blue and Gold Fleet` (ITP ID `41`). I just want to confirm that that is correct?
> 
> I also think that this would be related to some potentially complicated Airtable updates -- right now there is just one entry that is combined between a bunch of ferries. I can add this to the Airtable TODO doc if appropriate.

Yes, the Angel Island - Tiburon Ferry has split out from the old expired Trillium feed and is now listed as its own agency exclusively within the MTC feed. I'll update airtable to reflect this.

PR #1611, opened 2022-06-30
Update AVTA URL

evansiroky:
# Description

Updates the AVTA URL to a new URL that is on their website. (See https://www.avta.com/gtfs.php).

## Type of change

- [ ] New feature
- [ ] Documentation
- [x] agencies.yml

## How has this been tested?

Received communication to use the URL found on the referenced page, downloaded locally.

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -49,7 +49,7 @@ amtrak:
 antelope-valley-transit-authority:
   agency_name: Antelope Valley Transit Authority
   feeds:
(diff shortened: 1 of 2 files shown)


atvaccaro (approved):
lgtm, I re-triggered the failed action

evansiroky:
I have updated the PR with the URL noted on their website. I didn't do it at first because the feed at that URL was expired at first. After clarifying this with staff at the transit agency, they have fixed the issue so we can now move forward with this GTFS Schedule URL.

lauriemerrell (approved):

PR #1623, opened 2022-07-08
re-add RT limits

atvaccaro:


lauriemerrell (approved):
lgtm, talked about this offline with Andrew. We agreed on 1gb limit because we think throughput issues have been addressed so restarts may help again. We think setting request memory param is not directly applicable but doesn't hurt us.

Issue #1387, opened 2022-04-15
Dashboard modification request: filter out gtfs datasets based on deprecated data in agencies.yml vs airtable dashboard


evansiroky:
## User Story

*As a* transit data quality analyst using the [agencies.yml URLs vs airtable URIs dashboard](https://dashboards.calitp.org/dashboard/58-agencies-yml-urls-vs-airtable-uris),
*I want to* know which gtfs datasets in airtable have been deprecated,
*so that* I can make sure the agencies.yml file is up-to-date

## Acceptance criteria

*Given* data from the airtable gtfs datasets table
  *and* data of the agencies.yml file
*When* data is compared to determine what URIs exist in airtable that are missing in the agencies.ym file
*Then* only those records from the gtfs datasets that are not marked as deprecated should be considered

### Context

In the [agencies.yml URLs vs airtable URIs dashboard](https://dashboards.calitp.org/dashboard/58-agencies-yml-urls-vs-airtable-uris), all questions in the section `URIs in airtable gtfs datasets table that are not in agencies.yml` should be edited to filter out gtfs dataset records that are marked as deprecated.

This issue depends on #1343.

lauriemerrell:
Reopening until Metabase connection for this data is restored

lauriemerrell:
Re-closing now that v2 dashboard is once again available.

PR #1625, opened 2022-07-08
Airtable dbt mart initial work

### LINKED ISSUE #1159: Transit Technology Stacks

### LINKED ISSUE #1343: Airtable: Extract the new `deprecated date` field for GTFS datasets to warehouse
Per [this Slack post](https://cal-itp.slack.com/archives/C014Q6G3VCJ/p1649289639285859), we have a new `deprecated date` field in the GTFS Datasets Airtable data that should make it to the warehouse because it informs what data is valid at what times.

### LINKED ISSUE #1487: New airtable view: organizations + gtfs_service_data
### **Is your feature request related to a problem? Please describe.**
To compliment the `GTFS Agency-Use Reference` document in progress, I think a new airtable view should be created to compliment that resource and allow analysts to easily access data supporting that definition use-case.

### **Describe the solution you'd like**
A new view from airtable called something like `views.airtable_california_transit_organizations_x_gtfs_service_data_`, which includes a join of the following from airtable:

**table**: `gtfs_service_data`
* services (rename = 'mobility_service')
* agency_id
* network_id
* route_id
* category
* dataset_type
* gtfs_dataset

**table**: `organzations`
* name (rename = 'organization_name')
* mobility_services_managed (rename = 'mobility_service')
  * **This needs to be unnested**

### **Describe alternatives you've considered**
Describing in the docs how analysts can isolate that data themselves. But seeing as that would still require changes to the codebase to complete I believe it is worthwhile to just spin up a new view.

### Additional context
**Direction provided by E:**
> Provider’s service represented in GTFS = sum of “primary” gtfs-service relationships where service is managed by that provider

**Destination table preview (with changed column names)**:
Screen Shot 2022-05-09 at 1 21 23 PM

**Preview agency-use reference docs [DRAFT]**:
[What is an agency?](https://data-infra-1430--cal-itp-previews.netlify.app/warehouse/what_is_agency.html) preview document

### LINKED ISSUE #1630: Request for additional airtable data availability in metabaes
**Is your feature request related to a problem? Please describe.**

As a transit data quality analyst, I want to be able to view information about which operators are assessed and other relevant information.

**Describe the solution you'd like**

I would like the following additional fields from airtable available to use in metabase:

`organizations.website`
`services.operating_counties`
`services.funding_sources`
`gtfs_datasets.deprecated_date`

I would also like a new table `funding programs`. 
`funding programs.program`
`funding programs.funding_program_id`
`funding program.category`

I would also like to be able to link data from services to data in funding programs.

However, with `funding programs` and `services`, if I can merely use the value of `funding programs.program` inside a field within `services` that is an array instead of doing a big join across tables, then that would perhaps work even better.

**Describe alternatives you've considered**

Doing stuff with Jupyter notebooks.

lauriemerrell:
Only remaining TODO (shouldn't block PR review): checking legacy Airtable usage in Metabase.

# Description

This enables the tables in the `mart_transit_database` BQ dataset / dbt schema. Specifically, it creates new tables with data extracted from the Transit Tech Stacks Airtable base (#1159) and refactors the California Transit base data, creating new `dim_<thing>` tables. 

At this stage, the only existing data not migrated into the new format are a few mapping tables that I do not believe are used (ex. organizations <--> services operated). All the relevant data is available in staging tables, so if we need to re-create these, we can do so easily.

* Resolves #1159 
* Resolves #1343 (and thus unblocks #1387)
* Resolves #1487 
* Resolves #1630  

Also contains a few updates authored by @atvaccaro to change the dbt failure handling in the dbt run Airflow task -- specifically, to stop a failure within dbt (like a single table erroring) from stopping the docs from being updated, and to change the dependency on the dbt test task so that it will run as long as the prior task *finishes* (as opposed to *succeeds*). 

**New question for @atvaccaro: any concern with merging `legacy_airtable` as a source? I committed by accident (overzealous select), not sure if it's worth pulling it out of this PR given that there are still followups -- #1626.)** 

_After merge, I will need to re-create the existing agencies.yml / Airtable comparison Metabase dashboard to use these new warehouse tables instead of the old ones. I will do this as soon as the PR merges and the new tables run in the prod warehouse. Once that is done we can turn off the old Airtable pipeline (#1626)._

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested? 

`dbt run` & `dbt test`; confirmed all issues are unrelated to this work (ex. in GTFS schedule pipeline)

## Screenshots (optional)

**Resolved Questions for reviewers (specifically @atvaccaro):** 
* ~Do we think that mapping tables for many to many relationships (where the only columns are `key1`, `key2`, `name1`, and `name2`) need to have a separate primary key column defined? Their primary key currently is the composite of `key1` and `key2`. We can make a dedicated `key` column but it just feels like overkill to me, because you would only ever join with these via the existing keys.~
   * Resolved: no synthetic key for bridge tables
* ~How do we feel about using the `map_` naming convention? And specifically how do we feel about it for `services x products`, which has other columns? In this PR I am a bit inconsistent with `dim_provider_gtfs_data` as a dimensional table and `map_services_x_products` as a `map_`. `gtfs_service_data` is a mapping table that's defined in Airtable so I left that as a dimension but maybe we would want to do something else.~
    *  Resolved: using `bridge_` naming per Kimball


warehouse/models/mart/transit_database/_mart_transit_database.yml
```diff
@@ -9,10 +9,26 @@ models:
...
     description: '{{ doc("contracts_table") }}'
     columns:
       - *key
+      - name: contract_holder_organization_key
+        tests:
+          - relationships:
+             to: ref('dim_organizations')
+             field: key
+        meta:
+          metabase.semantic_type: type/FK
```
atvaccaro, line 24: I think dbt-metabase will already handle FKs based on relationships tests, so this is redundant

same in other places

warehouse/models/mart/transit_database/_mart_transit_database.yml
```diff
@@ -9,10 +9,26 @@ models:
         tests: &primary_key_tests
           - not_null
           - unique
+        meta:
+          metabase.semantic_type: type/PK
```
atvaccaro, line 13: I wonder if we should alter dbt-metabase to flag PKs based on not-null and unique tests existing?

warehouse/models/staging/transit_database/base/_base_transit_database.yml
```diff
@@ -1,7 +1,7 @@
 version: 2
 
 models:
-  - name: int_tts_services_ct_services_map
+  - name: base_tts_services_ct_services_map
```
atvaccaro, line 4: I don't think you can use multiple combination_of_columns keys under the same unique_combination_of_columns test; the YAML spec is to ignore dupes IIRC

warehouse/models/mart/transit_database/dim_service_components.sql
```diff
@@ -0,0 +1,68 @@
...
+        calitp_extracted_at
+    FROM latest
+),
+
+unique_service_components AS (
+    SELECT
+        * EXCEPT(rank)
+    FROM ranked_service_components
+    WHERE rank = 1
+),
```
atvaccaro, line 41: nice, apparently bigquery does support QUALIFY now, could use that here

https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#qualify_clause

warehouse/models/mart/transit_database/dim_properties_and_features.sql
```diff
@@ -1,7 +1,10 @@
 {{ config(materialized='table') }}
 
-WITH stg_transit_database__properties_and_features AS (
-    SELECT * FROM {{ ref('stg_transit_database__properties_and_features') }}
+WITH latest AS (
+    {{ get_latest_dense_rank(
```
atvaccaro, line 4: do these not need a partition-by?

warehouse/models/mart/transit_database/dim_contract_attachments.sql
```diff
@@ -1,18 +1,22 @@
 {{ config(materialized='table') }}
 
-WITH stg_transit_database__contracts AS (
-    SELECT * FROM {{ ref('stg_transit_database__contracts') }}
+WITH latest AS (
```
atvaccaro, line 3: nit: I think aliasing all of these latest CTEs with `latest` is a tad confusing but don't have to change now

I think something like `latest_contracts` would be clearer?

warehouse/models/mart/transit_database/dim_provider_gtfs_data.sql
```diff
@@ -0,0 +1,66 @@
...
+        AND osm.calitp_extracted_at = gsd.calitp_extracted_at
+    LEFT JOIN dim_gtfs_datasets AS gd
+        ON gsd.gtfs_dataset_key = gd.key
+        AND gsd.calitp_extracted_at = gd.calitp_extracted_at
+    WHERE gsd.category = "primary"
+),
+
+dim_provider_service_gtfs AS (
+    SELECT
+        {{ farm_surrogate_key(['organization_key', 'COALESCE(service_key,"_")', 'COALESCE(gtfs_service_data_key,"_")', 'COALESCE(gtfs_dataset_key,"_")']) }} AS key,
```
atvaccaro, line 49: nit: we could probably put this coalesce in the macro itself

atvaccaro:
> New question for @atvaccaro: any concern with merging `legacy_airtable` as a source? I committed by accident (overzealous select), not sure if it's worth pulling it out of this PR given that there are still followups -- #1626.)

I think it's fine. We can remove when we do 1626.

warehouse/models/mart/transit_database/dim_service_components.sql
lauriemerrell: awesome, updated: https://github.com/cal-itp/data-infra/pull/1625/commits/411d021ffabac2eeec92992bc0785b5908c6a8c3

warehouse/models/mart/transit_database/dim_properties_and_features.sql
lauriemerrell: Nope! I added the partition-by in this PR to support the get-latest-by-date operation where you partition by date and get the latest timestamp. But because we do that operation in the staging table, we just need to get everything from the latest date in the mart tables. 

But this did make me realize that time is just a string, I will need to cast it to make sure that sorting is correect, will do that

warehouse/models/mart/transit_database/_mart_transit_database.yml
atvaccaro: nope

warehouse/models/mart/transit_database/dim_contract_attachments.sql
lauriemerrell: ✅ https://github.com/cal-itp/data-infra/pull/1625/commits/d157f6f46ccd0be1ae446e6daf46a5887da79a66

warehouse/models/mart/transit_database/dim_provider_gtfs_data.sql
lauriemerrell: ✅ https://github.com/cal-itp/data-infra/pull/1625/commits/6c22784018b1faa1608ec4d48ece969c09a451b8

PR #1639, opened 2022-07-18
Expand GTFS schedule data published to CKAN

atvaccaro:
# Description

Adds additional CKAN IDs for the GTFS Schedule data set. Also updates dictionary and rounds lat/lon to 5 digits as required by the open data portal precision guidelines.

Resolves https://github.com/cal-itp/data-infra/issues/284

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Running locally for now.


tiffanychu90 (approved):
We can leave this for now, `4m` instead of putting it into `decimal degrees units`. They might come back with that tweak, but I can't find the exact language that it needs to be in that.

https://github.com/cal-itp/data-infra/blob/704a026e85737171e5acb85158cee0a119b993e5/warehouse/scripts/publish.py#L269-L270=

Issue #1534, opened 2022-05-26
GTFS Schedule Pipeline:  gtfs_loader validation updates


lauriemerrell:
Will need a new validation DAG task that (AC):

- [ ] Ingests the raw validation data from #1529 
- [ ] Outputs JSONs that are BQ-ready

**This task may not even be necessary if the raw data is BQ-readable.**

PR #1648, opened 2022-07-21
New schedule pipeline validation job

atvaccaro:
# Description

TODO as a follow-up: deploy updated images via CI/CD

Creates a new v2 GTFS Schedule validator job with new code, Docker image, and Airflow DAG. Runs daily to process the prior day's data.

One thing to note: I created a new Airflow macro `env_var()` so we can easily pass Airflow cluster env vars through to PodOperators running on the jobs cluster in the task YAML. I'm unsure whether I like this pattern more or less than explicit CLI arguments; I do think it's helpful when relying on the Pydantic type definitions for artifacts.

Resolves https://github.com/cal-itp/data-infra/issues/1534

Does NOT resolve: validation backfilled zipped data; @lauriemerrell not sure if we want to perform that zip backfill first, or just let this align with having v2 data?

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Testing locally in Airflow.

## Screenshots (optional)
An example validation:
```
[2022-07-22 15:53:51,150] {pod_launcher.py:149} INFO - downloading gs://test-calitp-gtfs-schedule-raw/schedule/dt=2022-07-20/base64_url=aHR0cHM6Ly93d3cubWVubG9wYXJrLm9yZy9Eb2N1bWVudENlbnRlci9WaWV3LzczMjAvZ29vZ2xlX3RyYW5zaXQuemlw/ts=2022-07-20T18:26:48.741978+00:00/google_transit_202205231753434232.zip to /tmp/tmp73w1k9gk/google_transit_202205231753434232.zip
[2022-07-22 15:53:51,288] {pod_launcher.py:149} INFO - executing schedule validator: java -jar /gtfs-validator-v2.0.0_cli --input /tmp/tmp73w1k9gk/google_transit_202205231753434232.zip --output_base /tmp/tmp73w1k9gk --feed_name us-na
[2022-07-22 15:53:52,123] {pod_launcher.py:149} INFO - saving validation notice to gs://test-calitp-gtfs-schedule-processed/validation_reports/dt=2022-07-20/base64_url=aHR0cHM6Ly93d3cubWVubG9wYXJrLm9yZy9Eb2N1bWVudENlbnRlci9WaWV3LzczMjAvZ29vZ2xlX3RyYW5zaXQuemlw/ts=2022-07-20T18:26:48.741978+00:00/validation_notices.jsonl.gz
```

Task completion:
```
[2022-08-17 19:05:08,941] {pod_launcher.py:149} INFO - downloading gs://test-calitp-gtfs-schedule-raw/schedule/dt=2022-08-17/base64_url=aHR0cHM6Ly9zdG9yYWdlLmdvb2dsZWFwaXMuY29tL2d0ZnMtZGF0YS9zY2hlZHVsZS1hc3NldHMvdGN0YS0yMDIxLTEyLTEwLnppcA==/ts=2022-08-17T17:55:36.584149+00:00/tcta-2021-12-10.zip to /tmp/tmppiql5l5v/tcta-2021-12-10.zip
[2022-08-17 19:05:09,029] {pod_launcher.py:149} INFO - executing schedule validator: java -jar /gtfs-validator-v2.0.0_cli --input /tmp/tmppiql5l5v/tcta-2021-12-10.zip --output_base /tmp/tmppiql5l5v --feed_name us-na
[2022-08-17 19:05:09,695] {pod_launcher.py:149} INFO - saving 3 validation notices to gs://test-calitp-gtfs-schedule-validation/validation_reports/dt=2022-08-17/base64_url=aHR0cHM6Ly9zdG9yYWdlLmdvb2dsZWFwaXMuY29tL2d0ZnMtZGF0YS9zY2hlZHVsZS1hc3NldHMvdGN0YS0yMDIxLTEyLTEwLnppcA==/ts=2022-08-17T17:55:36.584149+00:00/validation_notices.jsonl.gz
[2022-08-17 19:05:09,900] {pod_launcher.py:149} INFO - got 366 successes and 0 failures
[2022-08-17 19:05:09,900] {pod_launcher.py:149} INFO - saving 366 to gs://test-calitp-gtfs-schedule-validation/validation_job_results/dt=2022-08-17/results.jsonl
[2022-08-17 19:05:12,279] {pod_launcher.py:198} INFO - Event: validate-gtfs-schedule.30991057e36a44518ee9fe626a2636e5 had an event of type Succeeded
[2022-08-17 19:05:12,279] {pod_launcher.py:311} INFO - Event with job id validate-gtfs-schedule.30991057e36a44518ee9fe626a2636e5 Succeeded
[2022-08-17 19:05:12,396] {pod_launcher.py:198} INFO - Event: validate-gtfs-schedule.30991057e36a44518ee9fe626a2636e5 had an event of type Succeeded
[2022-08-17 19:05:12,396] {pod_launcher.py:311} INFO - Event with job id validate-gtfs-schedule.30991057e36a44518ee9fe626a2636e5 Succeeded
[2022-08-17 19:05:12,563] {taskinstance.py:1212} INFO - Marking task as SUCCESS. dag_id=unzip_and_validate_gtfs_schedule, task_id=validate_gtfs_schedule, execution_date=20220817T020000, start_date=20220817T184425, end_date=20220817T190512
```
image


jobs/gtfs-schedule-validator/gtfs_schedule_validator.py
```diff
@@ -0,0 +1,190 @@
...
+        output_dir=output_dir,
+        jar_path=jar_path,
+        verbose=verbose,
+    )
+
+
+@app.command()
+def validate_day(
+    day: datetime = typer.Argument(
+        default=pendulum.today(),
```
lauriemerrell, line 98: Do we have to define a default for this? I guess since it gets them based on partition it should be ok, but I am wondering about weirdness with DAG execution times... Kind of would prefer that this has to be explicitly given.

lauriemerrell (changes requested):
Some of these comments refer to the calitp-py-side implementation so I guess would be addressed there, but I am putting them here since I'm reviewing the whole thing as a unit
 
1. Can we actually fill out the README? One thing we should specifically do is credit the source for the validator JAR (I see that in the old version we were just `wget`ing it every time.) 
2. Looking at the actual `results` output file, I am kind of meh on the fact that the outcomes file reproduces the entire `input_record` data again in the `extract` field. Do we need to duplicate all that? 
3. Specifically, I am wondering about the fact that we include the entire `gtfs datasets` row in the `config` field... I can kind of see why this makes sense but it means that defining a schema to ingest the outcomes files will be a pain. I'm also a little bit worried about going down the same path we have now where we have basically-identical versions of the same metadata replicated all over the place. 

For the purposes of validation, can it not just save the partition labels for the extract that it's using (like date, base64 URL, time) instead of reproducing all the metadata here?

jobs/gtfs-schedule-validator/gtfs_schedule_validator.py
atvaccaro: yeah that's fair, I was mostly facilitating using the command directly via CLI but I'm not really convinced it's a great pattern yet... maybe we should be defining the actual typer entrypoints in `__main__.py`?

atvaccaro:
Addressing here.

> 1. Can we actually fill out the README? One thing we should specifically do is credit the source for the validator JAR (I see that in the old version we were just `wget`ing it every time.)

Yes, done.


> 2. Looking at the actual `results` output file, I am kind of meh on the fact that the outcomes file reproduces the entire `input_record` data again in the `extract` field. Do we need to duplicate all that?

So I've had mixed feelings on this... I've been worried about a "chain" of metadata that could be broken if a link in the chain is deleted/updated. Therefore, I was attempting to concatenate the lineage at each step, and keep the whole lineage on each object. However it's getting unwieldy for sure, and I think object versioning and retention may actually help out here. I'm going to change the type to accept a path and verify the path exists as a file; does that seem reasonable?

> 3. Specifically, I am wondering about the fact that we include the entire `gtfs datasets` row in the `config` field... I can kind of see why this makes sense but it means that defining a schema to ingest the outcomes files will be a pain. I'm also a little bit worried about going down the same path we have now where we have basically-identical versions of the same metadata replicated all over the place.

We can always tell Pydantic to ignore a given field, or just throw a `Dict[str, Any]` at it. It's not the worst solution in the world but I think I'm going to just simplify anyways.

> For the purposes of validation, can it not just save the partition labels for the extract that it's using (like date, base64 URL, time) instead of reproducing all the metadata here?

Going to do the full file path as above, I think that will work.

jobs/gtfs-schedule-validator/gtfs_schedule_validator.py
lauriemerrell: Kind of defer to you on that side, or maybe we can have a commented version for testing? Agree it's useful to leave testing breadcrumbs for local dev but just worry about leaving them in unmarked

atvaccaro:
> However it's getting unwieldy for sure, and I think object versioning and retention may actually help out here. I'm going to change the type to accept a path and verify the path exists as a file; does that seem reasonable?

Quick update on this as discussed, going to just exclude the underlying child fields as needed.

jobs/gtfs-schedule-validator/gtfs_schedule_validator.py
```diff
@@ -0,0 +1,263 @@
...
+class GTFSScheduleFeedExtractValidationOutcome(ProcessingOutcome):
+    extract: GTFSFeedExtractInfo = Field(..., exclude={"config"})
+    validation: Optional[GTFSScheduleFeedValidation]
+
+
+# TODO: this and DownloadFeedsResult probably deserve a base class
+class ScheduleValidationResult(PartitionedGCSArtifact):
+    bucket: ClassVar[str] = SCHEDULE_VALIDATION_BUCKET
+    table: ClassVar[str] = "validation_results"
+    partition_names: ClassVar[List[str]] = ["dt"]
```
lauriemerrell, line 74: making a note here -- I think we want to use timestamp too? technically if you ran validation once and then ran it again and a download occurred in the interval, you could overwrite with a different file.

lauriemerrell: basically, raises a question about whether `outcomes` is about a validation **action** or about the **latest/current state** of the related artifacts (like the inputs & outputs). if the latter, can keep date-only partition. if the former, should use timestamp

atvaccaro: I'm going to rename this, it's the file that represents the overall work done by a _job_ so it'll be date-partitioned since that's the Airflow data interval

maybe ScheduleValidationResult?

airflow/dags/parse_and_validate_rt/parse_rt_trip_updates.yml
```diff
@@ -11,7 +11,7 @@ arguments:
   - "trip_updates"
   # we run on a delay because of archiver write latency but then we process the prior _hour_
   - "{{ get_bucket() }}/rt/{{ execution_date.replace(minute=0, second=0).format('YYYY-MM-DDTHH*') }}"
-  - "{{ prefix_bucket('rt-parsed') }}"
+  - "{{ env_var('CALITP_BUCKET__RT_PARSED') }}"
```
lauriemerrell, line 14: This in particular is making me a little nervous... It seems like in general we are using the environment variables just for pipeline v2 (= archiver v3) content that:

a) is not yet canonical
b) is clearly separate from v1 (= archiver v2)

I guess for RT are we not going to have separate v2 buckets? It's the grayest area because we already did a lot of the stuff in the original parse jobs.... But wanted to ask

lauriemerrell (commented):
Generally looks good, just two process things, one inline below.

Other process thing (already discussed offline): I don't love merging stuff into main in data-infra using unmerged prereleases of calitp-py. Would prefer that we actually merge the necessary changes in calitp and reference a full release here.

airflow/dags/parse_and_validate_rt/parse_rt_trip_updates.yml
lauriemerrell: I guess specifically I think that this value is maybe set incorrectly in Composer? 
image

lauriemerrell: ^ but this is making me wonder if we need to manually (together?) go through each location that needs the environment variable set and make sure it's set correctly

atvaccaro: Nope this is a great catch; these should not have been changed. I will revise.

README.md
```diff
@@ -26,3 +26,13 @@ Documentation for this codebase lives at [docs.calitp.org/data-infra](https://do
 This repository uses black and pre-commit hooks to format code. To install locally, run
 
 `pip install pre-commit` & `pre-commit install` in the root of the repo.
+
+## Configuration via Environment Variables
+Generally we try to configure things via environment variables. In the Kubernetes
+world, these get configured via [Kustomize overlays](./kubernetes/overlays/).
```
lauriemerrell, line 32: This link / path doesn't seem to exist (`/kubernetes/overlays/`)? 

Also, sorry for belaboring this point / for what I am presumably missing, but where does this actually get set for the v3 RT archiver? Like I don't see a Kustomize file in this PR for the archiver. Would that happen in a subsequent PR?

airflow/docker-compose.yaml
```diff
@@ -79,6 +79,15 @@ x-airflow-common:
...
     GOOGLE_CLOUD_PROJECT: cal-itp-data-infra
 
+    CALITP_BUCKET__AIRTABLE: "gs://test-calitp-airtable"
+#    CALITP_BUCKET__GTFS_RT_RAW: "gs://test-calitp-gtfs-rt-raw"
+#    CALITP_BUCKET__GTFS_RT_PARSED: "gs://test-calitp-gtfs-rt-parsed"
+#    CALITP_BUCKET__GTFS_RT_VALIDATION: "gs://test-calitp-gtfs-rt-validation"
+    CALITP_BUCKET__GTFS_SCHEDULE_RAW: "gs://test-calitp-gtfs-schedule-raw"
+    CALITP_BUCKET__GTFS_SCHEDULE_VALIDATION: "gs://test-calitp-gtfs-schedule-validation"
+
+    GRAAS_SERVER_URL: $GRAAS_SERVER_URL
```
lauriemerrell, line 89: I think with #1690 we may not need this any more? maybe we can add a TODO, don't want to scope creep here too much without understanding ramifications but I literally just merged that

lauriemerrell (commented):
Confirmed that Airflow bucket environment variables seem appropriate... To be a real stickler we would maybe remove the RT ones until they're actually needed in the new RT parsing PR but I guess that's not necessary.

image

Two small inline questions

README.md
atvaccaro: Ah yeah let me link to the actual file.

lauriemerrell (approved):

PR #1651, opened 2022-07-22
give jupyter users 64gb instead of 8

atvaccaro:
Eric has hit the limit so let's bump it


atvaccaro:
waiting to confirm volume expansion will work

vevetron:
I'm pretty sure this modified the wrong values.

PR #1655, opened 2022-07-26
disable stop times upload to ckan for now

atvaccaro:
We're hitting a max string size limit when trying to upload this directly. Need to investigate if CKAN supports incremental uploads.


atvaccaro:
@lauriemerrell  already did this!

PR #1669, opened 2022-08-04
Fix CA Open Data Columns

lauriemerrell:
# Description

## Issue
Per email from Conveyal, the [latest GTFS open dataset on the CA open data portal](https://data.ca.gov/dataset/cal-itp-gtfs-ingest-pipeline-dataset) is missing the `calitp_itp_id` and `calitp_url_number` attributes. Upon inspection we were also publishing the `calitp_hash` and `<file name>_key` fields, which are not in the data dictionary and not intended to be published. 

## Resolution

This PR:
* Updates the column metadata so that `calitp_itp_id` and `calitp_url_number` *will* be published, and `calitp_hash` and `<file name>_key` will *not*
* Updates the Pydantic dbt Manifest parsing to make the `TestMetadata` attribute optional -- when I tried to dry-run the publish workflow, it failed because some [newly-defined](https://github.com/cal-itp/data-infra/pull/1418) [singular tests](https://docs.getdbt.com/docs/building-a-dbt-project/tests#singular-tests) lacked the `test_metadata` attribute which was defined as required in the Pydantic model. Per things like [this](https://github.com/dbt-labs/dbt-docs/blob/main/src/app/services/graph.service.js#L340), it appears that singular tests are like specifically identifiable by their lack of the `test_metadata` attribute. So I make that optional here.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

* Ran `dbt run` & `dbt test` and confirmed success
* Did a dry run (`poetry run python scripts/publish.py publish-exposure california_open_data --dry-run`) of the publish workflow and confirmed that the resulting files have only the desired columns -- at time of PR writing, this is still running because it's kind of slow, but I confirmed that two files were written with the correct columns

Once this merges, I will run the actual publish workflow to fix on the portal.

### DIFF
--- a/warehouse/models/gtfs_schedule_latest_only/_gtfs_schedule_latest_only.yml
+++ b/warehouse/models/gtfs_schedule_latest_only/_gtfs_schedule_latest_only.yml
@@ -14,19 +14,21 @@ models:
         description: '{{ doc("column_calitp_itp_id") }}'
         meta:
           metabase.semantic_type: type/FK
-          publish.ignore: true
       - &calitp_url_number
         name: calitp_url_number
         description: '{{ doc("column_calitp_url_number") }}'
         meta:
           metabase.semantic_type: type/FK
-          publish.ignore: true
       - &calitp_extracted_at
         name: calitp_extracted_at
         description: '{{ doc("column_schedule_calitp_extracted_at") }}'
+        meta:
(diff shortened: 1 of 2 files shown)


charlie-costanzo (approved):
LGTM!

PR #1671, opened 2022-08-04
fix(open data): suppress publishing stop_times because of size limit

lauriemerrell:
# Description

When attempting to publish an updated version of the GTFS latest-only data, my upload failed because the stop_times data is too large for the CKAN API. We will need to reach out to the CKAN admin folks for support (already have one inquiry out to them); in the meantime I am suppressing stop_times so we can re-publish everything else with the correct columns.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested? 

See #1669; we actually successfully re-published agency and routes but then stop_times failed because of data size.


charlie-costanzo (approved):
LGTM!

PR #1676, opened 2022-08-05
fix(airflow/gtfs_loader): replace non-utf-8 characters

### LINKED ISSUE #1675: Bug: Deal with special character in Foothill Transit routes.txt
**Describe the bug**

Starting on 2022-07-26, Foothill Transit started including a special character of some sort in their `routes.txt` file:
image

This caused the `gtfs_loader.gtfs_schedule_history_load` DAG task to fail with: `UnicodeDecodeError: 'utf-8' codec can't decode byte 0x96 in position 205: invalid start byte`

As a result, our type 2 merge logic has not run since 2022-07-26:

image

**To Reproduce**
1. Try to run the `gtfs_loader.gtfs_schedule_history_load` DAG task for 7/26/22
2. Wait for it to get to the `gs://gtfs-data/schedule/2022-07-26T00:00:00+00:00/112_0/routes.txt` file
3. See failure

**Expected behavior**
We need to be able to process all characters found in GTFS files. 

**Additional context**
Because this has jammed up the `depends_on_past` `merge_updates` DAG task, it is high priority.

lauriemerrell:
# Description

Fixes #1675 by ignoring non-UTF-8 characters in our copy operation (this is the task that copies new GTFS data from one CSV location to a second CSV location with only the known subset of columns.) 

I have chosen to use the `replace` pandas error handling, which replaces the illegal character with a defined unknown.

Alternatives would be:
- Trying to find an encoding that works for the Foothill Transit data on 2022-07-26. 
- Use `ignore` instead of `replace` (see [docs](https://docs.python.org/3/library/codecs.html#error-handlers)); this would replace with an empty space instead of a placeholder character

I liked the placeholder character because it preserves the fact that some data was present. 

Resolves #1675

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

- Tested in a notebook and manually confirmed that this option works for the affected data
- Ran Airflow locally 

## Screenshots 

Notebook:

image

Airflow:

image

(Production Airflow shows that that exact line is where the issue occurs without this fix:)
image

### DIFF
--- a/airflow/plugins/utils.py
+++ b/airflow/plugins/utils.py
@@ -48,7 +48,9 @@ def _keep_columns(
     # pandas.errors.ParserError, but the responsibility to catch this error is assumed
     # to be implemented in the code that calls this method.
     try:
-        df = pd.read_csv(read_gcfs(src_path), dtype="object", **kwargs)
+        df = pd.read_csv(
+            read_gcfs(src_path), dtype="object", encoding_errors="replace", **kwargs
+        )
     except EmptyDataError:
         # in the rare case of a totally empty data file, create a DataFrame
         # with no rows, and the target columns


edasmalchi (approved):
lgtm!

atvaccaro (approved):

PR #1695, opened 2022-08-18
Deploy Elavon SFTP ingest server into production

lottspot:
Deploys changes merged in #1692 into production


lottspot:
Archiver v3 changes should be of no concern for this merge, as its deployment is not CI-enabled

atvaccaro (approved):

PR #1703, opened 2022-08-23
Terraform integration

### LINKED ISSUE #1719: Enroll GCP infrastructure in Terraform
It seems like we're finally ready to start using Terraform to manage our infrastructure as code. Specifically, Terraform will let us manage our increasing number of external tables, user/service accounts, and K8s infrastructure. In addition, it would provide a good starting point for locking down our IAM permissions and applying appropriate permissions to existing users.

See https://github.com/cal-itp/data-infra/pull/1703 as the initial PR.

lottspot:
# Description

This changeset introduces a skeleton through which Terraform resources can be tracked and managed.

Resolves #1719

### DIFF
--- a/.github/workflows/terraform-deploy.yml
+++ b/.github/workflows/terraform-deploy.yml
@@ -1,0 +1,48 @@
+name: Deploy to a terraform target
+
+on:
+  push:
+    branches:
+      - 'main'
+      - 'targets/*'
+    paths:
+      - 'iac/terraform/*'
+
+jobs:
+
+  deploy:
+
+    runs-on: ubuntu-latest
+    steps:
+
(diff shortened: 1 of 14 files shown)


lottspot:
The directory structure needs some work I think... it currently feels like it requires too much backtracking in the tree structure to work with related resources

lottspot:
The workflows seemingly cannot be tested prior to a merge into main, but all of the machinery has been individually tested, and these workflows are ported/modified from a working project. Once this branch is merged, we should be able to begin enrolling infrastructure into terraform management while any required touch ups to the GitHub workflows get ironed out.

atvaccaro (approved):
thanks for walking us through this today!

PR #1708, opened 2022-08-24
Add RT VP vs Sched Table

natam1:
# Description

Adding a table that looks at the number of scheduled trips and trips with vehicle position data. This table currently has two operators (ITP_ID 300 Big Blue Bus and ITP_ID 290 SamTrans) and compares data for two months. 

The table currently lives in rt_views but can be changed! 

To come after GTFS rewrite: additional operators and more dates - NOT urgent. 

Works towards #[263](https://github.com/cal-itp/data-analyses/issues/263) in data-analyses

*note:* column agency_name not always merging right. looking into this. 

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Tested using dbt tables in natalie_views

## Screenshots (optional)
Screen Shot 2022-08-30 at 11 20 38 AM

### DIFF
--- a/warehouse/models/rt_views/gtfs_rt_vs_sched_routes.sql
+++ b/warehouse/models/rt_views/gtfs_rt_vs_sched_routes.sql
@@ -1,0 +1,74 @@
+WITH 
+gtfs_rt_vp_distinct_trips AS (
+    SELECT DISTINCT
+     calitp_itp_id,
+     calitp_url_number,
+     date,
+     trip_id,
+     trip_route_id,
+     vehicle_id
+    FROM {{ ref('stg_rt__vehicle_positions') }}
+    WHERE calitp_itp_id = 300
+        AND date between '2022-04-01' AND '2022-06-30'
+),
+
+vp_trips AS(
+SELECT 
+  date AS service_date, 
+  trip_id AS vp_trip_id, 
+  calitp_itp_id, 
+  calitp_url_number
+FROM gtfs_rt_vp_distinct_trips
+WHERE date BETWEEN '2022-04-01' AND '2022-06-30'
+),
+
+sched_trips AS(
+SELECT 
+  trip_id, 
+  route_id,
+  service_date, 
+  calitp_itp_id, 
+  calitp_url_number
+FROM {{ref('gtfs_schedule_fact_daily_trips')}}
+WHERE (calitp_itp_id=300
+    AND service_date BETWEEN '2022-04-01' AND '2022-06-30'
+    AND is_in_service = True)
+),
+rt_sched_joined AS(
+  SELECT
+  T1.calitp_itp_id,
+  T1.calitp_url_number,
+  T1.route_id,
+  T1.service_date,
+    COUNT(T1.trip_id) AS num_sched,
+    COUNT(T2.vp_trip_id) AS num_vp,
+    -- num_vp/num_sched AS pct_w_vp
+  FROM sched_trips AS T1
+  LEFT JOIN vp_trips AS T2
+    ON
+      T1.trip_id = T2.vp_trip_id
+      AND T1.calitp_itp_id = T2.calitp_itp_id
+      AND T1.calitp_url_number = T2.calitp_url_number
+      AND T1.service_date = T2.service_date
+  GROUP BY 1, 2, 3, 4
+),
+with_percent AS(
+  SELECT
+  T1.calitp_itp_id,
+  T2.agency_name,
+  T1.calitp_url_number,
+  T1.route_id,
+  T2.route_short_name,
+  T1.service_date,
+  T1.num_sched,
+  T1.num_vp,
+  num_vp/num_sched AS pct_w_vp
+  FROM rt_sched_joined AS T1
+  LEFT JOIN {{ref('gtfs_schedule_dim_routes')}} AS T2
+    ON
+      T1.route_id = T2.route_id
+      AND T1.calitp_itp_id = T2.calitp_itp_id
+      AND T1.calitp_url_number = T2.calitp_url_number
+)
+SELECT * 
+FROM with_percent
--- a/warehouse/models/rt_views/gtfs_rt_vs_sched_routes.sql
+++ b/warehouse/models/rt_views/gtfs_rt_vs_sched_routes.sql
@@ -1,27 +1,18 @@
 WITH 
-gtfs_rt_vp_distinct_trips AS (
+-- selecting the distinct trips from GTFS Vehicle Postions for two operators (SamTrans and Big Blue Bus) and two months
+vp_trips AS (
     SELECT DISTINCT
      calitp_itp_id,
      calitp_url_number,
-     date,
-     trip_id,
+     date AS service_date,
+     trip_id AS vp_trip_id,
      trip_route_id,
      vehicle_id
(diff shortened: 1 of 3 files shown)


warehouse/models/rt_views/gtfs_rt_vs_sched_routes.sql
```diff
@@ -0,0 +1,69 @@
+WITH 
+-- selecting the distinct trips from GTFS Vehicle Postions for two operators (SamTrans 290 and Big Blue Bus 300) and two months
+vp_trips AS (
+    SELECT DISTINCT
+     calitp_itp_id,
+     calitp_url_number,
+     date AS service_date,
+     trip_id AS vp_trip_id,
+     trip_route_id,
```
lauriemerrell, line 9: Just want to note that technically both `trip_id` and `route_id` are [only conditionally required](https://gtfs.org/realtime/reference/#message-vehicleposition) in vehicle positions based on kind of complicated logic... If we were to expand this table to cover all operators, we may need to add more logic to handle cases where `trip_id` is null. (This is something that we could just note in a comment here -- for example, could add a comment linking to the docs and noting that changes may be needed when generalizing.)

Also, it doesn't look like you use `route_id` from this table later, so you could possibly drop it from the `SELECT` statement here, although if we were to implement the more complicated conditional logic later we would need it then.

warehouse/models/rt_views/gtfs_rt_vs_sched_routes.sql
```diff
@@ -0,0 +1,69 @@
...
+     vehicle_id
```
lauriemerrell, line 10: If you don't use this column, I would drop it (BigQuery uses columnar storage so one of the easiest ways to improve query performance is to only select the set of columns that you specifically need.)

warehouse/models/rt_views/gtfs_rt_vs_sched_routes.sql
```diff
@@ -0,0 +1,69 @@
...
+  T1.service_date,
+  T1.num_sched,
+  T1.num_vp,
+  num_vp/num_sched AS pct_w_vp
+  FROM rt_sched_joined AS T1
+  LEFT JOIN {{ref('gtfs_schedule_dim_routes')}} AS T2
+    ON
+      T1.route_id = T2.route_id
+      AND T1.calitp_itp_id = T2.calitp_itp_id
+      AND T1.calitp_url_number = T2.calitp_url_number
```
lauriemerrell, line 66: I think you need a date check here -- possibly something like `AND T1.service_date BETWEEN T2.calitp_extracted_at AND T2.calitp_deleted_at` -- `dim_routes` can have multiple entries per `itp_id/url_number/route_id`, based on their effective dates.

warehouse/models/rt_views/gtfs_rt_vs_sched_routes.sql
```diff
@@ -0,0 +1,69 @@
...
+    ON
+      T1.trip_id = T2.vp_trip_id
+      AND T1.calitp_itp_id = T2.calitp_itp_id
+      AND T1.calitp_url_number = T2.calitp_url_number
+      AND T1.service_date = T2.service_date
+  GROUP BY 1, 2, 3, 4
+),
+
+-- getting the percent of scheduled trips with vehicle position data and adding the common route name
+with_percent AS(
```
lauriemerrell, line 50: This is very much just a style thing with no functional impact, but the dbt convention is usually that the final CTE would be called the name of the actual table so that the final `SELECT` statement is like `SELECT * FROM <actual_table_name>`

lauriemerrell (changes requested):
Thank you for taking a stab at this task that has been hanging out for ages! Generally looks good, just a few things (length of comment is not correlated with seriousness of comment 😅 -- I just wanted to explain everything in some detail for context because I know a lot of this is new.)

### Pre-commit

I think you need to run the linter/pre-commit checks locally, because the pre-commit check [is failing](https://github.com/cal-itp/data-infra/runs/8003921187?check_suite_focus=true). 
* These are checks that we have set up to ensure consistent code formatting (things like indents, capitalization of SQL, etc.) You can follow the steps [here](https://github.com/cal-itp/data-infra#pre-commit) to install them (you may need to do `python3 pip install` if `pip install` on its own doesn't work). 
* Then you can make a commit. For example, you could change `between` to `BETWEEN` (all caps) in your SQL file (that's one of the SQL checks that seems to be failing in the linter, based on the [online run](https://github.com/cal-itp/data-infra/runs/8003921187?check_suite_focus=true)), re-add, and re-commit. 
 * The checks should run and you'll see some failures. Failures in `flake8` and `black` will usually fix themselves (they will automatically do things like adding/removing empty newlines or adding spaces around commas) -- you can just `git add <files>` after the first commit attempt fails and the linters will have fixed those kinds of issues on their own.
 * `SQLFluff` which checks actual SQL syntax will **not** fix issues on its own, so those you'll have to fix manually. They are usually pretty specific and give the line number. The most annoying are the indentation ones, but otherwise they are usually pretty clear.

If you have issues with addressing the pre-commit checks, I think that @charlie-costanzo would be the best person to help you in more detail/pair on it. And don't worry if it takes a few attempts to pass them, sometimes when `flake8` and `black` fix a first round of formatting, they then create/uncover new issues so you may have to do it (re-add and re-commit) a few times before everything is addressed.

### dbt / YAML

There are a few dbt convention / process things it might be good to do:

* I do think that `rt_views` is a good place for this to live within the dbt structure. Given that, I would update the [`_rt_views.yml`](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/rt_views/_rt_views.yml) file to document this model. @charlie-costanzo can help if you want to talk through the process. At a minimum, I think we'd want to document: 
   1. Model description -- explicitly calling out that this is only 2 agencies for 2 months
   2. Ideally test primary key uniqueness & non-null-ness
* I would maybe change the table name slightly to something like gtfs_rt_vs_schedule_trips_sample (to call out very clearly that it is a sample) -- also would name it `trips` instead of `routes` because it seems like you're comparing trips by route, rather than comparing which routes are present?

### SQL

A few inline comments about queries, definitely generally looks good.

atvaccaro:
Thanks @lauriemerrell for the detailed review! I second everything.

natam1:
Thanks @lauriemerrell for this great review! I'll work on the changes today!

warehouse/models/rt_views/gtfs_rt_vs_sched_routes.sql
natam1: Made a note of it and removed route_id for now!

warehouse/models/rt_views/gtfs_rt_vs_sched_routes.sql
natam1: done!

warehouse/models/rt_views/gtfs_rt_vs_sched_routes.sql
natam1: ah makes sense!

natam1:
Changes made and checks have passed! Thank you again @lauriemerrell for all your help!!

lauriemerrell (approved):
Thanks @natam1! 🎉

PR #1712, opened 2022-08-27
Initial dbt models to support GTFS guidelines checks

### LINKED ISSUE #1471: Deal with two-row feed_info.txt
The current Foothill Transit (ITP ID `112`) feed's `feed_info.txt` has two rows with two sets of active dates (both have the same start date) and the same feed version identifier. 
image

This causes fanout in the warehouse, because we have two copies of the same feed in `gtfs_schedule_dim_feeds`. Specifically, a test is failing on `gtfs_schedule_fact_daily_trips` because we have two copies of the same trip. 

While this is clearly a bug in their feed, I am not entirely sure how to address this in the warehouse. I am inclined to put in a hacky condition in `gtfs_schedule_dim_feeds` or `feed_info_clean` to filter out one of these rows specifically because this should be rare and putting in logic that tries to programmatically address this kind of situation is unlikely to be robust.

AC:
- [ ] Only one row in `gtfs_schedule_dim_feeds` for the Foothill Transit feed from `2022-05-03`. 
- [ ] The unique combination of columns test on `gtfs_schedule_fact_daily_trips` should pass.

### LINKED ISSUE #1688: User Story: Data structure and/or dashboard to support GTFS Guideline check-level reporting
### Summary
As a Cal-ITP team member, I want an easy way to understand transit service providers' GTFS Guideline compliance, so I can quickly understand a provider's current GTFS needs in order to provide the support they need.

My goal is to have a single dashboard that summarizes, for a given day, the result of a series of about 30 GTFS checks for a given agency. Here's a prototype.

The checks are listed [here](https://docs.google.com/spreadsheets/d/1v5nPowMi0EzQpfDd9d0ZluWnKX8JKm8E2nGMmuUvDRE/edit#gid=1120242357) I'll describe the solution that I have in mind below, and I am open to other ways of solving this.

**An option for how to solve:**
Build a new fact table, containing one row per day, per agency, per check. Since each check is very different, relying on a combination of schedule, RT,  airtable data and more, much custom logic would need to live in the table-creation process. [Here's how I think the table would look](https://docs.google.com/spreadsheets/d/1v5nPowMi0EzQpfDd9d0ZluWnKX8JKm8E2nGMmuUvDRE/edit#gid=1164332414).

Once we have this fact table, dashboard creation in Metabase or another tool will be straightforward - and could be broken out into a separate task. An elegant part of this solution is that, when a new check is added to the fact table, it can flow directly into reporting without needing to push any reporting updates.

### Acceptance Criteria

**Example:**

1. An analyst is able to reproduce [this dashboard](https://user-images.githubusercontent.com/3301353/184032340-5d2252f0-91e3-4dad-b607-e68f69a99a91.png), including for yesterday's gtfs data, relying entirely on the output of this issue.

### Notes

Considering resource constraints, and the huge amount of checks involved, I'm open to being a very active developer on this task. As one option, (assuming you agree with my solution proposed above) perhaps Jarvus could design a fact table, including a single very simple check, and leave it up to me to add the remaining checks.

# Sprint Ready Checklist

lauriemerrell:
# Description

This PR addresses #1688 by creating an initial table following [the schema](https://docs.google.com/spreadsheets/d/1v5nPowMi0EzQpfDd9d0ZluWnKX8JKm8E2nGMmuUvDRE/edit#gid=1164332414) requested, with slight modification -- `agency-id` is implemented as the combination of `calitp_itp_id`, `calitp_url_number`, and `calitp_agency_name` (for readability).

Specifically, this table implements two guideline checks:
* GTFS schedule downloaded successfully: checks whether there was a download error for a given GTFS schedule feed 
* No validation errors in last 30 days: checks for number of validation notices with "error" severity in the last 30 days -- >0, fail, if 0, pass 

I'd like both @atvaccaro and @owades to review before merging. 

TODO after merge:
* Sync the `mart_gtfs_guidelines` dataset to Metabase in `Data Marts`. 

Notes for reviewers:
* The way I have this, each individual `check` is implemented as a staging table, and then the `mart` table just unions all those staging tables together. Does that seem appropriate? Do we want each individual check as its own mart table? 
* [As I noted](https://github.com/cal-itp/data-infra/issues/1688#issuecomment-1220038182) on the initial issue, we will definitely need to refactor a bit on the rewrite. Most particularly: right now, `dim_feeds` treats a given ITP ID's schedule and RT feeds as part of one shared entity (i.e., one "feed" has a schedule URL, a trip updates URL, a vehicle positions URL, and a service alerts URL), so right now the granularity of checks is `date/check/itp_id/url_number`, which means we can assess schedule and RT on the same "index". Post-rewrite, we won't necessarily have that same fully-native concept of a shared identifier across all four feed types. We can definitely reconstruct such a concept if needed. But I do think there's a question of what is the entity that is being assessed here -- is is the `service`, the `organization`, or just the `dataset`? If it's just the `dataset`, how do we want to link the 3 RT feed types together in the future?

NOTE: Currently, the uniqueness test on this table fails because of the two-row feed_info (#1471). I would like to leave the test and just fix #1471 separately.

Resolves #1688

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

Run locally (`dbt run` and `dbt test`). Have also manually inspected the resulting table, see below for screenshot. 

## Screenshot
image

Here is an example dashboard I made (don't want to open access in Metabase too much to avoid confusion -- this is populated by my namespaced staging version of the table):

image

### DIFF
--- a/warehouse/macros/define_gtfs_guidelines.sql
+++ b/warehouse/macros/define_gtfs_guidelines.sql
@@ -1,0 +1,13 @@
+-- declare checks
+{% macro static_feed_downloaded_successfully() %}
+"Static GTFS feed downloads successfully"
+{% endmacro %}
+
+{% macro no_validation_errors_in_last_30_days() %}
+"No validation errors in last 30 days"
+{% endmacro %}
+
+-- declare features
+{% macro compliant_on_the_map() %}
+"Compliant / On the Map"
+{% endmacro %}
--- a/warehouse/macros/define_gtfs_guidelines.sql
+++ b/warehouse/macros/define_gtfs_guidelines.sql
@@ -11,3 +11,15 @@
 {% macro compliant_on_the_map() %}
 "Compliant / On the Map"
 {% endmacro %}
+
+
+-- columns
+{% macro gtfs_guidelines_columns() %}
+date,
+calitp_itp_id,
+calitp_url_number,
+calitp_agency_name,
+check,
+status,
+feature
+{% endmacro %}
--- a/warehouse/macros/define_gtfs_guidelines.sql
+++ b/warehouse/macros/define_gtfs_guidelines.sql
@@ -9,7 +9,7 @@
 
 -- declare features
 {% macro compliant_on_the_map() %}
-"Compliant / On the Map"
+"Compliance"
 {% endmacro %}
 
 
--- a/warehouse/models/mart/gtfs_guidelines/gtfs_schedule_fact_daily_guideline_checks.sql
+++ b/warehouse/models/mart/gtfs_guidelines/gtfs_schedule_fact_daily_guideline_checks.sql
@@ -1,0 +1,18 @@
+{{ config(materialized='table') }}
+
+-- start query
+WITH stg_gtfs_guidelines__schedule_downloaded_successfully AS (
+    SELECT * FROM {{ ref('stg_gtfs_guidelines__schedule_downloaded_successfully') }}
+),
+
+stg_gtfs_guidelines__no_validation_errors_in_last_30_days AS (
+    SELECT * FROM {{ ref('stg_gtfs_guidelines__no_validation_errors_in_last_30_days') }}
+),
+
+gtfs_schedule_fact_daily_guideline_checks AS (
+    SELECT * FROM stg_gtfs_guidelines__schedule_downloaded_successfully
+    UNION ALL
+    SELECT * FROM stg_gtfs_guidelines__no_validation_errors_in_last_30_days
+)
+
+SELECT * FROM gtfs_schedule_fact_daily_guideline_checks
--- a/warehouse/models/mart/gtfs_guidelines/gtfs_schedule_fact_daily_guideline_checks.sql
+++ b/warehouse/models/mart/gtfs_guidelines/gtfs_schedule_fact_daily_guideline_checks.sql
@@ -10,9 +10,11 @@ stg_gtfs_guidelines__no_validation_errors_in_last_30_days AS (
 ),
 
 gtfs_schedule_fact_daily_guideline_checks AS (
-    SELECT * FROM stg_gtfs_guidelines__schedule_downloaded_successfully
+    SELECT {{ gtfs_guidelines_columns() }}
+    FROM stg_gtfs_guidelines__schedule_downloaded_successfully
     UNION ALL
-    SELECT * FROM stg_gtfs_guidelines__no_validation_errors_in_last_30_days
+        SELECT {{ gtfs_guidelines_columns() }}
+        FROM stg_gtfs_guidelines__no_validation_errors_in_last_30_days
 )
 
 SELECT * FROM gtfs_schedule_fact_daily_guideline_checks
--- a/warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__feed_guideline_index.sql
+++ b/warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__feed_guideline_index.sql
@@ -11,7 +11,8 @@ gtfs_schedule_dim_feeds AS (
 -- list all the checks that have been implemented
 checks_implemented AS (
     SELECT {{ static_feed_downloaded_successfully() }} AS check, {{ compliant_on_the_map() }} AS feature
-    UNION ALL SELECT {{ no_validation_errors_in_last_30_days() }}, {{ compliant_on_the_map() }}
+    UNION ALL
+    SELECT {{ no_validation_errors_in_last_30_days() }}, {{ compliant_on_the_map() }}
 ),
 
 -- create an index: all feed/date/check combinations
--- a/warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__feed_guideline_index.sql
+++ b/warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__feed_guideline_index.sql
@@ -1,0 +1,33 @@
+{{ config(materialized='table') }}
+
+WITH gtfs_schedule_fact_daily_feeds AS (
+    SELECT * FROM {{ ref('gtfs_schedule_fact_daily_feeds') }}
+),
+
+gtfs_schedule_dim_feeds AS (
+    SELECT * FROM {{ ref('gtfs_schedule_dim_feeds') }}
+),
+
+-- list all the checks that have been implemented
+checks_implemented AS (
+    SELECT {{ static_feed_downloaded_successfully() }} AS check, {{ compliant_on_the_map() }} AS feature
+    UNION ALL SELECT {{ no_validation_errors_in_last_30_days() }}, {{ compliant_on_the_map() }}
+),
+
+-- create an index: all feed/date/check combinations
+stg_gtfs_guidelines__feed_check_index AS (
+    SELECT
+        t2.calitp_itp_id,
+        t2.calitp_url_number,
+        t2.calitp_agency_name,
+        t1.date,
+        t1.feed_key,
+        t3.check,
+        t3.feature
+    FROM gtfs_schedule_fact_daily_feeds AS t1
+    LEFT JOIN gtfs_schedule_dim_feeds AS t2
+        USING (feed_key)
+    CROSS JOIN checks_implemented AS t3
+)
+
+SELECT * FROM stg_gtfs_guidelines__feed_check_index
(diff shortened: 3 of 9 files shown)


warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__schedule_downloaded_successfully.sql
```diff
@@ -0,0 +1,37 @@
...
+),
+
+static_feed_downloaded_successfully_check AS (
+    SELECT
+        feed_key,
+        date,
+        CASE
+            WHEN extraction_status = "success" THEN "PASS"
+            WHEN extraction_status = "error" THEN "FAIL"
+        ELSE null
```
atvaccaro, line 17: unsure about bigquery, but a lot of databases return `null` if none of the cases match

warehouse/macros/define_gtfs_guidelines.sql
```diff
@@ -0,0 +1,25 @@
+-- declare checks
```
atvaccaro, line 1: you could use https://docs.getdbt.com/docs/building-a-dbt-project/building-models/using-variables instead of macros potentially

warehouse/models/mart/gtfs_guidelines/gtfs_schedule_fact_daily_guideline_checks.sql
```diff
@@ -0,0 +1,20 @@
...
+stg_gtfs_guidelines__no_validation_errors_in_last_30_days AS (
+    SELECT * FROM {{ ref('stg_gtfs_guidelines__no_validation_errors_in_last_30_days') }}
+),
+
+gtfs_schedule_fact_daily_guideline_checks AS (
+    SELECT {{ gtfs_guidelines_columns() }}
+    FROM stg_gtfs_guidelines__schedule_downloaded_successfully
+    UNION ALL
+        SELECT {{ gtfs_guidelines_columns() }}
+        FROM stg_gtfs_guidelines__no_validation_errors_in_last_30_days
```
atvaccaro, line 17: nit: formatting

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__feed_guideline_index.sql
```diff
@@ -0,0 +1,33 @@
...
+),
+
+gtfs_schedule_dim_feeds AS (
+    SELECT * FROM {{ ref('gtfs_schedule_dim_feeds') }}
+),
+
+-- list all the checks that have been implemented
+checks_implemented AS (
+    SELECT {{ static_feed_downloaded_successfully() }} AS check, {{ compliant_on_the_map() }} AS feature
+    UNION ALL SELECT {{ no_validation_errors_in_last_30_days() }}, {{ compliant_on_the_map() }}
```
atvaccaro, line 14: nit: formatting

also do these need to be macros/variables anyways?

atvaccaro (approved):
I wish we could do these scrolling window aggregations in Metabase but I guess we can't?

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__schedule_downloaded_successfully.sql
lauriemerrell: yes, I think that is desirable... I can add a check that it's not null? but I don't want to default in case something weird happens, I think we'd want to investigate

warehouse/macros/define_gtfs_guidelines.sql
lauriemerrell: Is there a specific reason you think that would be better? I thought that macros were more accessible (I feel like we don't want average dbt users editing `dbt_project.yml` a lot? And this table is going to have a lot of iteration.)

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__feed_guideline_index.sql
lauriemerrell: I had them as macros/variables because we will use them in every table and they're just hard-coded strings.... Want it to be able to just tweak those in one place

lauriemerrell:
> I wish we could do these scrolling window aggregations in Metabase but I guess we can't?

We've discussed a bit offline, but I think that there's basically only one "check" (the validations one used here) that uses a scrolling window. 

I also think it's appropriate to create this actual table in the warehouse though given the importance of the GTFS guidelines and having them defined in a shared, canonical way, rather than making someone implement them in Metabase. But also yeah I don't think you can do window aggregations in the Metabase UI anyway.

warehouse/macros/define_gtfs_guidelines.sql
atvaccaro: IMO I think a macro for a string value is overkill, that's about it

lauriemerrell: Discussed offline -- my summary is: All the `check` values will be used at least two places (in their actual staging check table and in the construction of the index), I wanted to make it easy to be able to reference these hard-coded values if we do want to use them other places / make it so we can update one location and be confident it will propagate

warehouse/models/mart/gtfs_guidelines/gtfs_schedule_fact_daily_guideline_checks.sql
lauriemerrell: https://github.com/cal-itp/data-infra/pull/1712/commits/6ba71b6bec8fa5a6761e110e8395f11e17d858f0

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__feed_guideline_index.sql
lauriemerrell: https://github.com/cal-itp/data-infra/pull/1712/commits/6ba71b6bec8fa5a6761e110e8395f11e17d858f0

owades (commented):
_Focusing my review on the output_

This is looking great, and is so exciting to see. A few questions:
1. Can you confirm the schedule that this will run on?
2. In the demo Metabase dash, I see a column called pass_count in the top table. I'm guessing that should be "Check count", is that correct? No need to update it.
3. Would you mind renaming the feature "Compliant/On the map" to "Compliance"? It's been a bit of a moving target as we've settled on feature names, and this one will most likely stick (and in the future I can update).
4. Adding additional checks appears intuitive - but can you advise me as I do the first few? Happy for this to be in written documentation form or just explaining in a meeting/slack.

lauriemerrell:
@owades responses:

> Can you confirm the schedule that this will run on?

This table will be updated every time the dbt project runs, which currently is scheduled 1x/day at noon UTC (set [here](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/transform_warehouse/METADATA.yml#L2), if interested).

> In the demo Metabase dash, I see a column called pass_count in the top table. I'm guessing that should be "Check count", is that correct? No need to update it.

Ope yeah sorry that's a typo on my part. But yeah that's the denominator, total # checks.

> Would you mind renaming the feature "Compliant/On the map" to "Compliance"? It's been a bit of a moving target as we've settled on feature names, and this one will most likely stick (and in the future I can update).

✅ https://github.com/cal-itp/data-infra/pull/1712/commits/2fa72e27fc97539c9e249127c98cb1807c6830ac

> Adding additional checks appears intuitive - but can you advise me as I do the first few? Happy for this to be in written documentation form or just explaining in a meeting/slack.

Yes for sure! High level steps:
1. Define the new check's name in `macros/define_gtfs_guidelines.sql`. If it's part of a new feature (not `Compliance`), also define that.
2. In `models/staging/gtfs_guidelines/stg_gtfs_guidelines__feed_guideline_index.sql`, at line 16 you would add your new check & its associated feature following the two lines above. This will add them to the date/feed/check index.
3. Create a new model in `models/staging/gtfs_guidelines/` that performs your actual check, following the examples in this PR. Hopefully they are helpful examples because they demonstrate how this can be done based on a table that's keyed on `feed_key` and another that's keyed on `calitp_itp_id + calitp_url_number`. This staging model should have the number of rows you'd get from doing `SELECT * FROM stg_gtfs_guidelines__feed_guideline_index WHERE check = "new check name"`, i.e., you will be left-joining onto that subset of rows with your check information. 
4. In `models/mart/gtfs_guidelines/fact_daily_guideline_checks.sql`, you will add your new staging table at the top (after line 10), and then add it to the union (after line 19). 

One remaining question would be how to document that final mart model & exactly which checks are implemented (so, in general, this approach would not have you manually updating the docs in the `_gtfs_guidelines.yml` file as individual checks are added), but we can iterate on that. 

Hope that makes sense high level, definitely happy to advise ad hoc in Slack/meetings as needed!

owades (approved):
Amazing, thank you!

PR #1714, opened 2022-08-29
grant access to payments_rides for non-agency users

atvaccaro:
# Description

I didn't realize that row access policies (introduced in https://github.com/cal-itp/data-infra/pull/1697) meant 0 rows were shown if a user failed to match any criteria. This PR restores full access to `payments_rides` to the appropriate users.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
This has already been deployed and I've confirmed in Metabase.

### DIFF
--- a/warehouse/models/payments_views/payments_rides.sql
+++ b/warehouse/models/payments_views/payments_rides.sql
@@ -23,8 +23,9 @@
 " {{ create_row_access_policy(
     principals = ['serviceAccount:metabase@cal-itp-data-infra.iam.gserviceaccount.com',
                   'group:cal-itp@jarv.us',
-                  'user:natalie@calitp.org',
-                  'user:eric@calitp.org',
+                  'domain:calitp.org',
+                  'user:angela@compiler.la',
+                  'user:easall@gmail.com',
                   'user:jeremyscottowades@gmail.com',
                  ]
 ) }}",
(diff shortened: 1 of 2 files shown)


charlie-costanzo (approved):
LGTM, thanks for the fix!

warehouse/models/payments_views/payments_rides.sql
```diff
@@ -18,7 +19,16 @@
...
     principals = ['serviceAccount:clean-air-payments-user@cal-itp-data-infra.iam.gserviceaccount.com']
-) }}"
+) }}",
+" {{ create_row_access_policy(
+    principals = ['serviceAccount:metabase@cal-itp-data-infra.iam.gserviceaccount.com',
+                  'group:cal-itp@jarv.us',
+                  'domain:calitp.org',
+                  'user:angela@compiler.la',
+                  'user:easall@gmail.com',
+                  'user:jeremyscottowades@gmail.com',
```
mjumbewu, line 29: You mentioned possibly creating a group for payments. Is it possible to do that at least for these non-jarvus people? As time goes on and people shift off the project, I feel like things like this can become a security headache.

atvaccaro: Unfortunately I can't; right now, BigQuery only supports users (and service accounts), groups i.e. a group of emails in the same organization, and domains. I would prefer if we could use an IAM role but that's not supported (yet).

PR #1696, opened 2022-08-18
Gtfs schedule unzip v2

lauriemerrell:
# Description

This PR partially addresses #1533 -- it takes a zipfile, extracts its constituent files, and writes those files to GCS in hive-partitioned paths. This PR does **not** cover converting the raw data to JSONL. That will be a subsequent task/PR. 

Stray notes about implementation / PR contents:
* @atvaccaro also updated the new downloader to use the latest (as of 8/25) version of `calitp`
* In general here I am saving only the **path** of the prior / subsequent artifacts and not their entire metadata

Example outputs from this new DAG task:
* [Results file](https://console.cloud.google.com/storage/browser/_details/test-calitp-gtfs-schedule-unzipped/unzipping_results/dt%3D1901-01-01/results.jsonl;tab=live_object?project=cal-itp-data-infra&supportedpurview=project)
* [stops.txt file](https://console.cloud.google.com/storage/browser/_details/test-calitp-gtfs-schedule-unzipped/stops.txt/dt%3D1901-01-01/base64_url%3DaHR0cDovL21hcmludHJhbnNpdC5vcmcvZGF0YS9nb29nbGVfdHJhbnNpdC56aXA%3D/ts%3D1901-01-01T12:43:06.265686%2B00:00/stops.txt;tab=live_object?project=cal-itp-data-infra&supportedpurview=project)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Tested by running in local Airflow & inspecting results. Logging output attached below.

<details>
<summary>Airflow logs from successful local DAG run</summary>

```
[2022-08-25 15:31:05,491] {taskinstance.py:1115} INFO - Executing <Task(PythonOperator): unzip_gtfs_schedule> on 1901-01-01T00:00:00+00:00
[2022-08-25 15:31:05,568] {taskinstance.py:1252} INFO - Exporting the following env vars:
AIRFLOW_CTX_DAG_EMAIL=laurie.m@jarv.us,andrew.v@jarv.us,jameelah.y@jarv.us
AIRFLOW_CTX_DAG_OWNER=airflow
AIRFLOW_CTX_DAG_ID=unzip_and_validate_gtfs_schedule
AIRFLOW_CTX_TASK_ID=unzip_gtfs_schedule
[...]

Issue #1536, opened 2022-05-26
GTFS Schedule Pipeline: Warehouse updates


lauriemerrell:
Epic containing tasks related to item [#6 Warehouse updates from the GTFS Schedule Pipeline Proposal - May 2022 project](https://docs.google.com/document/d/1tPPpZMgPfRNviarF2BrSrwZYtSGnPzGcT75erHo-UZ0/edit#heading=h.a0ttns92ee8o).

atvaccaro:
Going to close, there's two outstanding issues but the vast majority is done!

PR #1738, opened 2022-09-01
Gtfs loader v2 jsonification

### LINKED ISSUE #1533: GTFS Schedule Pipeline: gtfs_loader core updates
Create a new gtfs_loader DAG that can (AC):

- [x] Read in zipped GTFS data (from #1528)
- [x] Unzip contents and save as JSONL in hive partitioned paths based on:
  - [x] The date and time that the code actually runs, not the Airtable execution date
  - [x] UUID (not ITP ID & URL number)
- [x] Outputs a single outcomes file summarizing download outcomes like what was implemented in https://github.com/cal-itp/data-infra/pull/1458 for RT
- [x] This DAG should run every day on all downloaded data (no pre-assessment of whether data changed)
Will need handling for CSV parse errors.

lauriemerrell:
# Description

Completes #1533, building on the work in #1696. Specifically, reads in files from unzipped GTFS downloads, converts them `jsonl`, and saves the `jsonl` files in the same hive partitions as the input file & creates accompanying outcomes file.

This PR does *not* create external tables for the JSONL data, that will be handled in #1536.

Resolves #1533

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

Run in local Airflow for 8/12 (date that has test data available.)

## Screenshots 

_Top two (existing) tasks manually marked as failures just to speed things up_
image


airflow/dags/unzip_and_validate_gtfs_schedule/convert_to_json/METADATA.yml
```diff
@@ -0,0 +1 @@
+tooltip: "Convert known GTFS files to JSONL"
```
atvaccaro, line 1: neat!

airflow/dags/unzip_and_validate_gtfs_schedule/convert_to_json/agency.yml
```diff
@@ -0,0 +1,5 @@
+operator: operators.GtfsGcsToJsonlOperator
+
+date: "{{ execution_date }}"
```
atvaccaro, line 3: is this needed? can operators get the execution date in the task context?

airflow/plugins/operators/gtfs_csv_to_jsonl.py
```diff
@@ -0,0 +1,165 @@
...
+    result = ScheduleParseResult(filename="results.jsonl", dt=day, outcomes=outcomes)
+    result.save(fs)
+
+    assert len(files) == len(
+        result.outcomes
+    ), f"ended up with {len(outcomes)} outcomes from {len(files)} files"
+
+
+class GtfsGcsToJsonlOperator(BaseOperator):
+    def __init__(self, date, input_table, gtfs_filename, *args, **kwargs):
```
atvaccaro, line 146: ah looks like `date` isn't actually used?

airflow/plugins/operators/gtfs_csv_to_jsonl.py
```diff
@@ -0,0 +1,165 @@
...
+        self.input_table = input_table
+        self.gtfs_filename = gtfs_filename
```
atvaccaro, line 148: is there a situation where these would ever differ except for removing the file extension?

atvaccaro (changes requested):
do we want to set a job failure threshold anywhere?

lauriemerrell:
> do we want to set a job failure threshold anywhere?

What do you mean? Like if >X% of outcomes are failures, we fail the job? Open to it but not sure what we'd set the threshold at

airflow/dags/unzip_and_validate_gtfs_schedule/convert_to_json/agency.yml
lauriemerrell: ope yeah good catch, this was a vestige from a prior attempt. fixed https://github.com/cal-itp/data-infra/pull/1738/commits/7b05d2e3fa8141381c9ee04ea1dd5265d84dcb18

airflow/plugins/operators/gtfs_csv_to_jsonl.py
lauriemerrell: yeah see above, fixed https://github.com/cal-itp/data-infra/pull/1738/commits/7b05d2e3fa8141381c9ee04ea1dd5265d84dcb18

airflow/plugins/operators/gtfs_csv_to_jsonl.py
lauriemerrell: yeah, the case where something has a wrong name in the input but we want to map it onto the valid place. so like `_stops.txt` in input

lauriemerrell: I don't know that this has actually happened, but it's something that this architecture specifically allows for

atvaccaro: Ok cool, I think providing a default could allow for less boilerplate but seems fine

atvaccaro:
> 

Yeah really it's to catch situations where a bug broadly impacted the job and/or we think that the overall success rate is oddly low and worth "failing" the job. It can occur after the results upload so it's primarily something to bubble up through alerting. The downloader is 95%, maybe this could be similar?

atvaccaro (approved):

airflow/plugins/operators/gtfs_csv_to_jsonl.py
lauriemerrell: yeah fair, https://github.com/cal-itp/data-infra/pull/1738/commits/66eda24220ed159db29a4b9892b86aadc5b21984

PR #1739, opened 2022-09-01
Add GTFS guideline check for wheelchair fields in trips.txt & stops.txt

owades:
# Description

Adds a new check to the [new gtfs guidelines check framework](https://github.com/cal-itp/data-infra/pull/1712), which has a result of PASS when:
1) All rows of trips.txt have a non-null value in the 'wheelchair_accessible" column
2) All rows of stops.txt have a non-null value in the 'wheelchair_boarding" column
If either are not met, it has a result of FAIL

Resolves # (No issue but I can create one if thats better for process/tracking)

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
I've manually verified the result of 6 agencies, including inspecting the count totals of the _feed_wheelchair_check_ table

Screen Shot 2022-09-02 at 10 58 31 AM


owades:
@lauriemerrell, I'd appreciate a detailed review when you have a chance! My main concerns are:
1. My use of _gtfs_schedule_index_feed_trip_stops_. Testing leads me to believe I'm using it correctly, but perhaps there exists a table I don't need to dedupe. I was looking for some kind of _dim feed trips_ table, but all I could find was something that was daily, which I don't think I want.
2. The big join at the bottom, to feed_guideline_index. Am I using extracted_at and deleted_at appropriately?

Also interested in style and best practice guidance. I also could perform much more rigorous testing than I've done so far.

lauriemerrell (commented):
Naming / formatting all look 💯 

Re: the actual query, I am adding this as a case study for our model storming agenda because this is exactly what we want to make easier. 

I think it might be conceptually more straightforward to do something like the following:

```
-- summarize to date range level by feed just using the dimensional table
summarize_trips AS (
   SELECT 
       calitp_itp_id,
       calitp_url_number,
       calitp_extracted_at,
       calitp_deleted_at,
       COUNTIF(wheelchair_accessible IS NOT NULL) AS ct_accessible_trips,
       COUNT(*) as ct_trips
    FROM dim_trips
    GROUP BY 1, 2, 3, 4),
-- now join that ranged summary to the actual daily index
trips_by_day AS (
  SELECT
    i.date
    i.calitp_itp_id,
    i.calitp_url_number,
    i.feed_key,
    i.check,
    i.feature,
    SUM(t.ct_accessible_trips) AS tot_accessible_trips,
    SUM(t.ct_trips) AS tot_trips
  FROM feed_guideline_index AS i
  LEFT JOIN summarize_trips AS t
      ON i.date >= t.calitp_extracted_at
      AND i.date < t.calitp_deleted_at
      AND i.calitp_itp_id = t.calitp_itp_id
      AND i.calitp_url_number = t.calitp_url_number
  GROUP BY
    i.date
    i.calitp_itp_id,
    i.calitp_url_number,
    i.feed_key,
    i.check,
    i.feature,
) 

-- then you can do the CASE WHEN for PASS/FAIL on whether tot_accessible_trips == tot_trips
```

Basically, I don't think that you have to use `gtfs_schedule_index_feed_trip_stops` -- `dim_trips` and `dim_stops` can be used to reconstruct a daily index if needed.

**However**, I think that these ranged joins (`i.date >= t.calitp_extracted_at`) can be more computationally expensive. Based on some testing for this specifically I don't think it's too bad because it looks like the first group by gets `dim_trips` down to a manageable size.

If you have time/interest and want to try the above and compare it with what you have here, I'd want to check both logical parity and execution comparison (if you run in the BQ UI, it will tell you information on the `execution details` tab about resource consumption). 

If you don't have time, I can try to run such a check, honestly partially for my own learning because I haven't experimented as much with this as I should.

owades:
Thanks for this @lauriemerrell. Your suggestion makes a lot of sense - I was using dim_feeds as a helpful abstraction, but I see that what I need is available in dim_stops and dim_trips. 

I ran a comparison and you are correct about resource consumption.
Original way:
```
Duration
 6 sec 
Bytes processed
 5.86 GB 
Bytes billed
 5.86 GB 
```
Your suggestion:
```
Duration
 2 sec 
Bytes processed
 240.51 MB 
Bytes billed
 241 MB 
```

Regarding logical parity - at first glance they are close but not identical. I need to come up with a better understanding of why that is next week.

owades:
@lauriemerrell I've been reviewing some more, and looking at instances in the output where value changes between PASS and FAIL. The next step for reviewing would be to manually look over the historic GTFS files to check that they line up with reality (for example SF Bay Ferry URL 1 should change from FAIL to PASS on 2022-07-12). I'm on the fence about whether this review is necessary (and not sure how to access the info without Transit.land membership), what do you think?

lauriemerrell:
@owades we can definitely check this in our data in GCS! I just checked SF Bay Ferry and confirmed that on 7/11/22, they did *not* have `wheelchair_accessible` set in `trips.txt` and on 7/12/22 they did! So that one seems good. 

Re: whether it's worth it in general, I think if you're comfortable checking this just in the warehouse vs. checking files manually, that would be one way forward? 

You can kind of confirm in the warehouse by comparing the results of this:

```
SELECT distinct trip_id, wheelchair_accessible
FROM `cal-itp-data-infra.views.gtfs_schedule_dim_trips`
WHERE calitp_itp_id = 280 and calitp_url_number = 1 and calitp_extracted_at = '2022-07-12'
```

(that's the new data on 7/12/22) 

with:

```
SELECT distinct trip_id, wheelchair_accessible
FROM `cal-itp-data-infra.views.gtfs_schedule_dim_trips`
WHERE calitp_itp_id = 280 and calitp_url_number = 1 and calitp_deleted_at = '2022-07-12'
```

(data that was deleted to make way for the new data)

And it confirms what I saw in the raw files

lauriemerrell (approved):

PR #1748, opened 2022-09-06
Add GTFS guideline check for feed_contact_email field in feed_info.txt

owades:
Adds a new check to the https://github.com/cal-itp/data-infra/pull/1712, which has a result of PASS when there is a non-empty value in the feed_contact_email field in the feed_info.txt file. When this is not the case, the check has a result of FAIL.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

Spot-check 5 agencies from results, ensuring that their feed_info.txt matches what's reported in the staging table.

## Screenshots (optional)
Screen Shot 2022-09-06 at 12 08 41 PM


owades:
@lauriemerrell can you take a look at this one as well, when you have a chance? Since this contains changes from #1739 I'll need to do some merge action before this is 100% ready. Can tag you for review at that point if you prefer.

owades:
This was too much noise, sorry reviewers. All set for review now.

lauriemerrell (approved):
Looks good to me! Two FYIs (no action needed but just for context, in case they help with subsequent checks):

* #1471 is even more relevant here, probably won't be fixed until rewrite, basically Foothill Transit had a bad `feed_info.txt` file with 2 rows
* In the current warehouse, `gtfs_schedule_dim_feeds` is a weird combination of metadata that we generate within the pipeline and information from the actual `feed_info.txt` files. Using `feed_info_clean` is fine (and probably cleaner), but I just wanted to note that the relevant "views" table would theoretically be `dim_feeds`. In the rewrite, I want to clean this up so that `feed_info` makes it into a mart in its own right and isn't comingled with metadata.

owades:
Thanks @lauriemerrell . I didn't think to check dim_feeds for this column - I'm happy going with feed_info_clean for now but also completely happy to switch over to dim_feeds if you recommend. Let me know.

lauriemerrell:
@owades No totally cool to stick with `feed_info_clean`, more just wanted to give context / explain why there isn't a `dim_feed_info` at this point!

PR #1751, opened 2022-09-06
shapes GTFS checks

owades:
# Description

Add two new GTFS Guideline Checks

1. Check that shapes.txt exists for the given day. PASS if so, FAIL if not
2. Check that there are no shape-related notices in the GTFS validator. PASS if none, FAIL if > 0
"Shape related notices":

- decreasing_shape_distance
- equal_shape_distance_diff_coordinates
- equal_shape_distance_same_coordinates
- stops_match_shape_out_of_order
- stop_too_far_from_shape
- stop_too_far_from_shape_using_user_distance
- stop_too_far_from_trip_shape
- decreasing_or_equal_shape_distance

Note that stop_too_far_from_trip_shape and decreasing_or_equal_shape_distance are the only two notices that have appeared for cal-itp feeds to this date.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
**Shape.txt is present check:**
- Confirmed that on days when the check fails, the shapes file appears in gtfs_schedule_fact_daily_feed_files with is_interpolated = true

**Shapes validation check:**
Confirm there are PASS and FAIL values for check, and no null values.


warehouse/macros/define_gtfs_guidelines.sql
```diff
@@ -11,6 +11,14 @@
 "Includes complete wheelchair accessibility data in both stops.txt and trips.txt"
 {% endmacro %}
 
+{% macro shapes_file_present() %}
+"Shapes.txt file is present"
+{% endmacro %}
+
+{% macro shapes_valid() %}
+"No shapes-related errors appear in the MobilityData GTFS Validator."
```
lauriemerrell, line 19: Looks like we haven't been using periods in general, may drop the one here for consistency

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__shapes_valid.sql
```diff
@@ -0,0 +1,49 @@
+WITH feed_guideline_index AS (
+    SELECT * FROM {{ ref('stg_gtfs_guidelines__feed_guideline_index') }}
+    WHERE check = {{ shapes_valid() }}
+),
+
+-- For this check we are only looking for errors and warnings related to shapes
+validation_fact_daily_feed_codes_shape_related AS (
+    SELECT * FROM {{ ref('validation_fact_daily_feed_codes') }}
+    WHERE code IN (
```
lauriemerrell, line 9: optional suggestion: This `where` clause could just be added to the next CTE to avoid having a whole separate CTE. Totally optional / style question

lauriemerrell (approved):
two very minor inline / style type things, up to you on whether to address

PR #1754, opened 2022-09-07
Airtable extracted ts

atvaccaro:
# Description

Resolves # [issue]

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## Screenshots (optional)
```
successfully fetched 208 of 210
Failures:
 404 Client Error: Not Found for url: https://urb.mecatran.com/urb/ws/feed/c2l0ZT1zbG90cmFuc2l0O2NsaWVudD1zZWxmO2V4cGlyZT07dHlwZT1ndGZzO2tleT0zZTMwMzM1OTRiMTE2NzA0N2IxNjQwNjA0ZjQwMGMzMzdiM2E1MTQ0
404 Client Error: Not Found for url: http://data.trilliumtransit.com/gtfs/taft-co-us/taft-ca-us.zip
Skipping since in development mode! Would have emailed 2 failures.
```


lauriemerrell:
Note: Should figure out what if anything we're doing about filepaths in unzip & parse steps.... I guess we can impute from the zipfile_files and zipfile_dirs attributes?

airflow/dags/download_gtfs_schedule_v2/download_schedule_feeds.py
```diff
@@ -17,23 +18,24 @@
...
 )
 from pydantic import validator
 
 GTFS_FEED_LIST_ERROR_THRESHOLD = 0.95
 
 
-class AirtableGTFSDataRecordProcessingOutcome(ProcessingOutcome):
-    airtable_record: AirtableGTFSDataRecord
+class GTFSDownloadOutcome(ProcessingOutcome):
+    config: GTFSDownloadConfig
```
lauriemerrell, line 38: I think this is what confuses me... If we are assuming that the file will have a valid config metadata, how is that different than actually backfilling valid config files? Like, when we backfill the actual schedule data (from the old downloader), how will we construct this config?

atvaccaro: We discussed this today; the tl;dr is that the config _extracts_ are a different thing than config _records_ as metadata on downloaded files. We plan to "backfill" the metadata but not the extracts themselves.

airflow/dags/airtable_loader_v2/generate_gtfs_download_configs.py
```diff
@@ -0,0 +1,109 @@
...
+    jsonl_content = gzip.compress("\n".join(record.json() for record in valid).encode())
+    GTFSDownloadConfigExtract(
+        filename="configs.jsonl.gz",
+        ts=extract.ts,
+        records=valid,
+    ).save_content(content=jsonl_content, fs=fs)
+
+
+if __name__ == "__main__":
+    convert_gtfs_datasets_to_download_configs(None, None)
```
lauriemerrell, line 109: how does this work if you can't `xcom_pull` to get the extract? (I assume this is present for testing, but like does it actually work?)

airflow/docker-compose.yaml
```diff
@@ -81,10 +81,11 @@ x-airflow-common:
 
     CALITP_BUCKET__AIRTABLE: "gs://test-calitp-airtable"
     CALITP_BUCKET__DBT_ARTIFACTS: "gs://test-calitp-dbt-artifacts"
+    CALITP_BUCKET__GTFS_CONFIG: "gs://test-calitp-gtfs-download-config"
     CALITP_BUCKET__GTFS_RT_RAW: "gs://test-calitp-gtfs-rt-raw"
```
lauriemerrell, line 85: should this also be updated?

airflow/plugins/operators/airtable_to_gcs.py
```diff
@@ -71,7 +71,7 @@ def fetch_from_airtable(self, api_key):
             f"Downloading airtable data for {self.air_base_name}.{self.air_table_name}"
         )
         all_rows = Table(api_key, self.air_base_id, self.air_table_name).all()
-        self.extract_time = pendulum.now("utc")
+        self.extract_time = pendulum.now()
```
lauriemerrell, line 74: wait so can we clarify the impact of this -- this is going to have `Z` timestamps? why are we doing this / why are we doing it here?

airflow/plugins/operators/gtfs_csv_to_jsonl.py
```diff
@@ -84,13 +96,17 @@ def parse_individual_file(
...
+                for line in lines
+            ).encode()
         )
 
         jsonl_file = GTFSScheduleFeedJSONL(
             ts=input_file.ts,
-            base64_url=input_file.base64_url,
+            extract_config=input_file.extract_config,
             filename=gtfs_filename + ".jsonl.gz",
             input_file_path=input_file.path,
```
lauriemerrell, line 111: do we want to drop this? I thought we don't want to record paths any more

airflow/plugins/operators/gtfs_csv_to_jsonl.py
```diff
@@ -25,7 +29,7 @@ class GTFSScheduleFeedJSONL(PartitionedGCSArtifact):
     bucket: ClassVar[str] = SCHEDULE_PARSED_BUCKET
     partition_names: ClassVar[List[str]] = GTFSScheduleFeedFile.partition_names
     ts: pendulum.DateTime
-    base64_url: str
+    extract_config: GTFSDownloadConfig
     input_file_path: str
```
lauriemerrell, line 33: do we want to drop this? I thought we don't want to record paths any more

airflow/requirements.txt
```diff
@@ -1,4 +1,4 @@
-calitp==2022.9.8
+calitp==2022.9.12a0
```
lauriemerrell, line 1: can we finalize https://github.com/cal-itp/calitp-py/pull/81, merge it, and do a proper full release to use?

jobs/gtfs-rt-parser-v2/pyproject.toml
```diff
@@ -6,7 +6,7 @@ authors = ["Andrew Vaccaro <atvaccaro@gmail.com>"]
 
 [tool.poetry.dependencies]
 python = ">=3.8,<3.10"
-calitp = "2022.9.1a3"
+calitp = "2022.9.12a0"
```
lauriemerrell, line 9: same q here, can we use a full release of calitp

jobs/gtfs-schedule-validator/gtfs_schedule_validator.py
```diff
@@ -186,90 +265,77 @@ def validate_day(
...
+        del pbar
+
     result = ScheduleValidationJobResult(
         filename="results.jsonl",
         dt=day,
         outcomes=outcomes,
     )
     typer.secho(
-        f"got {len(result.successes)} successes and {len(result.failures)} failures",
+        f"got {len(result.successes)} successes and {len(exceptions)} failures",
```
lauriemerrell, line 323: why use exceptions rather than `result.failures` here?

jobs/gtfs-schedule-validator/pyproject.toml
```diff
@@ -14,7 +14,7 @@ backoff = "^2.1.2"
 pydantic = "^1.9.1"
 pendulum = "^2.1.2"
 tqdm = "^4.64.0"
-calitp = "2022.8.25"
+calitp = "2022.9.12a0"
```
lauriemerrell, line 17: same q about calitp version

lauriemerrell (commented):
mostly nitpicks, most substantial thing is just how we want to handle the calitp changes. I would like us to do a proper release of calitp and merge the `GTFSDownloadConfig` before we merge this, if possible, unless we have a very specific reason not to (which we should document in #1774)

airflow/dags/airtable_loader_v2/generate_gtfs_download_configs.py
atvaccaro: oh I commented out the line and hard-coded 😂

airflow/dags/airtable_loader_v2/generate_gtfs_download_configs.py
```diff
@@ -0,0 +1,109 @@
...
+        if schedule_record_ids and len(schedule_record_ids) == 1:
+            first = schedule_record_ids[0]
+            if first in valid:
+                config.schedule_url_for_validation = valid[first][1].url
+
+    return list(valid[1] for valid in valid.values()), invalid
+
+
+def convert_gtfs_datasets_to_download_configs(task_instance, execution_date, **kwargs):
+    extract_path = task_instance.xcom_pull(task_ids="california_transit_gtfs_datasets")
```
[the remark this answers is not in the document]
atvaccaro: nope gusty handles it as long as the upstream task function returns some value

airflow/docker-compose.yaml
atvaccaro: the RT raw?

airflow/plugins/operators/airtable_to_gcs.py
atvaccaro: This _removes_ the explicit UTC timestamp i.e. no more Z but +00:00. This is the only place with an explicit Z and the problem is that pendulum doesn't handle the UTC properly anyways.

airflow/docker-compose.yaml
lauriemerrell: I thought we were moving to a v2 bucket even with just the metadata updates to make the backfill easier

airflow/plugins/operators/airtable_to_gcs.py
lauriemerrell: is this going to mess us up later for type 2 tables then (like, do Airtable external tables have wrong times)? just want to understand impact

atvaccaro: That's a good point, we should test how bigquery parses the timestamps. At worst we can fix the keys, but I think it should be fine. They're all UTC it just depends on whether the serialization used Z vs a 0 offset (I'm fuzzy on the difference, to be honest.)

airflow/docker-compose.yaml
atvaccaro: Ah yes I just wasn't sure what line you were referring to.

lauriemerrell: ope yeah sorry I see now it's not the right one

airflow/plugins/operators/airtable_to_gcs.py
lauriemerrell: I guess I meant more "pendulum doesn't handle the UTC properly anyways" -- do we think things are substantively wrong...?

jobs/gtfs-schedule-validator/gtfs_schedule_validator.py
atvaccaro: no real reason, they should be the same; I'll add an assert for that

lauriemerrell: mostly I've always kinda been 🤷 about the utility of the `results.<type>` methods and if they're there I feel like we should use them....?

airflow/dags/parse_and_validate_rt_v2/METADATA.yml
```diff
@@ -5,7 +5,7 @@ tags:
 default_args:
     owner: airflow
     depends_on_past: False
-    start_date: "2021-07-09"
+    start_date: "2022-09-14"
```
lauriemerrell, line 8: should these now be 9/15?

airflow/dags/unzip_and_validate_gtfs_schedule/METADATA.yml
```diff
@@ -5,7 +5,7 @@ tags:
 default_args:
     owner: airflow
     depends_on_past: False
-    start_date: "2021-07-09"
+    start_date: "2022-09-14"
```
lauriemerrell, line 8: same q, does 9/14 vs. 9/15 matter?

lauriemerrell (commented):
Sorry that I didn't remember this earlier, but there was a specific request to be notified by email about things that are failing the Airtable validation step... I would be ok deferring to a follow-up PR if you are willing to do that as soon as #1786 is merged. 

I think the scope would be: Base the email on line 91 in `generate_gtfs_download_configs` but say more specifically like `X records were excluded because they don't have the pipeline checkbox checked: <list the ones that didn't have checkbox checked>` and `Y failed validation: <list the ones that failed validation>`. I think we don't want to have to email people for any task in this DAG failing, so I wonder if we want an `email_failures` like what we have now in the downloader.

If you want me to move that ^ to a ticket I can, defer to you on whether it's doing here or not.

atvaccaro:
> I think the scope would be: Base the email on line 91 in `generate_gtfs_download_configs` but say more specifically like `X records were excluded because they don't have the pipeline checkbox checked: <list the ones that didn't have checkbox checked>` and `Y failed validation: <list the ones that failed validation>`. I think we don't want to have to email people for any task in this DAG failing, so I wonder if we want an `email_failures` like what we have now in the downloader.

Yes let's do this on a follow-up PR. Right now the task will fail if under 95% succeed which is probably OK as an alert mechanism for now.

airflow/plugins/operators/airtable_to_gcs.py
atvaccaro: No I think this is just a serialization issue. `datetime` picks up that they are UTC just fine.

lauriemerrell (approved):

PR #1758, opened 2022-09-07
remove rolling 30d from validator check

owades:
# Description

I learned today that the definition of this requirement has changed. Previously draft v3 guidelines said something like "No GTFS validator errors in the past month", which Laurie used (based my docs created by me) to create the original check. About a month ago, the guideline changed to "No recent GTFS validator errors." I discussed with E and we're comfortable just using the daily validator results.

I also took this opportunity to rename the macro "compliant_on_the_map()" to "compliance"

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Light testing, I went through the results and made sure we were getting true and false values, spot-checked the error counts.

## Screenshots (optional)
Screen Shot 2022-09-07 at 11 14 54 AM


lauriemerrell (approved):
LGTM!

PR #1766, opened 2022-09-08
5 new draft checks for rt feed

owades:
# Description

5 new checks:
**no_rt_critical_validation_errors**: If the feed has >0 RT files on the given day AND no critical errors in the RT validator, then PASS. If either condition isn't met, FAIL 
**trip_id_alignment**: If the feed has >0 RT files on the given day AND didn't not have an error in the validator related to mismatched trip_id's, then PASS. If either condition isn't met, FAIL. The specific conditions for the validator check are: "All trip_ids provided in the GTFS-rt feed must exist in the GTFS data, unless the schedule_relationship is ADDED"
**vehicle_positions_feed_present**, **trip_updates_feed_present**, & **service_alerts_feed_present**: If the given feed is present at least once on the day, PASS. else, FAIL

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
no_rt_critical_validation_errors:
manually cross-check output with values from gtfs_rt_fact_daily_validation_errors and gtfs_rt_fact_files_wide_hourly, for 5 rows of results.

trip_id_alignment:
manually cross-check output with values from gtfs_rt_fact_daily_validation_errors and gtfs_rt_fact_files_wide_hourly, for 5 rows of results.

vehicle_positions_feed_present:
manually cross-check with gtfs_rt_fact_files_wide_hourly for 5 rows

trip_updates_feed_present:
manually cross-check with gtfs_rt_fact_files_wide_hourly for 5 rows

service_alerts_feed_present:
manually cross-check with gtfs_rt_fact_files_wide_hourly for 5 rows

### DIFF
--- a/warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__no_rt_critical_validation_errors.sql
+++ b/warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__no_rt_critical_validation_errors.sql
@@ -3,7 +3,7 @@ WITH feed_guideline_index AS (
     WHERE check = {{ no_rt_critical_validation_errors() }}
 ),
 
--- It's called "hourly" but if you use the "file_count_day" it's a daily table
+-- gtfs_rt_fact_files_wide_hourly has one row per day per ID+URL
 gtfs_rt_fact_files_wide_daily AS (
 SELECT * FROM {{ ref('gtfs_rt_fact_files_wide_hourly') }}
 ),
--- a/warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__no_rt_critical_validation_errors.sql
+++ b/warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__no_rt_critical_validation_errors.sql
@@ -1,0 +1,66 @@
+WITH feed_guideline_index AS (
+    SELECT * FROM {{ ref('stg_gtfs_guidelines__feed_guideline_index') }}
+    WHERE check = {{ no_rt_critical_validation_errors() }}
+),
+
+-- It's called "hourly" but if you use the "file_count_day" it's a daily table
+gtfs_rt_fact_files_wide_daily AS (
+SELECT * FROM {{ ref('gtfs_rt_fact_files_wide_hourly') }}
+),
+
+gtfs_rt_fact_daily_validation_errors AS (
+    SELECT * FROM {{ ref('gtfs_rt_fact_daily_validation_errors') }}
+),
+
+gtfs_rt_validation_code_descriptions AS (
+    SELECT * FROM {{ ref('gtfs_rt_validation_code_descriptions') }}
+),
+
+rt_files_by_day AS (
+    SELECT
+        calitp_itp_id,
+        calitp_url_number,
+        date_extracted AS date,
+        COUNT(*) AS rt_files
+    FROM gtfs_rt_fact_files_wide_daily
+    GROUP BY 1,2,3
+),
+
+errors_by_day AS (
+    SELECT
+        t1.feed_key,
+        t1.date,
+        SUM(t1.occurrences) AS errors
+    FROM gtfs_rt_fact_daily_validation_errors AS t1
+    LEFT JOIN gtfs_rt_validation_code_descriptions AS t2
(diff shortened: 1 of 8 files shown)


warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__no_rt_critical_validation_errors.sql
```diff
@@ -0,0 +1,67 @@
...
+gtfs_rt_validation_code_descriptions AS (
+    SELECT * FROM {{ ref('gtfs_rt_validation_code_descriptions') }}
+),
+
+rt_files_by_day AS (
+    SELECT
+        calitp_itp_id,
+        calitp_url_number,
+        date_extracted AS date,
+        COUNT(*) AS rt_files
```
lauriemerrell, line 24: Just to confirm, I think that this is just going to return the count of the rows, so it will either be 1, 2, or 3. If you want the actual count of files, I think you'd want to do `SUM(file_count_day)`. For these purposes it works out to the same thing because a row will only be present if there was at least one file, but just wanted to confirm (if we were to ever change the implementation of this table to have a row per URL even if there was no data, that could change.)

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__trip_id_alignment.sql
```diff
@@ -0,0 +1,63 @@
+WITH feed_guideline_index AS (
+    SELECT * FROM {{ ref('stg_gtfs_guidelines__feed_guideline_index') }}
+    WHERE check = {{ no_rt_critical_validation_errors() }}
```
lauriemerrell, line 3: think this is the wrong `check`

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__no_rt_critical_validation_errors.sql
```diff
@@ -0,0 +1,67 @@
+WITH feed_guideline_index AS (
+    SELECT * FROM {{ ref('stg_gtfs_guidelines__feed_guideline_index') }}
+    WHERE check = {{ no_rt_critical_validation_errors() }}
+),
+
+-- gtfs_rt_fact_files_wide_hourly has one row per day per ID+URL
```
lauriemerrell, line 6: I see this comment repeated -- I think it may be worth updating to say `one row per day per ID+URL+feed type` (or clarifying what is meant by URL, normally I would just interpret as `calitp_url_number`, in which case this is not quite accurate)

lauriemerrell (changes requested):
A few small inline questions / suggestions, only `trip_id_alignment` definitely needs to be updated (I think it's currently incorrect, presumably just a missed copy/paste.)

This was a surprising (but not incorrect) way to see the wide daily files table used... I think we may want to make a cleaner/more dedicated table for you to use for these in v2.

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__no_rt_critical_validation_errors.sql
owades: Yep, this is my understanding - thinking that they get credit if they have at least one RT file present on a given day (I will also review the business logic with my peers). My thinking is that, since we will also score them on:
a) presence of each file
b) updates every 20sec for vehicle positions & trip updates
...it's OK to be lax on this check wrt what it means for an RT file to be present

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__trip_id_alignment.sql
owades: gah! thank you

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__no_rt_critical_validation_errors.sql
owades: this makes more sense, thanks!

lauriemerrell (approved):

Issue #1064, opened 2022-02-10
Bug: day transitions result in not-ready-yet data in GTFS Guidelines metabase dashboard


evansiroky:
**Describe the bug**

The GTFS Guidelines metabase dashboard has a date filter that points to yesterday. However, when the time changes to a new day this transition occurs at UTC time (which is 4pm during Pacific Standard Time, or 5pm during Pacific Daylight time) the date filter points to yesterday in UTC time. Since this is also when the daily DAGs begin to run, a lot of the data will not be ready yet and a lot of stuff may show up as blank.

**To Reproduce**

1. At 4:05pm Pacific Standard Time, go to https://dashboards.calitp.org/dashboard/26
2. Note blank entries for AC Transit URLs

**Expected behavior**

All of the metabase dashboard data should reflect the most recently available data and there shouldn't be any "downtime" of data.

**Additional context**

Maybe a way to resolve this is to set the internal timezone of the server that metabase is running on to be the America/Los_angeles timezone?

evansiroky:
That's an interesting note, @lauriemerrell. I think that solution works well for dashboards that are only focused on the most-up-to-date data, but may not always work for other ones where users want to select a particular date to look at.

I have gone ahead and changed the timezone of metabase, so we'll see if that ends up working out.

evansiroky:
Changing the metabase timezone did not resolve this.

holly-g:
Checked in with Evan (10/3) -- we can close this ticket.

PR #1775, opened 2022-09-12
Omit current date from fact_daily_guideline_checks

owades:
# Description

I observed that after 12am UTC (4pm pacific), guideline check dashboard in metabase shows skewed results, as it shows incomplete data from the current day.

This PR addresses this by preventing feed_guideline_index from ever including the current date.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

I ran this change on today's data, visible in `cal-itp-data-infra-staging.owades_mart_gtfs_guidelines.fact_daily_guideline_checks`. Verified that the most recent data is yesterday, while the main table contains data from today.

### DIFF
--- a/warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__feed_guideline_index.sql
+++ b/warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__feed_guideline_index.sql
@@ -24,7 +24,6 @@ checks_implemented AS (
 ),
 
 -- create an index: all feed/date/check combinations
--- we never want results from the current date, as data will be incomplete
 stg_gtfs_guidelines__feed_check_index AS (
     SELECT
         t2.calitp_itp_id,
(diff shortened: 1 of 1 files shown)


owades:
@lauriemerrell I'm interested in whether you'd suggest making this change upstream or downstream from here - thanks for reviewing!

lauriemerrell (approved):
> @lauriemerrell I'm interested in whether you'd suggest making this change upstream or downstream from here - thanks for reviewing!

Yeah, this is kind of a tricky one. I talked about it a bit with @atvaccaro... I think this is fine, and longer term we may want to actually look into the Metabase side of this too (already exists as #1064). 

The main concern here would just be that theoretically this solution may only work as intended because of quirks in the scheduled execution times of the various jobs, but I think for this use case it seems appropriate, and we probably should put more thought into how we handle this in the underlying `fact daily feeds` and `dim feeds`.

Defining this at the index seems good because it's a single point of intervention (i.e., not doing something different for each individual check) and I think it signals "this shouldn't even be evaluated" as opposed to truncating in the mart which could be interpreted as "we've evaluated this and we're just not displaying it". 

TLDR: LGTM.

owades:
Thank you!

PR #1784, opened 2022-09-13
add pathways validator check

owades:
# Description

New GTFS check:
If any of the following pathway/location related validation notices are present in the GTFS schedule validator, then FAIL the check. If not, then PASS.

Notices:
- pathway_to_platform_with_boarding_areas
- pathway_to_wrong_location_type
- pathway_unreachable_location
- pathway_dangling_generic_node
- pathway_loop
- missing_level_id
- platform_without_parent_station
- station_with_parent_station
- wrong_parent_location_type

## Type of change

- [X] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
I confirmed that 2 of these notices exist in the schedule validator table, and that the output of each subquery looks as expected. Aside from the list of eligible notices, this is exactly the same as the `shapes_valid` check.


lauriemerrell (approved):
Implementation looks fine, but I just had a thought that I should have had for the shapes check: do you want to check that the given feed actually had a pathways.txt file and, if it doesn't have the file, do we want to fail it? 

I think it didn't occur to me for shapes.txt because that file is more common, but I think a lot of our feeds don't even have pathways.txt yet so they may PASS this check by virtue of not having any pathways information at all. Or perhaps we want to add an N/A `status` to handle cases like this?

Defer to you, I have no issue merging this PR as-is -- IIRC there's a separate check for "provides pathways info" so people can probably figure it out by looking at the two checks together.

owades:
Thank you @lauriemerrell! It's a toughie, as such a small subset of transit operators use pathways (or have a need for pathways) in the first place. Maybe we do need an N/A value.

For now I will merge this, and more importantly I'm planning on teeing up the logic behind existing checks this week to E, Scott F & Olivia for their feedback. I expect I'll need to make a few adjustments once I hear from them.

As always, I appreciate the consideration into the actual usability of this reporting.

PR #1786, opened 2022-09-14
Use download configs in new RT archiver and carry metadata through parsing and validation

atvaccaro:
# Description

This was split off from the schedule PR to help in deployments.

Resolves https://github.com/cal-itp/data-infra/issues/1785

This has already been deployed to test; will deploy to prod once it is merged. There is a ready-to-go download config at `gs://calitp-gtfs-download-config/gtfs_download_configs/dt=2022-09-15/ts=2022-09-15T14:58:59.865018+00:00/configs.jsonl.gz`.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml


services/gtfs-rt-archiver-v3/docker-compose.yml
```diff
@@ -8,7 +8,8 @@ x-gtfs-rt-archiver-v3-common:
     CALITP_HUEY_REDIS_HOST: redis
     GOOGLE_APPLICATION_CREDENTIALS: /tmp/secret.json
     CALITP_BUCKET__AIRTABLE: "gs://test-calitp-airtable"
-    CALITP_BUCKET__GTFS_RT_RAW: "gs://test-calitp-gtfs-rt-raw"
+    CALITP_BUCKET__GTFS_CONFIG: "gs://test-calitp-gtfs-download-config"
+    CALITP_BUCKET__GTFS_RT_RAW: "gs://dev-calitp-gtfs-rt-raw"
```
lauriemerrell, line 12: when would this be bumped to `-v2`?

atvaccaro: For local running in docker-compose, never; since we have a live "test" version of this, I realized we should be using a `dev` bucket when running locally, for the raw data.

lauriemerrell: ahhh ok, I was not following that `test`/`dev` distinction (which I do think should be a larger conversation in general re: what is the purpose of the "live test")

PR #1811, opened 2022-09-19
create a priority class for dbt run pods and use it

atvaccaro:
May help prevent situations where dbt pods take too long to start.


lauriemerrell (approved):
if we ever make the schedule or Airtable download jobs into pod operators (as you've discussed doing), they should have highest priority right?

atvaccaro:
> if we ever make the schedule or Airtable download jobs into pod operators (as you've discussed doing), they should have highest priority right?

Yes good callout.

Issue #1718, opened 2022-08-29
User Story: Run `create_external_tables` DAG daily


lauriemerrell:
### Summary

As an analytics engineer, I want the `create_external_tables` DAG to run daily, so that:
1. Straightforward schema changes are picked up automatically
2. Issues are identified quickly & at the source, rather than requiring manual re-runs or causing problems in downstream dbt models 

### Acceptance Criteria

`create_external_tables` DAG is scheduled to run daily instead of once-only.

### Notes

The `gtfs_schedule_history` DAG being once-only (the main precedent in our codebase) makes a bit more sense because its schemas are based on actual GTFS (i.e., schema not changing much/at all), and all the tasks are actually related to each other. 

The `create_external_tables` DAG now contains a bunch of tables in different domain areas, with different schema guarantees/expectations. Having to manually run when there's a schema change in one table may result in triggering others to re-run and cause unexpected failures. 

Running regularly would increase stability and allow us to alert more proactively.

### Tester [Stakeholder]

1. @lauriemerrell, also @charlie-costanzo who ran into this in the aftermath of #1701 

# Sprint Ready Checklist
1. - [ ] Acceptance criteria defined
2. - [ ] Team understands acceptance criteria
4. - [ ] Team has defined solution / steps to satisfy acceptance criteria
5. - [ ] Acceptance criteria is verifiable / testable
6. - [ ] External / 3rd Party dependencies identified

PR #1814, opened 2022-09-20
run external tables daily and adjust emails

atvaccaro:
# Description

Closes https://github.com/cal-itp/data-infra/issues/1718

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
locally


lauriemerrell (approved):
my only question is on the "should run before dbt" -- I don't know how we feel about that being just enforced informally via execution time vs. a formal dependency. I guess we don't want a formal dependency because it's *not* an actual logical dependency and we don't want to block the dbt job, so I am ok with this approach, but just wanted to note it.

PR #1819, opened 2022-09-20
add an external table creation pool

atvaccaro:
I think firing off the 40+ tasks at once overloads our celery workers. I've created `external_table_pool` so we can limit but adjust vs a hard-coded concurrency limit.

image


lauriemerrell (approved):
🙏

PR #1821, opened 2022-09-21
actually check if netlify deploys succeeded

atvaccaro:
# Description

We should actually check if the docs deploy subprocess succeeded as the error message is easy to miss in the Airflow logs. I've already updated the k8s secret with the new Netlify token.

Closes https://github.com/cal-itp/data-infra/issues/1818

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Tested locally in docker-compose/k8s.

### DIFF
--- a/warehouse/scripts/run_and_upload.py
+++ b/warehouse/scripts/run_and_upload.py
@@ -55,24 +55,23 @@ def run(
 
     subprocess.run(get_command("compile")).check_returncode()
 
+    results_to_check = []
+
     if dbt_run:
         args = ["run"]
         if full_refresh:
             args.append("--full-refresh")
-        run_result = subprocess.run(get_command(*args))
+        results_to_check.append(subprocess.run(get_command(*args)))
     else:
         typer.echo("skipping run")
-        run_result = None
 
     if dbt_test:
-        test_result = subprocess.run(get_command("test"))
-    else:
-        test_result = None
+        results_to_check.append(subprocess.run(get_command("test")))
 
     if dbt_freshness:
-        freshness_result = subprocess.run(get_command("source", "snapshot-freshness"))
-    else:
-        freshness_result = None
+        results_to_check.append(
+            subprocess.run(get_command("source", "snapshot-freshness"))
+        )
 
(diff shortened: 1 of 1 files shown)


warehouse/scripts/run_and_upload.py
```diff
@@ -55,24 +55,23 @@ def get_command(*args) -> List[str]:
 
     subprocess.run(get_command("compile")).check_returncode()
```
lauriemerrell, line 56: so just to confirm, if compile fails, we don't do anything else? but if compile succeeds we will now attempt: run; docs deploy; metabase sync; freshness check (if configured) and will raise failures from all those steps at the end?

lauriemerrell (commented):
just want to confirm I understand the flow here (inline comment). is there a meaningful way to test this in test?

warehouse/scripts/run_and_upload.py
atvaccaro: There's no real logic change here except checking that the docs did deploy at the end. We've always checked for compile first, since a project that can't compile can't do anything else.

atvaccaro:
> is there a meaningful way to test this in test?

I've confirmed the deploy runs and the subprocess check succeeds on a successful deploy.

warehouse/scripts/run_and_upload.py
lauriemerrell: yeah sorry, I get that this isn't changing much here I mostly just wanted to make sure I understood the overall order of operations. and specifically that we will attempt ALL steps even if the dbt run has errors (which I think we *weren't* doing for a while before?)

lauriemerrell (approved):

warehouse/scripts/run_and_upload.py
atvaccaro: Ah yeah that was true originally and was wrong. A partial/errored run will still be reflected in the production database so we should be updating the docs/Metabase. We don't really have the bandwidth to do blue/green deploys.

Issue #1807, opened 2022-09-19
User Story: dbt-specific alerting that doesn't require accessing Airflow UI


lauriemerrell:
### Summary

As an analytics engineer / data product owner,
I want specific & visible alerting on dbt failures
So that I am aware of data availability issues without having to wade through unrelated alerts. 

Specific request prompted by @owades re: wanting to know about data freshness issues in `fact_daily_guideline_checks`, but reinforced by the fact that I had missed a significant ongoing issue with the `dbt test` DAG task because of alert fatigue on that specific DAG task failure. 

### Acceptance Criteria

Some dedicated location that calls out dbt table build/test failures without requiring people to look in Airflow directly.

Ideas:
* (highest effort / not sure this is worth it) We could parse the manifest or similar to make a table that could drive a Metabase dashboard showing people any tables that had issues.
* We could just drive alerts from the dbt-specific Airflow jobs to a more visible channel in Slack, and maybe reformat their alerts to be more informative. 
* We could make more informative email alerts for the dbt task failures and send those to interested parties.

### Tester [Stakeholder]

1. @owades 
2. me (@lauriemerrell)

# Sprint Ready Checklist
1. - [ ] Acceptance criteria defined
3. - [ ] Team understands acceptance criteria
4. - [ ] Team has defined solution / steps to satisfy acceptance criteria
5. - [ ] Acceptance criteria is verifiable / testable
6. - [ ] External / 3rd Party dependencies identified

lauriemerrell:
Closing this as addressed by Sentry.

PR #1822, opened 2022-09-21
Try to deploy sentry with the helm chart :)

atvaccaro:
# Description

Resolves # [issue]

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml


airflow/docker-compose.yaml
```diff
@@ -101,6 +101,7 @@ x-airflow-common:
     POD_LOCATION: "us-west2-a"
     AIRFLOW_ENV: "development"
     CALITP_USER: "pipeline"
+    SENTRY_DSN: "http://b5b82f68a88f4bbdb627fa774636cf62@sentry-nginx.sentry.svc.cluster.local/2"
```
themightychris, line 104: does this value work in local docker-compose envs?

kubernetes/apps/manifests/gtfs-rt-archiver-v3/archiver-app-vars.yaml
```diff
@@ -6,3 +6,4 @@ data:
   CALITP_HUEY_REDIS_HOST: redis
   CALITP_USER: pipeline
   GOOGLE_APPLICATION_CREDENTIALS: /secrets/gtfs-feed-secrets/google_application_credentials.json
+  SENTRY_DSN: "http://b5b82f68a88f4bbdb627fa774636cf62@sentry-nginx.sentry.svc.cluster.local/2"
```
themightychris, line 9: is this just a placeholder value or do we know `b5b82... ` will exist in fresh envs?

atvaccaro: This is configured at the Sentry project level so anything we want to log to this project (I named it `cal-itp`) will use this DSN, whether we're in the cluster or not.

airflow/docker-compose.yaml
atvaccaro: It's passed through to PodOperators that run in the cluster; the _Airflow_ running in docker-compose won't have access until we set up Ingress.

I'll duplicate the env var and add a comment.

lauriemerrell (commented):
How much additional work would it take to close #1807 with this PR? Also just to confirm, all the actual alerting settings (like saying that I should be notified for XYZ specific test failures) would be done in the Sentry UI, right? 

Is it possible for us to pull out the dbt table name that is causing the test or run failure in this PR? dbt test logging is.... really annoying and the table name is just like in the middle of the test name.

airflow/docker-compose.yaml
atvaccaro: This now uses the same DSN.

kubernetes/apps/manifests/gtfs-rt-archiver-v3/archiver-app-vars.yaml
atvaccaro: This is also updated to the new DNS.

atvaccaro:
> How much additional work would it take to close #1807 with this PR? Also just to confirm, all the actual alerting settings (like saying that I should be notified for XYZ specific test failures) would be done in the Sentry UI, right?

Yes, alerting is configured in the UI. For more granular/powerful dbt alert routing, we'd probably want to use CODEOWNERS or similar. https://docs.sentry.io/product/issues/issue-owners/ We just need to decide who needs to know about what. I'd like to get Sentry deployed as-is so we can start quieting Slack alerts and inviting users.

> Is it possible for us to pull out the dbt table name that is causing the test or run failure in this PR? dbt test logging is.... really annoying and the table name is just like in the middle of the test name.

It's a bit tricky since a test can potentially related to multiple tables (for example a relationships test) but it's possible. I will work on it.

lauriemerrell:
@atvaccaro ok thanks we can discuss further offline. I think we might be able to get away with alerts at the dataset level which you can get from the filepath that dbt logs in the failure I think (like the `see compiled at` part.)

atvaccaro:
> @atvaccaro ok thanks we can discuss further offline. I think we might be able to get away with alerts at the dataset level which you can get from the filepath that dbt logs in the failure I think (like the `see compiled at` part.)

In this case `run_results.json` drives the Sentry reporting, but we can look up the associated models via the manifest.

airflow/docker-compose.yaml
```diff
@@ -101,6 +101,7 @@ x-airflow-common:
     POD_LOCATION: "us-west2-a"
     AIRFLOW_ENV: "development"
     CALITP_USER: "pipeline"
+    SENTRY_DSN: "https://e498431022154366b0ff8b71cf2d93e0@sentry.calitp.org/2"
```
themightychris, line 104: is this DSN dedicated to local envs? Highly recommend using separate DSNs to segment traffic between envs as most of Sentry's early-stage filtering and flow control is at the DSN level. I've hosed DSNs before by letting them float around to local dev envs

kubernetes/apps/charts/sentry/values.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - secretName: metabase-tls
+        hosts:
+          - sentry.calitp.org
+  mail:
+    backend: smtp
+    useTls: true
+    username: "apikey"
+    port: 587
+    host: "smtp.sendgrid.net"
+    from: "hunter@compiler.la"
```
themightychris, line 29: this email is probably not ok to use anymore

kubernetes/apps/manifests/gtfs-rt-archiver-v3/archiver-app-vars.yaml
themightychris: won't new random ones get generated after we deploy to the cluster and instantiate the new sentry instance? or has it already been deployed manually and this PR is just catching the IaC up?

themightychris (approved):
none of my feedback is blocking, looking to @lottspot on the IaC bits though

kubernetes/apps/manifests/gtfs-rt-archiver-v3/archiver-app-vars.yaml
atvaccaro: This has already been deployed; I'm segmenting by environment rather than project.

kubernetes/apps/charts/sentry/values.yaml
atvaccaro: I will ask the services team; I think they had said it was OK last time we asked about Airflow emails coming from this.

airflow/docker-compose.yaml
atvaccaro: Does that imply separate projects? I was planning to segment by environment rather than project if possible. I can certainly split off a dev project.

services/gtfs-rt-archiver-v3/gtfs_rt_archiver_v3/consumer.py
```diff
@@ -12,14 +13,35 @@
...
+    return event
+
+
 def main(
     port: int = os.getenv("CONSUMER_PROMETHEUS_PORT", 9102),
     load_env_secrets: bool = False,
 ):
+    sentry_sdk.init(
+        environment=os.getenv("AIRFLOW_ENV"),
+        # TODO: add me back if fetch errors are being grouped together too much
```
lauriemerrell, line 42: per offline convo, should add this back in -- `ConnectionError` in particular is too generic and not actionable without URL

atvaccaro: Can do.

atvaccaro: This is done; it will already work for dbt because a custom exception is constructed with the unique_id and message, and the schedule download failures come from exception logs rather than unhandled exceptions (which we do want to maintain for RT so huey tasks exit with a failure).

lauriemerrell (approved):

airflow/docker-compose.yaml
atvaccaro: I'm going to deploy this as-is but make myself an issue to investigate separate envs/DSNs. Mainly I wonder how it affects team/member issue assignment and notifications.

Issue #1825, opened 2022-09-21
Bug: Feed aggregator checker doesn't assign correct `calitp_url_number`


lauriemerrell:
The feed aggregator checker does not assign URL numbers correctly. Buggy line is here I think: https://github.com/cal-itp/data-infra/blob/main/jobs/gtfs-aggregator-checker/gtfs_aggregator_checker/__init__.py#L51

This results in the behavior described below, which is incorrect and prevents us from being able to actually join the results from the feed aggregator checks with our other URL lists. 

Initial bug report from @owades: 
> Thank you @charlie-costanzo! I poked around today and I do have one observation/question - let me know if this is an OK place to discuss (versus slack, etc). I don't think the concept of calitp_url_number is being used correctly here. For example, in this table, a transit provider that has 4 feeds (1 schedule, 3 rt) would assign calitp_url_number 0, 1, 2, 3 respectively to those 4 feeds. I believe that elsewhere in Cal-ITP data tables, all 4 feeds would have calitp_url_number of 0 (and if there were an additional set of 4 feeds, they would get calitp_url_number of 1). So basically, I believe the calitp_url_number should refer to a set of up to 4 feeds.
![Screen Shot 2022-09-20 at 4 36 39 PM](https://user-images.githubusercontent.com/3301353/191383312-f0b77f6a-06a3-4299-a7fa-57c0b99fb514.png)
_Originally posted by @owades in https://github.com/cal-itp/data-infra/issues/1737#issuecomment-1253021585_

PR #1828, opened 2022-09-21
Add transitland check

owades:
# Description

New GTFS check. Based on [this slack thread](https://cal-itp.slack.com/archives/C01FNDG1ZPA/p1663779399139829), it is imperfect, but here is how it works:

For agencies with only one of the given GTFS feed: if the feed is on listed on transit.land, then PASS, else FAIL.
For agencies with more than one of the given feed, if ANY of the feed are listed on transit.land, then PASS, else FAIL. 

This odd behavior is a result of [this bug](https://github.com/cal-itp/data-infra/issues/1825). There is a more detailed solution possible, where we join based on the actual URL, but I don't think the juice is worth the squeeze, since it would be difficult to explain the additional complexity, and we expect to upgrade all of this soon.

This PR also updates docs slightly: removing markdown+html syntax that didn't work properly in dbt
## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

I went through a handful of agency results, comparing with what I saw on transitland. It checked out.

I also checked the markdown changes on a markdown preview tool.


warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__schedule_feed_on_transitland.sql
```diff
@@ -0,0 +1,38 @@
+WITH feed_guideline_index AS (
+    SELECT * FROM {{ ref('stg_gtfs_guidelines__feed_guideline_index') }}
+    WHERE check = {{ schedule_feed_on_transitland() }}
+),
+
+fact_daily_transitland_url_check AS (
+SELECT * FROM {{ ref('stg_gtfs_guidelines__fact_daily_transitland_url_check') }}
+),
+
+schedule_daily_check AS (
```
lauriemerrell, line 10: (this comment applies to all 4 checks) I think that you need to do some deduplication here (either `SELECT DISTINCT` or a group by or something else) -- I think that you might get fanout on this because (with the url number issue) this table can have multiple rows for one ITP ID that meet this condition.

(I am going to try to be more mindful of this, now that I've noticed some of the failing tests I messaged you about -- I suspect there are other checks that may have a version of this issue.)

lauriemerrell (changes requested):
One inline comment, which actually gave me an idea. It might be useful to refactor the catchall test that we have right now: https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/gtfs_guidelines/_gtfs_guidelines.yml#L13-L14, we could add individual check-level tests in the staging tables: https://github.com/cal-itp/data-infra/blob/main/warehouse/models/staging/gtfs_guidelines/_stg_gtfs_guidelines.yml -- I believe dbt allows you to specify an equal row count test with a `where` clause, so we could add tests in the YML to check that the resulting staging table has an equal count of rows with the portion of the index table that it started with. Does that make sense? If so, would you be interested in doing that here? If not, I can write an issue and see if we can do that on our side (and we'd ask you to add iteratively to that as you add new checks).

I'm just realizing that would give us more granular checks on where fanout is occurring. 🤔

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__schedule_feed_on_transitland.sql
owades: ah right, of course. yes I will dedupe here.

lauriemerrell:
> The tests are a good idea, and I can do it. I will try it out using the fact_daily_transitland_url_check as a model and see how far I get. Does dbt compile and/or dbt run automatically run the test?

Alas, no, you have to run them separately using [`dbt test`](https://docs.getdbt.com/reference/commands/test) -- apologies for not walking through that more specifically as part of dbt onboarding. Can use the same selection syntax as `dbt run` so you don't have to run tests on the whole warehouse (ex. `dbt test -s stg_gtfs_guidelines__feed_guideline_index`)

I actually wonder if you'd want to have *all* the tests that we currently have on the mart table applied on these staging tables instead, including the not-null and uniqueness, since those are currently also throwing a few failures in prod and I think the same considerations apply that it's better to figure out where they're coming from on the input side.

Thanks for bearing with us on all this, I think we're still feeling out our testing strategy in terms of which tests should be applied in staging vs. mart and such. 🙏

owades:
@lauriemerrell uniqueness and non-null checks implemented - let me know what you think! Once you approve/suggestion changes to this format, I can (slowly) go back and add tests to other checks

lauriemerrell (approved):
This all looks good to me! Thanks again for your flexibility with the changing guidance.

Just for reference, I had been wondering if we could also add a test like:
```
      - dbt_utils.equal_rowcount:
          compare_model: ref('stg_gtfs_guidelines__feed_guideline_index')
          where: check = {{ schedule_feed_on_transitland() }}
```

But it looks like dbt doesn't support macros in the YML in this fashion (see https://github.com/dbt-labs/dbt-core/discussions/5097, hat tip @atvaccaro for finding; I also tested locally and confirmed it doesn't seem to work). 

So, I think leaving the equal row count test on the final mart model is fine, and these uniqueness & not null tests should help. Since we're doing left joins onto the index in every case, I think it's fanout and not dropped rows that we're primarily worried about, and the uniqueness tests should catch that. 👍

Issue #1761, opened 2022-09-07
GTFS schedule warehouse refactor: Initial analytical tables


lauriemerrell:
Based on feedback from our [model storming conversation with analysts](https://docs.google.com/document/d/11goFDpVxvNa6TfiBcLwfww4bJgTJYW9iaff9Z09ArGQ), create the following tables (naming subject to change) in v2 of the warehouse:

- [x] `dim_feeds` -- feeds table with **feed-level** versioning information and associated organization(s) from Airtable -- see #1871
- [x] `fact_daily_feeds` -- `dim_feeds` made daily (i.e., what feeds were in effect on a given date) -- see #1871
- [x] `fact_daily_scheduled_trips` -- table of trips that were scheduled on a given date; should be actually incremental (address performance issues of prior version of this table); should be annotated with relevant data from RT (for example: did we get a cancelation in trip updates? was the trip observed in vehicle positions? -- these latter two questions will likely require their own tables to support)
- [x] (from prior model storming session with Eric several months ago) cleaned vehicle positions table that dedupes on header timestamp and perhaps pre-calculates start/stop at a given location (if lat/long don't change between subsequent pings)

PR #1836, opened 2022-09-22
First v2 fact table: `fct_vehicle_positions_messages`

lauriemerrell:
# Description

First new mart table 🎉  -- `fct_vehicle_positions_messages`. Also creates staging tables for download & unzip outcomes which will help create a `dim_feeds` in a follow up PR.

A few notes & questions:

* We will need to change the Airtable data keying to support historical versioning, but I did not want to tackle this in this PR. 
* How to check `wheelchairAccessible` -- there's a field in the vehicle positions vehicle message that we didn't include before. I added it here but I don't see any instances where it's populated in the external table. I'm wondering if I've specified the name incorrectly or if just no one is using it yet.

Partial progress towards #1761 

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

`dbt run` and `dbt test` run locally. There are some tests failing in other Airtable data but I believe they are unrelated to these changes. 

Also confirmed I can actually merge the fact table with GTFS datasets 😄 (date chosen because that's when test data is available -- leads to a few failed joins but I spot checked and I think those are because of actual Airtable record changes, i.e., they would be addressed by properly making the Airtable table historical, which is a known need.)

```
SELECT DISTINCT gtfs_dataset_key, dt, hour, t1._name, t1.base64_url, _extract_ts, _config_extract_ts, t2.key, t2.base64_url, t2._valid_from, t2._valid_to, t2.name
FROM `cal-itp-data-infra-staging.laurie_mart_gtfs.fct_vehicle_positions_messages` AS t1
LEFT JOIN `cal-itp-data-infra-staging.laurie_mart_transit_database.dim_gtfs_datasets` AS t2
  ON t1.gtfs_dataset_key = t2.key
WHERE t1.dt = '2022-09-12' and t1.hour = '2022-09-12T17:00:00+00' AND gtfs_dataset_key IS NULL
```


warehouse/models/mart/gtfs/_mart_gtfs.yml
```diff
@@ -0,0 +1,119 @@
...
+    columns:
+      - name: key
+        description: |
+          Synthetic primary key constructed from `base64_url`, `extract_ts`, and entity `id`.
+        tests:
+          - unique:
+              #TODO: make an RT test macro that does this date subtraction
+              where: dt >= DATE_SUB(CURRENT_DATE(), INTERVAL 2 DAY)
+          - not_null:
+              where: dt >= DATE_SUB(CURRENT_DATE(), INTERVAL 2 DAY)
```
atvaccaro, line 18: we may need to revisit these depending on how many bytes these bill... maybe we also pick a specific hour of the day to eliminate more partitions?

atvaccaro (approved):

warehouse/models/mart/gtfs/_mart_gtfs.yml
lauriemerrell: I forgot about this

lauriemerrell: I'll filter to a rush hour hour Pacific

PR #1837, opened 2022-09-22
Add check for missing shape_id's

owades:
# Description

Here is a new a GTFS check which confirms that there is a shape_id listed for every trip in trips.txt. If every trip has a non-null shape_id, this check has a result of "PASS". If even one row is missing a shape_id, it will have a result of "FAIL."

This check is testing for the **bold part** of the following guideline (from [this guideline draft](https://docs.google.com/document/d/1ao9dK0FyM_6mriz1ax_7QM84wg3HcADfepPV5C7-Cd4/edit#)):

> Shapes.txt is included within the feed, with valid **route shapes for each trip**. Shapes should be precise enough to show the right-of-way that the vehicle uses and not inaccurately exit the right-of-way.

Originally I was going to cross-check trips.txt with shapes.txt, to confirm each shape in trips.txt actually shows up in shapes.txt, but it occurred to me that instead we can rely on the gtfs-validator error: `foreign_key_violation`, would will be triggered by that case.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Manually reviewed several pass and fail results and confirmed the trips.txt files matched. This is also nearly identical to some existing checks, which have also been tested.


lauriemerrell (approved):
Looks good, my only question is about the `foreign_key_violation` -- I guess that that's a critical error / should be caught by the check for no actual errors from the validator, is the implication that you'd just rely on that other check to confirm that the shape IDs are valid? Just wanted to confirm that there's no actual check here for that validator error.

owades:
That's right, @lauriemerrell. It would be nice to capture foreign_key_violations specific to shapes.txt, but relying on the other validator check feels OK to me. A lot cleaner than reinventing the foreign key violation check (which I nearly did before backtracking). I will still run this by Olivia and E, though.

PR #1851, opened 2022-09-27
add ts to rt metadata

atvaccaro:
Laurie noticed I left out the actual RT extract timestamp from the final metadata, which we do need since the header timestamp is the last updated time (I think).


lauriemerrell (approved):
🙏 

And yes, just confirming we want this attribute so that we can:
1. Join between an input file and its outcome in the aggregation process
2. Monitor discrepancies between tick timestamp and header timestamp for quality (to assess the guideline on update frequency)

PR #1857, opened 2022-09-28
add VCTC GMV GTFS Schedule feed

### LINKED ISSUE #1855: GTFS Data: Update VCTC with GMV Schedule Feed
Update `agencies.yml` in order to add GMV VCTC GTFS Schedule feed to facilitate matching with RT feeds.

This will help support producing speedmaps and other metrics to support a Caltrans D5 grant application.

This feed is already tracked [in Airtable](https://airtable.com/appPnJWrQ7ui4UmIl/tbl5V6Vjs4mNQgYbc/viwx2oxkzkK7vOjPB/recrAG7e0oOiR6FiP?blocks=hide) and available at https://govcbus.com/gtfs

Update the following gtfs datsets:

- 380 ventura-county-transportation-commission

The current entry has the Trillium schedule feed url grouped with the GMV RT feed urls, I think the right thing to do is separate into 2 feeds, one with the existing schedule url and "null" rt urls and one with the RT urls and matching GMV GTFS Schedule (similar to Victor Valley Transit 360)

## Checklist

- [x] Manually made sure any new feeds have `itp_id`s that are not duplicative
- [x] Manually confirmed URIs are valid (to be automated in future per #908)
- [x] Made sure the Airtable database has consistent information

edasmalchi:
# Description

Quick agencies.yml change to add VCTC's GMV GTFS Schedule feed, needed for rt analysis to support Caltrans D5 ask. Feed already tracked in Airtable, see issue for additional context+checklist.

Resolves #1855 

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -1467,10 +1467,14 @@ vacaville-city-coach:
 ventura-county-transportation-commission:
   agency_name: Ventura County Transportation Commission
   feeds:
-    - gtfs_schedule_url: http://data.trilliumtransit.com/gtfs/vctc-ca-us/vctc-ca-us.zip
+    - gtfs_schedule_url: https://govcbus.com/gtfs
       gtfs_rt_vehicle_positions_url: https://govcbus.com/gtfs-rt/vehiclepositions
       gtfs_rt_service_alerts_url: https://govcbus.com/gtfs-rt/alerts
       gtfs_rt_trip_updates_url: https://govcbus.com/gtfs-rt/tripupdates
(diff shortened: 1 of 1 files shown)


lauriemerrell (approved):
This seems ok to me, though FYI it will also affect what URL is being used to generate the reports for this agency going forward (unless you've changed that logic re: URL 0). In particular when you're going from a feed with most of the optional files to a feed without. I'll leave it to you to merge to confirm you're comfortable with that (I think it's the right move; the main alternative would be leaving the old URL as URL 0 and moving the new URL and all the RT to URL 1, which I don't think is good because if one feed has RT and the other doesn't that should probably be URL 0.)

PR #1861, opened 2022-09-28
configure itp-id as FK on metabase

owades:
# Description

In order to apply a dashboard-level filter on a Metabase question using the itp-id field in `stg_gtfs_guidelines__fact_daily_transitland_url_check`, I need that field listed as a foreign key in the config yaml file.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
It hasn't been tested - relying on a review here! I followed [guidance provided by Laurie in Slack here](https://cal-itp.slack.com/archives/C02KH3DGZL7/p1663957100230209?thread_ts=1663953246.827349&cid=C02KH3DGZL7). Note that I didn't a YAML anchor because I don't expect to use this field again in this directory, as it mostly contains staging files.


lauriemerrell (approved):
Thanks for catching & updating these!

Issue #1868, opened 2022-09-30
GTFS schedule refactor -- warehouse: Make Airtable tables historical


lauriemerrell:
To ensure we can make the most of our GTFS data, we need type 2 Airtable data that contains prior data with effective dates (instead of only surfacing the current Airtable data, as we do now).

For this we should:
Change `gtfs_datasets`, `gtfs_service_data`, `services`, and `organizations` to: 

* Stop filtering to only one extract per day in staging
* Stop getting latest-only in mart 
* Start using a versioned key (instead of just using the Airtable record ID) -- surrogate key from record ID + extract timestamp. 
   * In the case of `gtfs datasets` in particular, we may want to use base64_url + timestamp, although this will introduce some complexity. We will need to handle historical cases where duplicate or null URLs were present.
* Update the bridge & synced table handling accordingly; this will be challenging because not all Airtable extracts will have the same timestamps

lauriemerrell:
Closing as not doing; have come to the assessment that most Airtable data updates are corrections rather than changes that should be captured from a continuity perspective. Will need to work out issues as they arise but approach proposed here doesn't seem valuable at this point.

lauriemerrell:
Reopening, we are doing this after all

PR #1871, opened 2022-10-03
Make v2 dim feeds

lauriemerrell:
# Description

Partially addresses #1761 by creating `dim_schedule_feeds` and `fct_schedule_feeds` in "v2 warehouse". (No longer `fact_daily_feeds` because we can version at a more granular level now.) 

Note that I am not pre-joining to organizations as described in that ticket because it's many-to-many and therefore a bit complicated, will have to keep iterating. 

I was unsure about the interdependency between the fact and dimension tables (wanting to build the dimensional table from the fact table but wanting to flag the fact table with the associated dimension record), ended up handling by creating an intermediate proto-fact table and driving both the mart tables from that. We may need to tweak that.

We have a ticking time bomb in the fact that currently `dim_gtfs_datasets` does not officially guarantee uniqueness by URL, so when we make that historical we are going to have to figure something out. This is already noted in ticket #1868.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

`poetry run dbt run -s +models/mart/gtfs` & `poetry run dbt test -s +models/mart/gtfs`, confirmed both succeed. Tables can be previewed at `cal-itp-data-infra-staging.laurie_mart_gtfs.fct_schedule_feeds` and `cal-itp-data-infra-staging.laurie_mart_gtfs.dim_schedule_feeds`

### DIFF
--- a/warehouse/models/docs/_docs_generic.md
+++ b/warehouse/models/docs/_docs_generic.md
@@ -1,0 +1,13 @@
+Docs macros that apply across several domains
+
+{% docs column_valid_from %}
+
+Starting timestamp (inclusive) from which this record is valid.
+
+{% enddocs %}
+
+{% docs column_valid_to %}
+
+Ending timestamp (inclusive) through which this record is valid.
+
+{% enddocs %}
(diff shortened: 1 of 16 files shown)


charlie-costanzo (approved):
LGTM! Thanks so much for taking care of this - I'll dive deeper as I get into the rewrite work but this looks like a great start

PR #1883, opened 2022-10-06
Remove notices from shapes validator check

owades:
# Description

Per team feedback, omit validator notices from shape-related validator check.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Hasn't been tested, but changes are minimal.


lauriemerrell (approved):
LGTM, only question would be maybe whether there's any chance that this was one of the tables causing fanout so running `dbt run` and `dbt test` could give insight into whether this is going to throw dbt test failures in prod. But it seems very unlikely so not blocking IMO, just wanted to note.

owades:
Thanks @lauriemerrell, I was a little lazy but you're definitely right. Just tested and all good (after re-running evvverrrything on my schema)

PR #1888, opened 2022-10-11
make gtfs schedule downloads run sooner (i.e. end of the UTC day)

### LINKED ISSUE #1886: Adjust when the GTFS schedule download normally runs.
Right now we're in the worst of both worlds; we run GTFS schedule downloads _before_ a new Airtable version is extracted, but we also run early in the day UTC so it takes a long time for unzipping to occur. The simplest improvement would be to just download feeds _after_ the Airtable extract.

atvaccaro:
# Description

Closes https://github.com/cal-itp/data-infra/issues/1886

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Locally.

### DIFF
--- a/airflow/dags/download_gtfs_schedule_v2/METADATA.yml
+++ b/airflow/dags/download_gtfs_schedule_v2/METADATA.yml
@@ -1,5 +1,5 @@
 description: "Download GTFS Schedule feeds as specified in Airtable"
-schedule_interval: "0 0 * * *"
+schedule_interval: "0 23 * * *"
 tags:
   - all_gusty_features
 default_args:


lauriemerrell (approved):
LGTM, we'll have to keep assessing this.

This change does mean that data will be unzipped immediately after download. It does not really change the fact that the gap between the Airtable extract and that Airtable extract's ingestion is longer than a business day.

atvaccaro:
closing in favor of https://github.com/cal-itp/data-infra/pull/1890

PR #1898, opened 2022-10-14
Deploy MVP Dask cluster and pull Jupyter singleuser image into this repo

atvaccaro:
# Description

Deploys a single Helm-managed Dask cluster to Kubernetes. The cluster is reachable with the following snippet.

```python
from dask.distributed import Client
client = Client("dask-scheduler.dask.svc.cluster.local:8786") # in jupyterhub the url is optional, there is a default set in a config file
client
```

Also creates a new Jupyter user image `ghcr.io/cal-itp/data-infra/jupyter-singleuser` that will be managed in this repo from now on, and replaces the `cal-itp` image. A dependency update within the new image solves the matplotlib issue [described here](https://stackoverflow.com/questions/73792756/plotnine-is-broken-with-matplotlib-3-6).

One thing to consider in the future is using the full daskhub Helm chart which would deploy both JupyterHub and Dask-Gateway for us.

There is an [imported Grafana dashboard](https://monitoring.k8s.calitp.jarv.us/d/a_khcwhWk/dask) to monitor the cluster.

## Type of change

- [x] New feature
- [x] Documentation
- [ ] agencies.yml

## How has this been tested?
Currently deployed.


kubernetes/apps/charts/dask/values.yaml
```diff
@@ -0,0 +1,35 @@
+dask:
+  scheduler:
+    image:
+      repository: "ghcr.io/cal-itp/data-infra/dask"
+      tag: "2022.10.13"
+    metrics:
+      enabled: true
+    serviceAnnotations:
+      prometheus.io/port: "8786"
+      prometheus.io/scrape: "true"
```
atvaccaro, line 10: this is not being picked up by prom

themightychris (approved):

PR #1899, opened 2022-10-14
18 is deprecated, just use latest for everything

atvaccaro:
# Description

Solves the failing action in https://github.com/cal-itp/data-infra/actions/runs/3250370484

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
This is hard to do without enabling image builds on PRs, which we could discuss.


lauriemerrell (approved):
link for posterity: https://github.com/actions/runner-images/issues/6002

PR #1894, opened 2022-10-13
Refactored GTFS schedule "base" tables

charlie-costanzo:
# Description
Continuation of work from #1884 

This PR includes refactored GTFS-schedule tables in the warehouse:
* Staging tables
* Pure dimensional tables
* Intermediate (deduplicating) tables
* Yml, docs, and tests updates

(*Latest-only views are not included in this PR*)

**The staging tables**:
* Are located in warehouse/staging/gtfs
* Have reimplemented whitespace & typecasting
* Are implemented for all the GTFS schedule tables that [have their own external tables](https://github.com/cal-itp/data-infra/tree/main/airflow/dags/create_external_tables/gtfs_schedule_v2)

**The pure dimensional tables**:
* Are located in warehouse/marts/gtfs
* Are constructed by inner joining the relevant staging table with dim_feeds to add the appropriate feed_key & effective dates

**The intermediate tables**:
* Are in warehouse/intermediate/gtfs
* Used for deduplication

Partially resolves (need to create latest only tables)
#1867 

## Type of change

- [x] New feature

Issue #1867, opened 2022-09-30
GTFS schedule refactor -- warehouse: GTFS schedule "base" tables


lauriemerrell:
Once a new `dim_feeds` table is created in #1761, we will need to create three batches of very consistent tables (tables that follow very similar patterns, replicated for all the different GTFS schedule files):

- Staging tables 
- Pure dimensional tables 
- Latest-only views on top of those dimensional tables, which can be used for things like the open data publishing to select only the most recent data

The staging tables will:
* Be located in `warehouse/staging/gtfs`
* Reimplement whitespace & typecasting handling currently handled in [`gtfs_views_staging`](https://github.com/cal-itp/data-infra/tree/main/warehouse/models/gtfs_views_staging)
* Be implemented for all the GTFS schedule tables that have their own external tables

The pure dimensional tables will be:
* Be located in `warehouse/marts/gtfs`
* Be constructed by inner joining the relevant staging table with `dim_feeds` to add the appropriate `feed_key` & effective dates (i.e., we will leverage feed-level versioning implemented in `dim_feeds` & will not re-implement versioning logic for each individual file)
* For stop_times in particular, we may need to explore making the table incremental -- we can check resource usage and defer this step if it's not too intensive 

The latest-only views on top of the dimensional tables will:
* Be located in `warehouse/marts/latest_gtfs`
* Be generated from a macro that selects only the latest data from each individual dimensional table 
* Replicate the existing functionality from the [`gtfs_schedule_latest_only`](https://github.com/cal-itp/data-infra/tree/main/warehouse/models/gtfs_schedule_latest_only) module

lauriemerrell:
Latest-only work, if needed, will happen in https://github.com/cal-itp/data-infra/issues/1940 -- closing this as completed by #1902, #1894, and #1884

PR #1902, opened 2022-10-14
Remaining base tables

lauriemerrell:
# Description

This PR does a few things:
* Adds the v2 `calendar`, `calendar_dates`, and `stop_times` staging and dimension (mart) tables 
* Removes `gtfs_dataset_key` from all dimension tables (was removed from macro in #1893 but that change was not incorporated in #1894 so that was causing column doesn't exist errors)
* **Stops deduplicating when GTFS primary key is not unique**: through testing with `stop_times`, I realized that deduplicating without sufficient care can lead to dangling references or other integrity issues. I am opting instead to allow for fanout and failing tests until we can have a bigger conversation about how we want to handle violations of the GTFS spec. 
* Adds an `is_future` flag to `fct_daily_schedule_feeds`, in preparation for work on daily service and daily trips (where future vs. past service flag will help determine what rows to update for incremental materializations)

Means that #1867 is mostly complete, pending:
* Creation of latest-only datasets (or other decision regarding where the open data will be published from)
* Decision about deduplicating 

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

`poetry run dbt run -s +models/mart/gtfs` and `poetry run dbt test -s +models/mart/gtfs`

**This PR intentionally reintroduces some failing tests, because I realized it is safer and more coherent than the way we were duplicating before. This is something we will plan to urgently address; we do not want these failing tests long term.**


atvaccaro (approved):
lgtm, there's some hanging empty lines from a find-replace but it's probably fine

PR #1917, opened 2022-10-24
fix the jupyter singleuser build, and also build dask

atvaccaro:
# Description

Resolves # [issue]

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml


tiffanychu90 (approved):

tiffanychu90:
@atvaccaro: Slightly tangential...but I think the dask client wouldn't have access to our Google BQ and GCS credentials? 

Testing right now whether I could write a partitioned parquet out without the serialization gives me a
`OSError: Forbidden: https://storage.googleapis.com/upload/storage/v1/b/calitp-analytics-data/o
Access denied.`

 Would adding `dask-bigquery==2022.05.0` help?

atvaccaro:
> @atvaccaro: Slightly tangential...but I think the dask client wouldn't have access to our Google BQ and GCS credentials?
> 
> Testing right now whether I could write a partitioned parquet out without the serialization gives me a `OSError: Forbidden: https://storage.googleapis.com/upload/storage/v1/b/calitp-analytics-data/o Access denied.`
> 
> Would adding `dask-bigquery==2022.05.0` help?

Broadly yes, I will add that lib; there's a couple ways we could have Dask do the access. https://docs.dask.org/en/stable/how-to/connect-to-remote-data.html#google-cloud-storage

I think the best option for right now is the `storage_options` param, so you authenticate as yourself. I'll worker on creating a new service account all the workers can use.

PR #1920, opened 2022-10-25
Add Torrance Transit Swiftly URLs and Header

aly-medina:
# Description

Adds Swiftly URL's to agencies.yml and authorization key to headers.yml

## Type of change

- [ ] New feature
- [ ] Documentation
- [x] agencies.yml

## How has this been tested?

Manual check to be done due to API Key.

### DIFF
--- a/airflow/data/agencies.yml
+++ b/airflow/data/agencies.yml
@@ -1380,6 +1380,10 @@ torrance-transit-system:
   agency_name: Torrance Transit System
   feeds:
     - gtfs_schedule_url: https://transit.torranceca.gov/home/showdocument?id=16673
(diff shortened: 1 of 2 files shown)


lauriemerrell:
Thank you @aly-medina for this! 

It looks like the linter is annoyed about some formatting, would you be able to run `pre-commit` locally (instructions [here](https://github.com/cal-itp/data-infra#pre-commit)) and push those updates and see if that allows the actions to succeed?

I also am just curious -- do we know if Torrance intends to keep (maintain) **both** sets of RT feeds? I see only the My Bus Info ones in Airtable. Asking because I think it's moderately unusual for us to have two full sets of RT feeds that reference the same schedule feed, but definitely seems possible.

lauriemerrell (approved):

PR #1922, opened 2022-10-25
Downgrade Dask and upgrade calitp/siuba!

atvaccaro:
# Description

Downgrades dask back down to 2022.8; the latest release of the helm chart is 2022.8 and it doesn't allow overriding the entrypoints, so it's using a CLI flag that was removed by dask 2022.10.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
This is deployed.


tiffanychu90 (commented):
Just to double check, is `graphviz` installed with both `pip` and `conda`? I thought I saw a `sudo apt-get install graphviz` before, but can't find it now.  This [StackOverflow](https://stackoverflow.com/questions/35064304/runtimeerror-make-sure-the-graphviz-executables-are-on-your-systems-path-aft) post did list either as an option.

atvaccaro:
> Just to double check, is `graphviz` installed with both `pip` and `conda`? I thought I saw a `sudo apt-get install graphviz` before, but can't find it now. This [StackOverflow](https://stackoverflow.com/questions/35064304/runtimeerror-make-sure-the-graphviz-executables-are-on-your-systems-path-aft) post did list either as an option.

Yes, it was in a prior PR, see here https://github.com/cal-itp/data-infra/blob/main/images/jupyter-singleuser/Dockerfile#L11

tiffanychu90 (approved):

PR #1932, opened 2022-10-27
Handle duplicate primary keys in trips

lauriemerrell:
# Description

This PR takes a first pass at flagging duplicate primary keys in the trips table.

It:
* Creates an intermediate incremental table, so that we can have a sparse history materialized as a table against which to do duplicate detection
* Adds a `warning_duplicate_primary_key` column in `dim_trips` to identify rows that have duplicate keys. This allows consumers the freedom to choose how to handle them, while making it obvious that there is an issue. We may iterate on this approach. 

Progress towards `fact_daily_scheduled_trips` in #1761.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

`poetry run dbt run -s stg_gtfs_schedule__trips int_gtfs_schedule__trips dim_trips` & 
`poetry run dbt test -s stg_gtfs_schedule__trips int_gtfs_schedule__trips dim_trips`.

Also paired with @atvaccaro to inspect the incremental materialization logic and confirmed it appeared solid.


warehouse/models/intermediate/gtfs/int_gtfs_schedule__incremental_trips.sql
```diff
@@ -0,0 +1,47 @@
...
+stg_gtfs_schedule__trips AS (
+    SELECT *
+    FROM {{ ref('stg_gtfs_schedule__trips') }}
+    {% if is_incremental() %}
+    WHERE _dt >= EXTRACT (DATE FROM TIMESTAMP('{{ max_ts }}'))
+    {% endif %}
+),
+
+int_gtfs_schedule__incremental_trips AS (
+    -- select distinct because there is at least one feed that
```
atvaccaro, line 25: cut off?

atvaccaro (approved):

warehouse/models/intermediate/gtfs/int_gtfs_schedule__incremental_trips.sql
lauriemerrell: whoops

Issue #1904, opened 2022-10-14
V2 validations & guidelines tables


lauriemerrell:
Now that we have some foundational v2 modeling in place, we need to start a v2 version of `fact_daily_guidelines_checks`. 

To get that work going, we need to:
* Make staging and fact tables for the raw `validations_notices` 
* Initialize the guideline index framework
* Implement an initial check or two as we did in #1712
* Work with @owades to start figuring out how to aggregate from a GTFS dataset to an organization (this part will probably spawn a follow-up ticket)

PR #1945, opened 2022-11-01
Create v2 validation/outcomes models, new index, and an initial check

atvaccaro:
# Description

Progress towards https://github.com/cal-itp/data-infra/issues/1904. Note there are some test failures currently, in situations where a daily feed does not have a corresponding successful validation result. I am working on the best way to fill those gaps where we can.

This PR adds the following
1. Staging tables for validation notices and outcomes for Schedule and RT
2. A new daily notices `fct_daily_schedule_feed_validation_notices` built from just the Schedule notices for now
3. A new daily checks index `int_gtfs_quality__schedule_feed_guideline_index` just for feed-level checks
4. An example intermediate feed-level check implementation `int_gtfs_quality__no_schedule_validation_errors`
5. A new daily checks model `fct_daily_feed_guideline_checks` built on top of the above; contains checked at the _feed_ level only

Notes recommended by Laurie
* I'm moving the checks themselves into intermediate (`int_`) tables which follows dbt's best practices of staging tables being 1:1 with sources
* Changing how the "checks implemented" work; there are new "intended" checks and "actual" checks models that are compared against one another; these are actual models rather than CTEs
* Moving away from itp_id/url_number as core pieces of the index; instead, my current plan is to allow checks to be implemented at the appropriate grain, and then roll things up to an "assessment" grain such as the new quartet concept

The new checks process
1. Add the check to `stg_gtfs_quality__intended_checks`; this will cause tests on `fct_implemented_checks` to fail
2. Build a new intermediate table, for example `int_gtfs_quality__no_critical_rt_validation_errors` could be a good potential test case
3. Add the new table to the list of unions in `fct_daily_feed_guideline_checks` (I am assuming we can use a single fact table for both Schedule and RT here)
4. Tests should now pass on `fct_implemented_checks`

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
A full dbt run passes. Also `poetry run dbt test -m +mart.gtfs_quality` passes.


warehouse/models/staging/gtfs/_stg_gtfs.yml
```diff
@@ -902,33 +902,3 @@ models:
         description: '{{ doc("gtfs_stop_times__timepoint") }}'
     meta:
       publish.gis_coordinate_system_epsg: WGS84
-  - name: stg_gtfs_rt__service_alerts_outcomes
```
lauriemerrell, line 905: do you need to rebase or is this intentional, I don't see a replacement

atvaccaro:
pre-commit is still having problems, it does pass locally though.

```
➜  warehouse git:(v2-validations-models) ✗ pre-commit run --all-files
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check yaml...............................................................Passed
check for added large files..............................................Passed
flake8...................................................................Passed
black....................................................................Passed
bandit...................................................................Passed
sqlfluff-lint............................................................Passed
```

warehouse/models/staging/gtfs/gtfs_guidelines_v2/stg_gtfs_guidelines__checks_implemented.sql
```diff
@@ -0,0 +1,38 @@
+{{ config(materialized='ephemeral') }}
+WITH stg_gtfs_guidelines__checks_implemented AS (
+    SELECT {{ static_feed_downloaded_successfully() }} AS check, {{ compliance() }} AS feature
+    UNION ALL
+    SELECT {{ no_validation_errors() }}, {{ compliance() }}
```
lauriemerrell, line 5: Shouldn't this be the only check that is listed as implemented in v2?

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__feed_guideline_index.sql
```diff
@@ -8,43 +8,8 @@ gtfs_schedule_dim_feeds AS (
...
-    SELECT {{ schedule_feed_on_transitland() }}, {{ feed_aggregator_availability() }}
-    UNION ALL
-    SELECT {{ vehicle_positions_feed_on_transitland() }}, {{ feed_aggregator_availability() }}
-    UNION ALL
-    SELECT {{ trip_updates_feed_on_transitland() }}, {{ feed_aggregator_availability() }}
-    UNION ALL
-    SELECT {{ service_alerts_feed_on_transitland() }}, {{ feed_aggregator_availability() }}
-    UNION ALL
-    SELECT {{ include_tts() }}, {{ accurate_accessibility_data() }}
+    SELECT * FROM {{ ref('stg_gtfs_guidelines__checks_implemented') }}
```
lauriemerrell, line 12: Wait so this is you refactoring how v1 is done? I don't think we want to use the same `checks_implemented` table across both, so @owades can actually track what is implemented where

lauriemerrell (commented):
Mostly good, I am just unsure about committing to some of the organizational stuff. A few must-changes including that we should not be using the same `checks_implemented` across v1 and v2 (the whole point of that table is to track what is actually implemented.)

Can you update the PR description to more clearly identify the changes in precedent between v1 and v2 that we will be asking for from @owades? 

For example:
- Using intermediate instead of staging tables (change in philosophy on our part / trying to adhere to the more generic dbt project structure)
- Creating the `checks_implemented` table as a separate table rather than just CTE (though see above, we need to disambiguate between v1 and v2)
- We need to clarify what the granularity of the index is; is it just schedule feeds and if so how that might change.... We may need to discuss that tomorrow

warehouse/models/staging/gtfs/gtfs_guidelines_v2/stg_gtfs_guidelines__checks_implemented.sql
atvaccaro: Discussed offline, I am re-configuring how this works.

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__feed_guideline_index.sql
atvaccaro: Discussed offline, I am re-configuring how this works.

warehouse/models/staging/gtfs/gtfs_quality/stg_gtfs_rt__service_alerts_validation_notices.sql
```diff
@@ -0,0 +1,30 @@
+WITH raw_validation_notices AS (
+    SELECT * FROM {{ source('external_gtfs_rt', 'service_alerts_validation_notices') }}
+),
+
+stg_gtfs_rt__service_alerts_validation_notices AS (
+    SELECT
```
lauriemerrell, line 6: is there any chance that it's worth having a macro for these columns to ensure consistency between the RT tables? I think I did that for the raw outcomes, not sure our opinion stylistically/organizationally.

warehouse/models/staging/gtfs/gtfs_quality/stg_gtfs_schedule__validation_notices.sql
```diff
@@ -0,0 +1,22 @@
+WITH raw_validation_notices AS (
+    SELECT * FROM {{ source('external_gtfs_schedule', 'validation_notices') }}
+),
+
+stg_gtfs_schedule__validation_notices AS (
```
lauriemerrell, line 5: see overall comment re: unnesting the notices at some point... would that be an intermediate table on top of this...? or a separate table for sample notices? 🤔

lauriemerrell (commented):
Thank you for tackling 🙏 

A few inline things below and then meta comments:
1. Can you update the PR description to clarify what is schedule only vs. not / generally make sure it aligns with PR's final state, I think it's changed from what's there now
2. Can you add a section specifically enumerating what tables @owades will need to update / create as he implements a new check (AFAICT it's: add the macros if it's a check not implemented in v1; add to the ephemeral checks implemented table; and an `int_` table with the check's actual logic; add to the list of unioned tables in the actual fact table) -- I think just a little more meta-documentation would be helpful because this PR does enough other stuff to be potentially confusing as a model to follow.
3. It would be helpful to have a table as unnested as this: https://github.com/cal-itp/data-infra/blob/main/airflow/dags/gtfs_schedule_history2/validation_notices_load.sql. If not in this PR, would like to do a follow up on that.

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__schedule_validation_severities.sql
```diff
@@ -0,0 +1,19 @@
+{{ config(materialized='table') }}
+
+WITH validation_notices AS (
+    SELECT *
+    FROM {{ ref('stg_gtfs_schedule__validation_notices') }}
+),
+
+int_gtfs_quality__schedule_validation_severities AS (
+    -- TODO: if a future version of the validator changes a codes severity
+    -- we will end up with multiple entries for code (our primary key)
```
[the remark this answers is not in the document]
atvaccaro: I can make a TODO to make that change when we update to 4.0, it doesn't seem to be an option in 3.1.1 at least

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__no_schedule_validation_errors.sql
```diff
@@ -0,0 +1,35 @@
...
+
+int_gtfs_quality__no_schedule_validation_errors AS (
+    SELECT
+        idx.date,
+        idx.feed_key,
+        {{ no_validation_errors() }} AS check,
+        {{ compliance() }} AS feature,
+        CASE
+            WHEN sum_total_notices > 0 THEN "FAIL"
+            WHEN sum_total_notices = 0 THEN "PASS"
```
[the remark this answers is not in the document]
atvaccaro: I think it's really about 99.9% and I imagine it's primarily cases of invalid zips, etc. I do think a `null` is appropriate because we can't always know exactly why we are missing output for the record. I've gone back and forth about sentinel values and over time I've come to dislike them, though I will defer to @owades on the specifics here.

warehouse/models/mart/gtfs/fct_daily_all_feeds.sql
```diff
@@ -0,0 +1,40 @@
...
+),
+
+rt_feeds AS (
+    SELECT * FROM {{ ref('fct_daily_rt_feed_files') }}
+),
+
+unioned AS (
+    SELECT
+        date,
+        feed_key,
```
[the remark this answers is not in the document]
atvaccaro: Sure, let me rename.

warehouse/models/mart/gtfs/_mart_gtfs.yml
```diff
@@ -1240,3 +1240,9 @@ models:
           - relationships:
               to: ref('dim_schedule_feeds')
               field: key
+  - name: fct_daily_all_feeds
+    description: |
+      Unioned model of Schedule and RT daily feeds.
```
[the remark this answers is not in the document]
atvaccaro: Right, I think the concept of a daily RT feed is ambiguous and one of the reasons why I don't think the "daily" models are super appropriate for use outside of validations. 🤔

warehouse/models/mart/gtfs_quality/fct_daily_schedule_feed_validation_notices.sql
```diff
@@ -0,0 +1,50 @@
...
+
+dim_schedule_feeds AS (
+    SELECT * FROM {{ ref('dim_schedule_feeds') }}
+),
+
+validation_codes AS (
+    SELECT * FROM {{ ref('int_gtfs_quality__schedule_validation_severities') }}
+),
+
+-- TODO: this is just schedule, add RT
```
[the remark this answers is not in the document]
atvaccaro: ah yeah ty

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__schedule_validation_severities.sql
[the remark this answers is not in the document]
lauriemerrell: Pretty sure you can do it in 3.1.1, but TODO is ok, probably makes sense to do as part of upgrade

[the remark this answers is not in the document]
lauriemerrell: Pretty sure you can do it in 3.1.1, but TODO is ok, probably makes sense to do as part of upgrade

warehouse/models/mart/gtfs_quality/fct_daily_schedule_feed_validation_notices.sql
```diff
@@ -0,0 +1,50 @@
...
+        outcomes.validation_success,
+        outcomes.validation_exception,
+        COALESCE(
+            SUM(total_notices),
+            CASE WHEN validation_success THEN 0 END
+        ) AS total_notices,
+    FROM fct_daily_schedule_feeds AS daily_feeds
+    LEFT JOIN dim_schedule_feeds AS dim_feeds
+        ON daily_feeds.feed_key = dim_feeds.key
+    LEFT JOIN validation_outcomes AS outcomes
```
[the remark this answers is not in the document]
atvaccaro: ah yeah that's a good call

warehouse/models/staging/gtfs/gtfs_quality/stg_gtfs_rt__service_alerts_validation_notices.sql
atvaccaro: Generally I try to avoid un-parameterized abstractions, i.e. functions/whatever without arguments. More macros will also increase our dbt project compilation time which isn't too bad right now but can get annoying over time. I don't feel a huge need to macro these 3 tables right now, but maybe if we have more versions of the notice schema in the future?

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__schedule_validation_severities.sql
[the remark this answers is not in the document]
atvaccaro: Ah I'd gotten an error with my snapshot ```java -jar ~/gtfs-validator-3.1.1-cli.jar -n
Exception in thread "main" com.beust.jcommander.ParameterException: The following option is required: -o, --output_base
	at com.beust.jcommander.JCommander.validateOptions(JCommander.java:321)
	at com.beust.jcommander.JCommander.parse(JCommander.java:283)
	at com.beust.jcommander.JCommander.parse(JCommander.java:265)
	at org.mobilitydata.gtfsvalidator.cli.Main.parseArguments(Main.java:66)
	at org.mobilitydata.gtfsvalidator.cli.Main.main(Main.java:39)```

warehouse/models/staging/gtfs/gtfs_quality/stg_gtfs_rt__service_alerts_validation_notices.sql
lauriemerrell: I guess for me, the biggest headache in reviewing PRs that touch RT is trying to check if we actually updated all the locations. So that's why I'd like some kind of abstraction. I am ok to not have it but that's where I am coming from and why I did it for the outcomes.

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__schedule_validation_severities.sql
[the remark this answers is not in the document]
atvaccaro: Ah well it still writes out a JSON schema file. 🤷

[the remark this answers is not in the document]
atvaccaro: Beat me to it, just realized that. 😵‍💫

warehouse/models/mart/gtfs_quality/_mart_gtfs_quality.yml
```diff
@@ -0,0 +1,85 @@
...
+        tests:
+          - dbt_utils.not_null_proportion:
+              at_least: 0.99
+
+  - name: fct_daily_schedule_feed_validation_notices
+    description: |
+      A daily model of validation notices found per feed. Data
+      is filled in so there is always a count of notices per day
+      per feed, even if that count is zero; null counts indicate
+      a lack of validation for that feed on that day, though
```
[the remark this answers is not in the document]
atvaccaro: @owades do you want to interpolate as part of this initial implementation?

warehouse/models/staging/gtfs/gtfs_quality/stg_gtfs_rt__service_alerts_validation_notices.sql
atvaccaro: Fair enough, I will adjust

warehouse/models/staging/gtfs/gtfs_quality/stg_gtfs_schedule__validation_notices.sql
atvaccaro: I think an intermediate table, because it's changing grain.

warehouse/models/mart/gtfs_quality/fct_daily_schedule_feed_validation_notices.sql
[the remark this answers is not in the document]
atvaccaro: Actually this won't work if we don't have any notices to join, I will make a TODO to ensure that the eventual versioned severity table is used here.

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__no_schedule_validation_errors.sql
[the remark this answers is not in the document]
owades: My opinion is it depends on what the most common reasons for validator failures would be. If a common issue is invalid zips, then we could leave the result here as null, and catch the invalid zips in a different check. 

Although, my preference would be for the validator to run and catch the invalid zip as an error.

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__schedule_feed_guideline_index.sql
```diff
@@ -0,0 +1,16 @@
...
+-- we never want results from the current date, as data will be incomplete
+int_gtfs_quality__schedule_feed_guideline_index AS (
+    SELECT
+        feed_key,
+        date,
+    FROM fct_daily_schedule_feeds
+    WHERE date < CURRENT_DATE
+)
+
+SELECT * FROM int_gtfs_quality__schedule_feed_guideline_index
```
owades, line 16: Is "schedule feed guideline index" an appropriate name for this? Since this table has a row per feed per day, it seems like perhaps "feed daily index" would be more accurate. This table doesn't seem to care about guidelines.

warehouse/models/mart/gtfs/_mart_gtfs.yml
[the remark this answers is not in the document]
owades: I'm trying to understand this back and forth - I'm not sure I understand since we also have fact daily tables for schedule and RT feeds. Could you clarify what the purpose of this unioned table is?

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__schedule_feed_guideline_index.sql
atvaccaro: The upstream model contains dates going into the future, whereas for guidelines we cannot check anything earlier than the current day, which I don't think is technically true for the data itself. Even if this model doesn't contain guidelines/checks, it's an index for those checks rather than for general data work. Is that reasonable?

warehouse/models/mart/gtfs_quality/_mart_gtfs_quality.yml
```diff
@@ -0,0 +1,85 @@
+version: 2
+
+models:
+  # TODO: once we have checks that occur above the feed level (e.g. organization)
+  #   we will need to create additional daily check fact tables at the
+  #   appropriate grain
+  - name: fct_daily_feed_guideline_checks
```
owades, line 7: How would an RT feed check fit into here?

warehouse/models/mart/gtfs/_mart_gtfs.yml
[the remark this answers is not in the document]
atvaccaro: Yeah I think I am going to remove this; I had a use case in mind but it didn't up being relevant.

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__schedule_feed_guideline_index.sql
owades: Yeah that's reasonable, thanks for explaining

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__include_tts.sql
```diff
@@ -16,8 +16,8 @@ tts_issue_feeds AS (
        calitp_deleted_at
     FROM dim_stops
     -- When there is no tts_stop_name field for a given stop_name, or tts_stop_name is identical to stop_name, we proceed to run a few tests
-    WHERE (tts_stop_name IS null OR tts_stop_name = stop_name) AND
-         (
+    WHERE (tts_stop_name IS null OR tts_stop_name = stop_name)
```
owades, line 19: Thanks for these changes - I agree with them but just curious - are these the result of linting or a manual check?

owades:
This looks great, I left a few questions.

One thing from the old model that is missing is the equivalent to [guideline_checks.md](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/gtfs_guidelines/guideline_checks.md). I'm happy to add that later, but also want to give you the opportunity to add it since you might do it better than me.

warehouse/models/mart/gtfs_quality/_mart_gtfs_quality.yml
atvaccaro: RT could fit in but wouldn't have a feed_key right now, since we really only have a schedule feed dimension. Once we produce the quartet concept, rows should also key to that.

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__include_tts.sql
atvaccaro: Just linting!

atvaccaro:
> One thing from the old model that is missing is the equivalent to [guideline_checks.md](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/gtfs_guidelines/guideline_checks.md). I'm happy to add that later, but also want to give you the opportunity to add it since you might do it better than me.

Ah I will add that, since I'm asking for us to accept all these structural changes.

warehouse/models/mart/gtfs_quality/_mart_gtfs_quality.yml
lauriemerrell: So, as we discussed, this framework kind of only applies to schedule feeds right now, but it should give enough to start implementing schedule feed checks in a way that will be directly applicable even as we make the quartet/dataset logic more robust (i.e., migrations towards this intermediate state are not wasted work). I am aiming to get my ITP ID mapping doc out today or tomorrow and that would move us towards the quartet concept.

warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
```diff
@@ -0,0 +1,47 @@
...
+    description: |
+      Daily index of v2 schedule feeds. Provides structure for
+      individual check models to be joined together by day.
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - feed_key
+            - date
+
+  - name: int_gtfs_quality__schedule_validation_severities
```
lauriemerrell, line 14: Should there be an equal row count test between each implemented check and the feed guideline index? Part of the point of the old model was to ensure that we weren't getting fanout OR missing rows, which is why there was an equal row count between the index and the final fact table AND there was the absolute not_null.

lauriemerrell (commented):
Do we need a follow up ticket to make `int_unnested_sample_notices`?

warehouse/models/mart/gtfs_quality/_mart_gtfs_quality.yml
owades: Ok thank you, and sorry for making you repeat yourselves!

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__include_tts.sql
owades: I gotta start linting

atvaccaro:
> Do we need a follow up ticket to make `int_unnested_sample_notices`?

yep! https://github.com/cal-itp/data-infra/issues/1959

warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
atvaccaro: Yes good call, will add rowcounts.

lauriemerrell (approved):

PR #1947, opened 2022-11-01
Fct daily gtfs rt feeds

lauriemerrell:
# Description

For my ITP ID mapping work I realized I needed a way to map RT feeds to Airtable, and we don't yet have a fact daily RT feeds concept or anything on which to test that mapping. This PR creates a summary of daily RT parsing outcomes to give something that can be used to test joining with Airtable.

Questions for reviewers:
- How do we feel about hard coding a one month range for RT full refresh....? Or how do we want to handle that kind of thing? 
- How do handle the join with Schedule data. This PR proposes going through Airtable instead of using the URL set as metadata on the raw input file. 

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
`poetry run dbt run -s fct_daily_rt_feed_files`
`poetry run dbt test -s fct_daily_rt_feed_files`


warehouse/models/mart/gtfs/fct_daily_rt_feed_files.sql
```diff
@@ -0,0 +1,85 @@
...
+
+pivoted_parse_outcomes AS (
+    SELECT
+        dt,
+        base64_url,
+        feed_type,
+        parse_success_file_count,
+        parse_failure_file_count
+    FROM grouped_parse_outcomes
+    PIVOT(SUM(file_count) FOR aggregation_outcome IN ("parse_success_file_count", "parse_failure_file_count"))
```
atvaccaro, line 61: yay

warehouse/models/mart/gtfs/fct_daily_rt_feed_files.sql
```diff
@@ -0,0 +1,85 @@
...
+fct_daily_rt_feed_files AS (
+    SELECT
+        parse.dt as date,
+        {{ dbt_utils.surrogate_key(['parse.dt', 'parse.base64_url']) }} AS key,
+        parse.base64_url,
+        parse.feed_type,
+        parse.parse_success_file_count,
+        parse.parse_failure_file_count,
+        url_map.gtfs_dataset_key,
+        datasets.schedule_to_use_for_rt_validation_gtfs_dataset_key,
```
atvaccaro, line 73: can this column name be shortened in some way

atvaccaro (approved):
how big does this end up being? RT parse outcomes are big files but maybe it's smaller once only some columns are selected

warehouse/models/mart/gtfs/fct_daily_rt_feed_files.sql
lauriemerrell: I mean, that's the name in the source table.... Inclined to leave it for now....?

lauriemerrell:
Not sure what you mean by this:
> how big does this end up being? RT parse outcomes are big files but maybe it's smaller once only some columns are selected

The only thing that is actually materialized here is an aggregation, so it's tiny (one row per day per feed). 

image

warehouse/models/mart/gtfs/fct_daily_rt_feed_files.sql
atvaccaro: works for me, I'm just unsure how usable this will be downstream

lauriemerrell: I am tentatively planning to use it for my current little ITP ID investigation jaunt

PR #1951, opened 2022-11-02
check for expired services in calendar.txt files

owades:
# Description

This check looks for expired services for each given feed. This is to measure compliance with the GTFS Best Practice:

> Remove old services (expired calendars) from the feed.

If the end date/date of the oldest service in calendar.txt/calendar_dates.txt is older than the current date, the feed FAILS the check. Otherwise, if either is greater than the current date, it passes. If there is no value for end_date in calendar.txt or date in calendar_dates.txt, it returns a result of N/A. (The N/A result has occurred for 14 agencies total, and is not currently true for any feed).

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
I checked 5 passing checks, 5 failing checks, and 5 with N/A, and confirmed that the feeds conformed to the business logic.
I ran `dbt test`


warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__no_expired_services.sql
```diff
@@ -0,0 +1,55 @@
...
+        calitp_url_number,
+        calitp_agency_name,
+        feed_key,
+        check,
+        feature,
+        end_date,
+        CASE
+            WHEN end_date < date THEN "FAIL"
+            WHEN end_date >= date THEN "PASS"
+            -- Else clause captures cases where there is no calendar.txt file, or it is empty. This is OK, as some feeds rely on calendar_dates.txt instead
```
lauriemerrell, line 49: Just out of curiosity, would you plan to incorporate `calendar_dates` into this check in the future? Like `max(calendar_dates.date) >= current_date`? Not sure how specifically the best practice relates only to the `calendar` file

lauriemerrell (approved):
Two non-blocking questions/notes, defer to you on both

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__no_expired_services.sql
owades: I think you're right. I had been thinking of soliciting feedback from the team before adding calendar_dates, but I'm going to add it now.

warehouse/models/staging/gtfs_guidelines/_stg_gtfs_guidelines.yml
```diff
@@ -156,3 +156,20 @@ models:
...
         tests:
           - not_null
+  - name: stg_gtfs_guidelines__no_expired_services
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - date
+            - calitp_itp_id
+            - calitp_url_number
+    columns:
```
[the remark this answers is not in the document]
owades: Done!

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__no_expired_services.sql
```diff
@@ -0,0 +1,72 @@
...
+   GROUP BY 1, 2, 3, 4
+),
+
+stalest_calendar_dates_services AS (
+   SELECT
+       calitp_itp_id,
+       calitp_url_number,
+       calitp_extracted_at,
+       calitp_deleted_at,
+       MIN(date) AS date
```
lauriemerrell, line 23: This one may be more complicated actually.... I think you may need to group by service ID, get the max by service ID, and then get the min over those maxes? Because I think as written it will just get the first date with service defined, and it is ok if that is in the past?

lauriemerrell (changes requested):

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__no_expired_services.sql
lauriemerrell: Would be fine if you want to just merge as it was (calendar only) and then iterate more on this, didn't mean to scope creep you 😅

owades: It's all good - doesn't hurt to get it right while it's top of mind.

owades: Thanks for thinking this through, @lauriemerrell. Yes, as always you are right. I under-thought about how calendar_dates works, and so my simple rule was "if a date in the past appears, it's bad". Your way of thinking of it reflects the guideline much better.

owades: I updated this, although the more I think about it I think it could go either way. Going to keep these changes at least for now, though.

lauriemerrell: I think that the old way would give you a ton of false failures, because they would mean that any feed using `calendar_dates` for service would get marked as expired because the first date is almost always going to be in the past. 

Agree that this implementation may still be buggy.... For example, for feeds that use *both* `calendar` and `calendar_dates`, this may cause them to fail the `calendar_dates` check for example if a holiday exception is defined for Labor Day on a service that's in effect from August to December. That service is not expired in `calendar` (ends in December) but is in `calendar_dates` (because only Labor Day in September will be listed). 

So I think getting max date by service ID across both may be even more robust (i.e., outer join your two checks and get the max date that appears in either table by service ID )? But defer to you on how far to go with this; I just suspect you may see still unduly high failure rates. I'll approve and defer to you.

lauriemerrell (approved):

owades:
Latest update passes tests and I think it's good. I still need to perform some manual checking, which I expect to do on MOnday

owades:
I checked edge cases in which a feed has calendar but not calendar_dates, and vice versa, and confirmed that this table worked as expected. Would you mind taking one more look, @lauriemerrell ?

lauriemerrell (approved):
Looks good to me! Thanks for muddling through on this 😓

PR #1953, opened 2022-11-03
update tech contact check to get rid of nulls and dedupe

owades:
# Description

I found two issues in `stg_gtfs_guidelines__technical_contact_listed.sql` that were impacting the downstream fact table:
- Because a dedupe wasn't performed, a fanout situation was created
- Because the check was defined prior to the join with `feed_guideline_index`, there existed unintentional nulll values in the status column

Note that this is one of the checks I built before I started including tests for each check.
## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
I spot checked several passing and failing values, and queried to make sure there are no longer null values being created.
I added a dbt test for this table, confirmed that it failed prior to the fix, and passed after the fix.
I wrote a query to compare the features statuses between the staging table on my schema and the main staging table, and found that the they were always equal.

### DIFF
--- a/warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__technical_contact_listed.sql
+++ b/warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__technical_contact_listed.sql
@@ -17,29 +17,16 @@ daily_technical_contact_check AS (
         t1.check,
         t1.feature,
         CASE
-            WHEN t2.feed_contact_email IS NOT null THEN "PASS"
+            WHEN LOGICAL_AND(t2.feed_contact_email IS NOT NULL) THEN "PASS"
             ELSE "FAIL"
-        END AS status,
-        COUNT(*)
+        END AS status
     FROM feed_guideline_index AS t1
     LEFT JOIN feed_info_clean AS t2
        ON t1.calitp_itp_id = t2.calitp_itp_id
       AND t1.calitp_url_number = t2.calitp_url_number
       AND t1.date >= t2.calitp_extracted_at
       AND t1.date < t2.calitp_deleted_at
-    GROUP BY 1,2,3,4,5,6,7
-),
-
-daily_technical_contact_check_dedupe AS (
-    SELECT
-        date,
-        calitp_itp_id,
-        calitp_url_number,
-        calitp_agency_name,
-        check,
-        feature,
-        status
-    FROM daily_technical_contact_check
+    GROUP BY 1,2,3,4,5,6
 )
 
-SELECT * FROM daily_technical_contact_check_dedupe
+SELECT * FROM daily_technical_contact_check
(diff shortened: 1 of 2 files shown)


owades:
@atvaccaro, are you able to advise on why the lint check is failing? i've been experiencing this off and on on a few different PRs this week

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__technical_contact_listed.sql
```diff
@@ -7,35 +7,39 @@ feed_info_clean AS (
...
-    LEFT JOIN feed_info_clean_check AS t2
+    LEFT JOIN feed_info_clean AS t2
        ON t1.calitp_itp_id = t2.calitp_itp_id
       AND t1.calitp_url_number = t2.calitp_url_number
       AND t1.date >= t2.calitp_extracted_at
       AND t1.date < t2.calitp_deleted_at
+    GROUP BY 1,2,3,4,5,6,7
+),
+
+daily_technical_contact_check_dedupe AS (
```
lauriemerrell, line 33: I am just wondering about calling this "dedupe" since it seems like a simple select statement -- is this intended to be a `select distinct`?

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__technical_contact_listed.sql
```diff
@@ -7,35 +7,39 @@ feed_info_clean AS (
...
     SELECT
         t1.date,
         t1.calitp_itp_id,
         t1.calitp_url_number,
         t1.calitp_agency_name,
         t1.check,
         t1.feature,
-        t2.status
+        CASE
+            WHEN t2.feed_contact_email IS NOT null THEN "PASS"
```
lauriemerrell, line 20: Because we have had some trouble with a two-row `feed_info` in the past, I think you could potentially do a `LOGICAL_AND(t2.feed_contact_email IS NOT NULL)` in the group by to make sure that you are actually aggregating within an individual instance of a file, if that makes sense?

lauriemerrell (commented):
Thanks for digging into the failures. Just two clarifications, not sure I am following the adjustment entirely.

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__technical_contact_listed.sql
owades: You're right, the language is weird there. The deduping is done by the COUNT(*) in the previous subquery, and so the subquery titled "...dedupe" does nothing other than (arguably) neaten it up. i'm just going to remove it.

warehouse/models/staging/gtfs_guidelines/stg_gtfs_guidelines__technical_contact_listed.sql
owades: wow, LOGCAL_AND was the function I didn't know I needed

lauriemerrell (approved):

owades:
@atvaccaro ready for you to force-merge

PR #1955, opened 2022-11-07
Stop times keying and incremental

lauriemerrell:
# Description

Progress towards https://github.com/cal-itp/data-infra/issues/1761, specifically a more useful trip summary table and thus reports trips/stops/routes table. 

Creates an `int_gtfs_schedule__incremental_stop_times` table to make stop times more performant. Updates `dim_stop_times` to consume this incremental table and updates `stg_gtfs_schedule__stop_times` to facilitate incremental table creation by adding the date column.

## Type of change

- [x] New feature

## How has this been tested?
`dbt run` and `dbt test`


warehouse/models/intermediate/gtfs/_int_gtfs.yaml
```diff
@@ -149,3 +149,50 @@ models:
         description: Service identifier from calendar and/or calendar_dates.
         tests:
           - not_null
+  - name: int_gtfs_schedule__incremental_trips
```
lauriemerrell, line 152: table name needs to be changed to `stop_times` here

warehouse/models/mart/gtfs/dim_stop_times.sql
```diff
@@ -31,9 +43,13 @@ dim_stop_times AS (
         continuous_drop_off,
         shape_dist_traveled,
         timepoint,
+        COALESCE(warning_duplicate_primary_key, FALSE) AS warning_duplicate_primary_key,
```
atvaccaro, line 46: nit: I think you can do this with a window function and without the CTE?

using produce from https://cloud.google.com/bigquery/docs/reference/standard-sql/window-function-calls#produce_table

```
WITH Produce AS
         (SELECT 'kale' as item, 23 as purchases, 'vegetable' as category
          UNION ALL
          SELECT 'banana', 2, 'fruit'
          UNION ALL
          SELECT 'cabbage', 9, 'vegetable'
          UNION ALL
          SELECT 'apple', 8, 'fruit'
          UNION ALL
          SELECT 'leek', 2, 'vegetable'
          UNION ALL
          SELECT 'lettuce', 10, 'vegetable')
SELECT *
     , COUNT(*) OVER (PARTITION BY category) > 1 AS warning_duplicate_primary_key
FROM Produce;

```

atvaccaro (approved):

atvaccaro (approved):
thanks!

PR #1956, opened 2022-11-07
add v2 mart_gtfs to metabase sync

lauriemerrell:
# Description

Want to sync v2 GTFS data to Metabase systematically (rather than just add manually via Metabase admin UI).

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Not sure if/how to test, @atvaccaro open to suggestions


atvaccaro (approved):
technically you could test locally with `poetry run dbt-metabase` and your metabase credentials but it should be fine

PR #1960, opened 2022-11-08
2 new v2 checks for feed expiration date

owades:
# Description

Two new checks:

- Fail if validator warning `feed_expiration_date_30_days` appears, pass if not
- Fail if validator warning `feed_expiration_date_37_days` appears, pass if not

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
` ./dbt.sh test --select int_gtfs_quality__no_30_day_feed_expiration+`
` ./dbt.sh test --select int_gtfs_quality__no_7_day_feed_expiration+`

I looked at aggregate output for both, and saw that vast majority pass, some fail, and very few have a null value (meaning there wasn't a row for that day in the validator output table).


lauriemerrell (changes requested):
One typo.

Structurally, I have one issue (which is on data services to figure out) which is that these feed notices were [only introduced in version 4.0.0](https://github.com/MobilityData/gtfs-validator/releases/tag/v4.0.0) of the validator, which we've only been running since #1973 merged 3 days ago.

So "the validator ran successfully and didn't return these validations" is only actually meaningful for the last 3 days, but I don't think our validation notices table takes account of that; basically there should be blanks for this specific notice for all feeds before 3 days ago, I think. I have recorded this structural issue in #1994, which I think @atvaccaro can maybe address when he's back next week.

I am ok to merge this PR for now and make fixes on our side, but just wanted you to be aware of this limitation and the fact that we will probably do a fix that will change the behavior of this check and result in a lot more nulls.

owades:
Thanks @lauriemerrell . Fixed the typo.

And your comment makes sense - feeds are passing in the past, when they should be null. This isn't a huge issue (would be worse if they were failing) but it's inaccurate and makes sense to fix.

lauriemerrell (approved):

PR #1968, opened 2022-11-10
migrate check for existence of shapes.txt

owades:
# Description

Check will PASS if the feed downloaded & unzipped successfully and shapes.txt was present
Check will FAIL if the feed downloaded & unzipped successfully and shapes.txt was NOT present
Check will return a null value if feed downloaded OR unzipped UNsuccessfully

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
I reviewed the output and confirmed that 5 passing feeds truly had a shapes file.
Only one feed failed - City of Wasco.


owades:
@lauriemerrell this check revealed that only one feed is missing shapes.txt - does this check out with you?

owades:
I added tests for this, and for this check in particular lowered the allowable null threshold to 80%. The idea is that more nulls are expected because a null result has meaning.

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__shapes_file_present.sql
```diff
@@ -0,0 +1,37 @@
+WITH feed_guideline_index AS (
+    SELECT * FROM {{ ref('int_gtfs_quality__schedule_feed_guideline_index') }}
+),
+
+daily_feed_shapes_present AS (
+    SELECT feed_key,
+           EXTRACT(date FROM ts) AS date,
+           -- zipfile_files is an array which may contain "shapes.txt" but may also contain a string like "google_transit/shapes.txt".
+           -- this long EXISTS statement solves for both
+           CASE WHEN EXISTS(SELECT * FROM UNNEST(zipfile_files) AS x WHERE LOWER(x) LIKE '%shapes.txt') THEN true
```
lauriemerrell, line 10: There is also the parse outcomes table: `stg_gtfs_schedule__file_parse_outcomes`, which would record if we had a totally corrupted CSV (and it has the benefit of being already "unnested" so that each individual file has its own row). So I would maybe consider using that instead, so you can check all three outcomes: download, unzip, and parse?

Also, instead of checking `LIKE %shapes.txt` I think it would be safer to just do like `TRIM(LOWER(x)) = 'shapes.txt'` because it is possible that feeds could have wacky other files in them. Unlikely but just in case. And even `TRIM` and `LOWER` gives them the benefit of the doubt if they have spaces or capital letters in their filenames, maybe we don't even want to do that.

warehouse/models/mart/gtfs_quality/fct_daily_feed_guideline_checks.md
```diff
@@ -12,4 +12,6 @@ Here is a list of currently-implemented checks:
 | ------------------------------------ |---------|------------ |
 | No errors in MobilityData GTFS Schedule Validator |Compliance |GTFS Schedule Validator produced no errors for the transit provider’s static feed. |
 |No shapes-related errors appear in the MobilityData GTFS Validator | Accurate Service Data | None of the following shapes-related errors appear in the GTFS Schedule Validator: decreasing_shape_distance, equal_shape_distance_diff_coordinates, decreasing_or_equal_shape_distance, decreasing_or_equal_shape_distance |
+|Shapes.txt file is present | Accurate Service Data | Static GTFS feed contains the file shapes.txt. This check will return a null value if the feed failed to download or unzip|
```
lauriemerrell, line 15: if you migrate to checking "parse" too, may want to update this

lauriemerrell (changes requested):
Few inline comments

> this check revealed that only one feed is missing shapes.txt - does this check out with you?

Yes, I double checked and that seems right. Also anecdotally I think folks have mentioned that CA has very good shapes coverage. 👍

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__shapes_file_present.sql
```diff
@@ -0,0 +1,37 @@
...
+                ELSE false
+           END AS has_shapes
+      FROM `cal-itp-data-infra-staging`.`owades_mart_gtfs`.`fct_schedule_feed_downloads`
```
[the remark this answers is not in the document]
owades: ahh oops!! thanks for catching

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__shapes_file_present.sql
owades: parse outcomes looks intriguing, I will check it out.
What I was solving for with `LIKE %shapes.txt` is situations where the filename was something like `google_transit/shapes.txt`, which is pretty common in the table I'm using

lauriemerrell: Ah yup I forgot about that -- I definitely think that https://dbt-docs.calitp.org/#!/model/model.calitp_warehouse.stg_gtfs_schedule__file_parse_outcomes is gonna be a better bet, because it has `gtfs_filename` already populated

owades:
Could you take a look, @atvaccaro? I did add a line break to the end of a yaml file that isn't mine, in order to past the linter

atvaccaro:
> Could you take a look, @atvaccaro? I did add a line break to the end of a yaml file that isn't mine, in order to past the linter

Yep thanks, I had fixed this on main I think if you want to rebase.

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__shapes_file_present.sql
```diff
@@ -0,0 +1,32 @@
...
+
+keyed_parse_outcomes AS (
+    SELECT * FROM {{ ref('int_gtfs_schedule__keyed_parse_outcomes')}}
+),
+
+daily_feed_shapes_present AS (
+    SELECT feed_key
+      FROM keyed_parse_outcomes
+     WHERE parse_success
+       AND filename = 'shapes.txt'
```
lauriemerrell, line 13: I think for this, it might be safer to use `gtfs_filename = shapes` (`gtfs_filename` is which GTFS file we have decided the data is. So if there was a feed that had a file called ` shapes.txt` with a space and we decided to include that as valid shapes data, it would have `gtfs_filename` = `shapes` even if `filename` had a different value.)

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__shapes_file_present.sql
```diff
@@ -0,0 +1,32 @@
...
+
+int_gtfs_quality__shapes_file_present AS (
+    SELECT
+        idx.date,
+        idx.feed_key,
+        {{ shapes_file_present() }} AS check,
+        {{ accurate_service_data() }} AS feature,
+        CASE
+            WHEN s.feed_key IS NOT null THEN "PASS"
+            ELSE "FAIL"
```
lauriemerrell, line 25: Not sure if you want to leave a null option for feeds that failed to download? (Would need to go back to joining in the overall feed download outcome to check for "feed present but doesn't have shapes" vs. "feed not present").

lauriemerrell (commented):
One suggestion one question

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__shapes_file_present.sql
owades: Thank you! I saw there were a few options here and was hoping you would guide me to the right one

owades: I mean, I could have also asked...

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__shapes_file_present.sql
lauriemerrell: In retrospect I may rescind this proposal. I think need two grains:

* Checks assessed against feeds that actually downloaded
* Checks assessed against URLs where download was attempted (this is basically just the "did it download" check) 

But I think maybe we want to just assume that the universe for any check that depends on a feed existing is "feeds that exist" and we can handle the non-existent case in a later join where we assemble the larger universe. If that makes sense

owades: I don't entirely understand but I'm happy to follow your advice for now. It might make sense for us to meet and discuss, especially since so many questions are coming up in migration.

lauriemerrell (approved):

PR #1969, opened 2022-11-11
migrate check for shapes-for-all-trips

owades:
# Description

Migrate existing check

Resolves # [issue]

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
I looked through several examples of PASS and FAIL to ensure the feeds looked as expected.
I created and ran a test on this int table.

### DIFF
--- a/warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
+++ b/warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
@@ -23,29 +23,24 @@ models:
             - severity
 
   - name: int_gtfs_quality__no_schedule_validation_errors
-    tests:
+    tests: &stg_gtfs_guideline_tests
       - dbt_utils.unique_combination_of_columns:
           combination_of_columns:
             - date
             - feed_key
       - dbt_utils.equal_rowcount:
           compare_model: ref('int_gtfs_quality__schedule_feed_guideline_index')
-    columns:
+    columns: &stg_gtfs_guideline_columns
       - name: status
         tests:
           - dbt_utils.not_null_proportion:
               at_least: 0.99
 
   - name: int_gtfs_quality__shapes_valid
-    tests:
-      - dbt_utils.unique_combination_of_columns:
-          combination_of_columns:
-            - date
-            - feed_key
-      - dbt_utils.equal_rowcount:
-          compare_model: ref('int_gtfs_quality__schedule_feed_guideline_index')
-    columns:
-      - name: status
-        tests:
-          - dbt_utils.not_null_proportion:
-              at_least: 0.99
+    tests: *stg_gtfs_guideline_tests
+    columns: *stg_gtfs_guideline_columns
+
+  - name: int_gtfs_quality__shapes_for_all_trips
+    tests: *stg_gtfs_guideline_tests
+    columns: *stg_gtfs_guideline_columns
+
--- a/warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
+++ b/warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
@@ -43,4 +43,3 @@ models:
   - name: int_gtfs_quality__shapes_for_all_trips
     tests: *stg_gtfs_guideline_tests
     columns: *stg_gtfs_guideline_columns
-
--- a/warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__shapes_for_all_trips.sql
+++ b/warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__shapes_for_all_trips.sql
@@ -1,0 +1,49 @@
+WITH feed_guideline_index AS (
+    SELECT * FROM {{ ref('int_gtfs_quality__schedule_feed_guideline_index') }}
+),
+
+dim_trips AS (
+    SELECT * FROM {{ ref('dim_trips') }}
+),
+
+summarize_trips AS (
+   SELECT
+       feed_key,
+       EXTRACT(date FROM _valid_from) AS valid_from,
+       EXTRACT(date FROM _valid_to) AS valid_to,
+       COUNTIF(shape_id IS NOT NULL) AS ct_shape_trips,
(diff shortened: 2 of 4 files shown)


warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
```diff
@@ -23,29 +23,23 @@ models:
...
-    tests:
+    tests: &stg_gtfs_guideline_tests
       - dbt_utils.unique_combination_of_columns:
           combination_of_columns:
             - date
             - feed_key
       - dbt_utils.equal_rowcount:
           compare_model: ref('int_gtfs_quality__schedule_feed_guideline_index')
-    columns:
+    columns: &stg_gtfs_guideline_columns
```
lauriemerrell, line 33: I think you made this change in your other new shape branch too, just flagging in case the merges get weird

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__shapes_for_all_trips.sql
```diff
@@ -0,0 +1,49 @@
...
+daily_trips AS (
+  SELECT
+    t1.date,
+    t1.feed_key,
+    SUM(t2.ct_shape_trips) AS tot_shape_trips,
+    SUM(t2.ct_trips) AS tot_trips
+  FROM feed_guideline_index AS t1
+  LEFT JOIN summarize_trips AS t2
+       ON t1.date >= t2.valid_from
+       AND t1.date <= t2.valid_to
```
lauriemerrell, line 29: Wondering if we want this to be a strict inequality, I think that's what we've generally done in v1

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__shapes_for_all_trips.sql
```diff
@@ -0,0 +1,49 @@
...
+dim_trips AS (
+    SELECT * FROM {{ ref('dim_trips') }}
+),
+
+summarize_trips AS (
+   SELECT
+       feed_key,
+       EXTRACT(date FROM _valid_from) AS valid_from,
+       EXTRACT(date FROM _valid_to) AS valid_to,
+       COUNTIF(shape_id IS NOT NULL) AS ct_shape_trips,
```
lauriemerrell, line 14: I am not sure if you want to do the extra step of checking if that shape ID actually can join with the shapes table, to check if they're all actually valid foreign keys? I suppose there is a separate check for dangling refs so maybe it doesn't matter, just wanted to ask.

lauriemerrell (approved):
Optional comments

warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
owades: yeah - in hindsight I should have done just one branch for all 3 changes. merging will be fun

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__shapes_for_all_trips.sql
owades: i was hoping you'd look at this - to me it's a question of how valid_to is designed. totally your call!

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__shapes_for_all_trips.sql
owades: this is tempting, I think I could argue either way on it

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__shapes_for_all_trips.sql
lauriemerrell: Oh, I was forgetting how the changes we made are supposed to work. I think the new guidance would just be to aggregate to the `feed_key` level and then left join onto the date index and ditch the date logic in the join entirely. Because we've done the ranged joins to construct the feed check date index already. (`valid_to` is now a timestamp and it's no longer a date-level operation, which is why the `feed_daily` table checks for what was in effect at noon Pacific and then calls that the feed for the whole day.)

lauriemerrell: I hope that makes sense

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__shapes_for_all_trips.sql
lauriemerrell: Totally defer to you, I don't really think one is clearly "better". 

If the intent is "they've listed a shape for every trip", I think leaving it as-is is probably simplest and has exactly one failure case (they didn't list a shape). Once you do the join you have to think about fanout and issues in the shapes table.

Doing the join would be more complete/comprehensive in terms of "there is an actual valid shape for every trip". But you might have to change the logic to check that there is exactly one match. And then you have multiple failure conditions.

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__shapes_for_all_trips.sql
owades: I think so! I'll just keep the join as-is, except that I'll remove date constraints entirely. Basically, feed_key gets me the date info I need. Is that right?

lauriemerrell: Yes. I am sorry for writing a paragraph without that as the exact TLDR 😅 

Also means I think you can drop the valid to and from columns from the select above but that is less important

owades:
@lauriemerrell can you take a quick look at changes? I ran tests and this passed.

lauriemerrell (approved):
🎉

PR #1974, opened 2022-11-14
V2 dim shapes geo

lauriemerrell:
# Description

Creates `dim_shapes_geo` in v2 warehouse, one of the key outstanding dependencies before we can migrate speedmaps.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

`poetry run dbt run -s int_gtfs_schedule__incremental_shapes dim_shapes dim_shapes_geo` and 
`poetry run dbt test -s int_gtfs_schedule__incremental_shapes dim_shapes dim_shapes_geo`

Also manually examined an example shape between v1 and v2 and made sure the first few coordinates were the same

### DIFF
--- a/warehouse/models/intermediate/gtfs/_int_gtfs.yaml
+++ b/warehouse/models/intermediate/gtfs/_int_gtfs.yaml
@@ -187,6 +187,7 @@ models:
             - shape_pt_sequence
     columns:
       - name: base64_url
+      - name: ts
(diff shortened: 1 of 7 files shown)


atvaccaro (approved):
are we still happy with `dim_shapes_geo` as a model name?

PR #1975, opened 2022-11-14
add shape array foreign key to scheduled trips

lauriemerrell:
# Description

Meant to include this in #1974 but forgot. Just adding the foreign key to the assembled shape geometry to the fact daily trip entry. 

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

`poetry run dbt run -s fct_daily_scheduled_trips` and `poetry run dbt test -s fct_daily_scheduled_trips`

### DIFF
--- a/warehouse/models/mart/gtfs/_mart_gtfs.yml
+++ b/warehouse/models/mart/gtfs/_mart_gtfs.yml
@@ -1204,6 +1204,12 @@ models:
           - relationships:
               to: ref('dim_routes')
               field: key
+      - name: shape_array_key
+        description: Foreign key to dim_shapes_arrays.
+        tests:
+          - relationships:
+              to: ref('dim_shapes_arrays')
+              field: key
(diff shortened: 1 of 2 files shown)


warehouse/models/mart/gtfs/_mart_gtfs.yml
```diff
@@ -1204,6 +1204,12 @@ models:
           - relationships:
               to: ref('dim_routes')
               field: key
+      - name: shape_array_key
```
atvaccaro, line 1207: should this be `shapes_array` key?

atvaccaro (approved):

warehouse/models/mart/gtfs/_mart_gtfs.yml
lauriemerrell: confirmed offline we are ok

PR #1976, opened 2022-11-15
make keyed parse outcomes intermediate table

lauriemerrell:
# Description

@owades asked how to join `stg_gtfs_schedule__file_parse_outcomes` to implement my proposal in #1968 and I realized that there isn't really a great way to do this. This PR creates a new intermediate table with individual parse outcomes joined with feed_key. PR also adds documentation for the existing table because it wasn't documented.

Questions for @atvaccaro:
- `dim_schedule_feeds` depends on parse outcomes via `joined_feed_outcomes` already, so this is creating a weird dependency graph.... How worried are we about that? I think at some point we will want to push a feed versioning table into the intermediate and have stuff inherit from that, because if we want to add any actual attributes to `dim_schedule_feeds` (like calendar date ranges etc.) that would require a circular dependency with what we have now, I think.
- Should this be in the mart? Maybe only the rows where `feed_key` is populated?
- This also made me realize that if we have a genuine flake where a file isn't parsed it's possible we won't do anything about that? I just checked and so far we don't have any rows where parsing failed. LMK if we should write an issue for this.

Resolves # [issue]

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

`poetry run dbt run -s int_gtfs_schedule__keyed_parse_outcomes` and `poetry run dbt test -s int_gtfs_schedule__keyed_parse_outcomes`

### DIFF
--- a/warehouse/models/intermediate/gtfs/_int_gtfs.yaml
+++ b/warehouse/models/intermediate/gtfs/_int_gtfs.yaml
@@ -216,3 +216,36 @@ models:
         - not_null
       - name: shape_dist_traveled
         description: '{{ doc("gtfs_shapes__shape_dist_traveled") }}'
+  - name: int_gtfs_schedule__keyed_parse_outcomes
+    description: |
+      All GTFS schedule file parse outcomes, with `feed_key` identifier joined
+      on to facilitate downstream use.
+    tests:
+      - dbt_utils.equal_rowcount:
+          compare_model: ref('stg_gtfs_schedule__file_parse_outcomes')
+    columns:
+      - name: feed_key
+        description: |
+          Foreign key to `dim_schedule_feeds`.
+        tests:
+          - relationships:
+              to: ref('dim_schedule_feeds')
+              field: key
+      - name: parse_success
+        description: '{{ doc("column_schedule_parse_success") }}'
+      - name: parse_exception
+        description: '{{ doc("column_schedule_parse_exception") }}'
+      - name: filename
+        description: '{{ doc("column_schedule_parse_filename") }}'
+      - name: _config_extract_ts
+      - name: feed_name
+        description: '{{ doc("column_schedule_parse_feed_name") }}'
+      - name: feed_url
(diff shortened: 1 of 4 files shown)


warehouse/models/intermediate/gtfs/_int_gtfs.yaml
```diff
@@ -216,3 +216,38 @@ models:
         - not_null
       - name: shape_dist_traveled
         description: '{{ doc("gtfs_shapes__shape_dist_traveled") }}'
+  - name: int_gtfs_schedule__keyed_parse_outcomes
```
atvaccaro, line 219: should this just be a fact table?

atvaccaro (approved):
lgtm!

atvaccaro:
> * `dim_schedule_feeds` depends on parse outcomes via `joined_feed_outcomes` already, so this is creating a weird dependency graph.... How worried are we about that? I think at some point we will want to push a feed versioning table into the intermediate and have stuff inherit from that, because if we want to add any actual attributes to `dim_schedule_feeds` (like calendar date ranges etc.) that would require a circular dependency with what we have now, I think.

Eh maybe not ideal but probably fine; graphs do end up weird when we join just to get FKs.

> * Should this be in the mart? Maybe only the rows where `feed_key` is populated?

I think so, to either.

> * This also made me realize that if we have a genuine flake where a file isn't parsed it's possible we won't do anything about that? I just checked and so far we don't have any rows where parsing failed. LMK if we should write an issue for this.

What does parsing mean here?

lauriemerrell:
> What does parsing mean here? I think for outcomes we should always have a row, right? Ideally our Airflow job will fail if we don't end up with 1 outcome per 1 input.

We will have an outcome row but if the file is not actually converted we'll be missing that data from that feed version until a new zipfile appears, see #1982 that I just wrote. This would only be an issue if it was a flake where the file was valid but somehow the parsing job failed and we didn't re-run it.

PR #1991, opened 2022-11-16
add text to speech check

owades:
# Description

tts check migration

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
`./dbt.sh test --select int_gtfs_quality__include_tts+`
looked at aggregate results to confirm that most fail, some pass=


owades:
@lauriemerrell using feed_key rather than dates allowed me to remove a deduping step...yay!

edasmalchi:
curious where the "no-no" words come from? Might be too regional, but I wonder if it would make any sense to add LA. It's usually La, as in La Brea, La Cienega, etc, but sometimes it's supposed to be the city, L.A. 

might be a tough check to add, just thinking about it since Transit app currently has me on a bus to "Downtown La"

lauriemerrell (approved):
LGTM, defer to you on business logic / @edasmalchi's question above

edasmalchi:
to be clear my comment is just a thought, don't mean to imply that it should be addressed in this PR

owades:
Thanks for this comment, @edasmalchi! And thanks for using "no-no words", the proper technical terminology 😉 

I used [this](https://docs.google.com/document/d/1LObjgDyiiE6UBiA3GpoNOlZ36li-KKj6dwBzRTDa7VU/edit#heading=h.puwtf1nwatif) as a primary reference (I thought I had included as a comment, but alas. It's definitely intended to be a growing list, and it makes sense to have a list that is  regionally accurate! I appreciate you looking at this and commenting.

I still need to run this check by teammates for business logic, and I'll include you on those comms. And i'll add the "la" as well.

owades:
I wanted to close the loop on this, @edasmalchi. I checked all stop names from all feeds, and it actually looks to me like LA isn't often used to represent "Los Angeles".

Here is the list - this was an educational exercise for me:
[bquxjob_22e25827_1855f5ffe50.csv](https://github.com/cal-itp/data-infra/files/10355814/bquxjob_22e25827_1855f5ffe50.csv)

Issue #1989, opened 2022-11-16
Bug: Metrolink download failing in both v1 and v2 pipelines even though it works in browser


lauriemerrell:
**Describe the bug**
We are getting `403` in both the v1  and v2 pipelines for Metrolink: `https://www.metrolinktrains.com/globalassets/about/gtfs/gtfs.zip`. However I am able to access the file via browser. 

I think it's an issue with `User-Agent`, but I can't figure out exactly what is happening. 

If I do this: 

```
r = requests.get(URL, headers = {
        "User-Agent": (
            "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_4)"
            "AppleWebKit/537.36 (KHTML, like Gecko)"
            "Chrome/83.0.4103.97 Safari/537.36"
        )
    })
```

In Jupyter, it succeeds (these header values are [what we specify in v1 prod](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/gtfs_downloader/download_data.py#L44-L50)). 

But when I run the [`download_url` function](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/gtfs_downloader/download_data.py#L38) with a breakpoint, verifying that this header is set, I still get a `403`. 😢 

I can't figure out why. 

Here's `r.request.headers` from my successful request: `{'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_4)AppleWebKit/537.36 (KHTML, like Gecko)Chrome/83.0.4103.97 Safari/537.36', 'Accept-Encoding': 'gzip, deflate, br', 'Accept': '*/*', 'Connection': 'keep-alive'}`.

Here's the request headers for the unsuccessful request using `download_url`: `{'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_4)AppleWebKit/537.36 (KHTML, like Gecko)Chrome/83.0.4103.97 Safari/537.36', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'Connection': 'keep-alive'}`

I checked using `Accept-Encoding` = `'gzip, deflate'` locally and that still worked, and that's the only difference I see between the successful and unsuccessful requests.

atvaccaro:
[...]

PR #1992, opened 2022-11-17
Dim schedule feeds better handle unzip

### LINKED ISSUE #1985: Bug: `dim_schedule_feeds` deeply counterintuitive behavior for cases where download succeeds but unzip fails
**Describe the bug**
Currently, `dim_schedule_feeds` is recording cases where we successfully downloaded a feed, regardless of whether the unzip operation was successful. Issue is here: https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/gtfs/dim_schedule_feeds.sql#L65. We should be also checking for `unzip_success` I think. 

Specific weird case identified by @owades in #1977 is Tulare (`base64_url = 'aHR0cHM6Ly90dWxhcmVjb2cub3JnL3RjYWcvZGF0YS1naXMtbW9kZWxpbmcvZ3Rmcy1kYXRhL3R1bGFyZS1jb3VudHktcmVnaW9uYWwtdHJhbnNpdC1hZ2VuY3ktdGNydGEtZ3Rmcw=='`). On November 6 and 7, it downloaded successfully but was not a parseable zipfile. 

Given that the parseable zipfiles on November 5 and 8 are identical, I think (?) we would want to consider that one continuous feed version and ignore the blips on the 6th and 7th.

However either way, I don't think "we downloaded data that we can't parse" belongs in `dim_schedule_feeds`; I think the intention for `dim_schedule_feeds` was "this is data that is actually available in the warehouse somewhere." So we should fix this so that:

* The unparseable zipfiles do not appear in `dim_schedule_feeds`
* For Tulare we have one continuous version across the Nov 6 and 7 blips 

Query to observe the downloads is:

```
SELECT *
FROM `cal-itp-data-infra.mart_gtfs.fct_schedule_feed_downloads`
WHERE base64_url = 'aHR0cHM6Ly90dWxhcmVjb2cub3JnL3RjYWcvZGF0YS1naXMtbW9kZWxpbmcvZ3Rmcy1kYXRhL3R1bGFyZS1jb3VudHktcmVnaW9uYWwtdHJhbnNpdC1hZ2VuY3ktdGNydGEtZ3Rmcw=='
ORDER BY ts DESC
```

lauriemerrell:
# Description

This PR addresses #1985 by changing the way that `dim_schedule_feeds` (v2 warehouse) handles download and unzip failures. Previously, we used `where download_success and unzip_success is not null` as a way to skip the most recent rows where data has been downloaded but not unzipped yet, but this led to the issue described in #1985.

This PR changes the logic to the following: 
```
 When data fails to download, we do not interpolate it unless an identical zipfile is subsequently
      downloaded. So for example, say we download zipfile A from URL X on January 1, then the download
      or unzip operation fails for any reason on January 2, and then we download zipfile A from URL X
      again on January 3. If you check this table on January 3 before the new data has appeared in the
      warehouse, it will show zipfile A has expired on January 2. But once the same zipfile is ingested
      again, if you check on January 4, we will show that zipfile A was continuously in effect
      January 1-3. The reason for this is that if a new zipfile version B were to be uploaded on January 3,
      we would want to correctly say that we don't know what was in effect on January 2.
```

Previously, if the most recent download attempt was a failure, we would just interpolate that to still be in effect. This is a more conservative interpretation.

In our current production tables, it results in only two changes:

* `Metrolink` (#1989) -- previously the feed downloaded on 11/4 was listed as still "in effect" into the future, now it ends on 11/5 (first day it failed to download). **If we get a successful download again and the zipfile is identical, we will link those up and the end date for this specific feed_key will be extended.**
* `Tulare` (issue identified in #1985) -- now, the two days that the file failed to parse are skipped over and we interpolate the same zipfile remaining in effect 11/5-11/8.

PR also consolidates some CTEs to make the `dim_schedule_feeds` construction more efficient. 

Resolves #1985

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

`dbt run -s dim_schedule_feeds` and `dbt test -s dim_schedule_feeds`, which does uncover some failing tests in the `dim_shapes` tables where the feed key relationship is failing. I refreshed those tables in my namespace and the tests pass, indicating that we don't need to run a full refresh after this merges; everything should work out on its own.

Also, I tested the `dim_schedule_feeds` table with an artificial time limit of November 7 and verified that the `Tulare` feed showed as `_valid_to` November 6 (the last successful parse) and then that extended to our placeholder end date when removing the time cap and running again. 

<details>
<summary> SQL used for testing </summary>
SQL to compare current prod with namespaced changes:
```
SELECT DISTINCT 
  changed.key,
  changed.base64_url,
  changed._valid_from,
  changed._valid_to,
  old.key,
  old.base64_url,
  old._valid_from,
  old._valid_to
FROM `cal-itp-data-infra-staging.laurie_mart_gtfs.dim_schedule_feeds` changed
FULL OUTER JOIN `cal-itp-data-infra.mart_gtfs.dim_schedule_feeds` old
USING(key, base64_url, _valid_from, _valid_to)
WHERE old.key IS NULL or changed.key IS NULL
ORDER BY changed._valid_from DESC
```

SQL used to inspect Tulare in particular:
```
SELECT *
FROM `cal-itp-data-infra-staging.laurie_mart_gtfs.dim_schedule_feeds`
WHERE base64_url = 'aHR0cHM6Ly90dWxhcmVjb2cub3JnL3RjYWcvZGF0YS1naXMtbW9kZWxpbmcvZ3Rmcy1kYXRhL3R1bGFyZS1jb3VudHktcmVnaW9uYWwtdHJhbnNpdC1hZ2VuY3ktdGNydGEtZ3Rmcw=='
ORDER BY _valid_from DESC
```

</details>

### DIFF
--- a/warehouse/models/intermediate/gtfs/_int_gtfs.yaml
+++ b/warehouse/models/intermediate/gtfs/_int_gtfs.yaml
@@ -175,7 +175,7 @@ models:
       - name: step
       - name: base64_url
       - name: ts
-  - name: int_gtfs_schedule__stop_times
+  - name: int_gtfs_schedule__incremental_stop_times
     description: |
       This table is an incremental, sparse history of GTFS schedule stop times data.
       It is a precursor to `dim_stop_times` with the goal of facilitating duplicate detection and missing key issues.
@@ -199,7 +199,8 @@ models:
       - name: stop_id
         description: '{{ doc("gtfs_stop_times__stop_id") }}'
         tests:
-        - not_null
+        - not_null:
+            severity: warn
       - name: stop_sequence
         description: '{{ doc("gtfs_stop_times__stop_sequence") }}'
         tests:
--- a/warehouse/models/mart/gtfs/_mart_gtfs.yml
+++ b/warehouse/models/mart/gtfs/_mart_gtfs.yml
@@ -3,8 +3,16 @@ version: 2
 models:
   - name: dim_schedule_feeds
     description: |
-      Each row is a "feed", representing a unique combination of a specific zipfile and associated
-      `gtfs_dataset` record (i.e., a specific consistent "version" of a GTFS dataset.)
+      Each row is a "feed", representing a unique zipfile version that was successfully downloaded and
+      has data in the warehouse.
+      When data fails to download, we do not interpolate it unless an identical zipfile is subsequently
+      downloaded. So for example, say we download zipfile A from URL X on January 1, then the download
+      or unzip operation fails for any reason on January 2, and then we download zipfile A from URL X
+      again on January 3. If you check this table on January 3 before the new data has appeared in the
+      warehouse, it will show zipfile A has expired on January 2. But once the same zipfile is ingested
+      again, if you check on January 4, we will show that zipfile A was continuously in effect
+      January 1-3. The reason for this is that if a new zipfile version B were to be uploaded on January 3,
+      we would want to correctly say that we don't know what was in effect on January 2.
       This table should be used to understand "versions" of constituent data like routes, trips, etc.
     tests:
       - dbt_utils.mutually_exclusive_ranges:
--- a/warehouse/models/mart/gtfs/_mart_gtfs.yml
+++ b/warehouse/models/mart/gtfs/_mart_gtfs.yml
@@ -3,8 +3,8 @@ version: 2
 models:
   - name: dim_schedule_feeds
     description: |
-      Each row is a "feed", representing a unique zipfile version that was successfully downloaded and
-      has data in the warehouse.
+      Each row is a "feed", representing a unique zipfile version associated with the same GTFS dataset record
+      that was successfully downloaded and has data in the warehouse.
       When data fails to download, we do not interpolate it unless an identical zipfile is subsequently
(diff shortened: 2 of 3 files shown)


owades:
This is a bit over my head at the moment, looking forward to Andrew reviewing 😅

lauriemerrell:
@owades all good, just wanted to make you aware that this was happening. We can discuss in our meeting on Monday

warehouse/models/intermediate/gtfs/_int_gtfs.yaml
```diff
@@ -175,7 +175,7 @@ models:
       - name: step
       - name: base64_url
       - name: ts
-  - name: int_gtfs_schedule__stop_times
+  - name: int_gtfs_schedule__incremental_stop_times
```
atvaccaro, line 178: was this just pointing at the old table?

atvaccaro (approved):

warehouse/models/intermediate/gtfs/_int_gtfs.yaml
lauriemerrell: This was not pointing at anything, broken reference, fixed here because I noticed it

PR #1997, opened 2022-11-18
Quartet draft

lauriemerrell:
# Description

Creates the concept of the "feed quartet", a set of schedule and up to 3 RT feeds that describe the same service. 

* Intermediate table that pivots the set of feeds associated with the same service
* Refactors `dim_provider_gtfs_data` to contain these quartets joined with organizations and services, rather than what it contained before ([validated this decision on Slack](https://cal-itp.slack.com/archives/C02H6JUSS9L/p1668792977184729))

## Type of change

- [x] New feature
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) -- AFAICT nothing is strucutrally using `dim_provider_gtfs_data` (no Metabase questions; confirmed on Slack that analysts are ok with change), but if there were something it would break from this 
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
`dbt run -s int_transit_database__service_datasets_pivoted dim_provider_gtfs_data`
`dbt test -s int_transit_database__service_datasets_pivoted dim_provider_gtfs_data`

### DIFF
--- a/warehouse/models/mart/transit_database/_mart_transit_database.yml
+++ b/warehouse/models/mart/transit_database/_mart_transit_database.yml
@@ -297,16 +297,20 @@ models:
              field: key
   - name: dim_provider_gtfs_data
     description: |
-      Each record links together an organization with a GTFS dataset,
-      where the organization manages a service for which that GTFS dataset is "primary".
-      Thus, each row represents a unique "primary" organization / service / gtfs_service_data /
-      gtfs_dataset relationship. This table can be used to determine what
-      GTFS dataset(s) represent a given provider's service(s); i.e., if
-      trying to determine whether a given "agency" "has" GTFS, this table can
-      be used to make that determination.
-
-      Note that GTFS datasets whose service relationships are null or unknown
-      will not be present here.
+      Each record links together an organization with a service and a set of up to four
+      GTFS datasets (schedule, RT service alerts, RT vehicle positions, and RT trip updates),
+      where the organization manages the service.
+      This table can be used to determine what GTFS dataset(s) represent a given provider's service(s);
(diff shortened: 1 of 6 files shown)


warehouse/models/mart/transit_database/_mart_transit_database.yml
```diff
@@ -297,42 +298,57 @@ models:
...
+      Each record links together an organization with a service and a set of up to four
+      GTFS datasets (schedule, RT service alerts, RT vehicle positions, and RT trip updates),
+      where the organization manages the service.
+      This table can be used to determine what GTFS dataset(s) represent a given provider's service(s);
+      i.e., this table can be used to determine whether a given "agency" "has" GTFS.
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - service_key
+            - organization_key
```
atvaccaro, line 310: what would be the implications of 2 rows with the same service/organization? is that a normal occurrence or a shouldn't-happen?

lauriemerrell: That is fine, we can have up to 3-4 sets of feeds for a given service if there are multiple "precursor" feeds. Bigger problem is same service with *different* organization. I will add a check that there is only one "primary" set per service. Test will fail until updates are made per my doc of issues

lauriemerrell: (If we have 3-4 sets of feeds for the service, we will have 3-4 rows in this table with the same service/org pair)

atvaccaro (approved):
lgtm!

PR #2000, opened 2022-11-21
migrate expired services check

owades:
# Description

Migrate existing check about expired services
## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
`./dbt.sh test --select int_gtfs_quality__no_expired_services+`
Looked at aggregate results and saw that about 2/3 of feeds pass this and 1/3 fail.


owades:
oops @lauriemerrell it was a draft when I tagged you before. ready now.

lauriemerrell (approved):

PR #2009, opened 2022-11-22
work on versioned schedule validation codes

### LINKED ISSUE #1994: Bug: Make `fct_daily_schedule_feed_validation_notices` properly validator version-aware
**Describe the bug**
Uncovered in review of #1960.

Our `fct_daily_schedule_feed_validation_notices` table does not properly account for validator versioning, because the [cross join](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/gtfs_quality/fct_daily_schedule_feed_validation_notices.sql#L44) does not check for which validator version we were actually running on a given date, and the way that we [`COALESCE`](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/gtfs_quality/fct_daily_schedule_feed_validation_notices.sql#L34-L37) the results means that we imply that if **any** validator ran that we can count 0 notices for all codes, even codes that did not exist in the validator version that was run on that day. 

There are at least two specific symptoms of this that we need to address:
1. `gtfs_validator_version` column is null for rows with zero notices. `gtfs_validator_version` should be populated for all rows to reflect what validator version actually ran against that feed on that date. 
image

2. Validations introduced in [validator version 4](https://github.com/MobilityData/gtfs-validator/releases/tag/v4.0.0) (which we [only started running this week](https://github.com/cal-itp/data-infra/pull/1973), in November) are listed for dates before that validator version started running in our pipeline. Validations should only be listed on dates for which their associated validator version was actually run. 
image

I think to fix this we may need to start listing the validator version that was run in the **validations outcome file**, not just in the validations files themselves, to account for cases where a feed returns 0 notices. We should be recording which validator version was actually run against each feed each day. Ambivalent on whether we can/should backfill this information if we can get the right outcome without doing so. 

We also need to fix the actual SQL to record this information. 

Acceptance criteria:
- The two specific symptoms noted above are both addressed
- Confirm that the check in #1960 shows nulls before November 14, 2022 
- Perhaps we should add a `gtfs_validator_version` not null check?

atvaccaro:
# Description

Closes https://github.com/cal-itp/data-infra/issues/1994

Adds seeds for version-aware validator notices; these will need to be updated as new versions of the validator are deployed. Downstream models then join on both the code and the validator version; versions exist in the index based on the dates with that version in outcomes.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml


atvaccaro:
this was done in https://github.com/cal-itp/data-infra/pull/2136

PR #2012, opened 2022-11-28
Downgrade GTFS Spec Tests

charlie-costanzo:
# Description

In moving our GTFS Schedule data to the new warehouse we added `not_null` error reporting for fields considered `required` according to the GTFS spec, which we now considered excessive.

This PR downgrades errors coming from those required fields in the GTFS spec to warnings in the following files:
* `_gtfs_schedule_latest_only.yml`
* `_mart_gtfs.yml`

**And removes them from the following staging-level tables in favor of reporting at the mart level:**
* `not_null` tests on GTFS spec fields in `_stg_gtfs.yml`

**No downgrades were made to:**
*  Any synthetic key tests
* Not null checks on `<something>_id` fields
* Any other fields that Cal-ITP has added outside the GTFS spec
* `dbt_utils_unique_combination_of_columns` tests
* `dbt_utils_mutually_exclusive_ranges` tests

**Added missing tests to:**
* `stg_gtfs_schedule__calendar_dates.exception_type`
  * Added not `null_test` severity `warn`

**Added dbt_utils.not_null_proportion to:**
* Any of the GTFS spec columns with < 30 results 
* (`<something>_id` fields included)

**Still erroring:**
* `gtfs_schedule_latest_only.stop_times`
  * `not_null_stop_times_stop_id` = 163 results

Resolves # [issue]

## Type of change

## How has this been tested?
`dbt run` and `dbt test`


warehouse/models/gtfs_schedule_latest_only/_gtfs_schedule_latest_only.yml
```diff
@@ -46,15 +46,21 @@ models:
...
         tests:
-        - not_null
+          - dbt_utils.not_null_proportion:
+              at_least: 0.99
+              severity: warn
       - name: agency_timezone
         description: '{{ doc("gtfs_agency__agency_timezone") }}'
         tests:
-        - not_null
+        - not_null:
```
lauriemerrell, line 61: Ok sorry this is an incredibly annoying suggestion, but I think it would be worth it to add a YAML anchor to name this like `&not_null_warn` and then use that below, to limit how much more verbose this makes the YAML. 

Could do the same for the .99 threshold: `&not_null_warn_99_threshold` or something

lauriemerrell (changes requested):
I have a question for @atvaccaro, maybe the three of us should chat. It seems a little silly to be doing identical `not_null` tests on models that are basically unchanged from staging through to the mart. I am wondering if we only want to have them in one location. Perhaps this is something to be deferred (I know we have a testing cost audit queued up) but this PR made me wonder.

atvaccaro:
> I have a question for @atvaccaro, maybe the three of us should chat. It seems a little silly to be doing identical `not_null` tests on models that are basically unchanged from staging through to the mart. I am wondering if we only want to have them in one location. Perhaps this is something to be deferred (I know we have a testing cost audit queued up) but this PR made me wonder.

Yeah I broadly agree; I'd mostly be OK with only enforcing not-null at the mart level. I do want uniqueness throughout.

lauriemerrell:
> Yeah I broadly agree; I'd mostly be OK with only enforcing not-null at the mart level. I do want uniqueness throughout.

This PR is only touching `not_null`. @charlie-costanzo -- I think we can remove all these tests at the staging level.

lauriemerrell (approved):

lauriemerrell:
Thanks so much @charlie-costanzo -- after this merges we should go through and resolve all of these in Sentry and figure out what still fires.

PR #2010, opened 2022-11-22
Better support for service/organization relationships to GTFS data

lauriemerrell:
# Description

This PR does several things to make `dim_provider_gtfs_data` more useful and lay the groundwork for traversing back and forth between GTFS and organization/service data:

- Adds the following columns to `dim_provider_gtfs_data`: `customer_facing`, `agency_id`, `network_id`, `route_id`, `ntd_id`, `hubspot_company_record_id` (in several case this requires adding fields in upstream locations, including adding `ntd_id` and `hubspot_company_record_id` to the `organizations` external table in the Airflow job)
- Adds `agency_id`, `network_id`, and `route_id` to `fct_daily_scheduled_trips`, which facilitates traversing directly between trips and service/organization relationships 

It also creates an intermediate daily RT URLs model that will be useful in an upcoming PR to create a few additional guideline indices. 

Progress towards #1904.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

`dbt build -s dim_provider_gtfs_data` and `dbt build fct_daily_scheduled_trips`.

There is one failing test in `dim_provider_gtfs_data` (`service_key` is currently not unique for `customer_facing`) which is a result of data entry issues in Airtable, I will follow up offline about those.

PR #2018, opened 2022-11-28
migrate checks for rt feed presence

owades:
# Description

Migrate 3 checks for RT feed presence.

Resolves # [issue]

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
`./dbt test` for each check and downstream
looked at aggregate results to confirm that they look normal


owades:
@atvaccaro @lauriemerrell , since these are my first RT checks I'd love a detailed review from one of you.

Specific callouts:
- Each check only runs for one of the three feeds, so I filter rt_feed_guideline_index accordingly before joining to it. I also don't run a test checking for equal rows to rt_feed_guideline_index, since it only should contain a subset of rows (ie only one feed)
- Separate from these checks, I modified the tests that @atvaccaro wrote for his first RT check, `no_rt_critical_validation_errors`, please take a look at this as I don't want to cause trouble!

warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
```diff
@@ -12,18 +12,15 @@ models:
...
             - date
             - base64_url
-    columns:
-      - name: status
-        tests:
-          - dbt_utils.not_null_proportion:
-              at_least: 0.99
+            - feed_type
+      - dbt_utils.equal_rowcount:
+          compare_model: ref('int_gtfs_quality__schedule_feed_guideline_index')
```
lauriemerrell: I don't think that this is necessarily expected to have an equal row count... We expect to have a moderately different number of schedule and RT feeds, I think?

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__feed_present_service_alerts.sql
```diff
@@ -0,0 +1,38 @@
+WITH feed_guideline_index AS (
+    SELECT * FROM {{ ref('int_gtfs_quality__rt_feed_guideline_index') }}
+     WHERE feed_type = 'service_alerts'
+),
+
+fct_daily_rt_feed_files AS (
```
lauriemerrell, line 6: Sorry to keep doing this (nuance/confusion about indexing on feeds/files vs. URLs/dataset records in Airtable), but in #2010 I have a daily URLs list for RT which might be better than checking files -- daily URLs will say whether we were *attempting* to hit a service alerts feed. Issue with `fct_daily_rt_feed_files` is that feeds with 0 files will not be listed, but if this is just checking for existence of a feed I think that the list of URLs is better.

lauriemerrell: But defer to you if you only want to give credit for feeds where we can get at least one file, I think that could make sense; just not sure where this check is meant to fall on the spectrum of success/validity vs. pure existence.

lauriemerrell (commented):
Couple of questions

warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
```diff
@@ -12,18 +12,15 @@ models:
             - date
 
   - name: int_gtfs_quality__no_rt_critical_validation_errors
-    description: |
-      Checks for any critical validation notices in RT feeds.
-    tests:
+    tests: &stg_gtfs_guideline_tests_rt
       - dbt_utils.unique_combination_of_columns:
```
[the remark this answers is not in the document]
owades: Nope, was just trying to be thorough. Thanks for explaining, makes a lot of sense.

lauriemerrell (approved):
Ok to merge with the current logic, just want to note that I think it's possible we'll want to revisit the indexing here if we ever want to change this to a "URL exists" check as opposed to a "files exist" check.

PR #2026, opened 2022-11-30
Add v2 check v2 trip alignment

owades:
# Description

Migrate trip-id-alignment check to v2
This checks that no trip_ids appear in RT that aren't in the associated schedule. Note that we also need to do a more in depth check that sees whether all trips in schedule appear in RT.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
./dbt test
Looked at aggregate data to confirm that most feeds pass and a subset fails

### DIFF
--- a/warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
+++ b/warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
@@ -63,7 +63,6 @@ models:
     tests: *stg_gtfs_guideline_tests
     columns: *stg_gtfs_guideline_columns
 
-
   - name: int_gtfs_quality__shapes_file_present
     tests: *stg_gtfs_guideline_tests
     columns: *stg_gtfs_guideline_columns
@@ -103,3 +102,5 @@ models:
   - name: int_gtfs_quality__no_expired_services
     tests: *stg_gtfs_guideline_tests
     columns: *stg_gtfs_guideline_columns
+
+# todo: add tests for this PR
--- a/warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
+++ b/warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
@@ -11,20 +11,6 @@ models:
             - feed_key
             - date
 
-  - name: int_gtfs_quality__no_rt_critical_validation_errors
-    description: |
-      Checks for any critical validation notices in RT feeds.
-    tests:
-      - dbt_utils.unique_combination_of_columns:
-          combination_of_columns:
-            - date
-            - base64_url
-    columns:
-      - name: status
-        tests:
-          - dbt_utils.not_null_proportion:
-              at_least: 0.99
-
   - name: int_gtfs_quality__rt_validation_outcomes
     description: |
       Unioned, incremental table of RT validation outcomes.
@@ -46,7 +32,7 @@ models:
             - severity
 
   - name: int_gtfs_quality__no_schedule_validation_errors
-    tests: &stg_gtfs_guideline_tests
+    tests: &stg_gtfs_guideline_tests_schedule
       - dbt_utils.unique_combination_of_columns:
           combination_of_columns:
             - date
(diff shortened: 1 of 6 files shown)


warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
```diff
@@ -11,20 +11,6 @@ models:
             - feed_key
             - date
 
-  - name: int_gtfs_quality__no_rt_critical_validation_errors
```
lauriemerrell, line 14: Is this deletion intentional?

warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
```diff
@@ -46,7 +32,7 @@ models:
             - severity
 
   - name: int_gtfs_quality__no_schedule_validation_errors
-    tests: &stg_gtfs_guideline_tests
+    tests: &stg_gtfs_guideline_tests_schedule
```
lauriemerrell, line 35: Just confirming that these are the same changes from #2018

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__trip_id_alignment.sql
```diff
@@ -0,0 +1,56 @@
...
+        idx.date,
+        idx.base64_url,
+        idx.feed_type,
+        {{ trip_id_alignment() }} AS check,
+        {{ fixed_route_completeness() }} AS feature,
+        rt_files,
+        errors,
+        CASE
+            WHEN COALESCE(errors, 0) = 0 THEN "PASS"
+            WHEN rt_files IS NOT NULL THEN "FAIL" -- we had no errors, but did have files
```
lauriemerrell, line 45: A bit confused by the comment here.... If `COALESCE(errors,0) != 0` then there *were* errors, right?

I also think this may need adjustment -- should the `rt_files` check be part of the first condition? I think as written that `rt_files IS NULL` will lead to `PASS` under this logic (because presumably if `rt_files` is null then `errors` will be null and then will be `COALESCED` to 0, if I am understanding correctly.)

Not to add more work to your plate, but I am wondering if at some point it will be necessary to audit null handling in the different checks and standardize at all 😅

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__trip_id_alignment.sql
```diff
@@ -0,0 +1,56 @@
...
+),
+
+critical_notices AS (
+    SELECT
+        date,
+        base64_url,
+        SUM(total_notices) AS errors,
+    FROM fct_daily_rt_feed_validation_notices
+    -- Description for code E003:
+    ---- "All trip_ids provided in the GTFS-rt feed must exist in the GTFS data, unless the schedule_relationship is ADDED"
```
lauriemerrell, line 29: This is a way more efficient way to check this, great point (relative to manually making a separate incremental table and checking)

lauriemerrell (changes requested):
Want to confirm that the check for `rt_files` is behaving as expected? Also think a rebase/merge of main might be helpful, diff seems a little funky

warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
owades: yep, same changes. I keep doing this with PRs, where I should probably just make it one PR so I'm not doing the same changes twice and needing to merge one and then merge into the other.

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__trip_id_alignment.sql
owades: No surprise that you are right. I borrowed this from the rt check that Andrew wrote...but what I really should have done is caught the issue in his PR! I've now updated both - erring on the side of specificity so that it's clear what the conditions are to pass and fail.

To pass: at least one RT file exists, and no errors detected
To fail: at least one RT file exists, more than zero errors detected
To have a null value: no RT files exist. Note that there are none of these results now, since the index table must not have a row for days when no RT files exist.

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__no_rt_critical_validation_errors.sql
```diff
@@ -39,8 +39,8 @@ int_gtfs_quality__no_rt_critical_validation_errors AS (
         rt_files,
         errors,
         CASE
-            WHEN COALESCE(errors, 0) = 0 THEN "PASS"
-            WHEN rt_files IS NOT NULL THEN "FAIL" -- we had no errors, but did have files
+            WHEN rt_files IS NOT NULL AND COALESCE(errors, 0) = 0 THEN "PASS"
+            WHEN rt_files IS NOT NULL AND errors > 0 THEN "FAIL" -- we had no errors, but did have files
```
lauriemerrell, line 43: Nit pick that I think the comment here is still inaccurate, but actual logic now correct

lauriemerrell (approved):
One comment-only update if you don't mind, but I don't need to check again

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__trip_id_alignment.sql
lauriemerrell: Yeah, I reviewed his PR too 🙈 Thanks for updating both!

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__no_rt_critical_validation_errors.sql
owades: thank you!

Issue #1948, opened 2022-11-02
Create a read-from-redis-duration metric in Huey


atvaccaro:
We had a minor v2 RT incident today in which our throughput plummeted leading to almost all tasks expiring; it was solved by restarting the consumers. It seems like this could have been due to slow Redis reads as task start delays jumped while task durations dropped. I'd like to try to add a metric to let us know how long Redis I/O is taking in the Huey consumers; it's possible we need to switch to another Huey Redis client type.

![image (13)](https://user-images.githubusercontent.com/4305366/199390579-43c77abe-39d5-471d-ba57-1e41317889e3.png)

atvaccaro:
I'm going to close this; originally I misinterpreted huey Consumers vs Workers and though Consumers read from redis and passed tasks to Workers. In reality, Workers read from redis directly, so the task delay metric is a good-enough proxy for us. There isn't anything "in between" reading and executing.

PR #2027, opened 2022-11-30
RT performance tweaks

atvaccaro:
# Description

1. Disables backoff and sets max_delay to 1, i.e. huey clients will always wait 1 second between task checks rather than scaling backoff
2. Switches from non-blocking workers to blocking; this means workers will block for up to 20 seconds (i.e. at minimum through the next tick) while reading from an empty queue; this should reduce the delay metric.

Also various clean-up such as making environment variable prefixes consistent and moving some hard-coded configs to env vars.

See https://github.com/cal-itp/data-infra/issues/1948#issuecomment-1334018195 for why I chose not to add a read-from-redis metric.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Deploying tweaks to test as I make them.

## Screenshots (optional)
image

### DIFF
--- a/kubernetes/apps/overlays/gtfs-rt-archiver-v3-prod/archiver-channel-vars.yaml
+++ b/kubernetes/apps/overlays/gtfs-rt-archiver-v3-prod/archiver-channel-vars.yaml
@@ -6,6 +6,5 @@ data:
   AIRFLOW_ENV: cal-itp-data-infra
   CALITP_BUCKET__GTFS_DOWNLOAD_CONFIG: "gs://calitp-gtfs-download-config"
   CALITP_BUCKET__GTFS_RT_RAW: "gs://calitp-gtfs-rt-raw-v2"
-  HUEY_CONSUMER_WORKERS: "16"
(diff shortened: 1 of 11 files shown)


kubernetes/apps/overlays/gtfs-rt-archiver-v3-prod/archiver-channel-vars.yaml
```diff
@@ -6,6 +6,5 @@ data:
   AIRFLOW_ENV: cal-itp-data-infra
   CALITP_BUCKET__GTFS_DOWNLOAD_CONFIG: "gs://calitp-gtfs-download-config"
   CALITP_BUCKET__GTFS_RT_RAW: "gs://calitp-gtfs-rt-raw-v2"
-  HUEY_CONSUMER_WORKERS: "16"
```
lauriemerrell, line 9: Thanks for the offline walkthrough, only remaining question is why we're removing this / why set to 4 (instead of 16) in the docker-compose below... Not sure I am following that part

lauriemerrell (approved):

PR #2037, opened 2022-12-02
Add v2 check for https, update feed-present check

owades:
# Description

I created a new check for whether an RT feed starts with "HTTPS". As we discussed in a meeting last week, the first action item for failing this check might be for Cal-ITP to try updating the URL in the warehouse.

In building this I realized I can generate multiple checks in one intermediate table. Having found this, I also updated the way I handle the "feeds_present" check to follow suit.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
`./dbt.sh build... +_table name_+`

The aggregate results for both tables make sense.
- Only 6 files are failing the HTTPS check, the rest pass
- For the feed_present check, most pass and some fail. Service alerts is most likely to fail, which makes sense as some feeds simply don't have any files (you could call this a bug, I think it's OK as long as we explain clearly).


warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__rt_https.sql
```diff
@@ -0,0 +1,30 @@
...
+             WHEN idx.feed_type = 'trip_updates' THEN {{ rt_https_trip_updates() }}
+             WHEN idx.feed_type = 'vehicle_positions' THEN {{ rt_https_vehicle_positions() }}
+        END AS check,
+        {{ best_practices_alignment() }} AS feature,
+        CASE
+            WHEN string_url LIKE 'https%' THEN "PASS"
+            WHEN string_url IS NOT null AND string_url NOT LIKE 'https%' THEN "FAIL"
+        END AS status,
+    FROM feed_guideline_index AS idx
+    JOIN rt_daily_url_index AS url_index
```
lauriemerrell, line 24: Do we want a `LEFT JOIN` here? Because otherwise this is going to drop days where the URL isn't present.... Hmm. (Especially if we aren't using an equal row counts test, wondering if we are incurring some risk about grain here.)

lauriemerrell (commented):
Just one substantial question about whether we want an `INNER` vs `LEFT` join when doing this combined 3-checks-in-1 table.

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__rt_https.sql
lauriemerrell: Ahh ok sorry for not following, I forgot about this. I was more concerned about cases like *since* November where there's a mismatch (like, say that URL A is in the guideline index on December 1, but not in the URL index.... Which now that I think about it I can't remember how possible that is).

Either way, I think that handling nulls with an NA is maybe preferable just because it's a bit more explicit? It's harder to troubleshoot a row that doesn't appear at all vs. a row that appears and explicitly has null data, if that makes sense

PR #2047, opened 2022-12-06
add check for proper lead time for schedule updates

owades:
# Description

If a change has been made to trips, stops, stop_times, calendars, or calendar_dates that affects service within 7 days of when the update was made, the check will FAIL. Otherwise it will pass.

This logic partially sits on top of the new table `int_gtfs_schedule__all_scheduled_service`, which allows me to compare different feed versions' effective schedule.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

Looked at aggregate results for 2 days ago:
PASS: 159
FAIL: 64

` ./dbt.sh build -s int_gtfs_quality__lead_time+  `

Looking at results manually, I confirmed that each infraction type is capturing correctly - including:
- trip added
- trip removed
- stop location changed
- stop time changed


warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__lead_time.sql
```diff
@@ -0,0 +1,249 @@
...
+),
+
+feed_update_count AS (
+    SELECT t1.date,
+           t1.feed_key,
+           SUM(t2.trip_added + t2.trip_removed + t2.trip_changed)
+               OVER (
+                   PARTITION BY t2.feed_key
+                   ORDER BY t2.date
+                   ROWS BETWEEN 30 PRECEDING AND CURRENT ROW
```
lauriemerrell, line 223: Just to confirm, the check for like September 30 is checking that nothing violated the 7 day rule in the last 30 days? (I was initially confused about why this is 30 and the actual check is 7 day warning)

lauriemerrell (changes requested):
At a high level this makes sense 👍 👏  I do have a few questions/suggestions, happy to have a call about this if it helps.

1. Have you checked how this works for feeds that **only** use `calendar_dates` and not `calendar` at all? One current such feed seems to be Yolobus, `feed_key = f8ce4306a3b172c6df6c9e9e5b992939`. 
   - I am just wondering if the separate trips vs. `calendar_dates` logic is going to work there, I kind of think you might need to take account of the full schedule information to check whether the trip is scheduled to occur anytime in the next 7 days... Maybe you could use the existing `int_gtfs_schedule__daily_scheduled_service_index` table ([docs](https://dbt-docs.calitp.org/#!/model/model.calitp_warehouse.int_gtfs_schedule__daily_scheduled_service_index)) for the logic of `service_id`s with service on a given date? 🤔 Like, maybe start with service IDs that have service in the next 7 days, and then follow changes out from there (whether those services themselves had any changes, or whether any of their trips did)? 
   
2. I think that it might be worth breaking this up into two models, to have the logic about changing data (the "was this trip/service added/removed/changed") in one intermediate model and then the guideline check references that -- I am kind of thinking that this "did these records change" logic could be useful for troubleshooting, outside of just this check so it would be nice to have it separately. I am also kind of wondering if there are any tests we'd want to put on the "did this change" logic on its own... Foreign key or row count or anything like that? Just because it's kind of a complicated operation, I'm wondering if there's anything we should be doing to validate it.
3. I am curious about the performance of this.... Can you check how much time / data processing it adds, running the mart fact table with and without this check? (Like, just looking at the output from the `dbt run` where it tells you bytes processed.) If this is very slow, I would wonder about maybe generating a hash of the grouped values for each trip/service ID and comparing that rather than comparing the full values when you do the version over version join.  
4. (definite change) I think you need to add this check to the union in the mart table.

And one inline clarification.

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__lead_time.sql
owades: That's correct. The idea is that schedule changes may happen infrequently, and so we want to look at all recent schedule changes to measure adherence to this best practice, rather than saying "did any changes you made on this date break the rule?".

owades:
> 1. Have you checked how this works for feeds that **only** use `calendar_dates` and not `calendar` at all? One current such feed seems to be Yolobus, `feed_key = f8ce4306a3b172c6df6c9e9e5b992939`.
>    * I am just wondering if the separate trips vs. `calendar_dates` logic is going to work there, I kind of think you might need to take account of the full schedule information to check whether the trip is scheduled to occur anytime in the next 7 days... Maybe you could use the existing `int_gtfs_schedule__daily_scheduled_service_index` table ([docs](https://dbt-docs.calitp.org/#!/model/model.calitp_warehouse.int_gtfs_schedule__daily_scheduled_service_index)) for the logic of `service_id`s with service on a given date? 🤔 Like, maybe start with service IDs that have service in the next 7 days, and then follow changes out from there (whether those services themselves had any changes, or whether any of their trips did)?

I haven't explicitly checked for feeds that only use calendar_dates, but I've checked that the calendar_dates and trips logic work independently. I'm intrigued by the suggestion to use daily_schedule_service_index, though I'd need to wrap my head around how that works for different feed versions. **Does it have daily data for multiple feed versions?** If so that could be a much easier comparison than what I did.

I did look at Yolobus, and from what I can see there's only one version of their feed on record (info below, thanks for the feed key), which means no comparison can be performed.
            base64_url: aHR0cHM6Ly93d3cueW9sb2J1cy5jb20vR1RGUy9nb29nbGVfdHJhbnNpdC56aXA
            feed_key: f8ce4306a3b172c6df6c9e9e5b992939
            _valid_from: 2022-10-26 03:00:23.360182 UTC

> 2. I think that it might be worth breaking this up into two models, to have the logic about changing data (the "was this trip/service added/removed/changed") in one intermediate model and then the guideline check references that -- I am kind of thinking that this "did these records change" logic could be useful for troubleshooting, outside of just this check so it would be nice to have it separately. I am also kind of wondering if there are any tests we'd want to put on the "did this change" logic on its own... Foreign key or row count or anything like that? Just because it's kind of a complicated operation, I'm wondering if there's anything we should be doing to validate it.

This suggestion makes sense. It would be even more helpful if it could produce the diff somehow, and show us _what_ changed. As it stands, if there's a change in trips, you would need to sift through a very long string containing lat and long for every stop, etc, to figure out what changed. It's very hard to test.

> 3. I am curious about the performance of this.... Can you check how much time / data processing it adds, running the mart fact table with and without this check? (Like, just looking at the output from the `dbt run` where it tells you bytes processed.) If this is very slow, I would wonder about maybe generating a hash of the grouped values for each trip/service ID and comparing that rather than comparing the full values when you do the version over version join.
I think you're right that it's a lot - `47.5 GB processed`, up from `32.2 GB processed`. Is there a hash function that you recommend? I haven't used them a ton, I see there are a few options.

> 4. (definite change) I think you need to add this check to the union in the mart table.

Done! Apparently trick is to omit it from both the "intended checks" table and the UNION table, that way it doesn't fail any tests 😉

lauriemerrell:
Ahh sorry, forgot about versioning -- for a `calendar_dates`-only example, how about Montebello, `base64_url = aHR0cHM6Ly9tYmwucmlkZXJhbGVydHMuY29tL2luZm9wb2ludC9ndGZzLXppcC5hc2h4`, which seems to have multiple versions and not use calendar? 

(Here's the query I am using to identify feeds in case it is helpful)
<details>
<summary>query to find feeds with multiple versions but no calendar</summary>
```
WITH multiple_versions AS (
  SELECT base64_url
  FROM `cal-itp-data-infra.mart_gtfs.fct_schedule_feed_downloads`
  GROUP BY base64_url
  HAVING COUNT(DISTINCT feed_key) > 1
)

SELECT EXTRACT(DATE FROM ts) AS dt, feed_key, t1.base64_url, name
FROM `cal-itp-data-infra.mart_gtfs.fct_schedule_feed_downloads` AS t1
LEFT JOIN `cal-itp-data-infra.mart_transit_database.dim_gtfs_datasets` AS t2
ON t1.gtfs_dataset_key = t2.key
INNER JOIN multiple_versions
  ON t1.base64_url = multiple_versions.base64_url
WHERE ('calendar.txt' NOT IN UNNEST(t1.zipfile_files)) AND (t1.zipfile_files IS NOT NULL) AND ts < '2022-12-07'
  AND download_success AND unzip_success
ORDER BY dt DESC
```
</details>

Re: 
> Does it have daily data for multiple feed versions? If so that could be a much easier comparison than what I did.

In its [current form](https://dbt-docs.calitp.org/#!/model/model.calitp_warehouse.int_gtfs_schedule__daily_scheduled_service_index) it is filtered to only the feed that was actually in effect on that day. But perhaps it's still worth reusing the logic from there -- we could refactor that to have an expanded calendar (basically, take the `daily_services` CTE but get rid of `fct_daily_schedule_feeds` join on line 59 which is what filters to only feeds in effect) and make that a standalone intermediate model that you could use here and we could filter separately over there to get only the active feeds.

My thinking is mostly just that the schedule logic is annoying/complicated, if we can reuse that it seems good, and it's my bad for not thinking of it when we talked about this before.

Re: the diffing, I think that perhaps the cleanest thing to do would be to generate a hash of the row in each file (stops, stop_times, trips, keyed on its GTFS record id like `stop_id, stop_hash`), and then in the big diffing table just check which record's hash has changed...? This is possibly going against things I said in the past, but I think you could basically make a "scheduled_events_diff" where each row is date/service/trip/stop_time/stop.... i.e., less grouped to trip ID. I think this would have worse performance (which is what I liked about the grouping) but be clearer? Because you could have more specific checks like stop_time_added, stop_changed. So maybe it depends how much you care about being able to granularly identify which record has changed. 

I know we already met about this but maybe it is worth meeting again, especially if we (data services) should help with breaking up the current form of `int_gtfs_schedule__daily_scheduled_service_index` or comparing performance of various options here.

owades:
Thank you, @lauriemerrell!

I think it makes sense to use a portion of `int_gtfs_schedule__daily_scheduled_service_index`, I will set up a meeting to discuss.

I'm not (yet) convinced we need to solve for the edge case of someone switching from cal -> cal_dates, but I think it makes sense to minimize the redundancy in our code. This doesn't look like a ton more work.

I think your hashing suggestion is a good idea, and it is worth building the discoverability into the table, because I expect I'll be needing to build that out later.

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__lead_time.sql
```diff
@@ -0,0 +1,174 @@
...
+         prev_trips.trip_stop_locations_hash AS prev_trip_stop_locations_hash
+    FROM scheduled_trips_version_history AS trips
+    FULL OUTER JOIN scheduled_trips_version_history AS prev_trips
+      ON trips.prev_feed_key = prev_trips.feed_key
+     AND trips.trip_id = prev_trips.trip_id
+     AND trips.service_date = prev_trips.service_date
+),
+
+-- We count each "infraction" type seperately, mostly for QA purposes
+----  Note that since all_scheduled_service table has a row per trip per service day,
```
lauriemerrell, line 126: Massive nit pick but `all_scheduled_service` is purely service ID (not trip) level

lauriemerrell (approved):
One optional comment nit pick. Congrats on the most complicated check ever 😅 👏

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__lead_time.sql
owades: i appreciate it nonetheless! updated the comment to be a bit more vague

owades:
@atvaccaro is this failing check something I need to worry about?

lauriemerrell:
> @atvaccaro is this failing check something I need to worry about?

He's out today, we are looking into it (not caused by your PR), I can merge if you want

PR #2044, opened 2022-12-05
consider situations with no RT entity to be successful parses

atvaccaro:
# Description

Fixes issue described in https://cal-itp.slack.com/archives/C0332307HKM/p1669134932042959 (closes https://github.com/cal-itp/data-infra/issues/2045)

This will necessitate a backfill to fix history, which we can do more fully when we do our typical backfill.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Locally, I've tested service alerts have a queryable header JSON.
```
select distinct JSON_VALUE(header, '$.timestamp')
from `cal-itp-data-infra-staging`.andrew_staging.stg_gtfs_rt__service_alerts_outcomes
where dt = '2022-12-01'
;
```

## Screenshots (optional)
```
[2022-12-05 22:52:31,160] {pod_launcher.py:149} INFO - WARNING: no parsed entity found in gs://test-calitp-gtfs-rt-raw-v2/service_alerts/dt=2022-10-13/hour=2022-10-13T14:00:00+00:00/ts=2022-10-13T14:59:20+00:00/base64_url=aHR0cHM6Ly9zbXJ0LnRyaXBzaG90LmNvbS92MS9ndGZzL3JlYWx0aW1lL3NlcnZpY2VBbGVydC9DQTU1OEREQy1EN0YyLTRCNDgtOUNBQy1ERUVBMTEzNEY4MjA=/feed
[2022-12-05 22:52:31,160] {pod_launcher.py:149} INFO - WARNING: no parsed entity found in gs://test-calitp-gtfs-rt-raw-v2/service_alerts/dt=2022-10-13/hour=2022-10-13T14:00:00+00:00/ts=2022-10-13T14:59:40+00:00/base64_url=aHR0cHM6Ly9zbXJ0LnRyaXBzaG90LmNvbS92MS9ndGZzL3JlYWx0aW1lL3NlcnZpY2VBbGVydC9DQTU1OEREQy1EN0YyLTRCNDgtOUNBQy1ERUVBMTEzNEY4MjA=/feed
[...]

PR #2049, opened 2022-12-07
fix issue introduced in #2044; use correct cte in final select

lauriemerrell:
# Description

Bug introduced in #2044, final select pulls raw data instead of the renamed stuff. Caused dbt failures overnight, will need to rerun the affected table's lineage once this merges.

image

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

`dbt run / dbt test` there are failing downstream tests but I do not think they are caused by this, talking to @atvaccaro offline


atvaccaro (approved):

atvaccaro:
bad find-replace on my part, sorry

PR #2057, opened 2022-12-07
Initial tables to count files by hour

### LINKED ISSUE #1895: User Story: Create monitoring dashboards for actual data availability for v2 archiver
### Summary

**As an** analytics engineer,  **I want** a Metabase dashboard showing RT file availability for the v3 archiver analogous to what we have for v2 (old) **so that** I get an get a more holistic picture of actual data availability than what Sentry alerts offer.

Sentry is helpful for identifying when outages start, but given the transient nature of most RT outages, it would be helpful to create a dashboard that shows overall file availability over time. This would also help show when new feeds are added. This would also help catch if there are failures that are causing issues in data creation that are not yet being caught in Sentry, which I suspect may have happened with SunLine (URL: https://infopoint.sunline.org/InfoPoint/GTFS-Realtime.ashx?Type=VehiclePosition) which had an expired certificate. I don't see any issues in Sentry related to the expired cert from the v3 archiver even though the v2 (old) archiver shows no data for a period and then the return of data. 

I think @atvaccaro might disagree with this request or think that at least part of it could be handled by improved Sentry settings, which may also be true.

### Acceptance Criteria
1. Tables that show:
   1.  overall file counts by day per URL, perhaps ideally also showing the first day that a URL appeared in our data to help troubleshoot 
   2. file counts by hour like our current wide by hour table?
3. A Metabase dashboard that consumes this table to visualize data availability similar to what we have for the current archiver

# Sprint Ready Checklist
3. - [ ] Team understands acceptance criteria
5. - [ ] Team has defined solution / steps to satisfy acceptance criteria
6. - [ ] Acceptance criteria is verifiable / testable
7. - [ ] External / 3rd Party dependencies identified

charlie-costanzo:
# Description

This PR adds new warehosue tables to facilitate:
* A replication of the [hourly v2 archiver Metabase Dashboard](https://dashboards.calitp.org/dashboard/48-gtfs-rt-files-hourly-view-by-day?date_filter=past1days) to assess the v3 archiver using v2 pipeline data
  * [New dashboard found here](https://dashboards.calitp.org/dashboard/98-v2-gtfs-rt-files-hourly-view-by-day?relative_date=yesterday)
  * This also includes the addition of a new table to that dashboard which accounts for file parse success percentages by hour
* A new dashboard to compare between v2 and v3 archivers

**TODO**: Add new table details to yml

New tables:
* fct_hourly_rt_feed_files
* fct_hourly_rt_feed_files_success

Partially resolves #1895 - still need to make new dashboard to compare between v2 and v3 archivers (only focusing on file presence, not creating new tables for v2 archiver for comparison)

## Type of change

- [x] New feature

## How has this been tested?
locally using `dbt_test` and `dbt_run`


warehouse/models/mart/gtfs/fct_hourly_rt_feed_files.sql
```diff
@@ -0,0 +1,109 @@
...
+),
+
+hourly_tot AS (
+
+    SELECT *
+    FROM
+        (SELECT
+
+            dt,
+            EXTRACT(HOUR FROM hour AT TIME ZONE "America/Los_Angeles") as download_hour,
```
lauriemerrell, line 56: I think we should leave this in UTC for consistency.... We've basically committed to having all system-generated timestamps in UTC (old dashboard is UTC)

warehouse/models/mart/gtfs/fct_hourly_rt_feed_files.sql
```diff
@@ -0,0 +1,109 @@
...
+
+),
+
+fct_hourly_rt_feed_files AS (
+    SELECT
+
+        {{ dbt_utils.surrogate_key(['pivoted_tots.dt', 'pivoted_tots.base64_url']) }} AS key,
+
+        url_index.dt,
+        url_index.string_url,
```
lauriemerrell, line 95: I would remove string URL and instead have the GTFS dataset key here, so that we can join in a human-readable record name (can do that join in Metabase I think)

warehouse/models/mart/gtfs/fct_hourly_rt_feed_files_success.sql
```diff
@@ -0,0 +1,154 @@
...
+    SELECT *
+    FROM int_gtfs_rt__unioned_parse_outcomes
+    {% if is_incremental() %}
+    WHERE dt >= DATE '{{ max_date }}'
+    {% else %}
+        WHERE dt >= DATE_SUB(CURRENT_DATE(), INTERVAL 1 MONTH)
+    {% endif %}
+),
+
+daily_tots AS (
```
lauriemerrell, line 32: I think this already exists in [fct_daily_rt_feed_files](https://dbt-docs.calitp.org/#!/model/model.calitp_warehouse.fct_daily_rt_feed_files), do we need it here?

warehouse/models/mart/gtfs/fct_hourly_rt_feed_files_success.sql
```diff
@@ -0,0 +1,154 @@
...
+
+hourly_tots AS (
+
+    SELECT
+
+        dt,
+        base64_url,
+        feed_type,
+        EXTRACT(
+            HOUR FROM hour AT TIME ZONE "America/Los_Angeles"
```
lauriemerrell, line 63: Same comment as above, lets leave this in UTC

warehouse/models/mart/gtfs/fct_hourly_rt_feed_files_success.sql
```diff
@@ -0,0 +1,154 @@
...
+
+            dt,
+            download_hour,
+            base64_url,
+            feed_type,
+            (success_file_count_hour / file_count_hour) * 100 AS pct_success_hour
+
+            FROM hourly_tots)
+    PIVOT(
+        SUM(pct_success_hour)
```
lauriemerrell, line 93: Just to confirm, we should not actually be summing here, right? There should only be one value per bucket? Just get nervous when I see summing of a percent/proportion, wondering if there is any other way to indicate this

warehouse/models/mart/gtfs/fct_hourly_rt_feed_files_success.sql
```diff
@@ -0,0 +1,154 @@
...
+    SELECT
+
+        joined_tots.* EXCEPT (dt, base64_url, feed_type),
+
+        url_index.dt,
+        url_index.string_url,
+        url_index.base64_url,
+        url_index.type AS feed_type,
+
+        TO_HEX(
```
lauriemerrell, line 134: Can use our [`farm_surrogate_key`](https://github.com/cal-itp/data-infra/blob/main/warehouse/macros/farm_surrogate_key.sql) macro for this instead

warehouse/models/mart/gtfs/fct_hourly_rt_feed_files_success.sql
```diff
@@ -0,0 +1,154 @@
...
+
+),
+
+fct_hourly_rt_feed_files_success AS (
+    SELECT
+
+        joined_tots.* EXCEPT (dt, base64_url, feed_type),
+
+        url_index.dt,
+        url_index.string_url,
```
lauriemerrell, line 130: Similar comment to above, I would key on `dt, URL, GTFS dataset key`

lauriemerrell (changes requested):
Overall looks good! I think that converting to Pacific Time might be part of what is causing the odd appearance in the dashboard of a data hole yesterday where I don't think one actually exists, wonder what dashboard will look like if you convert back to UTC. 

Otherwise, a few smaller comments. 👍

warehouse/models/mart/gtfs/fct_hourly_rt_feed_files_success.sql
charlie-costanzo: Yes you are right - I got tripped up because `PIVOT` requires an aggregation and I couldn't figure out how to get around it, so I just added the SUM() for the time being because I thought it would allow it to run and wouldn't change the results (and meant to note it in the PR). I just changed it to MAX() instead, which is less scary I think, but maybe we can talk about alternatives this afternoon. I bet this CTE and the one before it can be completely rethought...

warehouse/models/mart/gtfs/fct_hourly_rt_feed_files.sql
```diff
@@ -0,0 +1,82 @@
...
+        (0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10,
+            11, 12, 13, 14, 15, 16, 17, 18, 19,
+            20, 21, 22, 23)
+    )
+),
+
+fct_hourly_rt_feed_files AS (
+    SELECT
+
+        {{ farm_surrogate_key(['hourly_totals.dt', 'hourly_totals.base64_url', 'url_map.gtfs_dataset_key']) }} AS key,
```
lauriemerrell, line 61: Would drop `gtfs_dataset_key` from this

warehouse/models/mart/gtfs/fct_hourly_rt_feed_files_success.sql
```diff
@@ -0,0 +1,104 @@
...
+
+    SELECT *
+    FROM
+        (SELECT
+
+            dt,
+            download_hour,
+            base64_url,
+            feed_type,
+            (success_file_count_hour / file_count_hour) * 100 AS pct_success_hour
```
lauriemerrell, line 69: my bad for not being clearer earlier but I do think it might be safer to leave this as just a decimal proportion in the warehouse and lean on Metabase for percentage formatting? so then would change the name to `prop_success_hour` instead of `pct_`

lauriemerrell (changes requested):
Overall looks good, just some nitpicks on language and on reflection I do think proportion is maybe better than percentage (as you mentioned earlier)

lauriemerrell (approved):
👍 👏 🙏

charlie-costanzo:
@lauriemerrell okay ready for you to take a look again except one question - 

the key uniqueness tests are failing with one result for one key, `5396363564373334796`. is this problematic? do you think we should address it t or set up a test threshold or ...?

Screenshot 2022-12-12 at 6 23 45 PM

Issue #2060, opened 2022-12-08
Add new `blob_path` field to GTFS RT outcomes external, staging, intermediate, mart schemas


lauriemerrell:
In #2031, we changed the schema of parsing/validation outcome files to contain a `blob_path` field to account for blobs that exist but have no metadata. We need to update the following locations to account for this new field:

- `create_external_tables`: all the GTFS RT `_outcomes` tables (parsing and validation)
- In dbt:
   - `stg_gtfs_rt__*_outcomes` -- add the `blob_path` field
   - `int_gtfs_rt__unioned_parse_outcomes` -- add the `blob_path` field
   - `fct_daily_rt_feed_files` -- we need to figure out what to do here; should we try to impute the affected feed from the blob path and assign files without metadata to the correct feed, or not? (I think we probably should)

End result should resolve [this Sentry issue](https://sentry.calitp.org/organizations/sentry/issues/17580/) -- we need to either remove this `dataset_key_not_null` test, or do our best to impute affected feeds and then drop rows where dataset key cannot be assigned (I think this might be preferable).

PR #2062, opened 2022-12-08
bugfix: omit null values from rt_feed_guideline_index

owades:
# Description

Two new sentry issues have appeared, related to null values in rt_feed_guideline_index ("RT-FGI").
[1](https://sentry.calitp.org/organizations/sentry/issues/17583/?environment=cal-itp-data-infra&project=2&query=is%3Aunresolved+assigned%3Ascott.owades%40dot.ca.gov&statsPeriod=14d) & [2](https://sentry.calitp.org/organizations/sentry/issues/17582/?environment=cal-itp-data-infra&project=2&query=is%3Aunresolved+assigned%3Ascott.owades%40dot.ca.gov&statsPeriod=14d).

By my count, RT-FGI has 5 null rows for base64_url, and 5 null rows for feed_type. I can resolve by making this change to FGI. I have not explored going further upstream into `fct_daily_rt_feed_files`, but will accept that as a recommendation instead.

Resolves # [issue]

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
I successfully ran `dbt build [int table]+` on one of the RT int tables, with no errors downstream.

## Screenshots (optional)
The "before":
Screen Shot 2022-12-08 at 10 40 12 AM

### DIFF
--- a/warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__rt_feed_guideline_index.sql
+++ b/warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__rt_feed_guideline_index.sql
@@ -12,6 +12,8 @@ int_gtfs_quality__rt_feed_guideline_index AS (
         feed_type,
     FROM fct_daily_rt_feed_files
     WHERE date < CURRENT_DATE
+      AND base64_url IS NOT null
+      AND feed_type IS NOT null
 )
 
 SELECT * FROM int_gtfs_quality__rt_feed_guideline_index


lauriemerrell:
Ugh yeah sorry about this, I am going to defer to @atvaccaro here -- I **think** that this is caused by #2060 which would be the root cause fix. I am not sure if we want to merge this as a stopgap (and maybe make a comment there to revert once root cause addressed) or close this and just wait until we can fix that one.

atvaccaro:
> Ugh yeah sorry about this, I am going to defer to @atvaccaro here -- I **think** that this is caused by #2060 which would be the root cause fix. I am not sure if we want to merge this as a stopgap (and maybe make a comment there to revert once root cause addressed) or close this and just wait until we can fix that one.

I'd probably just prefer to let the test continue to fail than merge code we know we'll have to remove later, but I could be convinced otherwise.

owades:
> > Ugh yeah sorry about this, I am going to defer to @atvaccaro here -- I **think** that this is caused by #2060 which would be the root cause fix. I am not sure if we want to merge this as a stopgap (and maybe make a comment there to revert once root cause addressed) or close this and just wait until we can fix that one.
> 
> I'd probably just prefer to let the test continue to fail than merge code we know we'll have to remove later, but I could be convinced otherwise.

Nope I'm with you! I'll close this out.

PR #2064, opened 2022-12-08
new check for protobuf errors

owades:
# Description
This check confirms that each RT feed has less than 1% of requests getting a protobuf error.

I considered adding an approximate for fetch errors as well, but found that it was unreliable to divide the number of parse success & fails by 4,320 (there were several recent days where all feeds failed)

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Looking at aggregate results, I found a notable pattern: about 66% of checks failed every day, until a few days ago. Starting on 12/5, about 98% of checks pass. If this pattern doesn't make immediate sense to Laurie or Andrew, I will look deeper.

I ran `./dbt.sh test -s int_gtfs_quality__rt_protobuf_error+`, which yielded no errors or warnings.


warehouse/models/mart/gtfs_quality/fct_daily_rt_feed_guideline_checks.md
```diff
@@ -18,4 +18,7 @@ Here is a list of currently-implemented checks:
 | Service alerts RT feed uses HTTPS | Best Practice Alignment | The service alerts RT feed endpoint uses HTTPS instead of HTTPS to ensure feed integrity.|
 |Vehicle positions RT feed uses HTTPS | Best Practice Alignment | The vehicle positions RT feed endpoint uses HTTPS instead of HTTPS to ensure feed integrity.|
 | Trip updates RT feed uses HTTPS | Best Practice Alignment | The trip updates RT feed endpoint uses HTTPS instead of HTTPS to ensure feed integrity.|
+| Fewer than 1% of requests to Trip updates RT feed result in a protobuf error | Best Practice Alignment | On the given day, fewer than 1% of requests to Trip updates RT feed result in a protobuf error.|
```
lauriemerrell, line 21: This is incredibly nitpicky so feel free to disregard but "1% of **requests**" is perhaps not quite accurate here, because I think the denominator you're using is **files**. So if there's a URL where half of requests fail (no file returned) but the protos we do get are 100% parseable, this check will pass.

lauriemerrell (approved):
> Looking at aggregate results, I found a notable pattern: about 66% of checks failed every day, until a few days ago. Starting on 12/5, about 98% of checks pass. If this pattern doesn't make immediate sense to Laurie or Andrew, I will look deeper.

This is related to #2044. @atvaccaro I think we need to run a full refresh on `fct_daily_rt_feed_files`, because even though you ran the backfill this table is incremental and did not get the fixes (still showing parse failures before 12/5.).

warehouse/models/mart/gtfs_quality/fct_daily_rt_feed_guideline_checks.md
owades: How about: 
"On the given day, fewer than 1% of Trip updates RT feed downloads result in a protobuf error."

lauriemerrell: Yes sounds good!

PR #2076, opened 2022-12-12
Break up daily scheduled service index

lauriemerrell:
# Description

This PR:
- Refactors `int_gtfs_schedule__long_calendar` to be more useful by expanding to the actual day level & fixes a bug where Sundays were miscoded
- Creates `int_gtfs_schedule__all_scheduled_service`, which will support @owades on #2047 
- Fixes bugs in `int_gttfs_schedule__daily_scheduled_service_index`: one bug had Sundays miscoded so no `calendar` based Sunday service was appearing, and the other affected cases where a feed used both `calendar` and `calendar_dates` but had some `service_ids` that were unique to `calendar_dates`

TODO:
- [x] Talk to @owades 
- [x] If we want to do this, update the existing `_index` table accordingly 
- [x] Do we want to materialize as a table? 🤔 
- [x] Add YAML
- [x] ~Test warehouse image per @atvaccaro~ removed updates to poetry lockfile because it was causing unrelated issues

Resolves # [issue]

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
`dbt run -s +int_gtfs_schedule__all_scheduled_service+`
`dbt test -s +int_gtfs_schedule__all_scheduled_service+`


owades:
Thanks @lauriemerrell. Is the idea that this would go into the making of [int_gtfs_schedule__daily_scheduled_service_index](https://dbt-docs.calitp.org/#!/model/model.calitp_warehouse.int_gtfs_schedule__daily_scheduled_service_index)?

lauriemerrell:
@owades yes, the idea is that this creates a new `int_gtfs_schedule__all_scheduled_service` table (which is **not** filtered based on which feed is in effect on a given day) that your check and the existing index table can both consume. Your check can do the version over version comparison and the existing index table can just filter to only the feed versions we want to keep.

warehouse/models/intermediate/gtfs/int_gtfs_schedule__long_calendar.sql
```diff
@@ -1,30 +1,32 @@
...
-            CAST({{ dow[0] }} AS boolean) AS has_service
-        FROM dim_calendar
-    {% endfor %}
+    SELECT
+        feed_key,
+        {{ dbt_utils.surrogate_key(['feed_key', 'service_id', 'dt']) }} AS key,
+        service_id,
+        dt AS service_date,
+        EXTRACT(DAYOFWEEK FROM dt) AS day_num,
+        CASE
```
atvaccaro, line 16: nit: technically you could put the `case` inside the `cast` for a bit less verbosity

warehouse/models/intermediate/gtfs/_int_gtfs.yaml
```diff
@@ -88,30 +88,33 @@ models:
...
-        description: String name of day of the week like "monday"
-      - name: has_service
+      - name: service_date
+        description: |
+          Date on which this service was active (i.e., this date is betweem the
+          `start_date` and `end_date` for this service).
+      - name: day_num
+        description: |
+          Day of week as number (Sunday = 1, Saturday = 7).
+      - name: service_bool
```
atvaccaro, line 112: should this be not-null?

atvaccaro (approved):

owades (approved):
Thank you! Looks good.

warehouse/models/intermediate/gtfs/int_gtfs_schedule__long_calendar.sql
lauriemerrell: checked offline, decided not worth it

PR #2081, opened 2022-12-13
Lookback var rt feed files

lauriemerrell:
# Description

This PR updates a few RT archiver monitoring tables to use our incremental lookback variable.

Doing this because of behavior noted in #2064, which is that we did not `--full-refresh` after #2044 merged. After this PR merges I will rerun the `fct_daily_rt_feed_files` table with a lookback period to `2022-11-07`, which is how far the table currently goes.

image

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

`dbt run` locally


atvaccaro (approved):
we could consider making several for different categories of data

PR #2087, opened 2022-12-15
Assessed organization & service indices

### LINKED ISSUE #1904: V2 validations & guidelines tables

lauriemerrell:
# Description

**UPDATE: Per offline discussion with @atvaccaro, some of this needs to be refactored in favor of proper historical dimensions for Airtable. I have reopened #1868 and will do that refactor there. However, per @natam1 and @juliama0780 in Slack they would like to be able to start experimenting with the new tables, understanding that they are in draft form. So I'd like to merge this as-is and do a follow-on.**

This PR creates core functionality needed for the GTFS guideline checks work:
* A `fct_daily_guideline_checks` table that creates every possible assessed entity (organization / service / GTFS service data / GTFS dataset / URL / downloaded schedule feed) for a given date and appends results of all checks performed on that entity. 
* `fct_daily_organization_combined_guideline_checks` and `fct_daily_service_combined_guideline_checks` tables that aggregate those "whole entity" checks to the organization and service levels respectively

It also semi-incidentally:
- Removes the fake historicization in `dim_gtfs_datasets` (i.e., gets rid of the `key` / `airtable_record_id` distinction, removes the placeholder `_valid_from` and `_valid_to`)
- Fixes [this issue](https://sentry.calitp.org/organizations/sentry/issues/4516/) (just because this test failed while I was testing my changes and I realized the test was misdefined)

I am slightly punting on full documentation; deferred to #2092.

Resolves #1904 

**Questions for @atvaccaro in Scott's absence:**
- On 2022-11-11 the `gtfs service data` table did not extract successfully from Airtable, so right now that data is just misaligned for that date (GTFS datasets no longer match up with services and thus organizations on that date). I think the simplest answer is just to leave it, but we could try complicated imputation if we need. (That would take us more in the direction of constructing a more SCD type table.)
- How do we feel about historical `*_name` values in `fct_daily_guideline_checks`? Should it just be historical record IDs with current names? 🤔 Currently this PR just uses name as it was on the historical date. We have exactly one organization record and two services whose names have changed while keeping record ID the same.
image 
image

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

`dbt run` and `dbt test` on new tables

### DIFF
--- a/warehouse/models/staging/transit_database/stg_transit_database__gtfs_service_data.sql
+++ b/warehouse/models/staging/transit_database/stg_transit_database__gtfs_service_data.sql
@@ -13,6 +13,8 @@ stg_transit_database__gtfs_service_data AS (
         unnested_services AS service_key,
         unnested_gtfs_dataset AS gtfs_dataset_key,
         dataset_type,
+        -- only coalesce to false after the field had been created (November 23, 2022)
+        -- otherwise a null is genuinely a null
         CASE
             WHEN dt >= '2022-11-23' THEN COALESCE(customer_facing, FALSE)
         END AS customer_facing,
--- a/warehouse/models/staging/transit_database/stg_transit_database__gtfs_service_data.sql
+++ b/warehouse/models/staging/transit_database/stg_transit_database__gtfs_service_data.sql
@@ -13,7 +13,9 @@ stg_transit_database__gtfs_service_data AS (
         unnested_services AS service_key,
         unnested_gtfs_dataset AS gtfs_dataset_key,
         dataset_type,
-        COALESCE(customer_facing, FALSE) AS customer_facing,
+        CASE
+            WHEN dt >= '2022-11-23' THEN COALESCE(customer_facing, FALSE)
+        END AS customer_facing,
         category,
         agency_id,
         network_id,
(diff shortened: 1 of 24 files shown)


lauriemerrell:
Sort arrays alphabetically

warehouse/models/staging/transit_database/stg_transit_database__gtfs_service_data.sql
```diff
@@ -13,7 +13,9 @@ stg_transit_database__gtfs_service_data AS (
         unnested_services AS service_key,
         unnested_gtfs_dataset AS gtfs_dataset_key,
         dataset_type,
-        COALESCE(customer_facing, FALSE) AS customer_facing,
+        CASE
+            WHEN dt >= '2022-11-23' THEN COALESCE(customer_facing, FALSE)
```
atvaccaro, line 17: nit: could you add a comment

atvaccaro (approved):

atvaccaro:
Oh I assume the intermediate tables are setting the stage for historical Airtable models?

PR #2090, opened 2022-12-16
Add check for persistent ids in schedule

owades:
# Description

A new check to measure the GTFS Guideline:
> Maintain persistent identifiers (id fields) for stop_id, route_id, and agency_id across data iterations whenever possible.

To capture this, the check measures, for stops, routes & agencies, what percent of a given feed's IDs were not present in the previous version of the feed. If more than half of stop_ids, route_ids or agency_Ids were not present in the previous version, and this change happened anytime in the least 30 days, the check will fail. Otherwise it will pass.

My thought process is that it is uncommon for transit service to double their number of any of these items overnight, so seeing a large percent of new IDs is more likely going to be a failure to persist the ID values, rather than a real addition (I suppose there could be a much more detailed check to confirm some of this).

I expect the final business logic might shift when I review with teammates, and so I left some of the business logic calculation to one of the last queries so it's easy to update (I'm still interested in business logic feedback, though). 

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
`./dbt.sh build -s int_gtfs_quality__persistent_ids_schedule+`
`./dbt.sh build -s fct_implemented_checks`

I looked at edge case examples to see whether the logic was working correctly. Chasing down an example where 100% of agency_id's changed, I found an instance in feed `base64_url=aHR0cDovL2lwb3J0YWwuc2FjcnQuY29tL0dURlMvU1JURC9nb29nbGVfdHJhbnNpdC56aXA=` where on 10/7 the only agency_id was `SRTD` and on 10/8 the only agency was `Roseville ` (it was resolved on 10/12).


warehouse/models/mart/gtfs_quality/fct_daily_schedule_feed_guideline_checks.sql
```diff
@@ -17,7 +17,9 @@ unioned AS (
             ref('int_gtfs_quality__no_7_day_feed_expiration'),
             ref('int_gtfs_quality__no_30_day_feed_expiration'),
             ref('int_gtfs_quality__no_expired_services'),
+            ref('int_gtfs_quality__persistent_ids_schedule'),
```
lauriemerrell, line 20: Added twice?

warehouse/models/staging/gtfs/gtfs_quality/stg_gtfs_quality__intended_checks.sql
```diff
@@ -56,7 +56,11 @@ WITH stg_gtfs_quality__intended_checks AS (
     UNION ALL
     SELECT {{ passes_fares_validator() }}, {{ fare_completeness() }}
     UNION ALL
+    SELECT {{ persistent_ids_schedule() }}, {{ best_practices_alignment_schedule() }}
+    UNION ALL
     SELECT {{ lead_time() }}, {{ up_to_dateness() }}
+    UNION ALL
+    SELECT {{ persistent_ids_schedule() }}, {{ best_practices_alignment_schedule() }}
```
lauriemerrell, line 63: Also added twice here?

lauriemerrell (commented):
Two inline catches that I think need to be changed for sure (seems like the check is added twice?)

Otherwise, I think this does conceptually make sense but I think it could maybe be tightened up a bit for efficiency/readability (make sure we aren't pulling any extra columns, and perhaps try to make the big logic more space efficient, maybe using a macro.) 

My other thought here is -- is there any chance that it's worth persisting the `feed, next_feed` stuff, if you use the same thing here and in the 7 days warning check? Maybe that can be its own intermediate table instead of using same CTE twice in different checks. (In that case, the extra columns are ok because I think that you might need the `next_feed_valid_from` column for the 7 days check?)

Re: business logic, 50% maybe even seems high to me -- I would be cool with like 20-30%, but would also defer to the larger group of stakeholders on that.

warehouse/models/mart/gtfs_quality/fct_daily_schedule_feed_guideline_checks.sql
owades: oops, fixed!!

warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
```diff
@@ -232,3 +232,15 @@ models:
             - base64_url
       - dbt_utils.equal_rowcount:
           compare_model: ref('int_gtfs_quality__schedule_url_guideline_index')
+    columns: *stg_gtfs_guideline_columns
```
owades, line 235: looks like I had lost this line in an old merge

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__persistent_ids_schedule.sql
```diff
@@ -0,0 +1,55 @@
...
+    SELECT  * FROM {{ ref('dim_agency') }}
+),
+
+feed_version_history AS (
+    SELECT * FROM {{ ref('int_gtfs_quality__feed_version_history') }}
+),
+
+stop_id_comparison AS ( {{ ids_version_compare_aggregate("stop_id","dim_stops") }} ),
+route_id_comparison AS ( {{ ids_version_compare_aggregate("route_id","dim_routes") }} ),
+agency_id_comparison AS ( {{ ids_version_compare_aggregate("agency_id","dim_agency") }} ),
```
atvaccaro, line 24: only nit: I'd probably format this as the other top-level select statements

warehouse/macros/define_gtfs_guidelines.sql
```diff
@@ -191,3 +195,61 @@ check,
...
+           COUNT(CASE WHEN prev_id IS null THEN 1 END) AS id_added,
+           -- Previous id's removed
+           COUNT(CASE WHEN id IS null THEN 1 END) AS id_removed
+      FROM ids_version_compare
+     GROUP BY 1,2
+    HAVING ids_current_feed > 0
+{% endmacro %}
+
+-- For use in int_gtfs_quality__persistent_ids_schedule:
+{% macro max_new_id_ratio(table_name) %}
```
atvaccaro, line 248: this one is maybe a little short to justify a macro but I'm not opposed

lauriemerrell (approved):
Agree with Andrew's comments but I think they are both optional/don't require re-review

PR #2101, opened 2022-12-21
Added prod service account to payments_rides row access policy

charlie-costanzo:
# Description

In #2042 we added three dbt custom tests in to `payments_rides` to monitor daily, weekly, and monthly transaction deltas - but the row access policy on the table prevented the tests from being run correctly.

To resolve this, we added the prod service account to the row access policy so that it can query the table and run the tests appropriately.

## Type of change

## How has this been tested?
ran with prod service account

### DIFF
--- a/warehouse/models/payments_views/payments_rides.sql
+++ b/warehouse/models/payments_views/payments_rides.sql
@@ -33,7 +33,7 @@
 ]
 
 ) }}
--- TODO: In the last policy of the macro call above, see if we can get the prod service account out of context
+-- TODO: In the last policy of the macro call above, see if we can get the prod warehouse service account out of context
 
 WITH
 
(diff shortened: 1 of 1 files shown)


warehouse/models/payments_views/payments_rides.sql
```diff
@@ -22,6 +22,7 @@
 ) }}",
 " {{ create_row_access_policy(
     principals = ['serviceAccount:metabase@cal-itp-data-infra.iam.gserviceaccount.com',
```
charlie-costanzo, line 24: @atvaccaro is there any way to safely add a comment here? but also - I took a look at the dbt docs and [it looks like bigquery doesn't allow you to get user from the target context](https://docs.getdbt.com/reference/dbt-jinja-functions/target#bigquery)

atvaccaro: I would just add a SQL comment outside of the body of the macro call

atvaccaro (approved):

Issue #2109, opened 2023-01-02
Turn off/disable v1 pipeline for RT and schedule


atvaccaro:
## Services tasks

- [x] Tear down `gtfsrt-v3` node pool
- [x] Delete `gtfsrt-v3` k8s yaml (https://github.com/cal-itp/data-infra/pull/2110)
- [x] Delete `gtfsrt-v3` application code (https://github.com/cal-itp/data-infra/pull/2110)

## Airflow tasks

- [x] Pause legacy RT parsing/validation DAG
- [x] Pause legacy Schedule download/validation DAG
- [x] Delete legacy parsing/validation DAGs (https://github.com/cal-itp/data-infra/pull/2111)
- [x] Delete legacy external table tasks (https://github.com/cal-itp/data-infra/pull/2111)

## dbt tasks

- [ ] ~Disable legacy RT source and downstream models~
- [ ] ~Disable legacy Schedule source and downstream nodes~

We are just going to disable rather than delete for now; this should preserve the docs but indicate the data is stale. We cannot do this until publishing at least is migrated, since disabling them will raise compilation errors if they are still used.

lauriemerrell:
see also #1355

lauriemerrell:
Priority here is to reduce costs by turning off *ingest*, not to fully turn off access to v1 data (because data consumer migrations are not yet complete). 

Order of operations:
1. Turn off things in non-code ways (i.e., pause Airflow DAGs etc.)
2. One PR per heading in issue description actually *deleting* associated code (i.e., *delete* Airflow DAGs)
3. BigQuery stuff (#1355 and potentially moving tables to new locations) can wait

atvaccaro:
I think we can consider this done?! Everything is disabled/paused at this point.

PR #2111, opened 2023-01-02
Delete legacy parsing/validation DAGs

atvaccaro:
# Description

Progress towards https://github.com/cal-itp/data-infra/issues/2109

Deletes the following
1. Legacy RT parsing/validation DAG and associated image/CI
2. Legacy Schedule downloading/validation DAG and downstream DAGs
3. Legacy RT external tables

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Schedule DAGs are currently paused; waiting to pause RT until December is done.


docs/_toc.yml
```diff
@@ -89,7 +89,6 @@ parts:
       - file: services/overview
         sections:
         - file: services/gtfs-ckan-uploader
-        - file: services/gtfs-rt-archive
```
lauriemerrell, line 92: Do you need to rebase? I thought you already deleted this in #2110

lauriemerrell (changes requested):
Diff is kind of hard to parse compared with #2110, want to make it clearer which changes are actually new here vs. artifact of branching

atvaccaro:
> Diff is kind of hard to parse compared with #2110, want to make it clearer which changes are actually new here vs. artifact of branching (cited one inline example but I think there are others)

Ah yeah sorry, I forgot to rebase after merging the other PR

lauriemerrell (approved):
Seems ok to me, only question is whether there's stuff in `sandbox` that we want to delete also? Low priority

Issue #1355, opened 2022-04-07
User Story: Process to flag and/or delete obsolete tables


lauriemerrell:
## User stories

As a data scientist, I want to be able to delete unused tables in the BigQuery warehouse so that the production datasets only contain current, active tables (to reduce clutter and make it clear which tables are usable.)

Some considerations that must be decided:

- [ ] "Soft-delete": We likely want to make a holding dataset (for example `deprecated`) and some naming convention for how to move tables out of their prod locations
- [ ] "Hard delete": Do we want to set an expiration date on the soft-delete location so that unused tables are ultimately hard-deleted?

### Acceptance Criteria

To resolve this ticket we should,

- [ ] Define a process that can be used to remove deprecated tables from production BigQuery datasets
- [ ] Record that process in a central location (likely docs.calitp.org)
- [ ] Ensure that the process allows us to resolve #1026 (which can then be resolved one by one outside this ticket based on the new process)

# Sprint Ready Checklist
1. - [ ] Acceptance criteria defined
2. - [ ] Team understands acceptance criteria
3. - [ ] Team has defined solution / steps to satisfy acceptance criteria
4. - [ ] Acceptance criteria is verifiable / testable
5. - [ ] External / 3rd Party dependencies identified

atvaccaro:
Another consideration: what should we do about disabled/deleted models whose materialized artifacts may still be available in metabase or jupyter? We should probably flag them at minimum in some manner, such as Metabase description strings.

lauriemerrell:
Considerations I am aware of now (partial restate of @atvaccaro's point above):

* Models can be deleted in dbt project but persist in BQ warehouse. 
* Metabase can access tables in BQ that are not in dbt (Metabase sync is at the dataset/schema level, ex `mart_gtfs`). 
* JupyterHub can access tables in BQ that are not in dbt.


I think the goal here is to remove deprecated models from places that are accessible to Metabase / JupyterHub, not just to document. Perhaps there is a 2-tiered thing where we first document and then deprecate, but I think the goal is that non-maintained tables are not available in consumer-facing tools.

SorenSpicknall:
Here's a suggestion for basic process:

**Phase 1**
Identify deletion candidates and group into two categories - in active use, and not in active use. We can define an arbitrary activity cutoff to make this distinction, like two weeks since the last query.

**Phase 2**
[...]

PR #2112, opened 2023-01-02
Disable legacy dbt models and migrate tile publishing

atvaccaro:
# Description

Progress towards https://github.com/cal-itp/data-infra/issues/2109

TODO:

- [x] Change reports exposure when implemented (https://github.com/cal-itp/data-infra/pull/2204)
- [x] Test map tiles publishing
- [x] Change CKAN publishing when implemented (https://github.com/cal-itp/data-infra/pull/2199)
- [x] Move payments_rides over and test

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Tested map tiles work. payments_rides passes to the same extent as prod.

### DIFF
--- a/warehouse/models/payments_views/payments_rides.sql
+++ b/warehouse/models/payments_views/payments_rides.sql
@@ -37,14 +37,16 @@
 
 WITH
 
-gtfs_schedule_dim_routes AS (
-    SELECT *
-    FROM {{ ref('gtfs_schedule_dim_routes') }}
+dim_routes AS (
+    SELECT * FROM {{ ref('dim_routes') }}
+),
+
+dim_gtfs_datasets AS (
+    SELECT * FROM {{ ref('dim_gtfs_datasets') }}
 ),
 
 payments_feeds AS (
-    SELECT *
-    FROM {{ ref('payments_feeds') }}
+    SELECT * FROM {{ ref('payments_feeds') }}
 ),
 
 stg_cleaned_micropayments AS (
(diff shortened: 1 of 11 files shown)


warehouse/models/payments_views/payments_rides.sql
```diff
@@ -98,10 +100,11 @@ gtfs_routes_with_participant AS (
...
         g.calitp_deleted_at
 
-    FROM gtfs_schedule_dim_routes AS g
+    FROM dim_routes AS g
+    LEFT JOIN dim_gtfs_datasets AS d
+        ON g.base64_url = d.base64_url
     INNER JOIN payments_feeds AS p
-        ON g.calitp_itp_id = p.calitp_itp_id
-            AND g.calitp_url_number = p.calitp_url_number
+        ON d.key = p.gtfs_dataset_key
```
atvaccaro, line 107: this needs to change I think

lauriemerrell (approved):
I think it would be good for us to go through and actually add something to the table descriptions of the things that are no longer being updated to make it more explicit. I don't think you can see table last update time in Metabase. Not sure if it's worth doing that as this merges or if we should make a separate ticket. There's already #1355. I'm just worried that things are becoming stale in a way that may not be super visible to consumers.

atvaccaro:
> I think it would be good for us to go through and actually add something to the table descriptions of the things that are no longer being updated to make it more explicit. I don't think you can see table last update time in Metabase. Not sure if it's worth doing that as this merges or if we should make a separate ticket. There's already #1355. I'm just worried that things are becoming stale in a way that may not be super visible to consumers.

I'll modify that obsolete table ticket to include flagging deprecated/disabled tables in Metabase, if that seems reasonable.

Issue #2022, opened 2022-11-29
Create models and/or dashboards to track BigQuery usage and costs


atvaccaro:
**Is your feature request related to a problem? Please describe.**
There's some initial work towards visualizing our BigQuery costs, but they should be reviewed and tweaked to give us actionable insights.

Model - https://dbt-docs.calitp.org/#!/model/model.calitp_warehouse.stg_audit__cloudaudit_googleapis_com_data_access
Dashboard - https://dashboards.calitp.org/dashboard/76-bigquery-overview-dashboard

**Describe the solution you'd like**
Metabase dashboards should use a new fact table rather than staging; we should try to clean up data labeling and ensure that dashboards can be visualized by both user and model.

**Describe alternatives you've considered**
n/a

**Additional context**
dbt projects are visualizable using networkx and/or graphviz; it could be cool to show expensive nodes in the DAG.

PR #2116, opened 2023-01-03
Availability of Google Cloud data access logs in mart

SorenSpicknall:
# Description

In order to facilitate more detailed dashboarding for infra costs, this PR seeks to republish the entirety of `stg_audit__cloudaudit_googleapis_com_data_access` in mart, where it'll be usable by Metabase.

This contributes to, but does not yet fully resolve, #2022.

## Type of change

- [x] New feature
- [x] Documentation
- [ ] agencies.yml

## How has this been tested?

At the time of writing, table creation has been tested in the staging project via locally issued dbt commands.


warehouse/models/mart/audit/fct_bigquery_data_access.sql
```diff
@@ -0,0 +1,33 @@
...
+        job_type,
+        dbt_invocation_id,
+        create_disposition,
+        destination_table,
+        priority,
+        query,
+        statement_type,
+        write_disposition,
+        duration_in_seconds,
+        referenced_tables,
```
atvaccaro, line 20: can you either unnested in this model or create another table on top that unnests? I think seeing a rough breakdown of data volumes by queried table (i.e. read not write) may be useful; a table that is expensive to build but efficient to query is probably preferable for most of our uses cases.

atvaccaro (changes requested):

atvaccaro (changes requested):
I think even the unnested model seems like a fact? maybe `fct_bigquery_referenced_tables` and then you could add a caveat in the documentation about how it will overcount

SorenSpicknall:
> I think even the unnested model seems like a fact? maybe `fct_bigquery_referenced_tables` and then you could add a caveat in the documentation about how it will overcount

Totally right there - I gave it that name before actually deciding how to structure the table. I'll make that change in the next commit.

atvaccaro (approved):
lgtm!

PR #2117, opened 2023-01-04
Added data_quality_pipeline column to int_gtfs_rt__daily_url_index

charlie-costanzo:
# Description
In the new RT archiver dashboard tables we were pulling in feeds that are marked as "false" for the data quality pipeline, resulting in empty rows upon join. To resolve this, we added the `data_quality_pipeline` column to `int_gtfs_rt__daily_url_index`, and then filtered the two `fct_hourly_rt_feed_files*` tables where that column is TRUE, also updated yml files

Files changed:
* `int_gtfs_rt__daily_url_index.sql`
  * Added the column `use_in_data_quality_pipeline` from `stg_transit_database__gtfs_datasets`
* `fct_hourly_rt_feed_files.sql`
  * Filtered for the new column `use_in_data_quality_pipeline` on import of `int_gtfs_rt__daily_url_index.sql` where it is `TRUE`
* `fct_hourly_rt_feed_files_success.sql`
  * Filtered for the new column `use_in_data_quality_pipeline` on import of `int_gtfs_rt__daily_url_index.sql` where it is `TRUE`

yml docs updates:
* `_stg_transit_database.yml`
* `_int_gtfs.yaml`

## Type of change

## How has this been tested?
locally with `dbt run` and `dbt test`, and in a test Metabase dashboard

### DIFF
--- a/warehouse/models/mart/gtfs/fct_hourly_rt_feed_files.sql
+++ b/warehouse/models/mart/gtfs/fct_hourly_rt_feed_files.sql
@@ -12,7 +12,7 @@ WITH
 int_gtfs_rt__daily_url_index AS (
     SELECT *
     FROM {{ ref('int_gtfs_rt__daily_url_index') }}
-    WHERE use_in_data_quality_pipeline
+    WHERE data_quality_pipeline
 
 ),
 
--- a/warehouse/models/mart/gtfs/fct_hourly_rt_feed_files.sql
+++ b/warehouse/models/mart/gtfs/fct_hourly_rt_feed_files.sql
@@ -12,6 +12,8 @@ WITH
 int_gtfs_rt__daily_url_index AS (
     SELECT *
     FROM {{ ref('int_gtfs_rt__daily_url_index') }}
+    WHERE data_quality_pipeline IS TRUE
+
 ),
 
 int_gtfs_rt__unioned_parse_outcomes AS (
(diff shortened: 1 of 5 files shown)


warehouse/models/mart/gtfs/fct_hourly_rt_feed_files.sql
```diff
@@ -12,6 +12,8 @@ WITH
 int_gtfs_rt__daily_url_index AS (
     SELECT *
     FROM {{ ref('int_gtfs_rt__daily_url_index') }}
+    WHERE data_quality_pipeline IS TRUE
```
lauriemerrell, line 15: nit: because this is a boolean column you can just say `WHERE data_quality_pipeline` (don't need the `IS TRUE`). It's probably fine to leave as-is given that `data_quality_pipeline` doesn't have the clearest name (like, it might be better if it were called `use_in_data_quality_pipeline`, but that's an upstream issue)

lauriemerrell (approved):

lauriemerrell (commented):
I think in general we prefer to do column renaming in the staging table, just to avoid confusion (like, having the same column in multiple intermediate tables with different names). So I probably actually wouldn't actually rename here, sorry for being unclear about it (and I don't think it's worth it to rename the upstream column in this PR just in case it's used other places and it leads to a rabbit hole/scope creep)

lauriemerrell (approved):

lauriemerrell (approved):
🙏

PR #2118, opened 2023-01-04
Change check names from clickable blue to plain-text

owades:
# Description

As discussed with @lauriemerrell, try changing metabase semantic type so that check names won't be clickable.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

It hasn't - looking for guidance on whether this is OK to merge. Worst

## Screenshots
current state:
Screen Shot 2023-01-03 at 4 22 01 PM


lauriemerrell (approved):
I tested manually in Metabase (by just manually changing the data type to `title` for this) and confirmed this looks like it will work 👍 
image

PR #2121, opened 2023-01-04
add line number to schedule jsonl files

atvaccaro:
# Description

The schedule backfill has caused many dbt tests to start failing, for example [fare rules having duplicate rows](https://sentry.calitp.org/organizations/sentry/issues/4509/?environment=cal-itp-data-infra&project=2&query=is%3Aunresolved+dbt&statsPeriod=14d). We don't really want to filter these out immediately, but we do want to ensure that we're ending up with 1 record per row in the original extract file. Adding a line number to the CSV-to-JSONL conversion task will let us check for a unique combination of ts, url, and line number.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Local Airflow.

### DIFF
--- a/airflow/dags/create_external_tables/gtfs_schedule_v2/agency.yml
+++ b/airflow/dags/create_external_tables/gtfs_schedule_v2/agency.yml
@@ -9,6 +9,9 @@ hive_options:
   mode: CUSTOM
   source_uri_prefix: "agency/{dt:DATE}/{ts:TIMESTAMP}/{base64_url:STRING}/"
 schema_fields:
+  - name: _line_number
+    mode: NULLABLE
+    type: INTEGER
   - name: agency_id
     type: STRING
   - name: agency_name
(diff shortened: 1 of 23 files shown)


lauriemerrell (changes requested):
Can you add this to the external tables in this PR too? 

How long do you think it will take / cost to backfill this through all of history? I am still not very sold on backfill, I am wondering if we can add tests just from when this is added forward. If there is structural fanout that should still catch it.

atvaccaro:
> How long do you think it will take / cost to backfill this through all of history? I am still not very sold on backfill, I am wondering if we can add tests just from when this is added forward. If there is structural fanout that should still catch it.

I don't think it'll be too bad; the worst are stop_times (~30 min/day) and shapes (~10 min/day) and we don't have to re-run validate or unzip tasks. (stats from https://o1d2fa0877cf3fb10p-tp.appspot.com/duration?dag_id=unzip_and_validate_gtfs_schedule&days=30)

atvaccaro:
I think we can merge this and just selectively backfill when we want, from the day of merge. (Trying to close out PRs)

lauriemerrell (approved):

Issue #2043, opened 2022-12-05
User Story: mart tables for base schedule tables for analysts


tiffanychu90:
## User stories

A user story is implemented as well as it is communicated.
If the context and the goals are made clear, it will be easier for everyone to implement it, test it, refer to it.

---

### Summary

A user story should typically have a summary structured this way:

As an analyst, I want a pre-joined table for GTFS schedule data for the most common tables so I have an easy way to access assembled tables in a standardized way (rename these mart tables if needed): 
- [ ] `feed_key` to `organization name`: new under v2 warehouse
- [x] `trips`
- [x] `shapes`
- [x] `stops`
- [ ] `stop_times`...maybe..?

### Acceptance Criteria

This is something that can be verified to show that this user story is satisfied.

**Example:**

* [gtfs_utils for v2](https://github.com/cal-itp/data-analyses/blob/main/_shared_utils/shared_utils/gtfs_utils_v2.py)
* [Sample dbt SQL](https://github.com/cal-itp/data-analyses/blob/main/_shared_utils/shared_utils/dbt_for_gtfs_utils_v2.sql) -- I tried writing this similar to what I saw in `dbt-docs`, so hopefully it's mostly there? @lauriemerrell 

Also, here are a few points that need to be addressed:

### Notes

* could these incrementally updated everyday?
* https://github.com/cal-itp/data-analyses/pull/570

### Tester [Stakeholder]

1. @tiffanychu90 

# Sprint Ready Checklist
1. - [ ] Acceptance criteria defined
2. - [ ] Team understands acceptance criteria
3. - [ ] Team has defined solution / steps to satisfy acceptance criteria
4. - [ ] Acceptance criteria is verifiable / testable
5. - [ ] External / 3rd Party dependencies identified

holly-g:
Hi @tiffanychu90 , thanks for submitting this ticket! We'll tee this up for our next sprint (12/13 - 12/23). 
@lauriemerrell / @charlie-costanzo will reach out for a brief call with you to ask some clarifying questions before this gets picked up in the next sprint.

lauriemerrell:
Notes from scoping conversation with @tiffanychu90 and @charlie-costanzo : https://docs.google.com/document/d/1DJpRpvnCPXhwwDHbFuVtckRBI1TeUvQ7SinaeySGKlM/edit#

charlie-costanzo:
Closed in favor of continuing the remaining table in #2145

PR #2123, opened 2023-01-04
Schedule mart tables for analysts

charlie-costanzo:
# Description

V2 warehouse table additions at the request of analysts in issue #2043 

Tables added:
* `fct_daily_scheduled_stops`
* `fct_daily_scheduled_shapes`

Tables added:
* `fct_daily_scheduled_trips`
* `dim_stops`

## Type of change

- [x] New feature

## How has this been tested?
local `dbt run` and `dbt test`


lauriemerrell (changes requested):
Changes from our conversation look good, I think you need to add YAML for the new tables? (Sorry if this wasn't ready for review yet)

warehouse/models/mart/gtfs/fct_daily_scheduled_stops.sql
```diff
@@ -0,0 +1,76 @@
...
+),
+
+dim_routes AS (
+    SELECT *
+    FROM {{ ref('dim_routes') }}
+),
+
+
+stops_by_day AS (
+
```
lauriemerrell, line 34: Ok sorry I missed this before, I think that we may want to join with `dim_stops` here and group by `stop_key` rather than just by `stop_id` 🤔 Because that way people can do a versioned join if they need to? I kind of want @atvaccaro's opinion here

warehouse/models/mart/gtfs/fct_daily_scheduled_stops.sql
```diff
@@ -0,0 +1,107 @@
...
+        stops.stop_code,
+        stops.stop_name,
+        stops.stop_desc,
+        stops.location_type,
+        stops.stop_timezone,
+        stops.wheelchair_boarding
+
+    FROM pivot_to_route_type AS pivot
+    LEFT JOIN dim_stops AS stops
+        ON pivot.stop_id = stops.stop_id
```
lauriemerrell, line 100: You need `feed_key` here

warehouse/models/mart/gtfs/fct_daily_scheduled_stops.sql
```diff
@@ -0,0 +1,107 @@
...
+    )
+),
+
+fct_daily_scheduled_stops AS (
+    SELECT
+
+        pivot.service_date,
+        pivot.feed_key,
+        pivot.stop_id,
+
```
lauriemerrell, line 77: I would add `stops.key AS stop_key` here

warehouse/models/mart/gtfs/fct_daily_scheduled_stops.sql
```diff
@@ -0,0 +1,107 @@
...
+        stops_by_day.stop_event_count,
+
+        IFNULL(pivot.route_type_0, 0) AS route_type_0_count,
```
lauriemerrell, line 80: Per convo let's add route types 6, 7, 12

warehouse/models/mart/gtfs/fct_daily_scheduled_stops.sql
lauriemerrell: Per discussion we will add it later, not here

lauriemerrell (changes requested):
Changes we discussed

warehouse/models/mart/gtfs/_mart_gtfs.yml
```diff
@@ -1453,3 +1486,70 @@ models:
       - name: gtfs_dataset_key
         description: |
           The primary key for the record in `dim_gtfs_datasets` associated with this feed.
+  - name: fct_daily_scheduled_shapes
+    description: |
+      A daily table showing all shapes that were scheduled on a given date.
```
lauriemerrell, line 1491: I would say: `showing shapes that had trips scheduled on a given date`

warehouse/models/mart/gtfs/_mart_gtfs.yml
```diff
@@ -1453,3 +1486,70 @@ models:
...
+          Foreign key to the `dim_schedule_feeds` table.
+      - name: service_date
+        description: Date on which this shape was active.
+      - name: shape_array_key
+        description: Foreign key to dim_shapes_arrays.
+      - name: pt_array
+        description: Ordered array of WKT points that describe this shape.
+  - name: fct_daily_scheduled_stops
+    description: |
+      A daily table showing all stops that were scheduled on a given date.
```
lauriemerrell, line 1506: `showing all stops that had trip stop events scheduled on a given date`

lauriemerrell (commented):
two more comments, discussed live

warehouse/models/mart/gtfs/fct_daily_scheduled_stops.sql
```diff
@@ -0,0 +1,107 @@
...
+    PIVOT(
+        SUM(stop_event_count) AS route_type
+        FOR route_type IN
+        (0, 1, 2, 3, 4, 5, 11)
+    )
+),
+
+fct_daily_scheduled_stops AS (
+    SELECT
+
```
lauriemerrell, line 73: Let's add a surrogate key for this and shapes mirroring what we have for trips and add the same tests as well

lauriemerrell:
A few more comments based on investigation of failing uniqueness tests: 

* Sorry for any confusion earlier -- can we keep `shape_id` and `stop_id` *and* `*_key` on the final tables (not just `key` -- including `shape_id` on `fct_daily_scheduled_trips`)? Ideally people are not joining but it is helpful as a more human-readable identifier.
* As we discussed, it seems that the failing tests for shapes are because of `null` `shape_array_key` values. Requested solution: 
   * Add `shape_id` to `fct_daily_scheduled_trips` (add it from `dim_trips`, *not* from `dim_shapes_arrays` to account for cases where a shape ID was present but did not resolve to an array)
   * Add `shape_id` to the group by and filter out rows where shape ID is null: 
 ```
trips_counted AS (

    SELECT

        COUNT(DISTINCT trip_key) AS n_trips,
        feed_key,
        service_date,
        shape_id, 
        shape_array_key

    FROM fct_daily_scheduled_trips
    WHERE shape_id IS NOT NULL
    GROUP BY feed_key, service_date, shape_id, shape_array_key
)
```
* For the duplicates in `fct_daily_scheduled_stops`, you're going to need to either add a second grouping CTE to collapse the route types OR calculate `stop_event_count` by summing the pivoted route type columns. It seems like a second grouping CTE Is preferable. (Current `stops_by_day` CTE is actually grouping to the `stop/route_type` level rather than the pure `stop` level which is correct for what you need to pivot but for the final `stop_event_count` column you want pure `stop` without `route_type`. [Notes also in spreadsheet](https://docs.google.com/spreadsheets/d/1Aj4W7uUzhsA1lk_aNMdwOjkmfAFUIvXv3GVgQYzaCqM/edit#gid=0).)
* We should add the warning columns about duplicate primary keys from trips and stop times [like in `fct_daily_scheduled_service` ](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/gtfs/fct_daily_scheduled_service_summary.sql#L22-L30)

**Comment by lauriemerrell** [comment]
Per conversation, the cause here is duplicate primary keys in `dim_stops` so stops needs the duplicate warning treatment à la https://github.com/cal-itp/data-infra/pull/1932

File: warehouse/models/mart/gtfs/_mart_gtfs.yml [hunk 8]
```diff
@@ -1453,3 +1496,130 @@ models:
...
+          Rows with `true` in this column indicate that the columns in this table that are aggregated from
+          stop times data (`n_stops`, `n_stop_times`,  `trip_first_departure_sec`, `trip_last_arrival_sec`,
+          and `service_hours`) contain at least one row that had a duplicate primary key in the source stop times data.
+
+          I.e., at least one row being aggregated had a `trip_id` / `stop_sequence` pair that was not unique
+          in the input data. This indicates that data quality issues were present in the stop times data
+          that is being summarized here, and counts may be inflated due to multiple rows with identical identifiers.
+      - name: contains_warning_duplicate_trip_primary_key
+        description: |
+          Rows with `true` in this column have a duplicate primary key in dim_trips;
```
lauriemerrell, line 1589: This description should be refactored to look more like the `stop_times` one -- because trips are being aggregated here, this isn't an issue with the trip key, it's an issue with the number of rows that come from aggregating over those trips

warehouse/models/mart/gtfs/_mart_gtfs.yml
```diff
@@ -1453,3 +1496,130 @@ models:
...
+          i.e., `trip_id` is duplicated within an individual feed instance.
+          Treat these rows with caution.
+      - name: contains_warning_missing_foreign_key_stop_id
```
lauriemerrell, line 1592: If you're joining on `stop_id`, is it possible for this to be true? Are there rows in this table with a null stop_key? If so I think we need to rethink the merge order -- if this is supposed to be a stop-level fact, it shouldn't have missing keys at that level.

warehouse/models/mart/gtfs/_mart_gtfs.yml
```diff
@@ -1453,3 +1496,130 @@ models:
...
+      - name: service_date
+        description: Date on which this shape was active.
+      - name: shape_id
+        description: '{{ doc("gtfs_trips__shape_id") }}'
+      - name: shape_array_key
+        description: Foreign key to dim_shapes_arrays.
+      - name: pt_array
+        description: Ordered array of WKT points that describe this shape.
+      - name: contains_warning_duplicate_trip_primary_key
+        description: |
```
lauriemerrell, line 1524: See comment on `stops` -- because we are *aggregating* trips here, this should be refactored to explain the impact that this warning has on the counts. Rows with `true` here may have inflated `n_trips` because of duplicated primary keys.

lauriemerrell (changes requested):
I think that the code itself mostly looks good, only substantive question is about the handling of the null FK warning for stop times. 

Otherwise just docs updates 👍

warehouse/models/mart/gtfs/_mart_gtfs.yml
charlie-costanzo: okay @lauriemerrell so there were null `stop_key`s on this table, in my latest commit I just revised the join and the join order with `dim_stops` but would love your thoughts on that as a solution when you have a moment, I also revised the descriptions of the `warning_*` columns, and their use in tests

lauriemerrell: Yes, I think this is correct 👍

warehouse/models/mart/gtfs/fct_daily_scheduled_stops.sql
```diff
@@ -0,0 +1,140 @@
...
+    SELECT *
+    FROM {{ ref('dim_stops') }}
+),
+
+dim_routes AS (
+    SELECT *
+    FROM {{ ref('dim_routes') }}
+),
+
+stops_by_day AS (
```
lauriemerrell, line 24: I think it might be good to switch the order of this and the next CTE -- if you group by route type first, you can then just group that one to get daily total instead of grouping the full table. May want to double check that you get the same answer (are there any rows with null route_type? If so, may need to add handling for those, and in fact we may want to preserve them in the pivot too.)

If you do this, you'll need to add the `LOGICAL_OR` to the route type CTE as well.

lauriemerrell (commented):
This comment is not a huge deal but probably worth double checking the question of what happens if `route_type` is null (it is supposed to be required but just in case.)

Other fixes look good 👍

warehouse/models/mart/gtfs/fct_daily_scheduled_stops.sql
charlie-costanzo: @lauriemerrell okay cool I just pushed a commit that reorders the CTEs, and the results appear the same. 

So there **are** null `route_type`s, maybe we should chat very quickly about the handling/preservation of them.

lauriemerrell (approved):

Issue #1993, opened 2022-11-17
User Story: make gtfs schedule file modification date available in data pipeline v2


owades:
## User stories

---

### Summary

As Scott, I'd like to be able to measure the GTFS guideline:
```The web-server hosting GTFS data should be configured to correctly report the file modification date (see HTTP/1.1 - Request for Comments 2616, under Section 14.29).```

### Acceptance Criteria

I can build a check that shows, for every feed for every day, what the file modification date was when that feed was downloaded.

### Notes

I believe this information is available in `stg_gtfs_schedule__download_outcomes`, in the `download_response_headers` column.

Example value (look for "Last-Modified"): ```{"Cache-Control":"public","Content-Disposition":"attachment;filename=google_transit.zip","Content-Length":"608171","Content-Type":"application/x-zip-compressed","Date":"Wed, 16 Nov 2022 03:00:23 GMT","Expires":"Wed, 16 Nov 2022 11:30:00 GMT","Last-Modified":"Tue, 15 Nov 2022 18:48:44 GMT","Server":"Microsoft-IIS/10.0","X-AspNet-Version":"4.0.30319","X-Powered-By":"ASP.NET"}```

I see that this table feeds into `int_gtfs_schedule__joined_feed_outcomes`, which feeds into `fct_schedule_feeds`. Perhaps you can grab the modification date and pass it through?

You could probably convince me to do this, but as a best practice I will stay out of the staging tables unless advised.

# Sprint Ready Checklist
1. - [ ] Acceptance criteria defined
2. - [ ] Team understands acceptance criteria
3. - [ ] Team has defined solution / steps to satisfy acceptance criteria
4. - [ ] Acceptance criteria is verifiable / testable
5. - [ ] External / 3rd Party dependencies identified

lauriemerrell:
Need this for RT as well

PR #2124, opened 2023-01-04
pull through a parsed last-modified header

atvaccaro:
# Description

Quick parsing of `Last-Modified` headers.

Resolves https://github.com/cal-itp/data-infra/issues/1993

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml


lauriemerrell (approved):
May want to confirm with @owades what the best place to surface this for RT would be

owades (approved):
Looks great!

atvaccaro:
@owades I think you'll need to build a model similar to the stale feeds check; really any RT needs to be aggregated up in an incremental manner

PR #2127, opened 2023-01-05
create trip update mart table(s)

atvaccaro:
# Description

Expose trip updates in two mart models: `fct_trip_updates_messages` and `fct_stop_time_updates`

These are view-materialized and require partition filtering, but perform column renaming (🐫 to 🐍 ), record flattening, and array unnesting.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Tests do fail currently due to what I think is bad input data; I've chosen not to filter it out quite yet; changing the test may be better.


edasmalchi (approved):
lgtm, thanks!

lauriemerrell (approved):

PR #2128, opened 2023-01-05
Docs updates

owades:
# Description

I forgot to update the combined docs in the persistent-id-schedule check. This bodes well...

I ended up adding quite a few more copy updates to this PR.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml


lauriemerrell (approved):

lauriemerrell (approved):
I think it may be worth refactoring as we discussed to only list the detailed docs in the full check and just list what entity is being assessed, but defer to you (and that doesn't have to happen in this PR)

Issue #2134, opened 2023-01-06
Change GTFS Schedule external tables to require partition filters


atvaccaro:
- [ ] Require partition filters on the external tables
- [ ] Implemented a sane default lookback for full refreshes
- [ ] Consider having a specific lookback for dev (shorter time period than prod)

PR #2138, opened 2023-01-09
GTFS Schedule optimizations: require partition filters

atvaccaro:
# Description

~NOTE: must wait until https://github.com/cal-itp/data-infra/pull/2103 is merged~ this is merged

Closes https://github.com/cal-itp/data-infra/issues/2134

This is non-trivial to test for correctness because of a lack of true primary keys in the final dimension tables, but I've verified re-running this does not produce duplicate rows. The underlying macro first filters the _possible_ GTFS data by day, but still inner joins against feed versions.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Testing as best I can, though we may still have some queries that fail. We can monitor the query access logs for failures.

`-m +mart.gtfs_schedule_latest` run/test passes as much as we would expect. `dim_translations` fails, we may just want to disable it?

### DIFF
--- a/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_county_geography.yml
+++ b/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_county_geography.yml
@@ -12,5 +12,5 @@ source_format: NEWLINE_DELIMITED_JSON
 use_bq_client: true
 hive_options:
   mode: AUTO
-  require_partition_filters: false
+  require_partition_filter: false
   source_uri_prefix: "california_transit__county_geography/"
--- a/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_county_geography.yml
+++ b/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_county_geography.yml
@@ -12,4 +12,5 @@ source_format: NEWLINE_DELIMITED_JSON
 use_bq_client: true
 hive_options:
   mode: AUTO
+  require_partition_filters: false
   source_uri_prefix: "california_transit__county_geography/"
--- a/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_eligibility_programs.yml
+++ b/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_eligibility_programs.yml
@@ -12,4 +12,5 @@ source_format: NEWLINE_DELIMITED_JSON
 use_bq_client: true
 hive_options:
   mode: AUTO
+  require_partition_filters: false
   source_uri_prefix: "california_transit__eligibility_programs/"
--- a/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_eligibility_programs.yml
+++ b/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_eligibility_programs.yml
@@ -12,5 +12,5 @@ source_format: NEWLINE_DELIMITED_JSON
 use_bq_client: true
 hive_options:
   mode: AUTO
-  require_partition_filters: false
+  require_partition_filter: false
   source_uri_prefix: "california_transit__eligibility_programs/"
--- a/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_fare_systems.yml
+++ b/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_fare_systems.yml
@@ -12,7 +12,7 @@ source_format: NEWLINE_DELIMITED_JSON
 use_bq_client: true
 hive_options:
   mode: AUTO
-  require_partition_filters: false
+  require_partition_filter: false
   source_uri_prefix: "california_transit__fare_systems/"
 schema_fields:
   - name: reservations
--- a/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_fare_systems.yml
+++ b/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_fare_systems.yml
@@ -12,6 +12,7 @@ source_format: NEWLINE_DELIMITED_JSON
 use_bq_client: true
 hive_options:
   mode: AUTO
+  require_partition_filters: false
   source_uri_prefix: "california_transit__fare_systems/"
 schema_fields:
   - name: reservations
--- a/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_funding_programs.yml
+++ b/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_funding_programs.yml
@@ -12,4 +12,5 @@ source_format: NEWLINE_DELIMITED_JSON
 use_bq_client: true
 hive_options:
   mode: AUTO
+  require_partition_filters: false
   source_uri_prefix: "california_transit__funding_programs/"
--- a/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_funding_programs.yml
+++ b/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_funding_programs.yml
@@ -12,5 +12,5 @@ source_format: NEWLINE_DELIMITED_JSON
 use_bq_client: true
 hive_options:
   mode: AUTO
-  require_partition_filters: false
+  require_partition_filter: false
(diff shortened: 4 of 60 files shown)


warehouse/macros/make_schedule_file_dimension_from_dim_schedule_feeds.sql
```diff
@@ -1,12 +1,39 @@
 {% macro make_schedule_file_dimension_from_dim_schedule_feeds(dim_schedule_feeds, gtfs_file_table) %}
+{{ config(materialized='incremental') }}
```
lauriemerrell, line 2: It feels a bit weird to me to have the materialization in a macro. I guess it is ok? But I am a bit 🤔

lauriemerrell (commented):
I am still not sold on the need for this. I think that we can afford to run all of history for most schedule files with no problems, and I am particularly unsure that 120 days is enough for schedule. Like I think that when we do a full refresh we will basically always end up overriding to get all of history since `2021-04-16`. 

Can you add more documentation in the ticket or here about the rationale for doing this, and how you picked the defaults? Personally I think that having a mechanism for different lookback between prod and dev is actually more important than changing prod.

atvaccaro:
> > NOTE: must wait until #2138 is merged
> 
> Is this supposed to be self-referencing, or is there another PR that should be referenced as a dependency here?

Ah no good catch, it should've referenced https://github.com/cal-itp/data-infra/pull/2103

warehouse/macros/make_schedule_file_dimension_from_dim_schedule_feeds.sql
atvaccaro: Yeah it's a bit weird but ideally we just force anything using this macro to be incremental, since the macro itself only works in an incremental model.

Multiple calls to config are explicitly supported, luckily https://github.com/dbt-labs/dbt-core/pull/1150

atvaccaro:
So I'm coming around to wondering if remove `_valid_from/_valid_to` and just require joins to `dim_schedule_feeds`. Here's some example queries.

```
select *
from cal-itp-data-infra-staging.andrew_mart_gtfs.dim_stop_times
where '2023-01-19' between _valid_from and _valid_to; -- 460 gb

select *
from cal-itp-data-infra-staging.andrew_mart_gtfs.dim_stop_times_clustered_by_valids
where '2023-01-19' between _valid_from and _valid_to; -- 460 gb

select *
from cal-itp-data-infra.mart_gtfs.dim_stop_times st
inner join cal-itp-data-infra.mart_gtfs.dim_schedule_feeds sf
  on st.feed_key = sf.key
where '2023-01-19' between sf._valid_from and sf._valid_to; -- 461 gb

select *
from cal-itp-data-infra-staging.andrew_mart_gtfs.dim_stop_times_clustered_by_feed_key st
inner join cal-itp-data-infra.mart_gtfs.dim_schedule_feeds sf
  on st.feed_key = sf.key
where '2023-01-19' between sf._valid_from and sf._valid_to; -- 33 gb
```

warehouse/dbt_project.yml
```diff
@@ -22,6 +22,7 @@ clean-targets:         # directories to be removed by `dbt clean`
   - "dbt_packages"
 
 vars:
+  FULL_REFRESH_SCHEDULE_START: '2021-04-16'
```
lauriemerrell, line 25: nitpick: should this variable name say `GTFS_SCHEDULE`? just thinking that the word "schedule" for something like "full refresh" could maybe benefit from full context idk

warehouse/macros/get_latest_schedule_data.sql
```diff
@@ -5,8 +5,12 @@
 -- and non-public rows are excluded
 
 WITH {{ clean_table_name }}_latest AS (
-    SELECT * FROM {{ table_name }}
-    WHERE _is_current
+    SELECT * FROM {{ table_name }} t
+    WHERE EXISTS (
```
lauriemerrell, line 9: I guess it might be overkill but given that this is a macro (and therefore many birds with one stone)... any chance it's worth commenting that the rationale for this selection syntax is because the tables are clustered on feed key? like if we ever change that, this is the kind of thing that might be confusing

warehouse/models/mart/gtfs/_mart_gtfs.yml
```diff
@@ -406,20 +394,12 @@ models:
...
         description: '{{ doc("gtfs_areas__area_name") }}'
-      - *_valid_from
-      - *_valid_to
-      - *_is_current
+      - *_feed_valid_from
   - name: dim_attributions
     description: |
       Each row is a cleaned row from an attributions.txt file.
       Definitions for the original GTFS fields are available at:
       https://gtfs.org/reference/static#attributionstxt.
```
lauriemerrell, line 402: Missing `*_feed_valid_from` addition?

warehouse/models/mart/gtfs/_mart_gtfs.yml
```diff
@@ -450,20 +430,12 @@ models:
...
         description: '{{ doc("gtfs_attributions__attribution_phone") }}'
-      - *_valid_from
-      - *_valid_to
-      - *_is_current
+      - *_feed_valid_from
   - name: dim_fare_attributes
     description: |
       Each row is a cleaned row from a fare_attributes.txt file.
       Definitions for the original GTFS fields are available at:
       https://gtfs.org/reference/static#fare_attributestxt.
```
lauriemerrell, line 438: Missing `*_feed_valid_from` addition?

warehouse/models/mart/gtfs/_mart_gtfs.yml
```diff
@@ -490,20 +462,12 @@ models:
...
         description: '{{ doc("gtfs_fare_attributes__transfer_duration") }}'
-      - *_valid_from
-      - *_valid_to
-      - *_is_current
+      - *_feed_valid_from
   - name: dim_fare_leg_rules
     description: |
       Each row is a cleaned row from a fare_leg_rules.txt file.
       Definitions for the original GTFS fields are available at:
       https://gtfs.org/schedule/reference/#fare_leg_rulestxt.
```
lauriemerrell, line 470: Missing `*_feed_valid_from` addition?

warehouse/models/mart/gtfs/_mart_gtfs.yml
```diff
@@ -522,20 +486,12 @@ models:
...
         tests: *not_null_error
-      - *_valid_from
-      - *_valid_to
-      - *_is_current
+      - *_feed_valid_from
   - name: dim_fare_products
     description: |
       Each row is a cleaned row from a fare_products.txt file.
       Definitions for the original GTFS fields are available at:
       https://gtfs.org/schedule/reference/#fare_productstxt.
```
lauriemerrell, line 494: Missing `*_feed_valid_from` addition?

warehouse/models/mart/gtfs/_mart_gtfs.yml
```diff
@@ -554,20 +510,12 @@ models:
...
         tests: *not_null_warn_99_threshold
-      - *_valid_from
-      - *_valid_to
-      - *_is_current
+      - *_feed_valid_from
   - name: dim_fare_rules
     description: |
       Each row is a cleaned row from a fare_rules.txt file.
       Definitions for the original GTFS fields are available at:
       https://gtfs.org/reference/static#fare_rulestxt.
```
lauriemerrell, line 518: Missing `*_feed_valid_from` addition?

lauriemerrell (changes requested):
There's still YAML in `_int_gtfs` for `int_gtfs_schedule__incremental_stop_times` and the others (GH won't let me comment inline), can those be deleted?

warehouse/macros/get_latest_schedule_data.sql
lauriemerrell: @atvaccaro you did a 👍 react but I don't see a comment, did you want to do this?

lauriemerrell (commented):
see response on comment on the macro

lauriemerrell (approved):

PR #2140, opened 2023-01-09
exclude legacy models from dbt jobs

atvaccaro:
# Description

Pre-cursor to https://github.com/cal-itp/data-infra/pull/2112 which cannot be merged until the downstream exposures are either disabled or replaced/migrated.

This is a cost-saving measure since there is no new data upstream and we aren't changing these models.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml


lauriemerrell (approved):
Can you update the PR description? Per your message in Slack it seems like this is more of a cost savings in the interim, but we are not planning to disable/delete exposures -- we need to migrate them

PR #2136, opened 2023-01-06
Post-storm analysis dbt cleanup

atvaccaro:
# Description

Fix some small issues found during storm analysis work.

* Missing/incorrect resource requests caused dbt jobs to get crushed by validation jobs
* Validation notices need to be version-aware for key uniqueness

Resolves https://github.com/cal-itp/data-infra/issues/2131

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation
- [ ] agencies.yml

PR #2155, opened 2023-01-12
Fix fares validator check

atvaccaro:
# Description

Also adds some missing dev dependencies

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml


owades (approved):
I don't see any logic changes here besides adding a code, but there's nothing I object to so I approve. At some point could you explain what's changing here?

atvaccaro:
> I don't see any logic changes here besides adding a code, but there's nothing I object to so I approve. At some point could you explain what's changing here?

Yes, I think the problem is that only `duplicate_fare_rule_zone_id_fields` existed before v3? So we were ending up with more nulls once I fixed the code join in https://github.com/cal-itp/data-infra/pull/2136, which is actually correct in the current intermediate table design.

PR #2156, opened 2023-01-12
WIP: Migrate Open Data Publishing to V2 Warehouse Tables

### LINKED ISSUE #1940: Migrate open data publishing to v2 warehouse
We need to migrate the GTFS schedule open data publishing (https://data.ca.gov/dataset/cal-itp-gtfs-ingest-pipeline-dataset) to publish v2 rather than v1 data.

Acceptance criteria:

- [ ] Clarify whether we want to artificially keep ITP ID & (hopefully not) URL number in the published data; if yes, we need to find a way to assign them, if no, we need to update the data dictionary 
- [ ] Decide how we will take the latest-only slice from the v2 data
- [ ] Double check the v2 cleaning and columns (for example, in v2 we are adding some columns in the dimensional tables that we might not want to publish, like `warning_duplicate_primary_key`) -- we should make sure we are selecting only the columns that we want

SorenSpicknall:
# Description

Switches our open data publishing pipeline to use V2 tables, with underlying changes to how latest data slices are generated from those new tables. Note that this changes columns on mart tables in the warehouse _and_ columns on the published open data, so has potential to be disruptive.

Resolves #1940 

## Type of change

- [x] New feature
- [x] Documentation
- [ ] agencies.yml

## How has this been tested?

At the time of writing, testing is not yet adequate to merge.


warehouse/models/mart/gtfs_schedule_latest/dim_translations_latest.sql
```diff
@@ -0,0 +1,9 @@
+WITH
+dim_translations_latest AS (
```
lauriemerrell, line 2: AFAIK this is empty, have we checked how this shows up on the open data site? Just curious

lauriemerrell (approved):
I want to just note that simply deleting the SQL for the v1 `gtfs_schedule_latest_only` means that those tables will continue to get stale in BQ. Ok to merge but I think the urgency of #1355 continues to increase to ensure people are not using outdated data.

Related, I think we need to stop syncing the [`GTFS Schedule Feeds Latest` dataset](https://dashboards.calitp.org/browse/5-gtfs-schedule-feeds-latest) to Metabase?

warehouse/models/mart/gtfs_schedule_latest/dim_translations_latest.sql
SorenSpicknall: This luckily doesn't break anything; CKAN handles emptiness gracefully - it shows up with columns fully defined, and spits out a headings-only CSV when downloaded.

warehouse/scripts/publish.py
```diff
@@ -287,7 +287,7 @@ def _generate_exposure_documentation(
             description = resources[name].description
 
         row = MetadataRow(
-            dataset_name=name,
+            dataset_name=name.replace("dim_", "").replace("_latest", ""),
```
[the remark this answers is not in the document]
SorenSpicknall: Resolved in https://github.com/cal-itp/data-infra/pull/2156/commits/2024a4f1847cfc9f34b63a1205d385ec116600bb

PR #2161, opened 2023-01-13
new test for extreme daily check results

owades:
# Description

I created a new generic test, which will create a warning whenever a guidelines check has a 0% or 100% pass rate any day in the last 30 days. The goal is to proactively identify checks that are acting unexpectedly, rather than having stakeholders be the ones to notify me.

The reason for the 30 day filter is that there are different historic cutoffs for different checks, and as you go further back in time the results get, dare I say, less reliable.

This pr also reorganizes the tests a bit, for clarity.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
I ran the a few tests that use it, and confirmed that it caught unwanted behavior (mainly that `int_gtfs_quality__rt_feeds_present` is passing 100% right now, which I need to look into).


owades:
@lauriemerrell could you give this one a peek?

lauriemerrell (approved):

PR #2162, opened 2023-01-13
Use newly renamed columns

atvaccaro:
# Description

I'm not sure why we can't type this as JSON? @evansiroky renamed them so we can just reference the new ones!

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Ran Airflow and tested that `int_transit_database__services_history` would build on top.

### DIFF
--- a/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_services.yml
+++ b/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_services.yml
@@ -248,9 +248,9 @@ schema_fields:
   - name: __alerts
     type: INTEGER
     mode: NULLABLE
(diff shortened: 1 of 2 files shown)


lauriemerrell (commented):
what happens to data from before today if we make this change?

atvaccaro:
> what happens to data from before today if we make this change?

Evan indicated it wasn't maintained/production-ready so he was OK with ignoring it https://cal-itp.slack.com/archives/C040XD9UB33/p1673631575241449?thread_ts=1673625292.830799&cid=C040XD9UB33

lauriemerrell (approved):

PR #2171, opened 2023-01-14
remove newly-added code, add generous-N/A-assignment to faresV2 check

owades:
# Description

Unfortunately #2155 actually wasn't the right solution to the issue, as the newly-added code does measure Fares v2 (that code was actually recently removed from the validator).

I was trying to figure out the cause of the null values, and tried first time-restricting feed_guidline_index to be after 11/15/22 (the first date that any of the fares v2 codes appears in the fct table). Unfortunately there were still nulls, and I realized that
there are some days/feeds in `fct_daily_schedule_feed_validation_notices` where there is not a row for every single code (example query below). If this is by design, then I need to understand better. This PR suggests creating laxer requirements for an N/A value, basically assigning it always for null values. Not my favorite approach, but I'm hoping it will at least elicit a good conversation.

One hypothesis for what is happening is that, for feeds that don't have any Fares v2 files, the fares checks aren't run at all, and that's why they don't appear. If that's the case, that's clever, and I need to update my check because it should already be catching that situation.

```
SELECT * FROM `cal-itp-data-infra`.`mart_gtfs_quality`.`fct_daily_schedule_feed_validation_notices` 
    WHERE date = '2023-01-01'
     AND code IN ('fare_transfer_rule_duration_limit_type_without_duration_limit',
                    'fare_transfer_rule_duration_limit_without_type',
                    'fare_transfer_rule_invalid_transfer_count',
                    'fare_transfer_rule_missing_transfer_count',
                    'fare_transfer_rule_with_forbidden_transfer_count',
                    'invalid_currency_amount'
                    )
          AND feed_key = '195f0eb3ea4076ee55d54034b552a6e2'
```

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
`./dbt.sh build -s int_gtfs_quality__passes_fares_validator`


lauriemerrell:
I suspect this is actually a bug in @atvaccaro's fix, because `fct_daily_schedule_feed_validation_notices` is supposed to be a Cartesian product to all checks / feeds, so it shouldn't matter if the validator doesn't run those checks. I am wondering more if the versioning of individual codes is still not quite right.

PR #2188, opened 2023-01-19
check for rt updates every 20sec

owades:
# Description

New check to measure the best practice:
```
Vehicle Positions & Trip Updates feeds contains updates at least once every 20 seconds, including updated timestamps and data for each trip and vehicle in service.
```
This measures just the first half, without checking each vehicle & trip in service. I may followup with that part.

This is a very tough check to pass, as just one lagging update in the day (out of 4,320 total) will cause a fail. We could consider setting a higher threshold, but this is what the best practice states.

## Type of change

- [X] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
`./dbt.sh build -s int_gtfs_quality__rt_20sec_tu+`
`./dbt.sh build -s int_gtfs_quality__rt_20sec_vp+`
Manually looked at row-by-row results, to confirm that first and last updates of the day don't cause anomalous values.


atvaccaro:
Should this be checking the message headers rather than our extract time? It's tricky since we are responsible for fetching every 20 seconds in order to do the check, but technically we could be fetching an outdated message which would be identified by the header.

owades:
I think that's right @atvaccaro. Going back to draft - I think I need to switch to using `stg_gtfs_rt__vehicle_positions` so I can use the `header_timestamp` field, and use similar logic with that field. Gaps in our extracts could lead to missing out on agency gaps, but our extract frequency should provide decent coverage.

atvaccaro (approved):
lgtm!

We need to have some discussions around trip updates (and vehicle positions to a lesser extent) transforms very soon; do you expect any further checks on top of actual trip updates data?

Issue #1773, opened 2022-09-09
User Story: Include manual GTFS check data from Airtable in the warehouse


owades:
### Summary

As a Cal-ITP team member interfacing with agencies, I want to understand where they stand on a series of manually-performed GTFS checks, so that I can be completely up to date on the status of their GTFS feed and provide support accordingly.

While we are automating every check possible, there remain ~14 checks which can only be performed manually (listed below). These checks mostly already are occurring, and are tracked in Airtable. We need to get these checks into the warehouse, so they can be added to fact_daily_guideline_checks.

I believe the two options are:
A) Connect the existing tables to the warehouse
B) Create a better Airtable data structure, and connect this to the warehouse.

As addition context for option A, the checks are listed in 3 separate tables (I may be using incorrect Airtable lingo): `gtfs-dataset check data`, `gtfs-service check data`, and `provider check data`. It seems straightforward to consolidate these 3 tables to one, if that were deemed necessary. There's also a table called `gtfs checks` which provides info about each check. As I understand, these tables are linked to the source of truth transit-provider table in Airtable, so hopefully clean joins are possible.

List of needed manual checks, just for extra context:
- Static feed ingested by Google Maps and/or a combination of Apple Maps, Transit App, Bing Maps, Moovit or local Open Trip Planner services.
- Realtime feed ingested by Google Maps and/or a combination of Apple Maps, Transit App, Bing Maps, Moovit or local Open Trip Planner services.
- Static and RT feeds are representative of all fixed-route transit services under the transit providers' purview;
- Static and RT feeds are representative of all deviated-fixed route transit services under the transit providers' purview
- Shapes in shapes.txt are precise enough to show the right-of-way that the vehicle uses and not inaccurately exit the right-of-way.
- Include tts_stop_name entries in stops.txt for stop names that are pronounced incorrectly in most mobile applications.
- Passes Grading Scheme v1
- Follow all GTFS Schedule Best Practices (version 1.0 or later) and "Should" statements in specification.
- Consistent with GTFS Realtime Best Practices and "Should" statements in specification.
- GTFS Schedule link is posted on website
- GTFS Realtime link is posted on website
- Authentication process follows guidelines
- Includes an open license that allows commercial use
- Technical contact is listed on website

### Acceptance Criteria

I am able to added addition GTFS checks based off of the Airtable data tables listed above. Data remains up to date.

### Tester [Stakeholder]

1. Scott Owades

# Sprint Ready Checklist
[...]

PR #2197, opened 2023-01-20
add new manual check fields and relevant check indexes

atvaccaro:
# Description

Adds new fields from Airtable concerning manual checks; progress towards https://github.com/cal-itp/data-infra/issues/1773

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Ran extract, created, tables, ran/tested dbt models locally.


warehouse/models/staging/transit_database/stg_transit_database__gtfs_datasets.sql
```diff
@@ -87,6 +87,13 @@ stg_transit_database__gtfs_datasets AS (
         authorization_header_parameter_name,
         header_secret_key_name,
         url_to_encode,
+        manual_check__link_to_dataset_on_website,
+        manual_check__accurate_shapes,
+        manual_check__data_license,
+        manual_check__authentication_acceptable,
+        manual_check__stable_url,
+        manual_check__localized_stop_tts,
```
lauriemerrell, line 95: What his (stop TTS) already in the external table? Don't see it added here

lauriemerrell (commented):
Would you want to add these in the downstream tables later? (Intermediate and dimensional tables.) Or are you thinking that these do not go into the dimensions and instead go directly into like a daily checks fact table?

warehouse/models/staging/transit_database/stg_transit_database__gtfs_datasets.sql
atvaccaro: I think it's here https://github.com/cal-itp/data-infra/pull/2197/files#diff-967c2a19f93fc05bf18c018f3cebf57245be1b197007fbf96e57653c978df589R166

atvaccaro:
> Would you want to add these in the downstream tables later? (Intermediate and dimensional tables.) Or are you thinking that these do not go into the dimensions and instead go directly into like a daily checks fact table?

Yeah I was just thinking about the checks themselves; I can pull them all the way through if you think we should.

lauriemerrell:
> Yeah I was just thinking about the checks themselves; I can pull them all the way through if you think we should.

We need to get them usable for @owades, so I think we need to pull them one more step. We don't currently have dedicated organization, service, GTFS service data, (maybe?) GTFS dataset indices for these entities' checks, so we will need to make those too.

warehouse/models/staging/transit_database/stg_transit_database__gtfs_datasets.sql
lauriemerrell: Oh sorry my bad, I missed it because it wasn't in the list he requested. Good catch to you

atvaccaro:
> > Yeah I was just thinking about the checks themselves; I can pull them all the way through if you think we should.
> 
> We need to get them usable for @owades, so I think we need to pull them one more step. We don't currently have dedicated organization, service, GTFS service data, (maybe?) GTFS dataset indices for these entities' checks, so we will need to make those too.

I think for this PR I want to merge so we can use the production external tables for developing the checks models. I'll open up a new PR with the indices and a couple checks.

lauriemerrell (approved):

PR #2200, opened 2023-01-20
make trip update quality checks cheaper

atvaccaro:
# Description

Doing this in advance of two other checks as well, that will query trip updates. Since the `stop_times_updates` are a large portion of trip update messages, it's relatively performant to incrementally materialize the rest of the fields and use that model for any quality checks that need trip updates. Otherwise, each check would re-read the raw data for each check model, even if those individual check models are made incremental.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Local dbt run; main tests pass on the downstream models


owades (approved):

lauriemerrell (approved):
Can you expand on the `tbd` in testing?

PR #2201, opened 2023-01-23
Create new check indexes and one example check

atvaccaro:
# Description

I've created the initial indexes and one check for organizations all the way to the unioned daily fact.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
dbt tests pass on the actual check and the daily models


warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__contact_on_website.sql
```diff
@@ -0,0 +1,26 @@
...
+
+int_gtfs_quality__contact_on_website AS (
+    SELECT
+        idx.date,
+        idx.organization_key,
+        -- TODO: use real macros
+        'Organization has contact info on website' AS check,
+        'Has contact info' AS feature,
+        CASE manual_check__contact_on_website
+            WHEN 'Yes' then 'PASS'
```
owades, line 19: I think we need `WHEN 'NO' THEN 'FAIL'` as well

owades (approved):
This looks good to me. My only question is if it flows far enough - should this also be included in `fct_daily_organization_combined_guideline_checks` and perhaps also in `fct_daily_guideline_checks`?

atvaccaro:
yeah let me do that here

Issue #2098, opened 2022-12-20
Reports data updates for v2 migration


lauriemerrell:
Updated version of #1762, which I am closing in favor of this with clearer requirements.

We need to make four tables to enable the reports site migration, per conversation between me and @edasmalchi today:
- [ ] An "organizations assessed" index analogous to the work in #2087 but with a logic change such that regional precursor feeds are used instead of the combined regional feed for MTC 511 agencies. This may be an intermediate table. 
- [ ] A reports index table like [reports_gtfs_schedule_index](https://dbt-docs.calitp.org/#!/model/model.calitp_warehouse.reports_gtfs_schedule_index)
- [ ] An organization-level summarized version of [fct_daily_scheduled_service_summary](https://dbt-docs.calitp.org/#!/model/model.calitp_warehouse.fct_daily_scheduled_service_summary)
- [ ] A table showing ID changes within the month like [gtfs_schedule_fact_route_id_changes](https://dbt-docs.calitp.org/#!/model/model.calitp_warehouse.gtfs_schedule_fact_route_id_changes) and [gtfs_schedule_fact_stop_id_changes](https://dbt-docs.calitp.org/#!/model/model.calitp_warehouse.gtfs_schedule_fact_stop_id_changes)

PR #2203, opened 2023-01-23
Reports month id changes

charlie-costanzo:
THIS PR HAS BEEN REPLACED BY #2224 

# Description

Recreating reports tables `gtfs_schedule_fact_route_id_changes` and `gtfs_schedule_fact_stop_id_changes` using V2 warehouse tables as `fct_monthly_route_id_changes` and `fct_monthly_stop_id_changes`

These tables also use the macro `ids_version_compare_aggregate()`, but in order to do so splits part of it into a new macro called `id_aggregate()` in the same file for use in `int_gtfs_quality__persistent_ids_schedule`

To-do:
* yml documentation, tests
* Communicate out proposed changes to other stakeholders

This PR handles the last bullet point in #2098 

## Type of change
- [x] New feature

## How has this been tested?
Locally using `dbt run` and `dbt test`


charlie-costanzo:
Suggested changes to macro ids_version_compare_aggregate():
* Keep up to and including ids_version_compare CTE
* Move last query to int_gtfs_quality__persistent_ids_schedule for each *_id 
  * (or turn into new macro?)

Then add to new tables fct_monthly_*_id_changes
* Incorporate a dim_metric_date likeness, join in
* Join in a likeness to gtfs_schedule_dim_feeds

warehouse/models/mart/gtfs_quality/fct_monthly_route_id_changes.sql
```diff
@@ -0,0 +1,68 @@
...
+    SELECT DISTINCT date_start, date_end, publish_date FROM {{ ref('idx_monthly_reports_site') }}
+),
+
+month_start AS (
+    SELECT
+        base64_url,
+        feed_key,
+        id,
+        publish_date
+    FROM route_id_comparison
```
lauriemerrell, line 23: I don't follow why `route_id_comparison` is needed here -- it doesn't look like you are keeping any of the route comparison columns.

lauriemerrell (commented):
I am not 100% following the approach here, we can maybe discuss in our meeting later today?

One thing just to confirm is that I think we don't necessarily want to use Scott's actual comparison logic (or his feed version history) here because it compares version over adjacent version and we want to define month start and month end versions which may not be exactly the same thing.

PR #2204, opened 2023-01-23
Reports v2 tables

lauriemerrell:
# Description

This PR:

- Creates three new `mart_gtfs_quality` tables intended to be consumed by the reports site directly:
   - `idx_monthly_reports_site`: Replaces v1 table `reports_gtfs_schedule_index` -- indicates what organizations should have a report produced each month, and contains the summary statistics that appear in the top section of reports on reports.calitp.org  
    - `fct_daily_reports_site_organization_scheduled_service_summary`: Daily service total by organization, to populate the service hours by day type charts
    -  `fct_monthly_reports_site_organization_validation_codes`: All validation codes observed for a given organization's feeds within the given month
    - `fct_monthly_reports_site_organization_file_checks` to check for file presence of specific files
- Refactors and renames `fct_daily_feed_scheduled_service_summary` (formerly just `fct_daily_scheduled_service_summary`) to include rows for feeds that have *no* scheduled service on the given day (so that we can report on days where an organization has no service scheduled)
- Adds the `reports_site_assessed` flag to `int_gtfs_quality__daily_assessment_candidate_entities`, which necessitated creating a separate table `int_gtfs_quality__naive_organization_service_dataset_full_join` because BigQuery said that `int_gtfs_quality__daily_assessment_candidate_entities` has gotten too complicated 😅 
- Removes the old (v1) reports site dbt exposure and creates a new one, defined in `_mart_gtfs_quality.yml`
- Renames the `original_record_id` field to `source_record_id` for Airtable records (realized that `original_record_id` may sound like it is associated with a specific version, when this is the *unversioned* identifier)
- Removes two unused paths from `dbt_project.yml` 

Note that I am also drafting a ["release notes" doc ](https://docs.google.com/document/d/1RMYSAbiiq1a9zAnQm_MIHAC5Hbc19G4LG-uEOk1wdKk/edit) which I will share in Slack with the relevant parties to help with the overall behavior changes between v1 and v2.

<details> 
<summary>Completed PR to do list</summary>

TODO:
- [x] YAML for everything 
   - [x] YAML for service changes
- [x] Move the org/feed mapping to an intermediate table 
- [x] Filter down to only feeds that are used for reports for the organization service summary, move that to the GTFS quality mart instead of GTFS 
- [x] Add summary information to the reports index table
- [x] Update reports site exposure
- [ ] Investigate row count difference in fact daily guideline checks -- I am skipping this, assuming it is a difference in incremental tables
- [x] Validations table
- [x] ITP ID as integer
- [x] AC Transit -- why no service on December 4?  
</details> 

Resolves most of #2098 except the ID changes that @charlie-costanzo is doing in #2203

## Type of change

- [x] New feature
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) -- renames a table 
- [x] Documentation
- [ ] agencies.yml

## How has this been tested?
I have run the affected lineages locally and examined the output of `idx_monthly_reports_site` in particular to confirm that December results align with the existing published reports. 

`poetry run dbt test -s 1+exposure:calitp_reports_site` has no failures.


lauriemerrell:
Ope, forgot to add the checks for file presence

warehouse/models/mart/gtfs_quality/fct_monthly_reports_site_organization_file_checks.sql
```diff
@@ -0,0 +1,79 @@
+{{ config(materialized='table') }}
+
+WITH files_of_interest AS (
+    SELECT 'shapes.txt' AS filename, 'shapes' AS gtfs_filename, 'Visual display' AS reason
```
atvaccaro, line 4: nit: there are enough lines it might be worth using the [array method](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/payments_views/payments_feeds.sql)

warehouse/models/mart/transit_database/bridge_schedule_dataset_for_validation.sql
```diff
@@ -16,7 +16,7 @@ bridge_schedule_dataset_for_validation AS (
         LEAST(datasets._valid_to, sched_ref._valid_to) AS _valid_to
     FROM datasets
     INNER JOIN datasets AS sched_ref
-        ON datasets.schedule_to_use_for_rt_validation_gtfs_dataset_key = sched_ref.original_record_id
+        ON datasets.schedule_to_use_for_rt_validation_gtfs_dataset_key = sched_ref.source_record_id
```
atvaccaro, line 19: this reads weirdly but `schedule_to_use_for_rt_validation_gtfs_dataset_key` is just an aliased column

warehouse/models/mart/gtfs_quality/fct_monthly_reports_site_organization_validation_codes.sql
```diff
@@ -0,0 +1,43 @@
...
+    FROM fct_daily_schedule_feed_validation_notices AS notices
+    INNER JOIN int_gtfs__organization_dataset_map AS orgs
+        ON notices.date = orgs.date
+        AND notices.feed_key = orgs.schedule_feed_key
+    INNER JOIN idx_monthly_reports_site AS idx
+        ON notices.date BETWEEN idx.date_start AND idx.date_end
+        AND orgs.organization_source_record_id = idx.organization_source_record_id
+    WHERE total_notices > 0
+    QUALIFY validation_validator_version = MAX(validation_validator_version)
+        OVER (PARTITION BY publish_date, organization_source_record_id)
```
atvaccaro, line 39: does the `QUALIFY` "execute" before the `DISTINCT`?

atvaccaro (approved):
this is very exciting

I can take a stab at fixing daily notices, or you can

warehouse/models/mart/transit_database/bridge_schedule_dataset_for_validation.sql
lauriemerrell: Yes we should probably rename it, I got lazy with key vs ID at some point. But I don't think it's worth doing here

warehouse/models/mart/gtfs_quality/fct_monthly_reports_site_organization_validation_codes.sql
lauriemerrell: Seems so, and anecdotally I did see notices from a prior version omitted

https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#qualify_clause

image

warehouse/models/mart/gtfs_quality/fct_monthly_reports_site_organization_file_checks.sql
lauriemerrell: sure, done

Issue #2063, opened 2022-12-08
User Story: Create trip ID coverage analysis tables 


lauriemerrell:
### Summary

A user story should typically have a summary structured this way:

1. **As a** data analyst or data quality assessor
2. **I want** to be able to see information about which trip IDs appeared in schedule vs RT data
3. **So that** I can understand trip ID discrepancies between RT and schedule and perhaps show agencies detailed information about their coverage 

### Acceptance Criteria

- [ ] Incremental tables that summarize a trip ID's presence in each type of RT data by day:
   - [ ] For vehicle positions, first and last message times / count of messages
   - [ ] For service alerts, distinct message text?, distinct cause/effect?, count of messages?
   - [ ] For trip updates, schedule relationship, perhaps max delay?, number of cancelled stops?, count of messages?

### Tester [Stakeholder]

1. @natam1 
2. Maybe @owades for the trip ID alignment guideline check

# Sprint Ready Checklist
1. - [ ] Acceptance criteria defined
3. - [ ] Team understands acceptance criteria
4. - [ ] Team has defined solution / steps to satisfy acceptance criteria
5. - [ ] Acceptance criteria is verifiable / testable
6. - [ ] External / 3rd Party dependencies identified

owades:
Yes please!

lauriemerrell:
I'm going to call this completed based on `fct_observed_trips`, which doesn't do the comparison with schedule but has a key to enable that join. There is also separate issue #2685 for the comparison functionality

PR #2205, opened 2023-01-23
Create "small" materialized trip updates model, and initial observed trips fact on top

atvaccaro:
# Description

Progress towards https://github.com/cal-itp/data-infra/issues/2063 (joining RT and Schedule for validity analysis is coming)

Creates an observed trips fact based on TripDescriptor. Also creates `int_gtfs_rt__trip_updates_no_stop_times` an incremental model to serve multiple downstream models with cheaper trip updates data.

Question: should we join this against daily scheduled trips?

- [x] For vehicle positions, first and last message times / count of messages
- [x] For service alerts, distinct message text?, distinct cause/effect?, count of messages?
- [x] For trip updates, schedule relationship, perhaps max delay?, number of cancelled stops?, count of messages?

`int_gtfs_rt__trip_updates_no_stop_times` is intended to be generically useful; `stop_times_updates` is approximately 90% of the size of a trip update record, so if we exclude them, we can materialize the rest of the record in dbt (for now). Under the current dbt DAG, every additional trip_updates model means another 150 GB read per day even if those models are incremental. With `int_gtfs_rt__trip_updates_no_stop_times` we eat that cost once, and build more models on top while also benefitting from columnar storage.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Models build up through `fct_observed_trips` as expected


warehouse/models/mart/gtfs/fct_observed_trips.sql
```diff
@@ -1,41 +1,88 @@
...
+    LEFT JOIN fct_daily_rt_feed_files
+        ON service_alerts.dt = fct_daily_rt_feed_files.date
+        AND service_alerts.base64_url = fct_daily_rt_feed_files.base64_url
+),
 
 fct_observed_trips AS (
     SELECT
         {{ dbt_utils.surrogate_key([
             'dt',
             'associated_schedule_gtfs_dataset_key',
```
owades, line 75: i think it might be cleaner to use the gtfs_dataset_key, rather than the schedule key

lauriemerrell: Because he's trying to join across separate datasets (trip updates / vehicle positions / service alerts), there are a few things that they have in common:
* `associated_schedule_gtfs_dataset_key` -- this is the schedule dataset that they use for validation (what Andrew is currently using)
* `service` / `organization` (traversing farther through the tree of associations)

I told him to use `associated_schedule_gtfs_dataset_key` because it's the "closest" -- it uses only datasets rather than going out to service/organization.

But these records do not have`gtfs_dataset_key` itself in common, because they are different feeds (trip updates vs. vehicle positions means different `gtfs_dataset_key`, different `base64_url`, etc.). So `associated_schedule_gtfs_dataset_key` is the "nearest" thing that these records would have in common.

owades: Of course of course, sorry to make you explain again! I keep going back and forth on remembering what gtfs_dataset_key means, I keep thinking it means "quartet key".

lauriemerrell: Yeah, it is definitely confusing -- as soon as I finish #2204 documentation is my next priority 😅

atvaccaro: Thanks! Going to resolve this and put this up for review in a sec.

owades (approved):

atvaccaro:
I'm going to use dim_provider_gtfs_data for the join in fct_observed_trips rather than the schedule URL from the metadata

lauriemerrell (changes requested):
I requested the change to `dim_provider_gtfs_data` offline, LMK when that is done & I will review

warehouse/models/mart/gtfs/fct_observed_trips.sql
```diff
@@ -0,0 +1,130 @@
...
+    SELECT
+        {{ dbt_utils.surrogate_key([
+            'dt',
+            'schedule_to_use_for_rt_validation_gtfs_dataset_key',
+            'trip_identifier',
+        ]) }} as key,
+        dt,
+        schedule_to_use_for_rt_validation_gtfs_dataset_key,
+        trip_identifier,
+        COALESCE(tu.trip_id, vp.trip_id, sa.trip_id) AS trip_id,
```
lauriemerrell, line 106: Can you add these columns to the YAML?

lauriemerrell (changes requested):
Just a few things, only the request for first/last header/extract timestamps is blocking from my perspective.

warehouse/models/intermediate/gtfs/int_gtfs_rt__vehicle_positions_trip_summaries.sql
```diff
@@ -0,0 +1,54 @@
...
+            'trip_start_time',
+            'trip_start_date',
+        ]) }} as key,
+        dt,
+        base64_url,
+        trip_id,
+        trip_route_id,
+        trip_direction_id,
+        trip_start_time,
+        trip_start_date,
```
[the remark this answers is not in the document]
atvaccaro: Ah that's interesting, and yeah can do this with `ARRAY_AGG`.

https://docs.snowflake.com/en/sql-reference/functions/min_by.html would be neat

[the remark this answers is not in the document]
lauriemerrell: Not sure if it's blocking per se but if it won't take too long may be nice to have

warehouse/models/mart/gtfs/fct_observed_trips.sql
```diff
@@ -0,0 +1,130 @@
...
+    SELECT * FROM {{ ref('int_gtfs_rt__trip_updates_summaries') }}
+),
+vehicle_positions AS (
+    SELECT * FROM {{ ref('int_gtfs_rt__vehicle_positions_trip_summaries') }}
+),
+service_alerts AS (
+    SELECT * FROM {{ ref('int_gtfs_rt__service_alerts_trip_summaries') }}
+),
+
+-- get each of these distinct
```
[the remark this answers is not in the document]
atvaccaro: I think this is fine, it's not 100% ideal but it's better than using `dim_provider_gtfs_data`. I just want us to try to keep these sorts of patterns in mind and maybe build an ephemeral model to help.

warehouse/models/intermediate/gtfs/int_gtfs_rt__vehicle_positions_trip_summaries.sql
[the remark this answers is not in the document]
atvaccaro: I did it

lauriemerrell (approved):

PR #2206, opened 2023-01-23
Add service-level check and fct table

owades:
# Description

New check to see whether this guideline is met:
```
100% of trips marked as “Scheduled,” “Canceled,” or “Added” within the Trip Updates feed are represented within the Vehicle Positions feed.
```

I might be able to incorporate the observed_trip data structure proposed [here](https://github.com/cal-itp/data-infra/pull/2205) to reduce the cost of my big join.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Not yet


owades:
@lauriemerrell , could you take a look at this when you are back? I created a new fct table as you suggested, but I'm still using service_key as a unique ID. I think I need some high-level assistance from you - for example, should I be grouping by "provider" rather than "service"? Thank you!

lauriemerrell:
Ok update I missed that @atvaccaro did already make this https://github.com/cal-itp/data-infra/blob/main/warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__services_guideline_index.sql

owades:
Thank you! I think I can get this updated today based on your feedback here.

owades:
@lauriemerrell, this is taking quite a long time to run, I wonder if the join with the BETWEEN statement is slowing stuff down.

owades:
I think I will hold off on this PR until #2205 is done, as it will speed things up dramatically for me.

lauriemerrell:
@owades re:
>  I wonder if the join with the BETWEEN statement is slowing stuff down.

The tables in my sample SQL above are both tiny (~MB), so I don't think that step is the issue -- `fct_vehicle_positions_messages` is a view with no tables in between it and the raw data, so it's going into the filesystem, agree that #2205 will be the big timesaver here.

owades:
starting fresh on new branch

PR #2207, opened 2023-01-24
Add Redwood Coast Transit Swiftly RT URLs

evansiroky:
# Description

Adds Swiftly RT URLs for Redwood Coast Transit

## Type of change

- [ ] New feature
- [ ] Documentation
- [x] agencies.yml

## How has this been tested?

Downloaded locally


atvaccaro:
@lauriemerrell correct me if I'm wrong, but I think everything consuming agencies.yml is turned off at this point, though some of the code referencing it is still hanging around (e.g. CI). @evansiroky would you be OK with stopping updates to agencies.yml going forward? I don't think there's a compelling reason to continue with updates; they'd be reflected in git history but that's about it.

lauriemerrell:
Agree, inclined to close.

atvaccaro:
going to close this per slack!

PR #2210, opened 2023-01-24
Surface Feed Fetch Errors in Warehouse

### LINKED ISSUE #2152: User Story: keep track of RT feed fetch errors in warehouse
## User stories

---
### Summary

As a part of data quality reporting, I'm trying to measure the following [Best Practice](https://gtfs.org/realtime/best-practices/):
```
The server hosting GTFS Realtime data should be reliable and consistently return validly-formatted protobuf-encoded responses. Fewer than 1% of responses should be invalid (protobuf errors or fetching errors).
```
I'm able to track protobuf errors using the existing warehouse, but fetch errors are currently not tracked. This request is to track them.

### Acceptance Criteria

This is something that can be verified to show that this user story is satisfied.

**Example:**

1. I can build a gtfs check that tells whether fewer than 1% of requests lead to a fetch error

### Tester [Stakeholder]

1. Scott Owades

# Sprint Ready Checklist

SorenSpicknall:
# Description

This PR brings some feed fetch error information from Sentry into the data warehouse for further analysis. It is loosely based on [our work bringing Airtable data into BigQuery](https://github.com/cal-itp/data-infra/blob/main/airflow/plugins/operators/airtable_to_gcs.py), and is intended to be extensible from one [Sentry Issue](https://docs.sentry.io/product/issues/) to another. One example Sentry loader operator instance based on Sentry issue 431, our most frequently alerting, is provided, and others can be added ad hoc.

Partially resolves #2152 - this makes our baseline data available, but doesn't yet format it to be easily joinable on other tables. An additional PR will address that need.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

Airflow and dbt runs have been completed for all the relevant resources in staging environments.


SorenSpicknall:
@atvaccaro or @lauriemerrell, I would love a first pass review on it in order to orient my continuing work. I know I still need to do the following things:

- Split up large file uploads to GCS or date-limit the API calls so that batching isn't necessary
- Better define how external table reading handles new file versions with different partition values

What I don't know at this juncture:

- Whether the data cleaning functions in the middle of the sentry_to_gcs.py, adapted from airtable_to_gcs.py, are necessary or desirable for this data or are redundant/unnecessary/specifically for Airtable/etc.
- How we generally approach tests for external tables (it'd be great to have an example to draw from)
- Whether my code changes to move from BaseModel (used by airtable_to_gcs.py) to the newer PartitionedGCSArtifact need cleanup to reflect best practices for using PartitionedGCSArtifact
- What else I don't know

lauriemerrell:
>Whether the data cleaning functions in the middle of the sentry_to_gcs.py, adapted from airtable_to_gcs.py, are necessary or desirable for this data or are redundant/unnecessary/specifically for Airtable/etc.

Decent chance they aren't needed -- I'd examine the structure of the responses and see whether Sentry returns a lot of arrays? Airtable's API responses are based on the user-defined tables within Airtable so they required a decent amount of cleaning to make even ingestible. I don't think there's really a harm to having them here but I assume the Sentry API is better behaved/more structured so it might be safe to remove them if they aren't needed to make BQ-safe files. (Sorry if that's a non-answer, the TLDR would be that I don't think we assume these are needed by default -- another comparison would be https://github.com/cal-itp/data-infra/blob/main/airflow/plugins/operators/amplitude_to_flattened_json.py which has less cleaning.)

>How we generally approach tests for external tables (it'd be great to have an example to draw from) 

I think by default we actually don't really test external tables (in a code-testing sense), partially for cost reasons? The only testing I do is to run the create external tables DAG task in local Airflow pointed at the test bucket and confirm I can select from the external table.

> Whether my code changes to move from BaseModel (used by airtable_to_gcs.py) to the newer PartitionedGCSArtifact need cleanup to reflect best practices for using PartitionedGCSArtifact

Defer to @atvaccaro on this he is the 👑 of `PartitionedGCSArtifact`

> Better define how external table reading handles new file versions with different partition values

I am unsure what you mean by this -- if you mean different partition values meaning you are partitioning on different keys, then those would need to be in different buckets (see https://cloud.google.com/bigquery/docs/hive-partitioned-loads-gcs). If you mean just different files in different partitions, you generally shouldn't need to modify anything in the external table operator to get this to work.

atvaccaro:
> * Whether the data cleaning functions in the middle of the sentry_to_gcs.py, adapted from airtable_to_gcs.py, are necessary or desirable for this data or are redundant/unnecessary/specifically for Airtable/etc.

It'll depend on whether Sentry has similar data output such as nulls in arrays, invalid BigQuery column names, etc. Since this is an API and not a human-edited spreadsheet, I think column names should be fine. Cleaning arrays is probably fine but also probably unnecessary.

> * How we generally approach tests for external tables (it'd be great to have an example to draw from)

In Airflow, there's also a post-hook in the external table operator you can use to test with a select statement. On the dbt side, you can use use macros in dbt source YAML to point at the staging external tables while testing locally; there's already a top-level variable called `SOURCE_DATABASE` that is semi-consistently used.

> * Whether my code changes to move from BaseModel (used by airtable_to_gcs.py) to the newer PartitionedGCSArtifact need cleanup to reflect best practices for using PartitionedGCSArtifact

Use pendulum types for any temporal partitions; usually `ts: pendulum.DateTime` and then `dt` is a property that just calls `ts.date()` to get a `pendulum.Date`. To be ultra-nitpicky, I'd rename `extract_time` to `extract_ts` as I've tried to keep it obvious when something is a Date, Time, or DateTime/Timestamp. `issue_id` should probably be an `int` if possible since you're probably expecting it to be an integer; it's often useful to keep Pydantic types strict within reason.

Also I probably wouldn't put any scraping logic in a subclass of PartitionedGCSArtifact; just use it to represent the output artifact and interaction with GCS. The rest of the code should be fine in regular functions.

> * What else I don't know

Be sure that your artifacts end up with idempotent keys, that's probably the most important thing as far as maintainability goes; ensures you can always fix and backfill stuff later as long as you aren't actually changing the keying.

lauriemerrell:
@atvaccaro we contradicted each other on external table testing -- I thought we were explicitly moving away from using the post-hook, am I wrong?

atvaccaro:
> @atvaccaro we contradicted each other on external table testing -- I thought we were explicitly moving away from using the post-hook, am I wrong?

Yeah that's fair, you _can_ still use it if you're willing to potentially deal with required partition filtering, but yes I do think that checking that you can query the table/pointing dbt at the table are the better options.

Side note please require partition filters on this external table, it'll be easy to make incremental in dbt and it could be a lot of data; didn't think about that until just now, sorry.

warehouse/models/mart/gtfs_quality/fct_rt_feed_fetch_errors.sql
```diff
@@ -0,0 +1,24 @@
+{{ config(materialized='table') }}
+
+WITH fct_rt_feed_fetch_errors AS (
```
atvaccaro, line 3: I think it might be useful to extract the URL from the tags and cluster on that; also I'm now realizing we don't actually tag with the `tick` which will make joins harder than they should be 🤔 unless we just care about top-level counts anyways

airflow/plugins/operators/sentry_to_gcs.py
```diff
@@ -0,0 +1,164 @@
...
+
+    extract.data = cleaned_df.to_json(orient="records", lines=True).encode()
+
+    return extract
+
+
+class SentryExtract(PartitionedGCSArtifact):
+    bucket: ClassVar[str] = CALITP_BUCKET__SENTRY_LOGS
+    table: ClassVar[str] = "events"
+    dt: pendulum.Date
```
atvaccaro, line 107: You can make `dt` a property since it's derived from `ts`

airflow/plugins/operators/sentry_to_gcs.py
```diff
@@ -0,0 +1,164 @@
...
+
+    print(f"Downloading Sentry event data for project {extract.project_slug}")
+    all_rows = iterate_over_sentry_records(extract, auth_token)
+    extract.extract_ts = pendulum.now()
+
+    raw_df = pd.DataFrame([{**make_arrays_bq_safe(row)} for row in all_rows])
+
+    cleaned_df = raw_df.rename(make_name_bq_safe, axis="columns")
+
+    extract.data = cleaned_df.to_json(orient="records", lines=True).encode()
```
atvaccaro, line 99: this would need to be compressed somwhere to be a valid `.gz` file

airflow/plugins/operators/sentry_to_gcs.py
```diff
@@ -0,0 +1,164 @@
...
+            bucket (str): GCS bucket where the scraped Sentry issue will be saved.
+            project_slug (str): The identifier for the Sentry project being examined.
+            auth_token (str, optional): The auth token to use when downloading from Sentry.
+                This can be someone's personal auth token. If not provided, the environment
+                variable of `CALITP_SENTRY_AUTH_TOKEN` is used.
+        """
+        self.ts = pendulum.now()
+        self.dt = self.ts.date()
+        self.bucket = bucket
+        self.extract = SentryExtract(
```
atvaccaro, line 149: I'd call this stuff in the execute so the timestamp is accurate; technically this constructor gets called on import I think? I could be wrong

warehouse/models/staging/rt/_stg_rt.yml
```diff
@@ -24,3 +24,22 @@ models:
...
+      - name: event_type
+      - name: groupid
+      - name: eventid
+      - name: projectid
+      - name: message
+      - name: title
+      - name: location
+      - name: culprit
+      - name: user
+      - name: tags
```
atvaccaro, line 40: I'd probably do the JSON lookups here

airflow/plugins/operators/sentry_to_gcs.py
```diff
@@ -0,0 +1,164 @@
...
+
+    return extract
+
+
+class SentryExtract(PartitionedGCSArtifact):
+    bucket: ClassVar[str] = CALITP_BUCKET__SENTRY_LOGS
+    table: ClassVar[str] = "events"
+    dt: pendulum.Date
+    ts: pendulum.DateTime
+    partition_names: ClassVar[List[str]] = ["dt", "ts"]
```
atvaccaro, line 109: I might recommend adding `project_slug` as a partition if you aren't adding it to the individual records; it looks like the records just contain the numeric id in `projectid` right now

atvaccaro (changes requested):
lmk if you want to discuss anything synchronously!

SorenSpicknall:
Ready for re-review. I've narrowed the default window down to a deterministic one-day span of events that can be toggled via `sentry_loader.yml` and reproduced on subsequent runs, updated the dbt models to efficiently load incrementally (while leaving an easy path to drop days based on `dt` or `ts` if we need to do a backfill or correct issues with a previous load), moved `ts` initialization into the execute phase of the operator rather than its initialization, and made other requested tweaks.

airflow/dags/create_external_tables/sentry_error_logs.yml
```diff
@@ -0,0 +1,42 @@
...
+bucket: "{{ env_var('CALITP_BUCKET__SENTRY_EVENTS') }}"
+suffix_bucket: true
+source_objects:
+  - "*.jsonl.gz"
+destination_project_dataset_table: "external_sentry_error_logs.events"
+source_format: NEWLINE_DELIMITED_JSON
+use_bq_client: true
+hive_options:
+  mode: CUSTOM
+  source_uri_prefix: "events/{dt:DATE}/{ts:TIMESTAMP}/"
```
atvaccaro, line 11: Do you have thoughts on partitioning this by event date rather than extract date? Otherwise if we run a backfill, we'll end up with duplicate data rather than overwriting our already-extracted data.

I think this actually dovetails with our payments v2 discussion as well, in that we have to deal with the potential for multiple extracts of the same historical data depending on backfills, retries, etc.

airflow/plugins/operators/sentry_to_gcs.py
```diff
@@ -0,0 +1,200 @@
...
+    def execute(self, **kwargs):
+        ts = pendulum.now()
+        dt = ts.date()
+        extract = SentryExtract(
+            project_slug=self.project_slug,
+            dt=dt,
+            ts=ts,
+            filename=f"{self.project_slug}.jsonl.gz",
+        )
+        target_date = self.target_date or ts.subtract(days=1).to_date_string()
```
atvaccaro, line 195: It seems like `target_date` should really be `execution_date`, I think? Each task execution should fetch the same set of data, to facilitate backfills as well as catchups from the `start_date`.

airflow/plugins/operators/sentry_to_gcs.py
atvaccaro: I think this is still relevant; I don't see anything in individual records other than the numeric `projectid`.

Actually taking this with partitioning by _event_ date, I think the best set of partitions might be `event_dt, project_slug` so data is clearly labeled by project in GCS as well. `extract_ts` could either be the final partition or the filename, just some way to distinguish between multiple extracts of the same data.

airflow/dags/sentry_loader/sentry_loader.yml
```diff
@@ -0,0 +1,6 @@
+operator: operators.SentryToGCSOperator
+bucket: "gs://test-calitp-sentry"
+hive_options:
+  mode: CUSTOM
+  require_partition_filter: true
```
atvaccaro, line 5: nit: these can be removed

atvaccaro (changes requested):
I think the data tagging still needs adjustment, unless I'm missing something the lookback isn't aligned by task schedule/execution_date (currently it's either by prior day, or a fixed target date).

airflow/plugins/operators/sentry_to_gcs.py
```diff
@@ -179,22 +177,22 @@ def __init__(
...
+        execution_ts = pendulum.from_format(context.get("ts"), "YYYY-MM-DDTHH:mm:ssZ")
+        execution_dt = execution_ts.date()
         extract = SentryExtract(
             project_slug=self.project_slug,
-            dt=dt,
-            ts=ts,
-            filename=f"{self.project_slug}.jsonl.gz",
+            execution_dt=execution_dt,
+            execution_ts=execution_ts,
+            filename=f"{execution_ts}.jsonl.gz",
```
atvaccaro, line 190: Have you thought about backfills and whether we want to always overwrite?

Also it looks like you end up setting `execution_ts` to a different value later; could you just delay instantiating the extract object until after the API scrape? I think the function signature could look something like:

```python
def fetch_and_clean_from_sentry(target_date: pendulum.Date, auth_token: str) -> Tuple[pendulum.DateTime, Optional[List[Dict]]]:
```

And then you create the extract object just before saving data.

airflow/plugins/operators/sentry_to_gcs.py
```diff
@@ -133,12 +133,11 @@ def fetch_and_clean_from_sentry(extract, target_date, auth_token):
 class SentryExtract(PartitionedGCSArtifact):
     bucket: ClassVar[str] = CALITP_BUCKET__SENTRY_EVENTS
     table: ClassVar[str] = "events"
-    dt: pendulum.Date
-    ts: pendulum.DateTime
-    partition_names: ClassVar[List[str]] = ["dt", "ts"]
+    execution_dt: pendulum.Date
+    execution_ts: pendulum.DateTime
```
atvaccaro, line 137: I think `dt` is still an OK name, since it's the data's actual date. `execution_ts` is confusing because it's set to a different value within the task, but the filename uses the value from the Airflow context. Is it intended to be an identifier of the task instance or the actual wall clock time?

airflow/dags/sentry_loader/sentry_loader.yml
```diff
@@ -0,0 +1,3 @@
+operator: operators.SentryToGCSOperator
+bucket: "{{ env_var('CALITP_BUCKET__SENTRY_EVENTS') }}"
```
atvaccaro, line 2: I don't think the `bucket` arg is used in the operator anymore

airflow/dags/sentry_loader/sentry_loader.yml
```diff
@@ -0,0 +1,3 @@
...
+project_slug: "{{ env_var('CALITP__SENTRY_PROJECT_SLUG') }}"
```
atvaccaro, line 3: I think this should actually be hard-coded since it's part of the task definition; for example if we wanted to also sync the benefits sentry project, we'd have another task YAML with `project_slug: benefits`

Also I'm now realizing we haven't discussed Sentry environments but I think we can filter in the warehouse if needed.

SorenSpicknall:
Unexpected error reporting conditions resulting from a transit agency GTFS feed outage on 2/25 and 2/26 revealed an edge case that greatly encumbers this approach, so closing in favor of a solution that routes data into BQ more directly from the underlying Clickhouse datastore, rather than the Sentry API.

PR #2211, opened 2023-01-24
Test daily guideline checks

atvaccaro:
# Description

Adds a test to help flag potential bugs in `fct_daily_guideline_checks`; tests that checks can only come from one "source" and that the entities match up as we expect. For example, a check for schedule feeds ends up living on a schedule feed entity grained record, etc.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
dbt runs and passes on the changed tables

### DIFF
--- a/warehouse/macros/define_gtfs_guidelines.sql
+++ b/warehouse/macros/define_gtfs_guidelines.sql
@@ -21,6 +21,18 @@
 'organization'
 {%- endmacro %}
 
+{% macro organization() -%}
+'service'
+{%- endmacro %}
+
+{% macro organization() -%}
+'gtfs_service_data'
+{%- endmacro %}
+
+{% macro organization() -%}
+'gtfs_dataset'
+{%- endmacro %}
+
 --
 -- CHECK NAMES
 --
--- a/warehouse/macros/define_gtfs_guidelines.sql
+++ b/warehouse/macros/define_gtfs_guidelines.sql
@@ -21,15 +21,15 @@
 'organization'
 {%- endmacro %}
 
-{% macro organization() -%}
+{% macro service() -%}
 'service'
 {%- endmacro %}
 
-{% macro organization() -%}
+{% macro gtfs_service_data() -%}
 'gtfs_service_data'
 {%- endmacro %}
 
-{% macro organization() -%}
+{% macro gtfs_dataset() -%}
 'gtfs_dataset'
 {%- endmacro %}
 
--- a/warehouse/macros/define_gtfs_guidelines.sql
+++ b/warehouse/macros/define_gtfs_guidelines.sql
@@ -1,4 +1,29 @@
+--
+-- ENTITIES
+--
+{% macro schedule_feed() %}
+'schedule_feed'
+{% endmacro %}
+
+{% macro schedule_url() %}
+'schedule_url'
+{% endmacro %}
+
+{% macro rt_feed() %}
+'rt_feed'
+{% endmacro %}
+
+{% macro rt_url() %}
+'rt_url'
+{% endmacro %}
+
+{% macro organization() %}
+'organization'
+{% endmacro %}
+
+--
 -- CHECK NAMES
+--
 {% macro static_feed_downloaded_successfully() %}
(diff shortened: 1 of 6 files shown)


warehouse/models/mart/gtfs_quality/fct_daily_guideline_checks.sql
```diff
@@ -66,13 +66,27 @@ fct_daily_guideline_checks AS (
...
-        ) AS status
+        ) AS status,
+        CASE
+            WHEN schedule_feed_checks.status IS NOT NULL THEN idx.entity = {{ schedule_feed() }}
+            WHEN rt_feed_checks.status IS NOT NULL THEN idx.entity = {{ rt_feed() }}
+            WHEN schedule_url_checks.status IS NOT NULL THEN idx.entity = {{ schedule_url() }}
+            WHEN rt_url_checks.status IS NOT NULL THEN idx.entity = {{ rt_url() }}
+            WHEN organization_checks.status IS NOT NULL THEN idx.entity = {{ organization() }}
+        END AS matches_entity,
+        CASE WHEN schedule_feed_checks.status = 'PASS' THEN 1 ELSE 0 END
```
owades, line 84: shouldn't this be `WHEN schedule_feed_checks.status IS NOT null THEN 1 ELSE 0 `?

owades (approved):
I left one question, guessing you have a good answer but perhaps it needs to be changed.

warehouse/macros/define_gtfs_guidelines.sql
```diff
@@ -1,4 +1,29 @@
...
+{%- endmacro %}
+
+{% macro rt_url() -%}
+'rt_url'
+{%- endmacro %}
+
+{% macro organization() -%}
+'organization'
+{%- endmacro %}
+
```
lauriemerrell, line 23: Between you and @owades what is the plan for adding "service"? Wondering if @owades needs to do that on #2206 or not

warehouse/models/mart/gtfs_quality/_mart_gtfs_quality.yml
```diff
@@ -207,6 +212,9 @@ models:
       - *check
       - name: status
       - name: feature
+      - name: matches_entity
```
lauriemerrell, line 215: If this is being surfaced on this mart table can you add a short description indicating that this is just for testing/auditing? Same for below

lauriemerrell (commented):
Just two logistical/docs questions. Thanks for tackling this

warehouse/macros/define_gtfs_guidelines.sql
atvaccaro: I'll add the macros here; then @owades just has to use them in intended_checks

warehouse/macros/define_gtfs_guidelines.sql
```diff
@@ -1,4 +1,41 @@
...
+{% macro organization() -%}
```
owades, line 24: fix macro names (id "service" macro is called "organization()")

atvaccaro: 🤦 thank you

warehouse/macros/define_gtfs_guidelines.sql
```diff
@@ -21,15 +21,15 @@
...
 'service'
 {%- endmacro %}
 
-{% macro organization() -%}
+{% macro gtfs_service_data() -%}
 'gtfs_service_data'
 {%- endmacro %}
 
-{% macro organization() -%}
+{% macro gtfs_dataset() -%}
```
owades, line 32: there's also those dash symbols

atvaccaro: yes that's for stripping whitespace

owades: now I know!

lauriemerrell (approved):
👍 I do think maybe a *slightly* more verbose description of the testing fields wouldn't hurt, but it is fine / not blocking IMO

PR #2215, opened 2023-01-26
Upgrading k8s

atvaccaro:
# Description

I also updated AIrflow to point at the new Sentry domain.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml


SorenSpicknall (approved):
Approving with one note: there was [some internal discussion](https://cal-itp.slack.com/archives/C015B2EBHU0/p1674840274359109?thread_ts=1674834181.640799&cid=C015B2EBHU0) about a setup for services that takes jarv.us altogether out of the equation. It wasn't for immediate action, but wanted to flag that we may need to bring these two conversations together and make other changes later.

PR #2218, opened 2023-01-26
Add v2 manual checks

owades:
# Description

Add many manual checks and supporting infrastructure. We will need to test these when we actually have data (could consider adding test values now).

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
./dbt.sh build on all relevant files


warehouse/macros/define_gtfs_guidelines.sql
```diff
@@ -209,7 +210,83 @@
...
+{% macro stable_url_tu() %}
+"Trip updates feed is published at a stable URI (permalink) from which it can be “fetched” automatically by trip-planning applications"
+{% endmacro %}
+
+{% macro stable_url_sa() %}
+"Service alerts feed is published at a stable URI (permalink) from which it can be “fetched” automatically by trip-planning applications"
+{% endmacro %}
+
+{% macro grading_scheme_v1() %}
+"Passes Grading Scheme v1"
```
lauriemerrell, line 257: FWIW I feel like this is gonna be all null for a very long time 😅

warehouse/macros/define_gtfs_guidelines.sql
```diff
@@ -209,7 +210,83 @@
...
+{% endmacro %}
+
+{% macro link_to_dataset_on_website_schedule() %}
+"GTFS Schedule link is posted on website"
+{% endmacro %}
+
+{% macro link_to_dataset_on_website_vp() %}
+"Vehicle positions link link is posted on website"
```
lauriemerrell, line 265: `link link`

warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
```diff
@@ -347,3 +351,77 @@ models:
...
+            - gtfs_dataset_key
+      - dbt_utils.equal_rowcount:
+          compare_model: ref('int_gtfs_quality__gtfs_dataset_guideline_index')
+    columns: *stg_gtfs_guideline_columns
+
+  - name: int_gtfs_quality__stable_url
+    tests: *stg_gtfs_guideline_tests_dataset
+    columns: *stg_gtfs_guideline_columns
+
+  - name: int_gtfs_quality__grading_scheme_v1
```
lauriemerrell, line 389: Does this actually apply to RT feeds? I think you have a mismatch here maybe in terms of the `# ----- All 4 feeds` comment and how things are actually implemented below in terms of which index they use

warehouse/models/mart/gtfs_quality/fct_daily_gtfs_dataset_guideline_checks.md
```diff
@@ -0,0 +1,32 @@
+{% docs fct_daily_gtfs_dataset_guideline_checks %}
+
+Each row represents a date/guideline check/gtfs dataset combination, with pass/fail
+information indicating whether that feed complied with that check on that date.
+Only contains checks that are performed at the gtfs_dataset
```
lauriemerrell, line 5: Can we clarify documentation for which are schedule-only vs. both? Technically that is kind of mixing grains here but I think it's probably ok

lauriemerrell (changes requested):
Just one typo and a few clarifications. I admittedly skimmed the individual actual checks a little bit, looked mostly at overall infra.

lauriemerrell:
Also to respond to comment in Slack, `gtfs service data` represents the mapping between a service and a dataset. So it contains one column for a service, a column for its dataset, and the columns for selectors to choose what rows within that dataset correspond to that service and then other columns that apply to the relationship (for example, this is where the `customer_facing` flag is set, because that flag applies to whether a dataset is customer-facing for a given service.)

warehouse/macros/define_gtfs_guidelines.sql
owades: Lol yes

warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
owades: regarding the grading scheme, you're right! just switched it to use the above "schedule only" index table.

Regarding the rest, I'm not totally sure what you mean: I think those checks do apply to all 4 feeds. But surely I could be missing something.

warehouse/macros/define_gtfs_guidelines.sql
owades: thanks thanks

warehouse/models/intermediate/gtfs/gtfs_quality/_int_gtfs_quality.yml
lauriemerrell: Sorry I just meant "below" like "in the files themselves" referring to GH diff display order 😓 But yeah I think this was the one that stood out to me

warehouse/models/mart/gtfs_quality/fct_daily_gtfs_dataset_guideline_checks.md
owades: Clarified a bit!

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__data_license.sql
```diff
@@ -0,0 +1,27 @@
+WITH
+
+idx AS (
+    SELECT * FROM {{ ref('int_gtfs_quality__gtfs_dataset_schedule_guideline_index') }}
```
lauriemerrell, line 4: Ok sorry for not catching this before -- I just double checked the guidelines directly and I think there are both RT & schedule versions of this -- so maybe it should be in the "all 4" category

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__demand_responsive_routes_match.sql
```diff
@@ -0,0 +1,27 @@
+WITH
+
+idx AS (
+    SELECT * FROM {{ ref('int_gtfs_quality__gtfs_service_data_guideline_index') }}
```
lauriemerrell, line 4: Another sorry I should have caught before -- I think that this one only applies to schedule feeds. 
So unfortunately we may need a separate `gtfs_service_data_schedule` index?

lauriemerrell (commented):
Sorry, two grain issues I missed at first

lauriemerrell (approved):
Split remaining comments into https://github.com/cal-itp/data-infra/issues/2219, approving this PR for now to keep it moving

Issue #1670, opened 2022-08-04
Bug: Investigate dbt test failures


j-meelah:
**Describe the bug**
Multiple dbt tests are failing (see #data-infra-alerts). 

**Work Involved**
Spend ~5 hours investigating it, fixing any low-hanging fruit, and then documenting remaining issues as standalone tickets.

atvaccaro:
closed by https://github.com/cal-itp/data-infra/pull/2222

Issue #1831, opened 2022-09-21
Investigate failing dbt tests (GTFS & Airtable)


lauriemerrell:
We have several tests that are failing for GTFS schedule, RT, & Airtable data in dbt right now. The purpose of this ticket is to capture an **investigation** task that we need to do to identify the root cause and propose next steps. If the fixes for any of these will take more than ~30 min to implement, they can be deferred to a follow-up ticket.

**Acceptance criteria**:

Proposed resolution (demote to warn; change test if test is wrong; fix actual underlying data) for each of the following failing tests:
- [ ] table: `gtfs_schedule_fact_daily_feed_stops` / failing check: `dbt_utils_unique_combination_of_columns_gtfs_schedule_fact_daily_feed_stops_feed_key__stop_key__date`
- [ ] table: `stops_clean` / failing check: `unique_stops_clean_stop_key`
- [ ] table: `stg_transit_database__gtfs_service_data` / failing check: `dbt_utils_unique_combination_of_columns_stg_transit_database__gtfs_service_data_key__calitp_extracted_at`
- [ ] table: `dim_contract_attachments` / failing check: `unique_dim_contract_attachments_key`
- [ ] table: `stg_rt__vehicle_positions` / failing check: `unique_stg_rt__vehicle_positions_key`

atvaccaro:
https://github.com/cal-itp/data-infra/pull/2222 solves the GTFS dim errors; I've put the Airtable tests on our Friday agenda.

lauriemerrell:
`unique_dim_contract_attachments_key` issue remains but is documented in #2909

PR #2222, opened 2023-01-30
Resolve dbt test failures

atvaccaro:
# Description

Progress towards https://github.com/cal-itp/data-infra/issues/1670
Progress towards https://github.com/cal-itp/data-infra/issues/1831

May resolve these given some investigation into Airtable failures, but these are the higher impact ones.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Rebased on top of the new incremental tables, tests pass as expected.

Note: Moving to line numbers as intra-file IDs would make the special null handling unnecessary.


lauriemerrell (approved):
This looks ok to me. If as part of working on this PR you did an audit of our failing tests in Sentry and you can summarize roughly what % this resolves, that would be helpful (can discuss outside the body of this PR, but I thought we have over 30 failing tests and I assume this doesn't cover all of them.)

I also think we should just close #1670 here, I don't think that ticket is serving a meaningful purpose at this point. (It was a task scoped for a specific sprint and I don't think we'd resurrect it in its current form.)

atvaccaro:
> This looks ok to me. If as part of working on this PR you did an audit of our failing tests in Sentry and you can summarize roughly what % this resolves, that would be helpful (can discuss outside the body of this PR, but I thought we have over 30 failing tests and I assume this doesn't cover all of them.)
> 
> I also think we should just close #1670 here, I don't think that ticket is serving a meaningful purpose at this point. (It was a task scoped for a specific sprint and I don't think we'd resurrect it in its current form.)

This covers the easily fixable and most annoying ones, i.e. GTFS primary key tests with lots of failing records, that I assumed would have the most impact on daily work.

I need to dig more into the payments failures and we'll probably need to discuss Airtable failures on Friday; I will add it to the agenda.

lauriemerrell:
> This covers the easily fixable and most annoying ones, i.e. GTFS primary key tests with lots of failing records, that I assumed would have the most impact on daily work.

I think that none of these tables are consumed anywhere right now (except I'm about to use feed_info for reports in one context), so they're noisy but I don't think they're actually affecting much practically.

PR #2224, opened 2023-01-30
Monthly reports ID changes

charlie-costanzo:
# Description

This PR replaces the now-closed #2203 

Recreating reports tables `gtfs_schedule_fact_route_id_changes` and `gtfs_schedule_fact_stop_id_changes` using V2 warehouse tables as `fct_monthly_route_id_changes` and `fct_monthly_stop_id_changes`

These tables also use the macro `ids_version_compare_aggregate()`, but in order to do so splits part of it into a new macro called `id_aggregate()` in the same file for use in `int_gtfs_quality__persistent_ids_schedule`

To-do:
* yml documentation, tests
* Communicate out proposed changes to other stakeholders

This PR handles the last bullet point in #2098 

## Type of change
- [x] New feature

## How has this been tested?
Locally using `dbt run` and `dbt test`


warehouse/models/mart/gtfs_quality/fct_monthly_route_id_changes.sql
```diff
@@ -0,0 +1,68 @@
+WITH dim_routes AS (
+    SELECT  * FROM {{ ref('dim_routes') }}
+),
+
+feed_version_history AS (
```
lauriemerrell, line 5: I think this can be removed, see below

warehouse/models/mart/gtfs_quality/fct_monthly_route_id_changes.sql
```diff
@@ -0,0 +1,68 @@
...
+    SELECT * FROM {{ ref('int_gtfs_quality__feed_version_history') }}
+),
+
+route_id_comparison AS (
```
lauriemerrell, line 9: As discussed offline, I think you don't need to use this

lauriemerrell (changes requested):
As we discussed offline, I think there are a few changes remaining:

Refactoring: 
1. Start with `idx_monthly_reports_site` as you do now
2. Use`int_gtfs_quality__organization_dataset_map` as to get the organization/URL/feed version combination at the beginning and end of the month 
3. Do your ID change assessment on beginning/end of month level by feed
4. Then group that ID change summary to the organization level


Two other comments we didn't discuss live but that I am seeing now:
* I think we need YAML for these new guys
* We should add the final tables to the new reports exposure that I added in my PR #2204 (but shout out to you because it's the one you sketched out AGES ago and then we had to comment until this was all implemented for real)

lauriemerrell (commented):
Just a few suggestions

warehouse/models/mart/gtfs_quality/fct_monthly_route_id_changes.sql
```diff
@@ -0,0 +1,86 @@
...
+
+),
+
+month_comparison AS (
+    SELECT
+        * EXCEPT (route_id),
+        t1.route_id AS start_table_source_id,
+        t2.route_id AS stop_table_source_id,
+        CASE
+            WHEN t2.route_id IS NULL AND t1.route_id IS NOT NULL THEN 'Removed'
```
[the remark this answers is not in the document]
lauriemerrell: Sorry, I just meant: I would aim to make it so that you can't have any rows where *both* are null, so ideally `t2.route_id IS NULL` can only happen if `t1.route_id` is populated and you shouldn't have to manually check that `t1.route_id IS NOT NULL`.... Does that make sense?

warehouse/models/mart/gtfs_quality/_mart_gtfs_quality.yml
```diff
@@ -549,6 +549,32 @@ models:
           Boolean for whether this file was present for all feeds for this organization
           on `date_checked`.
 
+  - name: fct_monthly_route_id_changes
+    description: |
+      A table that quantifies how many routes in an organization's schedule were added, removed, or unchanged
+      over the course of a month by comparing the routes scheduled on the first day of the month
```
lauriemerrell, line 555: nitpick here: it's technically not `routes scheduled on the first day of the month` but `routes present in the feed that was active on the first of the month`.... def a subtle difference but could make a difference ex. if the first of the month is a weekend the scheduled routes will be fewer but the routes present will stay the same

warehouse/models/mart/gtfs_quality/fct_monthly_route_id_changes.sql
```diff
@@ -0,0 +1,83 @@
...
+),
+
+month_comparison AS (
+    SELECT
+        t1.* EXCEPT (route_id),
+        t1.route_id AS start_table_source_id,
+        t2.route_id AS stop_table_source_id,
+        CASE
+            WHEN t2.route_id IS NULL THEN 'Removed'
+            WHEN t1.route_id IS NULL THEN 'Added'
```
lauriemerrell, line 61: So, this looks right to me but somehow it doesn't seem to be getting the `added` rows for LA Metro last month 🤔 
image

lauriemerrell (approved):
🎉

PR #2228, opened 2023-01-31
all trip updates trips in vehicle positions feed check

owades:
# Description

Measures adherence to the best practice:
```
100% of trips marked as “Scheduled”, “Canceled”, or “Added” within the Trip Updates feed are represented within the Vehicle Positions feed
```
Adds a field to `int_gtfs_rt__trip_updates_summaries` and `fct_observed_trips` to assist in filtering for scheduled, canceled & added trips.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
./dbt.sh build int_gtfs_quality__all_tu_in_vp+
Looked at aggregate results, confirmed that only a few services pass this strict check daily. Past few days of data:
The ones with N/A are due to having no trip updates feed that day - I'm curious for @lauriemerrell or @atvaccaro 's reaction to this figure (~700 TU RT feeds without results per day)
Screen Shot 2023-01-31 at 1 07 06 PM


warehouse/models/intermediate/gtfs/int_gtfs_rt__trip_updates_summaries.sql
```diff
@@ -53,6 +53,7 @@ int_gtfs_rt__trip_updates_summaries AS (
         MAX(trip_update_timestamp) AS max_trip_update_timestamp,
         MAX(trip_update_delay) AS max_delay,
         COUNT(DISTINCT CASE WHEN schedule_relationship = 'SKIPPED' THEN stop_id END) AS num_skipped_stops,
+        COUNT(DISTINCT CASE WHEN schedule_relationship IN ("SCHEDULED","CANCELED”,“ADDED") THEN 1 END) AS num_scheduled_canceled_added,
```
atvaccaro, line 56: I think you want `then stop_id` rather than `then 1` if you want to count the distinct stops; also maybe `num_scheduled_canceled_added_stops`?

Just FYI we'll have to run a full refresh of this model; by default dbt will ignore new columns ([docs](https://docs.getdbt.com/docs/build/incremental-models#what-if-the-columns-of-my-incremental-model-change)) but it is something we can change if we find ourselves doing this enough!

Also formatting nitpick: spaces between elements of the list, and single-quotes rather than double

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__all_tu_in_vp.sql
```diff
@@ -0,0 +1,56 @@
...
+        THEN 1 END) AS tu_present
+    FROM services_guideline_index AS idx
+    LEFT JOIN dim_provider_gtfs_data AS quartet
+    ON idx.service_key = quartet.service_key
+    AND idx.date BETWEEN EXTRACT(DATE FROM quartet._valid_from) AND EXTRACT(DATE FROM quartet._valid_to)
+    LEFT JOIN fct_observed_trips AS f
+    ON quartet.associated_schedule_gtfs_dataset_key = f.schedule_to_use_for_rt_validation_gtfs_dataset_key
+    AND idx.date = f.dt
+    AND f.tu_num_trips_scheduled_canceled_added > 0
+    GROUP BY 1,2
```
atvaccaro, line 37: nit: formatting and I think I usually like converting the date to a timestamp more than extracting the date from both `between` columns?

I might double-check with @lauriemerrell on this join in general, I don't quite understand `dim_provider_gtfs_data` fully yet

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__all_tu_in_vp.sql
```diff
@@ -0,0 +1,56 @@
...
+int_gtfs_quality__all_tu_in_vp AS (
+    SELECT
+        service_key,
+        date,
+        {{ all_tu_in_vp() }} AS check,
+        {{ fixed_route_completeness() }} AS feature,
+        vp_and_tu_present,
+        tu_present,
+        CASE
+            WHEN vp_and_tu_present * 1.0 / NULLIF(tu_present,0) = 1 THEN "PASS"
```
atvaccaro, line 49: can you compare the counts directly?

atvaccaro (changes requested):

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__all_tu_in_vp.sql
owades: haha, come to think of it that would be more straightforward. I guess I was setting myself up to use a percentage other than 100%

warehouse/models/intermediate/gtfs/int_gtfs_rt__trip_updates_summaries.sql
owades: Thank you!

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__all_tu_in_vp.sql
```diff
@@ -0,0 +1,56 @@
...
+       COUNT(
+        CASE WHEN
+            f.tu_num_distinct_message_ids IS NOT NULL
+        THEN 1 END) AS tu_present
+    FROM services_guideline_index AS idx
+    LEFT JOIN dim_provider_gtfs_data AS quartet
+    ON idx.service_key = quartet.service_key
+    AND TIMESTAMP(idx.date) BETWEEN quartet._valid_from AND quartet._valid_to
+    LEFT JOIN fct_observed_trips AS f
+    ON quartet.associated_schedule_gtfs_dataset_key = f.schedule_to_use_for_rt_validation_gtfs_dataset_key
```
lauriemerrell, line 34: @atvaccaro is adding actual foreign keys to `fct_daily_observed_trips` in #2231 and I think once that merges you'll want to use that -- like actually join on the quartet components (join on quartet trip updates key and vehicle positions key in addition to associated schedule, because one schedule can technically have multiple quartets associated.)

lauriemerrell (commented):
I think it makes sense to stick with this for now with just a few small tweaks noted below (some of which are on data services in terms of providing better interfaces) and it may need a slight refactor in future as our "quartet" handling matures, but I don't think it's worth going overboard on that here.

(TLDR: service/quartet relationship can cause some fanout, the steps I'm proposing below are meant to mitigate that, I think it's possible there will still be some secret weirdness but I think it will cover the 80% cases.)

warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__all_tu_in_vp.sql
owades: done!!

lauriemerrell (approved):
I think this seems ok -- if we see weird behavior on this check, we may need to revisit.

atvaccaro (approved):

PR #2230, opened 2023-02-01
new check: "100% of scheduled trips on a given day are represented within the Trip Updates feed"

owades:
# Description

Measures this guideline:
```
100% of scheduled trips on a given day are represented within the Trip Updates feed. This includes canceled trips, which should be accounted for by either marking a trip as canceled or adjusting the estimated arrival times.
```

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
` ./dbt.sh build -s int_gtfs_quality__scheduled_trips_in_tu_feed+     `
Confirmed that aggregate results look normal:
Screen Shot 2023-01-31 at 6 49 36 PM


warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__scheduled_trips_in_tu_feed.sql
```diff
@@ -0,0 +1,63 @@
...
+    SELECT
+       idx.date,
+       idx.service_key,
+       COUNT(scheduled_trips.trip_id) AS scheduled_trips,
+       COUNT(observed_trips.trip_id) AS observed_trips,
+    FROM services_guideline_index AS idx
+
+    LEFT JOIN dim_provider_gtfs_data AS quartet
+    ON idx.service_key = quartet.service_key
+    AND TIMESTAMP(idx.date) BETWEEN quartet._valid_from AND quartet._valid_to
```
lauriemerrell, line 29: Same comments as #2228 here -- I think we need a filter on `guidelines_assessed` here on `dim_provider_gtfs_data` on both joins

lauriemerrell (commented):
Most of the same comments as #2228 apply here, I think

lauriemerrell (approved):
Same caveat as #2228 I am optimistic that this will work but I also suspect there may be quirks that will need attention.

PR #2231, opened 2023-02-01
handle incremental schedule dimensions if the existing table is empty

atvaccaro:
# Description

Resolves failing dbt model from last night; have to handle incremental schedule dimensions if the existing table is empty e.g. dim_translations.

Also adds FKs to fct_observed_trips for use in quality checks as requested by @lauriemerrell 

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Both full refresh and incremental pass locally.

### DIFF
--- a/warehouse/models/mart/gtfs/_mart_gtfs_fcts.yml
+++ b/warehouse/models/mart/gtfs/_mart_gtfs_fcts.yml
@@ -1037,19 +1037,25 @@ models:
       # foreign keys
       - name: tu_gtfs_dataset_key
         tests:
-          - relationships:
+          - dbt_utils.relationships_where:
               to: ref('dim_gtfs_datasets')
               field: key
+              to_condition: "type = 'trip_updates'"
+      - name: tu_base64_url
       - name: vp_gtfs_dataset_key
         tests:
-          - relationships:
+          - dbt_utils.relationships_where:
               to: ref('dim_gtfs_datasets')
               field: key
+              to_condition: "type = 'vehicle_positions'"
+      - name: vp_base64_url
       - name: sa_gtfs_dataset_key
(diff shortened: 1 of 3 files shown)


warehouse/models/mart/gtfs/_mart_gtfs_fcts.yml
```diff
@@ -1030,6 +1034,22 @@ models:
         description: Part of a TripDescriptor.
       - name: trip_start_date
         description: Part of a TripDescriptor.
+      # foreign keys
+      - name: tu_gtfs_dataset_key
+        tests:
+          - relationships:
```
lauriemerrell, line 1040: Do we have a version of `dbt utils` that supports `relationships where`? could add a secondary check that it has the right data type but probably not really necessary

lauriemerrell (commented):
Do you think it would make sense to also have the URL for each one? I'm just wondering if we will want to use that as an unversioned foreign key for GTFS feed stuff over time 🤔 Is there a downside to adding it?

warehouse/models/mart/gtfs/_mart_gtfs_fcts.yml
atvaccaro: Ah good call, will do

atvaccaro:
> Do you think it would make sense to also have the URL for each one? I'm just wondering if we will want to use that as an unversioned foreign key for GTFS feed stuff over time 🤔 Is there a downside to adding it?

Yeah I think it's somewhat an anti-pattern to keep both the natural and versioned key everywhere but it's probably fine for now?

lauriemerrell (approved):

Issue #2208, opened 2023-01-24
Payments Validations Tests tables and dashboard updates


charlie-costanzo:
## Description
The foundation of the Payments Transaction Validation dashboard is complete, but we need to make a few updates from here:
- [x] Add date spine to transaction delta calculations (Addressed in #2232, #2240, #2241)
- [x] Add line charts and other formatting changes for transaction delta calculations by provider
- [x] Re-format dashboard
- [x] Route notifications through Slack
- [x] Remove old Sentry alerts

PR #2232, opened 2023-02-01
Payments tests date spine

charlie-costanzo:
# Description

This PR adds a date spine to `payments_daily_transaction_deltas` so that it records `n = 0` for days that a participant had no recorded rides

Partially addresses #2208 

## Type of change

## How has this been tested?
local `dbt run`

### DIFF
--- a/warehouse/models/payments_views/payments_daily_transaction_deltas.sql
+++ b/warehouse/models/payments_views/payments_daily_transaction_deltas.sql
@@ -5,6 +5,10 @@ WITH payments_rides AS (
 
 ),
 
+payments_tests_date_spine AS (
+    SELECT * FROM {{ payments_tests_date_spine() }}
+),
+
 extract_count_date AS (
 
     SELECT
@@ -42,12 +46,16 @@ payments_daily_transaction_deltas AS (
 
     SELECT
 
-        participant_id,
-        transaction_date,
-        ridership_count,
-        relative_difference
+        t1.participant_id,
+        t1.day,
 
-    FROM calculate_relative_difference
+        t2.ridership_count,
+        t2.relative_difference
+
+    FROM payments_tests_date_spine AS t1
+    LEFT JOIN calculate_relative_difference AS t2
+        ON (t1.day = t2.transaction_date)
+            AND (t1.participant_id = t2.participant_id)
 
 )
 


warehouse/macros/payments_tests_date_spine.sql
```diff
@@ -0,0 +1,40 @@
...
+            GENERATE_DATE_ARRAY(min_max_dates.min_date, min_max_dates.max_date, INTERVAL 1 DAY)
+    ) AS day_history
+
+),
+
+payments_tests_date_spine AS (
+    SELECT
+        participant_id,
+        day_history
+    FROM distinct_providers, date_range
```
atvaccaro, line 34: nit: could you explicitly call `CROSS JOIN` here

warehouse/macros/payments_tests_date_spine.sql
```diff
@@ -0,0 +1,40 @@
+{% macro payments_tests_date_spine() %}
```
atvaccaro, line 1: I would consider making this an ephemeral model rather than a macro; IMO macros without arguments tend to be code smell

warehouse/macros/payments_tests_date_spine.sql
```diff
@@ -0,0 +1,40 @@
...
+        MAX(DATE(transaction_date_time_pacific)) AS max_date
+    FROM payments_rides
+),
+
+date_range AS (
+
+    SELECT day_history
+    FROM min_max_dates,
+        UNNEST(
+            GENERATE_DATE_ARRAY(min_max_dates.min_date, min_max_dates.max_date, INTERVAL 1 DAY)
```
atvaccaro, line 25: nit: there's a dbt utils macro for this

atvaccaro (approved):

warehouse/models/payments_views/payments_tests_date_spine.sql
```diff
@@ -0,0 +1,31 @@
...
+
+min_max_dates AS (
+    SELECT
+        MIN(DATE(transaction_date_time_pacific)) AS min_date,
+        MAX(DATE(transaction_date_time_pacific)) AS max_date
+    FROM payments_rides
+),
+
+create_date_range AS (
+        {{ dbt_date.get_base_dates(start_date="2022-01-01", end_date="2022-02-01", datepart="day") }}
```
charlie-costanzo, line 20: @atvaccaro okay I hardcoded these dates into the macro here to make sure it works as expected but I'm having a hard time using the `max`/`min` values from the previous CTE for `start_date` and `end_date` - it seems to want to explicit dates? this is probably a dumb question but whats the best way to use the calculated values here?

atvaccaro: Ah yeah you'd have to get the values as strings in Jinja using `get_column_values` and template them in. I think it's fine to leave as you had it, unless you want to practice using macros.

atvaccaro: Actually on second though, I think you may be able to just select them as long as the value is properly parenthetical? see https://github.com/dbt-labs/dbt-utils#date_spine-source

PR #2238, opened 2023-02-02
Reports v2 fixes

### LINKED ISSUE #2098: Reports data updates for v2 migration

lauriemerrell:
# Description

This PR covers the outstanding dependencies needed to migrate the reports site generation to v2 and handles a bug introduced in #2204:

* Adds `organization_website` to `idx_monthly_reports_site` after conversation with @edasmalchi (rather than`feed_publisher_url` which is used in v1).
* Adds `earliest_feed_end_date` to `idx_monthly_reports_site`
* Fixes bug in `fct_monthly_reports_site_organization_validation_codes` that was causing validations from non-assessed feeds to appear (artificially inflating validation counts). 
* Adds `human_readable_description` column to `fct_monthly_reports_site_organization_validation_codes`
   * This required scraping the descriptions fairly manually from the RULES.md files in the MD repo. See https://github.com/MobilityData/gtfs-validator/issues/1324 for some context. Files related to this:
      *   `runbooks/scrape_validator_rules.ipynb` - notebook used for initial scrape
      * updates in `seeds/` -- added CSV files for each validator version that we have used with version, codes, and descriptions. Descriptions had manual cleaning done.
      * Made new `int_gtfs_quality__schedule_validator_rule_details_unioned` table to union all the new seeds together
      * Using the new unioned table, refactor the existing severities table: this improves performance considerably (24 GB to 11 KB) -- **this means that ALL validation codes will appear in the daily validations table, even ones that we have never observed in our data (whereas previously we only "knew about" codes we had observed at least once). I think this is an improvement but it does mean a noticeable expansion in the number of rows.** I am not marking this as a breaking change because we already have rows for total_notices = 0, this just expands the number of such rows but it's not fundamentally changing the grain/schema. 

Resolves #2098 for good I think

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Have run & tested affected tables locally.


warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__schedule_validator_rule_details_unioned.sql
```diff
@@ -0,0 +1,18 @@
...
+            ref('gtfs_schedule_validator_rule_details_v3_1_1'),
+            ref('gtfs_schedule_validator_rule_details_v4_0_0'),
+        ],
+    ) }}
+),
+
+int_gtfs_quality__schedule_validator_rule_details_unioned AS (
+    SELECT
+        {{ dbt_utils.surrogate_key(['code', 'version',]) }} AS key,
+        * EXCEPT(_dbt_source_relation)
```
atvaccaro, line 14: not a fan of _dbt_source_relation?

atvaccaro (approved):
🙏 

would it be worth deleting/disabling the old seeds or do you just want to minimize change right now

Issue #2141, opened 2023-01-10
Should `fct_daily_schedule_feed_validation_notices` have future dates?


lauriemerrell:
Currently the [`fct_daily_schedule_feed_validation_notices`](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/gtfs_quality/fct_daily_schedule_feed_validation_notices.sql) table contains validations for future dates. I am a bit iffy about this because this table also contains columns called `validation_success`, which I feel implies success in validating that feed **for that date**.

Wondering if we should either remove that column, at least add an explicit `is_future` column to indicate feeds that haven't been validated for the given date yet, or truncate this table to only past dates. One specific reason would be that date-aware notices like `feed_expiration_notice` will not be accurate for future dates.

PR #2249, opened 2023-02-03
stop validation notices from showing past next version's implementation

atvaccaro:
# Description

Waiting on https://github.com/cal-itp/data-infra/issues/2329 closure first

Finally stops showing validation notices in the future as well as past the next version's start. Adds a custom test to check there is only 1 validator version present per feed per day.

Closes https://github.com/cal-itp/data-infra/issues/2141

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Tests pass after rebase!

### DIFF
--- a/warehouse/models/mart/gtfs_quality/fct_daily_schedule_feed_validation_notices.sql
+++ b/warehouse/models/mart/gtfs_quality/fct_daily_schedule_feed_validation_notices.sql
@@ -45,7 +45,7 @@ first_outcome_per_feed_per_version AS (
 outcomes_with_end_dt AS (
     SELECT
         *,
-        COALESCE(LEAD(DATE_SUB(extract_dt, INTERVAL 1 DAY)) OVER (PARTITION BY feed_key ORDER BY extract_dt), feed_valid_to_dt) AS outcome_valid_to
+        COALESCE(LEAD({{ make_end_of_valid_range_date('extract_dt') }}) OVER (PARTITION BY feed_key ORDER BY extract_dt), feed_valid_to_dt) AS outcome_valid_to
     FROM first_outcome_per_feed_per_version
 ),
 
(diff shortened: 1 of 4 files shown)


lauriemerrell (commented):
Not sure that this is working as expected? (I pulled your branch & ran off that)
image
image

warehouse/models/mart/gtfs_quality/fct_daily_schedule_feed_validation_notices.sql
```diff
@@ -37,6 +41,14 @@ first_outcome_per_version AS (
     ) = 1
 ),
 
+-- each validation outcome is valid until the next version, or the end of the feed's validity
+outcomes_with_end_dt AS (
+    SELECT
+        *,
+        COALESCE(LEAD(DATE_SUB(extract_dt, INTERVAL 1 DAY)) OVER (PARTITION BY feed_key ORDER BY extract_dt), feed_valid_to_dt) AS outcome_valid_to
```
lauriemerrell, line 48: In general I do not necessarily love hard-coding `Interval 1 day` kind of things. Why not `{{ make_end_of_valid_range(LEAD(extract_dt) ... ) }}`?

lauriemerrell (commented):
Just one question about how you're populating the end date

warehouse/models/mart/gtfs_quality/fct_daily_schedule_feed_validation_notices.sql
atvaccaro: Ah just because I was dealing with dates and not timestamps; I can make a duplicate macro for dates.

lauriemerrell (approved):
I guess it seems weird that this has date-level valid from/to instead of timestamp but ok

atvaccaro:
> I guess it seems weird that this has date-level valid from/to instead of timestamp but ok

Yeah that's fair, I just did it since everything is done at the date level in this model.

atvaccaro:
> I guess it seems weird that this has date-level valid from/to instead of timestamp but ok

Yeah that's fair, I just did it since everything is done at the date level in this model.

PR #2250, opened 2023-02-03
prep for calitp grafana domain

atvaccaro:
# Description

I think Grafana might be the last service we don't have a CNAME for; the current domain is https://monitoring.k8s.calitp.jarv.us. I've looked at Composer and unfortunately they don't seem to support custom domains still https://stackoverflow.com/questions/65354788/custom-domain-for-google-cloud-composer.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml


thekaveman (approved):
CNAME created

PR #2252, opened 2023-02-03
resolve bug in check

owades:
# Description
By joining on both the vp and tu keys, I was filtering out non-null values in each. This was fine for the tu feed, since the check only cares about trips that appear in the tu feed. I removed the join to the vp feed, and instead joined on the associated schedule feed - I believe this should yield the same result.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Confirmed that some checks both pass and fail.
Ran ` ./dbt.sh build -s int_gtfs_quality__all_tu_in_vp+ `

### DIFF
--- a/warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__all_tu_in_vp.sql
+++ b/warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__all_tu_in_vp.sql
@@ -18,13 +18,13 @@ joined AS (
        idx.service_key,
        COUNT(
         CASE WHEN
-            f.tu_num_distinct_message_ids IS NOT null
+            f.tu_num_distinct_message_ids IS NOT NULL
             AND
-            f.vp_num_distinct_message_ids IS NOT null
+            f.vp_num_distinct_message_ids IS NOT NULL
         THEN 1 END ) AS vp_and_tu_present,
        COUNT(
         CASE WHEN
-            f.tu_num_distinct_message_ids IS NOT null
+            f.tu_num_distinct_message_ids IS NOT NULL
         THEN 1 END) AS tu_present
     FROM services_guideline_index AS idx
 
--- a/warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__all_tu_in_vp.sql
+++ b/warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__all_tu_in_vp.sql
@@ -40,7 +40,7 @@ joined AS (
     ON idx.date = f.dt
     AND quartet.trip_updates_gtfs_dataset_key = f.tu_gtfs_dataset_key
     -- Joining on vp_gtfs_dataset_key would ensure that vp_num_distinct_message_ids was never null
-    -- The below join condition may be unnecessary, though it could rule out potential cases where one TU feed is mapped to multiple scheduel feeds
+    -- The below join condition may be unnecessary, though it could rule out potential cases where one TU feed is mapped to multiple schedule feeds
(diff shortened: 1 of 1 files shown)


warehouse/models/intermediate/gtfs/gtfs_quality/int_gtfs_quality__all_tu_in_vp.sql
```diff
@@ -39,7 +39,9 @@ joined AS (
     LEFT JOIN fct_observed_trips AS f
     ON idx.date = f.dt
     AND quartet.trip_updates_gtfs_dataset_key = f.tu_gtfs_dataset_key
-    AND quartet.vehicle_positions_gtfs_dataset_key = f.vp_gtfs_dataset_key
+    -- Joining on vp_gtfs_dataset_key would ensure that vp_num_distinct_message_ids was never null
+    -- The below join condition may be unnecessary, though it could rule out potential cases where one TU feed is mapped to multiple scheduel feeds
```
lauriemerrell, line 43: Massive nit: typo in `scheduel`

lauriemerrell (commented):
Just to confirm: Is the intention that if there's a trip updates feed with no vehicle positions feed at all, that should fail? And if there's a vehicle positions feed with no trip updates, is the result supposed to be `N/A`? I wonder if it's worth just double checking each of those cases to ensure it's coming out as expected.

Here's a query identifying this (you can swap tu and vp to check the other one):

```
SELECT *
FROM `cal-itp-data-infra.mart_transit_database.dim_provider_gtfs_data`
WHERE vehicle_positions_gtfs_dataset_key IS NULL AND trip_updates_gtfs_dataset_key IS NOT NULL AND _is_current
ORDER BY organization_name
```

Apologies for misleading, I guess I had assumed that the intention was "if both trip updates and vehicle positions are present, this condition should be met" (which, annoying as it is, is what the indices help clarify: what is our universe for each check... because this is implemented as a `service` check, we can kind of do any combination)

owades:
> Just to confirm: Is the intention that if there's a trip updates feed with no vehicle positions feed at all, that should fail? And if there's a vehicle positions feed with no trip updates, is the result supposed to be `N/A`? I wonder if it's worth just double checking each of those cases to ensure it's coming out as expected.
> 
> Here's a query identifying this (you can swap tu and vp to check the other one):
> 
> ```
> SELECT *
> FROM `cal-itp-data-infra.mart_transit_database.dim_provider_gtfs_data`
> WHERE vehicle_positions_gtfs_dataset_key IS NULL AND trip_updates_gtfs_dataset_key IS NOT NULL AND _is_current
> ORDER BY organization_name
> ```
> 
> Apologies for misleading, I guess I had assumed that the intention was "if both trip updates and vehicle positions are present, this condition should be met" (which, annoying as it is, is what the indices help clarify: what is our universe for each check... because this is implemented as a `service` check, we can kind of do any combination)

Thank you for looking through this and thinking critically about it as usual. I think the original design may have been correct, and that we might need updates to `fct_observed_trips`. In hindsight this PR is basically a hack to get directionally accurate results.

The issue is that as it stands, if there are no vehicle positions for a given trip for a given day, `fct_observed_trips` will have a null value for the associated `vp_gtfs_dataset_key`. I think we need to switch the order of the joins in the `vehicle_positions_with_associated_schedule` [subquery](https://github.com/cal-itp/data-infra/blob/f97681a8d808cce6a58a6f5cb3ca2220219926ba/warehouse/models/mart/gtfs/fct_observed_trips.sql#L69). @atvaccaro

lauriemerrell (approved):

lauriemerrell:
Talked to @owades offline, we are just going to merge this and revisit again later

PR #2265, opened 2023-02-07
Sentry/Grafana infra improvements

atvaccaro:
# Description

* Create backup cronjob for Sentry postgres
* Deploy postgres for use by Grafana
* Create backup cronjob for Grafana postgres
* Add monitoring.calitp.org as host for Grafana

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

### DIFF
--- a/kubernetes/apps/charts/postgresql/.helmignore
+++ b/kubernetes/apps/charts/postgresql/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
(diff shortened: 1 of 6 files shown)


themightychris:
The postgres server here isn't namespaced to indicate it's scoped to Grafana—is there an idea of this postgres server being a central shared resource for config backends?

If so, Metabase has a scoped postgres server running for its config backend already, would be good to think about folding that one in eventually if central is the direction. Otherwise if the intention is this only be for grafana it should be scoped as such

atvaccaro:
> The postgres server here isn't namespaced to indicate it's scoped to Grafana—is there an idea of this postgres server being a central shared resource for config backends?
> 
> If so, Metabase has a scoped postgres server running for its config backend already, would be good to think about folding that one in eventually if central is the direction. Otherwise if the intention is this only be for grafana it should be scoped as such

Ah no I'm planning on keeping them separate; are you referring to the helm chart itself? My plan was to keep a generic chart and then use namespaces to deploy them per "thing". For example, in retrospect it could've been useful to use this same pattern for Sentry as well rather than the Sentry Helm-managed Postgres. I'm more than happy to split up the chart per-application if you prefer.

lottspot:
What is the thinking in introducing new charts vs using the previously existing resources embedded in the candidate branch?

atvaccaro:
> What is the thinking in introducing new charts vs using the previously existing resources embedded in the candidate branch?

Can you clarify on which resources? I thought the Grafana chart only had sqlite, and then we don't have anything in the cluster template for Sentry IIRC.

lottspot:
> Can you clarify on which resources?

Total space out on my part... I forgot that the metabase DB is actually embedded with the chart and is not a standalone chart. This is a great addition, thanks for including it!

lottspot (approved):

Issue #2235, opened 2023-02-01
Update Metabase dim_ntd_agency_info NTD Data to 2021


juliama0780:
Update `dim_ntd_agency_info` to 2021 data. The NTD dataset is this [dataset](https://dashboards.calitp.org/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7ImRhdGFiYXNlIjoyLCJ0eXBlIjoicXVlcnkiLCJxdWVyeSI6eyJzb3VyY2UtdGFibGUiOjI0NTJ9fSwiZGlzcGxheSI6InRhYmxlIiwidmlzdWFsaXphdGlvbl9zZXR0aW5ncyI6e319) in Metabase.

We will be demo-ing the GTFS Features dashboard mid-February and would appreciate an update from the 2018 data to 2021 data, from [this file.](https://www.transit.dot.gov/ntd/data-product/2021-annual-database-agency-information)

atvaccaro:
Discussed this during sprint planning. Going to implement a very minimal loader, save a few years of data to GCS, and make available in the warehouse via external tables. Not going to automate anything or load other datasets yet, but should provide a pattern for future work.

PR #2269, opened 2023-02-07
NTD extract script and dbt models

atvaccaro:
# Description

Resolves https://github.com/cal-itp/data-infra/issues/2235

This PR sets up a framework for further NTD extracts, but doesn't actually set up any orchestration or implement beyond the requested data. We performed an extract and results are stored at `gs://calitp-ntd-data-products/annual-database-agency-information/dt=2023-02-08/ts=2023-02-08T17:22:59.824755+00:00/year=2021/annual-database-agency-information.jsonl.gz`. We used the latest extract of the Airtable NTD table to create a seed mapping from `ntd_id` to our organizations by name rather than Airtable record ID; `name` is currently unique at the organization level.

There's now a new dimension table called `mart_ntd.dim_annual_database_agency_information` which is a type 2 dimension of this particular NTD report at the year level; i.e. the type 2 windows are at the `year, ntd_id` level. This will let us support multiple extracts of the yearly data. We've provided an unversioned key on `dim_organizations` called `annual_database_agency_information_2021_key` that points to the latest extracted version of the 2021 data present in `dim_annual_database_agency_information`.

There are two main follow-ups to this.
1. The type 2 versioning on `dim_annual_database_agency_information` is naive and does not handle deletions.
2. For eventual historical lookups, we'll need a bridge table between `dim_organizations` and `dim_annual_database_agency_information`.

We left these out of scope since @juliama0780 only needs the current version of the 2021 data.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Tested all the way down through the dimensions in dbt.

## Screenshots (optional)
image


warehouse/seeds/_seeds.yml
```diff
@@ -81,3 +81,16 @@ seeds:
       - name: severity
         tests:
           - not_null
+
+  - name: ntd_agency_to_organization
+    description: Extract of Airtable mapping from NTD agency info to transit database organizations.
```
lauriemerrell, line 86: Can you add a comment about versioning / how & when this matching was done? And why are you using organization name rather than `source_record_id`?

warehouse/models/mart/ntd/_mart_ntd.yml
```diff
@@ -0,0 +1,33 @@
+version: 2
+
+models:
+  - name: dim_annual_database_agency_information
+    tests:
```
lauriemerrell, line 5: Please add a description that explicitly identifies what year / table data this is, preferably with a pretty specific link to the source data and perhaps a note about how to join with organizations.

script/scrape_ntd.py
```diff
@@ -0,0 +1,72 @@
+"""
+
```
lauriemerrell, line 2: Might be worth a contextualizing comment -- is this mostly ready to become an Airflow operator and/or DAG task once we're ready...?

script/requirements.txt
```diff
@@ -1,8 +1,70 @@
-# This requirements.txt file is used to install the necessary packages to run the python scripts in
-# this scripts folder
-jinja2==2.11.3
-PyYAML==5.4.1
-jsonschema==3.2.0
-pandas==1.1.4
-requests==2.22.0
-markupsafe==2.0.1
+aiohttp==3.7.4.post0; python_version >= "3.8" and python_version < "3.11"
```
lauriemerrell, line 1: Why did this expand so much / why do we have both requirements.txt and poetry in here? 😅

airflow/dags/create_external_tables/ntd_data_products/annual_database_agency_information.yml
```diff
@@ -0,0 +1,139 @@
+operator: operators.ExternalTable
+bucket: gs://calitp-ntd-data-products
+prefix_bucket: true
+post_hook: |
```
lauriemerrell, line 4: Do we want to keep using these (post-hooks)?

warehouse/seeds/ntd_agency_to_organization.csv
```diff
@@ -0,0 +1,231 @@
+ntd_id,legacy_ntd_id,organization_name
```
lauriemerrell, line 1: Did you guys do any spot checks on this? If we could run a summary on # organizations with GTFS data but no NTD ID, I wonder if we can make a to-do item for someone to check on those. The coverage on this is one of the things I am most nervous about for analyst usage.

lauriemerrell (commented):
Mostly fairly nitpicky but we do definitely need to add a useful, analyst usage-oriented description for the new mart table.

warehouse/seeds/_seeds.yml
atvaccaro: I wanted this to be more human-maintainable and I didn't expect raw Airtable record IDs to be particularly user-friendly; also name has more semantic meaning here. I think my main question is whether you would prefer that the record ID or the org name be the join identifier?

lauriemerrell: Joining on name-y strings always makes me a bit squirrelly about things like punctuation and capitalization (especially because organizations have a name within the NTD source that is not guaranteed to the be the same as our organization name). So I feel like record ID is safer, but I understand it's not very user-friendly. Sorry that's not conclusive.

script/requirements.txt
atvaccaro: ah I just wanted to do this since it helps with enforcing a given Python version; I exported to requirements.txt to make the existing CI work although it's going to be made immediately redundant by https://github.com/cal-itp/data-infra/pull/2262

airflow/dags/create_external_tables/ntd_data_products/annual_database_agency_information.yml
atvaccaro: I'm ambivalent, I think they're useful quick tests on small data sets

warehouse/seeds/ntd_agency_to_organization.csv
lauriemerrell: Sorry I meant more like spot checks on what % of organizations is that, and how many organizations with GTFS don't match, probably most importantly how many organizations in `fct_daily_organization_combined_guideline_checks` don't match

warehouse/seeds/_seeds.yml
atvaccaro: I'll keep both in the seed but join on record_id for now.

warehouse/models/mart/ntd/_mart_ntd.yml
```diff
@@ -0,0 +1,41 @@
+version: 2
+
+models:
+  - name: dim_annual_database_agency_information
+    description: |
+      Versioned extracts of the NTD Annual Database Agency Information.
```
lauriemerrell, line 6: Doesn't this only contain 2021 right now? I think it is better to be clear about what is actually in here. If we automate this we can update it but presumably that will require other updates anyway.

atvaccaro: Ah I was going to import some others too; specifically 2018 at least since that's what is in Airtable. The key on `dim_organizations` only points to 2021. If you want to limit this to 2021, I can do that.

lauriemerrell: We already have `mart_transit_database.dim_ntd_agency_info` containing the 2018 data. And I guess maybe that makes me realize we should just add this in there...? But that is confusing because right now that is structured as an Airtable-sourced table. If you bring 2018 data in this table let's deprecate `mart_transit_database.dim_ntd_agency_info`

atvaccaro: Right I was planning on deprecating the Airtable table; I don't think we actually want/need it in the broader Airtable framework.

atvaccaro:
Discussed briefly offline; going to merge as-is and do @lauriemerrell's requested org checks against production data. Also marking `dim_ntd_agency_info` as deprecated.

lauriemerrell:
Talked offline, I am ok with merging this as-is and then doing some follow up to clean up 2018 (by deleting the existing table and then using this new seed file to get the NTD ID on tables like the assessed entity candidates) after testing in prod.

warehouse/seeds/ntd_agency_to_organization.csv
atvaccaro: https://github.com/cal-itp/data-infra/issues/2272

lauriemerrell (approved):

PR #2304, opened 2023-02-21
Assorted minor guidelines cleanup

### LINKED ISSUE #2223: We are not assessing entities (org/service/dataset combos) that are missing a component
In the course of working on #2204 I noticed that in `int_gtfs_quality__daily_assessment_candidate_entities` we are `COALESCE`ing the assessment statuses to `false` in such a way that if there is an entity that is missing that component, it will be marked as not assessed (see how in the precursor `full join` each individual assessment determination -- for organizations, datasets, services, and service data is coalesced separately to false, meaning that if there's an entity that doesn't have that component, the assessment status will be false).

So an organization/service pair that has no GTFS data at all will be marked as not assessed.  

I need to check with @owades and others for what we want: do we want to be able to generate an assessment report in these cases? If the answer is no, I think this is fine. If we do want to be able to generate assessments in these cases then we need to re-work this coalescing logic.

This is mostly going to affect [fct_daily_organization_combined_guideline_checks](https://dbt-docs.calitp.org/#!/model/model.calitp_warehouse.fct_daily_organization_combined_guideline_checks) -- each individual column (for org, service, dataset) is carried forward into `fct_daily_guideline_checks` so we can change this without too much trouble I think (may depend on need for full refreshes.)

We should go through each combination:
* Do we want to assess an organization...
   * With no service?
   * With service but no GTFS?
 * Do we want to assess a service...
    * With no organization?
    * With no GTFS?
 * Do we want to assess GTFS data...
    * With no organization?
    * With no service?

### LINKED ISSUE #2263: GTFS guidelines refactor N/As & assorted cleanup
Talked to @owades and @atvaccaro today in the course of working on #2257 and identified a bundle of various small fixes needed for GTFS guidelines work:

- [x] Add not null % checks on the `status` column in guideline check tables
- [x] Move `fct_daily_guideline_checks` to an intermediate table (out of the mart)
- [x] Add organizations and services with no GTFS data to `fct_daily_guideline_checks` (stop coalescing GTFS service data assessment data to FALSE in assessed entities intermediate table)
- [x] Improve N/A logic in `fct_daily_guideline_checks` so that it's more rational: see [proposal doc](https://docs.google.com/document/d/1L9djR4EZnxLTnqM6bcLGax_mCQZm9aDhf9_gCCpatyU/edit#heading=h.57pi2w5i9ev7)
   - [x] Includes updating a few models to use different indices 
- [x] Make a macro for the `N/A` status values instead of hard-coding them (and maybe we should do PASS and FAIL too while we are at it?)
- [ ] ~Add something to `fct_implemented_checks` that indicates the first date that a given check was checked (see [here](https://github.com/cal-itp/data-infra/pull/2276#discussion_r1102019471)) and use that in `fct_daily_guideline_checks` so we don't have checks listed for days on which they were not assessed (or, alternatively, create a third `N/A` status for "before check was implemented")~ Laurie 3/29/23: skipping for now, did implement a new NA too early status

After working on #2304, here is my plan for the remaining to do items here:

- [x] Ditch the individualized indices entirely; just add entity flags to the candidate entities table with which checks apply
- [x] Use the general index for all tests
- [x] For rows with no applicable entity, use the new not applicable value to indicate as such
- [x] Delete the `fct_daily_*_guideline_checks` tables (except `organization_combined` and `service_combined`)
   - [x] This will require refactoring fct implemented checks  
- [x] Add date handling at the check level.... May require another new `N/A` status for `check not in effect yet`
- [x] Swap order of manual check needed and pass in the guideline aggregation logic
- [x] Fix inline no QA notes

Other TODOs here:
- [x] test row count test performance & decide on testing approach
- [ ] remove unused macros once done  
- [x] Handle new checks in #2339
- [x] Reimplement fct_implemented_checks

lauriemerrell:
# Description

This PR contains a bunch of mostly-stylistic changes towards #2263.

* Rearranges file structures for GTFS quality so that:
   * `gtfs_quality` is no longer under `gtfs` in `staging` and `intermediate`
   * Creates new `intermediate/gtfs_quality/guideline_checks` folder (and YAML file) so that checks are a bit more separate in the file structure and in particular this lets you run them all separately from other GTFS quality tables. 
* Moves `fct_daily_guideline_checks` to intermediate table `int_gtfs_quality__guideline_checks_long` (out of the mart)
* Adds organizations and services with no GTFS data to `int_gtfs_quality__guideline_checks_long`
* Creates & swaps in macros for all check statuses (instead of hard-coding strings), including two new `N/A` statuses; only the `N/A Check Logic` one is actually used here 
* Creates a macro for the guidelines aggregation logic 
* Linter updates on most of the guideline check models 

While working on this I realized that the more structural stuff probably deserves a dedicated PR and some separate discussion. 

Resolves some of #2263, also resolves #2223 that I forgot I had written 

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

`poetry run dbt run -s models/intermediate/gtfs_quality/guidelines_checks/` & ` poetry run dbt test -s models/intermediate/gtfs_quality/guidelines_checks/`; manually tested the organizations & services changes


warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__lead_time.sql
```diff
@@ -53,7 +53,7 @@ trips_expanded AS (
       LEFT JOIN dim_stops t3
         ON t3.stop_id = t2.stop_id
        AND t3.feed_key = t2.feed_key
-     GROUP BY 1,2,3
+     GROUP BY 1, 2, 3 -- noqa
```
owades, line 56: what is "-- noqa"?

warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__lead_time.sql
```diff
@@ -129,7 +129,7 @@ daily_improper_trips_updates AS (
    WHERE
          -- Service date is between 0 and 7 days from now
          days_until_service_date > 0 AND days_until_service_date <= 7
-   GROUP BY 1,2,3
+   GROUP BY 1, 2, 3 -- noqa
```
owades, line 132: same question about noqa

warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__persistent_ids_schedule.sql
```diff
@@ -15,20 +14,20 @@ dim_agency AS (
     SELECT  * FROM {{ ref('dim_agency') }}
 ),
 
-feed_version_history AS (
+feed_version_history AS ( --noqa
```
owades, line 17: one more

owades (changes requested):
Just requesting explanation about the "noqa" comments in a few places.

owades:
@lauriemerrell could you also explain the impact of 
`Adds organizations and services with no GTFS data to int_gtfs_quality__guideline_checks_long`?

warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__lead_time.sql
lauriemerrell: It's to suppress [SQLFluff](https://docs.sqlfluff.com/en/stable/) warnings on this line. @atvaccaro said he had to turn off SQLFluff in CI so it hasn't been running. But thanks for flagging, I think this suppression is probably too intense and I should be more specific to turn off specific rules

lauriemerrell:
> @lauriemerrell could you also explain the impact of
Adds organizations and services with no GTFS data to int_gtfs_quality__guideline_checks_long?

Yeah this is what we discussed in [this doc](https://docs.google.com/document/d/1L9djR4EZnxLTnqM6bcLGax_mCQZm9aDhf9_gCCpatyU/edit#heading=h.zdt3vnfxqu8f), basically just that now an organization or service can appear in the final tables even if they don't have GTFS data (previously, we were suppressing all organizations without GTFS data). 

I think we will need to add a follow up for a specific check that these will fail, but in the interim these will appear with a lot of nulls. Documented this need here: https://github.com/cal-itp/data-infra/issues/2311

Looking for examples here has caused me to find some potential bugs though so I will look into those (suspiciously, `int_gtfs_quality__all_tu_in_vp` now only contains `N/A` results since 2/15).

lauriemerrell:
Notes to self for tomorrow to keep investigating -- compare these two:

```
SELECT status, COUNT(*)
FROM `cal-itp-data-infra-staging.laurie_staging.int_gtfs_quality__guideline_checks_long`
WHERE check = '100% of trips marked as Scheduled, Canceled, or Added within the Trip updates feed are represented within the Vehicle positions feed' AND date >= '2023-02-15'
GROUP BY status
```
And
```
SELECT status, COUNT(*)
FROM `cal-itp-data-infra.mart_gtfs_quality.fct_daily_guideline_checks`
WHERE check = '100% of trips marked as Scheduled, Canceled, or Added within the Trip updates feed are represented within the Vehicle positions feed' AND date >= '2023-02-15'
GROUP BY status
```

* My branch has ~300 more N/As, when it should probably have nulls? 
* The organization and service combined checks need to have the `NULLIF`s updated to use the NA macros instead of the string `N/A`

lauriemerrell:
Ok, this is ready for review again -- quoting myself:

> My branch has ~300 more N/As, when it should probably have nulls?

This was actually correct; because the `tu in vp` check is implemented as a `service` level check, services without RT get a `N/A - check logic` flag even though they probably should be `N/A - no entity`. We can possibly clean that up later but I think the behavior here is correct for now.

I confirmed that in the `organization_combined` table we are getting nulls for organizations with no GTFS on the GTFS-specific checks. 

>The organization and service combined checks need to have the `NULLIF`s updated to use the `NA` macros instead of the string `N/A`

Done, I created a macro for this to simplify in future.

warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__persistent_ids_schedule.sql
```diff
@@ -15,20 +14,22 @@ dim_agency AS (
     SELECT  * FROM {{ ref('dim_agency') }}
 ),
 
-feed_version_history AS (
+-- SQLFluff thinks this CTE isn't used, but it is required for the ids_version_compare_aggregate macro
+feed_version_history AS ( -- noqa: disable=L045
```
atvaccaro, line 18: nit: would making the macro take a CTE name or `ref()` fix that?

warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__lead_time.sql
```diff
@@ -129,7 +131,9 @@ daily_improper_trips_updates AS (
    WHERE
          -- Service date is between 0 and 7 days from now
          days_until_service_date > 0 AND days_until_service_date <= 7
-   GROUP BY 1,2,3
+   -- SQLFluff doesn't like numeric aliasing here alongside explicit naming in window functions
+   GROUP BY 1, 2, 3 -- noqa: disable=L054
+   -- noqa: enable=L054
```
atvaccaro, line 136: does `-- noqa: L054` at the end of the line not work here?

warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__lead_time.sql
```diff
@@ -53,7 +53,9 @@ trips_expanded AS (
       LEFT JOIN dim_stops t3
         ON t3.stop_id = t2.stop_id
        AND t3.feed_key = t2.feed_key
-     GROUP BY 1,2,3
+   -- SQLFluff doesn't like numeric aliasing here alongside explicit naming in window functions
+   GROUP BY 1, 2, 3 -- noqa: disable=L054
+   -- noqa: enable=L054
```
atvaccaro, line 58: does `-- noqa: L054` at the end of the line not work here?

atvaccaro (approved):

owades (approved):

PR #2310, opened 2023-02-22
Fix dbt-metabase sync in warehouse job

atvaccaro:
# Description

dbt-metabase has changed behavior so it raises exceptions if any errors are thrown during a sync, for example if a column in YAML does not exist in the warehouse. While we could just ignore/suppress these exceptions, I'm just fixing the YAML.

A couple lessons from this: I think  ephemeral models do not play nicely with dbt-metabase; they aren't instantiated in the database so there's nothing for FKs to point at. Technically this applies to anything that isn't referenced in a FK test that isn't in Metabase; in this specific case it was actually a staging table.

Also I realized we don't actually set a schema for the new latest GTFS models, so I've done that here.

## Type of change

- [ ] New feature
- [x] Documentation
- [ ] agencies.yml

## How has this been tested?
I've tested all 8 active schemas that we publish (adjusted them in this PR), and all sync to Metabase successfully now. I'm also trying to ensure `docker-compose run airflow tasks test deploy_dbt_docs sync_dbt_docs_to_metabase 2023-02-22` works so we can actually just update docs as needed, but that will necessitate a full refresh in staging first.


atvaccaro:
I do think I need reviews from both @lauriemerrell and @charlie-costanzo since this changes models they created.

warehouse/models/mart/gtfs/_mart_gtfs_fcts.yml
```diff
@@ -263,9 +263,10 @@ models:
       - name: vehicle_license_plate
         description: |
           See: https://gtfs.org/realtime/reference/#message-vehicledescriptor.
-      - name: vehicle_wheelchair_accessible
-        description: |
-          See: https://gtfs.org/realtime/reference/#message-vehicledescriptor.
+        # Should this exist? It does not currently in the model.
```
lauriemerrell, line 266: Yes, can you write a ticket to add it? This is part of the spec and given our interest in accessibility information we should have it. 
image

warehouse/models/mart/gtfs_quality/_mart_gtfs_quality.yml
```diff
@@ -30,9 +30,11 @@ models:
         name: check
         tests:
           - not_null
-          - relationships:
-              to: ref('stg_gtfs_quality__intended_checks')
-              field: check
+          # This is a useful test but dbt-metabase fails since
```
lauriemerrell, line 33: I kind of think the answer here is going to be to instantiate the intended checks table in my refactor of the guidelines indexing stuff

warehouse/models/mart/transit_database/_mart_transit_database.yml
```diff
@@ -487,16 +468,14 @@ models:
           * "Regional Precursor Feed" - This category applies to agencies whose data appears in a
             combined regional feed but who produce some other independent dataset.
             The independent dataset may be an input into the combined regional feed.
-      - name: agency_id
```
lauriemerrell, line 490: These columns should be here 👀 If they got removed I think we should add them back in the actual model?

warehouse/scripts/run_and_upload.py
```diff
@@ -264,34 +264,34 @@ def get_command(*args) -> List[str]:
...
+        results_to_check.append(
+            subprocess.run(
+                [
+                    "dbt-metabase",
+                    "models",
+                    "--dbt_manifest_path",
+                    "./target/manifest.json",
+                    "--metabase_database",
+                    "Data Marts (formerly Warehouse Views)",
+                    "--dbt_schema_excludes",
```
lauriemerrell, line 276: I am a little nervous about changing from an inclusion list to an exclusion list, doesn't this mean that any new schema will be included by default? Is that what we want?

airflow/dags/deploy_dbt_docs/sync_dbt_docs_to_metabase.yml
```diff
@@ -25,19 +26,21 @@ env_vars:
...
+  MB_HOST: dashboards.calitp.org
 
 secrets:
   - deploy_type: volume
     deploy_target: /secrets/jobs-data/
     secret: jobs-data
     key: service-account.json
   - deploy_type: env
-    deploy_target: METABASE_USER
+    deploy_target: MB_USER
```
lauriemerrell, line 39: Just curious about the reason for changing this?

lauriemerrell (commented):
Just a few questions.

airflow/dags/deploy_dbt_docs/sync_dbt_docs_to_metabase.yml
atvaccaro: Ah dbt-metabase will pick up the new ones directly, rather than us getting them from `os.environ` and passing as CLI args in the subprocess call.

warehouse/scripts/run_and_upload.py
atvaccaro: Hmm that's a good point; my assumption was that we'd really only have new mart schemas and we'd generally sync them to the data marts DB in Metabase. (Also it's only 1 CLI call instead of N). I was trying to avoid the silent failure case of someone creating a new schema, adding it in Metabase, and not seeing docs update. Do you have thoughts on that vs erroring when a new schema is added, and explicitly ignoring if it isn't set in Metabase?

warehouse/models/mart/gtfs_quality/_mart_gtfs_quality.yml
atvaccaro: I think in this case it's also because the intended checks model would end up in `staging` which isn't in Metabase. I wasn't able to find a flag to tell dbt-metabase to ignore specific columns, only a whole model, which might work.

warehouse/scripts/run_and_upload.py
lauriemerrell: (we are discussing this offline)

lauriemerrell: Ok, I misunderstood what this controls. This just syncs documentation & column metadata annotations. The core setting of what schemas' **data** to actually sync to Metabase is exclusively in the Metabase UI. So there isn't a risk here of **data** being exposed in Metabase unintentionally.

warehouse/models/mart/gtfs/_mart_gtfs_fcts.yml
atvaccaro: I'll just do it here, it's just not in the mart.

lauriemerrell (approved):

PR #2332, opened 2023-02-27
split failure emails into separate task again

atvaccaro:
# Description

Airflow emails via Postmark are not working right now; I'm trying to debug but it's hard locally, so I want to be able to retry just this task. Also, we shouldn't be retrying the actual download task because the failure email failed to send. The v1 pipeline had separate tasks and we just didn't split them up in v2.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Locally.

## Screenshots (optional)
image

### DIFF
--- a/airflow/dags/download_gtfs_schedule_v2/download_schedule_feeds.py
+++ b/airflow/dags/download_gtfs_schedule_v2/download_schedule_feeds.py
@@ -8,7 +8,6 @@ import logging
 from typing import ClassVar, List, Optional
 
 import humanize
-import pandas as pd
 import pendulum
 import sentry_sdk
 from calitp_data_infra.auth import get_secrets_by_label
@@ -165,7 +164,9 @@ def download_all(task_instance, execution_date, **kwargs):
         )
         task_instance.xcom_push(
             key="download_failures",
-            value=pd.DataFrame(f.dict() for f in result.failures),
+            value=[
+                json.loads(f.json()) for f in result.failures
+            ],  # use the Pydantic serializer
         )
 
     success_rate = len(result.successes) / len(configs)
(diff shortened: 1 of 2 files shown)


airflow/dags/download_gtfs_schedule_v2/download_schedule_feeds.py
```diff
@@ -167,39 +162,11 @@ def download_all(task_instance, execution_date, **kwargs):
...
-                "evan.siroky@dot.ca.gov",
-                "hunter.owens@dot.ca.gov",
-                "jameelah.y@jarv.us",
-                "olivia.ramacier@dot.ca.gov",
-            ],
-            html_content=html_content,
-            subject=(
-                f"Operator GTFS Errors for {datetime.datetime.now().strftime('%Y-%m-%d')}"
-            ),
+        task_instance.xcom_push(
```
lauriemerrell, line 165: `xcom` 😅

airflow/dags/download_gtfs_schedule_v2/email_download_failures.py
```diff
@@ -0,0 +1,56 @@
...
+            task_ids="download_schedule_feeds", key="download_failures"
+        )
+    )
+    if failures_df.empty:
+        html_content = f"All feeds were downloaded successfully on {execution_date}!"
+    else:
+        html_report = failures_df.to_html(border=False)
+
+        html_content = f"""\
+    NOTE: These failures come from the v2 of the GTFS Schedule downloader.
```
lauriemerrell, line 30: we can prob remove this now

lauriemerrell (approved):

airflow/dags/download_gtfs_schedule_v2/download_schedule_feeds.py
atvaccaro: I think this is the appropriate use case 😆

PR #2334, opened 2023-02-27
Bugfix: update manual check logic based on values entered in airtable

owades:
# Description

When I set up the manual checks last month, we didn't have any results in airtable. Now that we do, I've gone through and updated the logic to reflect the actual values for these checks. I did this by looking at distinct values for each manual check and confirming that each case was covered by the corresponding intermediate check table.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
I confirmed that each intermediate table runs successfully.

### DIFF
--- a/warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__data_license.sql
+++ b/warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__data_license.sql
@@ -23,9 +23,13 @@ int_gtfs_quality__data_license AS (
             WHEN gtfs_datasets.type IN ("vehicle_positions", "trip_updates", "service_alerts") THEN {{ compliance_rt() }}
         END AS feature,
         CASE
-            WHEN manual_check__data_license LIKE ('%Permissive%') THEN {{ guidelines_pass_status() }}
+            WHEN manual_check__data_license LIKE ('%Permissive%')
+                 OR manual_check__data_license LIKE ('%CC-BY%')
+                 OR manual_check__data_license LIKE ('%ODL-BY%')
+                THEN {{ guidelines_pass_status() }}
             WHEN manual_check__data_license LIKE ('%Restrictive%')
-                 OR manual_check__data_license = 'Missing' THEN {{ guidelines_fail_status() }}
+                 OR manual_check__data_license = 'Missing'
+                THEN {{ guidelines_fail_status() }}
             WHEN manual_check__data_license = 'N/A - dataset is not public-facing' THEN {{ guidelines_na_check_status() }}
             ELSE {{ guidelines_manual_check_needed_status() }}
         END AS status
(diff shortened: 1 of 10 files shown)


warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__data_license.sql
```diff
@@ -22,9 +22,11 @@ int_gtfs_quality__data_license AS (
             WHEN gtfs_datasets.type = "schedule" THEN {{ compliance_schedule() }}
             WHEN gtfs_datasets.type IN ("vehicle_positions", "trip_updates", "service_alerts") THEN {{ compliance_rt() }}
         END AS feature,
-        CASE manual_check__data_license
-            WHEN 'Yes' THEN {{ guidelines_pass_status() }}
-            WHEN 'No' THEN {{ guidelines_fail_status() }}
+        CASE
```
owades, line 25: @evansiroky could you confirm this logic checks out? Basically anything containing the  word "Permissive" passes, and containing "Restrictive" with the exact text "Missing" fails

owades:
@lauriemerrell could you review? (now that you are auto-tagged i can't directly request a review. let me know if there's a better way to ask and/or if i don't need to ask due to autotag)

warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__trip_planner_rt.sql
```diff
@@ -17,6 +17,7 @@ int_gtfs_quality__trip_planner_rt AS (
         CASE manual_check__gtfs_realtime_data_ingested_in_trip_planner
             WHEN 'Yes' THEN {{ guidelines_pass_status() }}
             WHEN 'No' THEN {{ guidelines_fail_status() }}
+            WHEN 'N/A - no fixed-route service' THEN {{ guidelines_na_check_status() }}
```
lauriemerrell, line 20: I think this is fine for these purposes (like, within this PR) but ideally perhaps we might check that this aligns with that's in the warehouse/maybe there would be some kind of automated QC in Airtable to make sure that this isn't selected in error

lauriemerrell (approved):
LGTM assuming Evan agrees

warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__data_license.sql
evansiroky: There are also some options of `CC-BY` and `ODL-BY` that should also have a passing status. Otherwise, it looks good.

owades: Thanks, I'm updating it to include those two values (in caps)

warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__trip_planner_rt.sql
owades: do you mean that we would actually query the distinct values from the airtable datatable and assign them here, rather than just writing them as text?

owades:
added additional passing status values, ran it to confirm it still works. merging.

warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__trip_planner_rt.sql
lauriemerrell: Ope sorry should have been more specific -- I just meant that `no fixed-route service` should probably have an automated check to confirm it aligns with the new `fixed-route` checkbox / data structure (i.e., make sure that this value is not selected for a service that we elsewhere say has fixed-route service)

Issue #1897, opened 2022-10-13
Bug: Need to adjust RT fetch exception fingerprinting


atvaccaro:
**Describe the bug**
The default string representation of some connection exceptions include the default Python object repr which has a hexadecimal memory reference to the object. It means we are not grouping events like the following:

https://sentry.calitp.org/organizations/sentry/issues/4111/?environment=cal-itp-data-infra&project=2&query=is%3Aunresolved+mybusinfo&statsPeriod=14d
https://sentry.calitp.org/organizations/sentry/issues/4106/?environment=cal-itp-data-infra&project=2&query=is%3Aunresolved+mybusinfo&statsPeriod=14d

The fingerprint should probably use the top-level exception only, and not the underlying exception.

**To Reproduce**
Check any exception referencing an HTTPConnection object.

**Expected behavior**
These events are grouped going forward.

**Additional context**

atvaccaro:
This is still a current issue, see https://sentry.calitp.org/organizations/sentry/issues/39602/?environment=cal-itp-data-infra&project=2&query=is%3Aunresolved

PR #2336, opened 2023-02-27
remove object memory locations from exception string representation

atvaccaro:
# Description

Resolves https://github.com/cal-itp/data-infra/issues/1897

The exception string looks like this post-replacement: `"HTTPSConnectionPool(host='avl.yctd.org', port=443): Max retries exceeded with url: /RealTime/GTFS_TripUpdates.pb (Caused by ConnectTimeoutError(<urllib3.connection.HTTPSConnection object at ...>, 'Connection to avl.yctd.org timed out. (connect timeout=30)')) (https://avl.yctd.org/RealTime/GTFS_TripUpdates.pb)"`.

I don't want to just use the exception type by itself because the same exception could be caused by different underlying exceptions. Also, this avoids interacting with the underlying exception(s) except for the purpose of Sentry fingerprinting.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Running locally, will deploy to test.

### DIFF
--- a/services/gtfs-rt-archiver-v3/gtfs_rt_archiver_v3/consumer.py
+++ b/services/gtfs-rt-archiver-v3/gtfs_rt_archiver_v3/consumer.py
@@ -4,6 +4,7 @@ Huey's startup hooks are per _worker_ and not the overall consumer process.
 """
 import logging
 import os
+import re
 import sys
 
 import sentry_sdk
@@ -15,6 +16,8 @@ from prometheus_client import start_http_server
 
 from .tasks import RTFetchException, huey
 
+compiled_regex = re.compile("0x[0-9a-fA-F]+")
+
 
 def set_exception_fingerprint(event, hint):
     if "exc_info" not in hint:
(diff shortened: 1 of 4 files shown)


services/gtfs-rt-archiver-v3/gtfs_rt_archiver_v3/consumer.py
```diff
@@ -15,6 +16,8 @@
 
 from .tasks import RTFetchException, huey
 
+compiled_regex = re.compile("0x[0-9a-fA-F]+")
```
SorenSpicknall, line 19: I don't know rigorously how memory addresses are allocated to HTTPSConnection objects, but in practice it seems that the object reference always ends up `0x7f...` in our environment. I considered that we could tighten this filter to reflect that, but not knowing whether there's a hard and fast rule preventing spillover into 0x8... space or 0x7e... space, I don't actually advise doing so unless you have supporting information.

SorenSpicknall (approved):
One small note on the regex, but probably not actionable.

PR #2338, opened 2023-02-28
dataset URL table for reports site

edasmalchi:
# Description

A new table to be queried during the reports generation process to get the correct dataset names and URLs for each organization and month

Supports https://github.com/cal-itp/reports/issues/231 

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

Looks OK in my staging tables


edasmalchi:
pending documentation, also does anybody know the pros/cons of trying to parse the URLs in this table vs. post-query?

lauriemerrell (commented):
This approach looks good to me! Just two comments:

1. If you want to provide the human-readable string URL, you can use [this macro](https://github.com/cal-itp/data-infra/blob/main/warehouse/macros/url_safe_base64.sql) -- could just add on line 20: `{{ from_url_safe_base64('map.base64_url') }} AS string_url`. I would have a preference for calculating that here rather than decoding in reports site code just for validation/check-ability. 

2. I think having this as a separate table is probably fine but another way to do this would be to just add this (the list of URLs) as a column in the [existing index table](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/gtfs_quality/idx_monthly_reports_site.sql). You could take `dataset_map` (might require refactoring the CTEs a little), create a `month_start` column, and then use the following to get an array of distinct URLs for that month, which could then be joined onto the index. Only reason this would be nice is to keep that table as kind of a one stop shop for that top/metadata section. But I don't think it's a dealbreaker. 
```
SELECT
   month_start,
   organization_itp_id,
   ARRAY_AGG(DISTINCT {{ from_url_safe_base64('map.base64_url') }}) 
FROM dataset_map
GROUP BY month_start, organization_itp_id
```

edasmalchi:
Revised to use arrays and structs included in [existing table](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/gtfs_quality/idx_monthly_reports_site.sql) per discussion with @lauriemerrell

I think this is ready for review!

warehouse/models/mart/gtfs_quality/idx_monthly_reports_site.sql
```diff
@@ -104,7 +104,28 @@ summarize_feed_info AS (
     GROUP BY 1, 2
 ),
 
-idx_monthly_reports_site AS (
+dataset_map AS (
```
lauriemerrell, line 107: This is mostly a dbt conventions thing but I would actually:
* Move this up towards the top of the file because it uses a `ref` 
* Because we actually already reference this table once above, I would move this to the very top and then refactor `assessed_entities` to select from this with its end of month filter applied

(So, have just one CTE that directly references `ref('int_gtfs_quality__organization_dataset_map')` and then have the second CTE filter that first CTE.)

Rationale is to make "import" refs visible at the top of the file and show clearly which elements are using the same data as opposed to referencing twice. I think sometimes there can be actual query performance impacts too but in this case this is a pretty small table so I think that's less relevant.

lauriemerrell (changes requested):
Basically looks good to me, just one code style / conventions thing!

warehouse/models/mart/gtfs_quality/idx_monthly_reports_site.sql
edasmalchi: makes sense, made the change

lauriemerrell (approved):
🎉

PR #2339, opened 2023-02-28
add basic service level check for whether a feed is present

### LINKED ISSUE #2311: Create service-level guidelines check for has RT/schedule
Per discussion on #2304, we need to make sure that organizations and services without GTFS data have some discrete check that they fail that makes it obvious that they don't have GTFS data. 

We should add four new service-level checks for:

- Has schedule
- Has VP
- Has TU
- Has SA

owades:
# Description

Create service-level check that says whether each feed is listed.

Resolves #2311 

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
ran dbt build on all four, and everything down stream.
Looked at aggregate results for 2 days ago:
33% of services pass the schedule check
17% of services pass the trip updates check
17% of services pass the service alerts check
17% of services pass the vehicle positions check
(the above 3 all round to 17&, but are not equactly equal)

### DIFF
--- a/warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__feed_listed_sa.sql
+++ b/warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__feed_listed_sa.sql
@@ -19,7 +19,7 @@ joined AS (
     LEFT JOIN dim_provider_gtfs_data AS quartet
         ON idx.service_key = quartet.service_key
        AND TIMESTAMP(idx.date) BETWEEN quartet._valid_from AND quartet._valid_to
-       AND quartet.customer_facing
+       AND quartet.gtfs_service_data_customer_facing
     GROUP BY 1,2
 ),
 
--- a/warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__feed_listed_sa.sql
+++ b/warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__feed_listed_sa.sql
@@ -1,0 +1,38 @@
+WITH
+
+idx AS (
+    SELECT * FROM {{ ref('int_gtfs_quality__services_guideline_index') }}
+),
+
+dim_provider_gtfs_data AS (
+    SELECT * FROM {{ ref('dim_provider_gtfs_data') }}
+),
+
+joined AS (
+    SELECT
+        idx.date,
+        idx.service_key,
+        -- Some services are mapped to multiple service_alerts_gtfs_dataset_key values at the same time.
+        -- This is one way to address that, open to better solutions!
+        ANY_VALUE(service_alerts_gtfs_dataset_key) AS gtfs_dataset_key
+    FROM idx
+    LEFT JOIN dim_provider_gtfs_data AS quartet
+        ON idx.service_key = quartet.service_key
+       AND TIMESTAMP(idx.date) BETWEEN quartet._valid_from AND quartet._valid_to
+       AND quartet.guidelines_assessed
+    GROUP BY 1,2
+),
+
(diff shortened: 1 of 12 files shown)


warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__trip_planner_rt.sql
```diff
@@ -14,15 +24,24 @@ int_gtfs_quality__trip_planner_rt AS (
...
         {{ trip_planner_rt() }} AS check,
         {{ compliance_rt() }} AS feature,
-        CASE manual_check__gtfs_realtime_data_ingested_in_trip_planner
-            WHEN 'Yes' THEN {{ guidelines_pass_status() }}
-            WHEN 'No' THEN {{ guidelines_fail_status() }}
-            WHEN 'N/A - no fixed-route service' THEN {{ guidelines_na_check_status() }}
+        CASE
+            -- If there are no RT feeds listed, the service should not fail this check, instead it should be "N/A"
+            -- Note that we're only looking at VP and TU, since that's what it being checked for in this manual check.
+            WHEN tu.status = 'FAIL' AND vp.status = 'FAIL' THEN {{ guidelines_na_check_status() }}
```
lauriemerrell, line 30: I think that this will be improved by the indexing changes I am making but it is fine for now (I may end up refactoring a bit but don't think there's a better thing to be done in the meantime short of those larger changes, if that makes sense)

warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__feed_listed_sa.sql
```diff
@@ -0,0 +1,38 @@
...
+        idx.date,
+        idx.service_key,
+        -- Some services are mapped to multiple service_alerts_gtfs_dataset_key values at the same time.
+        -- This is one way to address that, open to better solutions!
+        ANY_VALUE(service_alerts_gtfs_dataset_key) AS gtfs_dataset_key
+    FROM idx
+    LEFT JOIN dim_provider_gtfs_data AS quartet
+        ON idx.service_key = quartet.service_key
+       AND TIMESTAMP(idx.date) BETWEEN quartet._valid_from AND quartet._valid_to
+       AND quartet.guidelines_assessed
```
lauriemerrell, line 22: So, this makes me a bit nervous, because of how the guidelines automation & the reports site are using different definitions of "assessed". This definition means we are making a service-level assessment based on guidelines-assessed data which may not be selected in the reports site. In practice I think this will largely be fine because of how the MTC 511 data is structured, but can we add a `TODO` comment here indicating that this might be fragile when the reports site eventually starts surfacing this data? 

I can also see if in the reindexing work I can think of a different way to handle 🤔

lauriemerrell (approved):
Only actual request is to add a comment, otherwise I think there may be more follow up here down the road but nothing to do differently at this point with what we have available now

warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__feed_listed_sa.sql
owades: Since we're using the ANY_VALUE() function, this logic is pretty forgiving. If it makes you more comfortable, we can remove the guidelines_assessed filter and it will pretty much work the same. What do you think?

lauriemerrell: Let's maybe just replace with `where customer_facing` instead? More forgiving but I don't want to give people credit if we have access to private feeds for example

lauriemerrell (approved):

PR #2357, opened 2023-03-06
Fix some bugs in `dim_provider_gtfs_data`

### LINKED ISSUE #2350: Bug: `dim_provider_gtfs_data` is incorrectly separating some schedule and RT datasets
**Describe the bug**
While working on #2340 I noticed that 

```
SELECT *
FROM `cal-itp-data-infra.mart_transit_database.dim_provider_gtfs_data`
WHERE _is_current
AND service_name = "Muni Metro" 
```

Yields 2 rows for the MTC 511 regional feed, as does 

```
SELECT *
FROM `cal-itp-data-infra.mart_transit_database.dim_provider_gtfs_data`
WHERE _is_current
AND service_name = "Duarte Transit" 
```
For Foothill Transit.

In both of these cases the RT feeds are not being correctly grouped together in the same row as the schedule feed.

**To Reproduce**
See queries above.

**Expected behavior**
RT & schedule feeds should appear on the same row in these cases.

lauriemerrell:
# Description
Resolves #2350 by:
* Fixing a bug where we were marking the MTC 511 **schedule** feed as not reports-assessed but not handling the RT feeds. This led to an issue where the schedule and RT feeds were not ending up on the same line in the pivot because they had different values for `reports_site_assessed`.
* Fixing a bug with how `route_id, agency_id, network_id` were included in the pivot. These values are generally not included for RT GTFS dataset records so this also was preventing RT feeds from being correctly grouped with their schedule feeds. Per a conversation with @evansiroky in Slack, I am removing these fields from the mart in general since they are not maintained. I am however keeping them in some intermediate tables in case we need to restore them (i.e., they are included for versioning purposes but not actually exposed.)

Note that this change now combines the regional alerts feed with the other regional feeds for MTC 511, even though the regional alerts feed does not support record selectors. I think that this change is actually probably desirable?

## Type of change

- [ ] New feature
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) -- removes `route_id, network_id, agency_id` columns -- I **think** no one is using them but could be breaking if they are used
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

`poetry run dbt run -s dim_gtfs_service_data+` & `poetry run dbt test -s dim_gtfs_service_data+`

Did a bunch of poking at `dim_provider_gtfs_data` in particular and verified that the only feed that is substantially affected is Elk Grove Transit Service service's relationship with Sacramento data in 2022, which I think was essentially buggy anyway. 

Number of rows is decreasing from `15,100` to `14,299` which I think shows that this was a significant but not catastrophic change, especially since MTC 511 agencies were implicated.


lauriemerrell:
Do not merge until people have had time to respond in Slack to the logic here.

warehouse/models/intermediate/gtfs_quality/int_gtfs_quality__daily_assessment_candidate_entities.sql
```diff
@@ -136,10 +132,11 @@ int_gtfs_quality__daily_assessment_candidate_entities AS (
                 AND backdated_regional_feed_type = 'Regional Subfeed'
                 AND has_guidelines_assessed_schedule_feed) THEN TRUE
             -- finally, confirm we have at least one schedule feed and that the overall entity is assessed
-            -- and we suppress the MTC regional combined feed from being used in reporting
+            -- and we suppress the MTC regional combined feed (schedule and all 3) from being used in reporting
```
atvaccaro, line 135: nit: `all 3 RT`?

atvaccaro (approved):

warehouse/models/intermediate/gtfs_quality/int_gtfs_quality__daily_assessment_candidate_entities.sql
lauriemerrell: haha I literally was already fixing

PR #2360, opened 2023-03-07
Add all exception_ columns to Sentry loader null value processing

SorenSpicknall:
# Description

Based on data used in testing, we initially designated a group of about half a dozen columns to filter null values from within arrays for BigQuery safety. When implemented in production with a start date of 1/01/2023 (rather than the shorter time period used for testing), our task encountered several other columns that met the same conditions. Though Sentry documentation is not 100% clear about the underlying data conditions, it now seems clear that all the array columns prefixed with `exception_` have potential for null values within their arrays. This PR adds all of those columns to our list for processing.

There is an argument to be made for implementing an iterator over the columns that looks for `exception_` in the name of a Sentry Clickhouse column and makes these substitutions, rather than relying on an explicit list. I decided against that approach initially for a couple reasons:

1. We don't actually know for certain whether we can universally treat such columns the same way down the road, e.g. if we upgrade Sentry and get new columns in the events table or if we bring in additional tables from Sentry's Clickhouse instance. Applying the behavior automatically might break on new columns, especially if any new columns come in that meet the naming pattern but _aren't_ arrays.
2. Retaining the explicit column list (for this loop and for the `cols_with_unix_timestamps` loop below it) makes it easy to assess which columns are being modified during the task, in the code, at a glance.
3. The explicit column list minimizes the number of loop iterations vs. going through all the columns and looking for matches on which to act, though this is an ultimately tiny efficiency concern on the scale of the full script.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
The updated script was run for all dates from 01/01/2023 forward, and that data brought through successfully to their final tables via dbt runs against both the testing and the production environments.


atvaccaro (approved):
why not run this on all lists present in the data?

SorenSpicknall:
> why not run this on all lists present in the data? i.e. ignore the name and just handle anything of type `list`

Two initial reasons: the first is the tiny time savings of not iterating over each column to detect its type, which is only a minor concern. The second is that, in the interest of getting this merged in time for the overnight task run while also devoting part of my workday to Elavon data extraction, routing additional columns through existing logic seemed less time-intensive than writing the extra couple of lines to detect and act upon column types and then test that new logic. I'm happy to implement that tomorrow, but would prefer to put this simple list addition in place for today in order to retain my afternoon's focus on other work. Is that acceptable?

atvaccaro:
> > why not run this on all lists present in the data? i.e. ignore the name and just handle anything of type `list`
> 
> Two initial reasons: the first is the tiny time savings of not iterating over each column to detect its type, which is only a minor concern. The second is that, in the interest of getting this merged in time for the overnight task run while also devoting part of my workday to Elavon data extraction, routing additional columns through existing logic seemed less time-intensive than writing the extra couple of lines to detect and act upon column types and then test that new logic. I'm happy to implement that tomorrow, but would prefer to put this simple list addition in place for today in order to retain my afternoon's focus on other work. Is that acceptable?

works for me; do you mind putting a TODO comment in the code?

PR #2361, opened 2023-03-08
do not retry transform_warehouse and do not fail if dbt-metabase fails

atvaccaro:
# Description

Follow-up ticket to do next sprint maybe: https://github.com/cal-itp/data-infra/issues/2362

Removes retry on regular transform job, and stops failing the job if dbt-metabase fails.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Tested locally


lauriemerrell (approved):
I guess this makes me a bit nervous because it means if dbt-Metabase fails it will fail silently right? Can you add the follow up ticket to the sprint doc to make sure it gets done?

PR #2364, opened 2023-03-08
Sentry Pipeline List Cleaning Automation and Docs Updates

SorenSpicknall:
# Description
Two sets of changes contained within:

1. Updates the list cleaning behavior of the Sentry loader script to detect which columns contain list objects, rather than relying on a manual column list as discussed [here](https://github.com/cal-itp/data-infra/pull/2360). Note that we don't just check the datatype of the Pandas column, which interprets lists as an ambiguous `object` type - we check the first element of the column to determine whether it's a list type.
2. Updates the Sentry table docs to fully align with the current schema - there were some leftover outdated docs on the `fct_` table from when we were bringing data through Sentry's API rather than Clickhouse.

## Type of change

- [x] New feature
- [x] Documentation
- [ ] agencies.yml

## How has this been tested?
The script was tested via local runs into a real bucket, using several example days of Sentry data (including a day that had empty results for data).

### DIFF
--- a/jobs/sentry-error-loader/sentry_error_loader.py
+++ b/jobs/sentry-error-loader/sentry_error_loader.py
@@ -60,24 +60,9 @@ def fetch_and_clean_from_clickhouse(project_slug, target_date):
 
     cleaned_df = all_rows.rename(make_name_bq_safe, axis="columns")
 
-    # TODO: convert to smarter detection of which columns have arrays
-    cols_with_nulls_in_arrays = [
-        "exception_frames_abs_path",
-        "exception_frames_colno",
-        "exception_frames_filename",
-        "exception_frames_function",
-        "exception_frames_lineno",
-        "exception_frames_in_app",
-        "exception_frames_package",
-        "exception_frames_module",
-        "exception_frames_stack_level",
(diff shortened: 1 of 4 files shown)


jobs/sentry-error-loader/sentry_error_loader.py
```diff
@@ -60,24 +60,9 @@ def fetch_and_clean_from_clickhouse(project_slug, target_date):
...
-        "exception_frames_stack_level",
-        "exception_stacks_mechanism_type",
-        "exception_stacks_mechanism_handled",
-        "exception_stacks_type",
-        "exception_stacks_value",
-    ]
-
-    for col_name in cols_with_nulls_in_arrays:
+    # Determine which columns contain list types, and process BQ-disallowed nulls in lists
+    array_cols = [x for x in list(cleaned_df) if isinstance(cleaned_df[x][0], list)]
```
atvaccaro, line 64: nit: I have no idea if the clickhouse client ever returns something like a tuple but `hasattr(cleaned_df[x][0], '__iter__')` might be more robust

atvaccaro (approved):

Issue #2328, opened 2023-02-27
dbt docs homepage won't load


lauriemerrell:
The page at: https://dbt-docs.calitp.org/#!/overview will not load, just hangs like this indefinitely:

![image](https://user-images.githubusercontent.com/55149902/221605684-c516f43f-b241-4675-ab90-980c936af1ca.png)

I first noticed the issue on Friday, wanted to wait over weekend to see if it persisted. I think it's everything under `/overview` because this has the issue too: https://dbt-docs.calitp.org/#!/overview/calitp_warehouse.

For that reason I wondered if I caused it with #2293. I checked that with locally generated and served docs both `/overview` and `/overview/calitp_warehouse` load correctly.

PR #2365, opened 2023-03-09
use top-level overview doc instead of for project

atvaccaro:
# Description

Closes https://github.com/cal-itp/data-infra/issues/2328

I'm not sure why it's only erroring on the project-specific overview.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Works locally.


lauriemerrell (approved):
I am ok with this but would prefer to preserve the existing overview content

atvaccaro:
> I am ok with this but would prefer to preserve the existing overview content

Yes we should revisit at some point; it does seem like a dbt docs bug.

PR #2369, opened 2023-03-09
Littlepay payments v2 sync and parse

atvaccaro:
# Description

I'm using one bucket each for raw and parsed data, partitioned by instance. Files are also versioned by extract time, and a new version is only created if the etag or last-modified timestamp changes (avoids having to read the whole file and compare hashes, but assumes one of those two fields will change). All external tables use string types, and the staging models will safe_cast to the appropriate type.

Follow-up PR is https://github.com/cal-itp/data-infra/pull/2398

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Tested locally.

## Screenshots (optional)

### DIFF
--- a/airflow/plugins/operators/littlepay_raw_sync.py
+++ b/airflow/plugins/operators/littlepay_raw_sync.py
@@ -4,7 +4,7 @@ import os
 import traceback
 from concurrent.futures import Future, ThreadPoolExecutor
 from datetime import datetime
-from typing import ClassVar, Dict, List
+from typing import ClassVar, Dict, List, Optional
 
 import boto3
 import pendulum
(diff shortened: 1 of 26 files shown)


airflow/plugins/operators/littlepay_raw_sync.py
```diff
@@ -0,0 +1,217 @@
...
+LITTLEPAY_RAW_BUCKET = os.getenv("CALITP_BUCKET__LITTLEPAY_RAW")
+
+
+class LittlepayFileKey(str):
+    @classmethod
+    def __get_validators__(cls):
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, v):
```
SorenSpicknall, line 35: Could this maybe have a more specific name, like validate_filename, to avoid the implication that something about the file _contents_ is being validated if/when this is called from other places?

airflow/plugins/operators/littlepay_raw_sync.py
```diff
@@ -0,0 +1,217 @@
...
+        except KeyError:
+            print(f"metadata missing on {fileinfo.name}")
+            raise
+        prior = RawLittlepayFileExtract(**json.loads(metadata_str))
+        save = (
+            file.s3object.LastModified != prior.s3object.LastModified
+            or file.s3object.ETag != prior.s3object.ETag
+        )
+    except FileNotFoundError:
+        save = True
```
SorenSpicknall, line 112: What do we end up saving if we reach this, in practice? Should it go to a different place than the results of successful tries?

airflow/plugins/operators/littlepay_to_jsonl.py
```diff
@@ -0,0 +1,132 @@
...
+            "micropayment-adjustments",
+            "micropayment-device-transactions",
+            "micropayments",
+            "product-data",
+            "refunds",
+            "settlements",
+        ]
+
+        with logging_redirect_tqdm():
+            for entity in tqdm(entities):
```
SorenSpicknall, line 106: It seems like there should be a way to run this on a chosen subset of entities if desired, but not necessary for initial implementation.

SorenSpicknall (approved):
Just a few small things.

airflow/plugins/operators/littlepay_raw_sync.py
atvaccaro: This condition is hit whenever we have never synced a source file before, so it's a happy path and probably the most common path.

airflow/plugins/operators/littlepay_to_jsonl.py
atvaccaro: I had thought about that but wasn't sure if 10 separate tasks was worth it right now, as our list of files is more static/known. For context @lauriemerrell created separate tasks per GTFS file in the Schedule pipeline since we only want to process a whitelist of known files from whatever's in zip files (RT does this as well for orchestration purposes). 

I'll add a TODO note.

PR #2374, opened 2023-03-10
Reports vendor info

edasmalchi:
# Description

A new table (or intermediate table to be joined to idx_monthly_reports_site) providing a high-level summary of vendors each organization uses for GTFS-schedule related tasks and GTFS-RT related tasks.

Supports 

- https://github.com/cal-itp/reports/issues/194
- https://github.com/cal-itp/reports/issues/181

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

Successful build of `fct_monthly_reports_site_organization_gtfs_vendors` on Eric's staging.

## Screenshots (optional)
Screenshot 2023-03-10 at 13 50 37


warehouse/models/mart/gtfs_quality/fct_monthly_reports_site_organization_gtfs_vendors.sql
```diff
@@ -0,0 +1,104 @@
...
+    SELECT
+    organization_name, organization_source_record_id, organization_itp_id,
+    date_start, reports_vendor_type,
+    ARRAY_AGG(DISTINCT reports_vendor_name) AS vendor_names
+    FROM to_aggregate
+    WHERE reports_vendor_name IS NOT NULL
+    GROUP BY 1, 2, 3, 4, 5
+),
+
+-- is there a more elegant way to do this?
```
lauriemerrell, line 78: I think you might be able to just `PIVOT`? https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#pivot_operator

warehouse/models/mart/gtfs_quality/fct_monthly_reports_site_organization_gtfs_vendors.sql
```diff
@@ -0,0 +1,104 @@
...
+    WHERE component_name IN
+        ('GTFS generation', 'Scheduling (Fixed-route)', 'Real-time info',
+        'Arrival predictions', 'GTFS-rt vehicles/trips', 'GTFS Alerts Publication')
+),
+
+dim_orgs AS (
+    SELECT
+        source_record_id, name, organization_type
+    FROM {{ ref('dim_organizations') }}
+    WHERE _is_current
```
lauriemerrell, line 22: Just noting that by adding this filter, you *might* risk having a failed join if an organization record for a vendor has been deleted by was active in the past. Don't think it's a huge risk because IIRC the service components tables aren't fully historical yet (which we may want to just double check on..... If there is anyone we know has changed vendors, I think we may need to look into making those fully historical so we aren't assessing a current vendor on a prior vendor's data?)

warehouse/models/mart/gtfs_quality/fct_monthly_reports_site_organization_gtfs_vendors.sql
```diff
@@ -0,0 +1,104 @@
...
+        gtfs_generation_components.product_vendor_organization_source_record_id
+    FROM assessed_entities
+    LEFT JOIN gtfs_generation_components
+        ON assessed_entities.service_key = gtfs_generation_components.service_key
+),
+
+orgs_joined AS (
+    SELECT
+        entities_components_joined.*,
+        dim_orgs.name, dim_orgs.organization_type
```
lauriemerrell, line 40: Stylistically, I would probably alias these to `dim_orgs.name AS vendor_name`, `dim_orgs.organization_type AS vendor_organization_type` just for clarity

warehouse/models/mart/gtfs_quality/fct_monthly_reports_site_organization_gtfs_vendors.sql
```diff
@@ -0,0 +1,104 @@
...
+    FROM entities_components_joined
+    LEFT JOIN dim_orgs
+        ON entities_components_joined.product_vendor_organization_source_record_id = dim_orgs.source_record_id
+),
+
+to_aggregate AS (
+    SELECT
+        *,
+        CASE
```
lauriemerrell, line 49: Totally optional stylistic point: You could also do this `CASE WHEN` on the initial `gtfs_generation_components` CTE above. My reason to do so would just be that I think that this doesn't rely on any of the joined data so that makes it very clear where it comes from and kind of pre-states "this is what I am going to be using from this table". I think if you do that you could also possibly remove this CTE and thus shorten the query just a bit. 

But it makes no functional difference.

lauriemerrell (commented):
Just a few comments, mostly style / kind of hypothetical (the least helpful kinds of comments? 😅 )

One specific thing to do once ready is to add YAML documenting the existence of this table. If you haven't already tested joining with `idx_monthly_reports_site` I would maybe just double check that to make sure your grain is ok. 

I think it's fine to have this in a separate table, I saw you asked in Slack about having to turn this into JSON if it were joined into `idx_monthly_reports_site`. If you want to be able to use this to filter on the site, I suspect you may need to go that route (join and turn it into JSON on that table), but I could be wrong -- would defer to @ryon and @charlie-costanzo about that. (And just to be clear, that is a specific limitation of a specific adapter that we use in the reports site generation that couldn't handle arrays of structs; I still think that was the elegant way to go on that in the warehouse.) 

Even if you do want to join it in, given how much logic there is in this, I think it could make sense to keep this as a separate table (intermediate in that case, as you note in the PR description) and then join it into the index table.

warehouse/models/mart/gtfs_quality/fct_monthly_reports_site_organization_gtfs_vendors.sql
lauriemerrell: Put another way, I think we might eventually need to make a vendor/components bridge table that uses the versioned `key`s that you could use here 🤔

lauriemerrell: So, I think this is probably ok for now, but maybe we need a ticket to make it more robust in future?

warehouse/models/mart/gtfs_quality/fct_monthly_reports_site_organization_gtfs_vendors.sql
edasmalchi: couldn't quite figure out how to make that work, perhaps we can just leave this part as-is for now?

edasmalchi:
@lauriemerrell is there a preferred ordering for entries in `_mart_gtfs_quality.yml`?

lauriemerrell:
> is there a preferred ordering for entries in _mart_gtfs_quality.yml?

It's pretty YOLO, I'd just add it in the end 😅

edasmalchi:
tested join with `idx_monthly_reports_site` and lookin' OK (some nulls where no vendor data but I think that's alright)

```
SELECT * FROM `cal-itp-data-infra-staging.eric_mart_gtfs_quality.idx_monthly_reports_site` as t1
INNER JOIN `cal-itp-data-infra-staging.eric_mart_gtfs_quality.fct_monthly_reports_site_organization_gtfs_vendors` as t2
USING(organization_name, organization_source_record_id,
organization_itp_id, date_start)
WHERE publish_date = '2023-03-01'
LIMIT 1000
```

edasmalchi:
I think this could be ready! Would appreciate a docs build/review since I can't do that using the Hub yet

lauriemerrell (approved):
LGTM 🚀 , I do think we might want a follow-up ticket to figure out the organization versioning here (probably low priority, I don't think that organization records get deleted often, but if a vendor changes names I think this logic would show the updated name.... That could maybe be desirable? But basically it just might be worth exploring further to confirm)

Issue #2347, opened 2023-03-03
Bug(?): Add Pacific Time activity date/time attributes to fct_observed_trips


lauriemerrell:
Currently, `service_date`, `activity_date`, and the start and end times in `fct_daily_scheduled_trips` are generally implicitly in Pacific Time because they come directly from GTFS (strings converted to dates). 

However `dt` in `fct_observed_trips` is in UTC because it comes from our processing time, and `header_timestamp` is also in UTC. `fct_observed_trips.trip_start_date` and `fct_observed_trips.trip_start_time` are strings directly from GTFS that are likely implicitly in Pacific Time but they are not guaranteed to be populated.

The upshot of all this is that there is not a good canonical way to actually join between `fct_daily_scheduled_trips` and `fct_observed_trips` because the date columns may not align. Trips that are after 4pm Pacific will be tagged as the following day in `fct_observed_trips`.

We should add Pacific Time dates & times to `fct_observed_trips` so that we can actually join these two tables.

In both trips tables we should also have a naming convention making it clear what time zone each column is in.

lauriemerrell:
Here's an example of a trip that appears twice on the same `dt` but with different `trip_identifier`s 🤔 

```
SELECT *
FROM `cal-itp-data-infra.mart_gtfs.fct_observed_trips`
WHERE schedule_to_use_for_rt_validation_gtfs_dataset_key = 'e124e85f1ee935165fb29f172ab459c2' AND dt = '2023-02-04' AND trip_id = '101-4'
ORDER BY trip_id
```

PR #2375, opened 2023-03-10
Fix scheduled in TU check bug #2373

### LINKED ISSUE #2373: Bug: issue in new version of int_gtfs_quality__scheduled_trips_in_tu_feed
**Describe the bug**
I started looking at examples where my version of the scheduled_trips_in_tu_feed passed and Laurie's new version failed, and here's the first fruit of that investigation.

I think the counting and aggregation logic is causing some double-counting of observed trips.

When I look at the new version of `int_gtfs_quality__scheduled_trips_in_tu_feed`, for Commerce Municipal Bus Lines for 2023-02-04, I see the check failed because schedule has 112 trips while observed/tripupdates has 154 trips.

Logically this shouldn't be possible, since the query is left joining observed onto schedule - if they aren't equal, observed should be less, not greater.

I think the issue is that the aggregation logic is aggregating some duplicate rows, which double-counts trips. Admittedly I could go deeper into why there are duplicate rows, but I feel like Laurie can understand and/or explain this intuitively without an example. Let me know though, and I can go deeper.

An idea that I have to improve the `compare_trips` subquery is that, instead of the line:
`COUNTIF(observed_trips.trip_id IS NOT NULL) AS observed_trips,`
we could have the line:
`COUNTIF(observed_trips.trip_id IS null) AS missing_observed_trips,`

The check would pass if missing_observed_trips = 0, and fail if missing_observed_trips > 0

Here are the queries I used to confirm that both the schedule and trip updates feed showed 112 trip_id's on the given day:

```
SELECT count(distinct trip_id)  FROM `cal-itp-data-infra.mart_gtfs.fct_daily_scheduled_trips` t
LEFT JOIN `cal-itp-data-infra.mart_transit_database.dim_gtfs_datasets` d
ON t.gtfs_dataset_key = d.key
  AND CAST(t.service_date AS TIMESTAMP) BETWEEN d._valid_from AND d._valid_to
WHERE t.service_date = "2023-02-04"
AND d.base64_url = 'aHR0cHM6Ly9jaXR5Y29tbWJ1cy5jb20vZ3Rmcw=='
```
```
SELECT count(distinct trip_id) FROM `cal-itp-data-infra.mart_gtfs.fct_observed_trips` t
LEFT JOIN `cal-itp-data-infra.mart_transit_database.dim_gtfs_datasets` d
ON t.tu_gtfs_dataset_key = d.key
  AND CAST(t.dt AS TIMESTAMP) BETWEEN d._valid_from AND d._valid_to
WHERE t.dt = "2023-02-04"
    AND t.tu_base64_url IS NOT NULL
AND d.base64_url = 'aHR0cHM6Ly9jaXR5Y29tbWJ1cy5jb20vZ3Rmcy1ydC90cmlwdXBkYXRlcw=='
```

I did a spot-check and believe the 112 trips in each are the same trips.

lauriemerrell:
# Description
Fixes #2373:

* The specific issue described in the ticket was actually caused by trips on the same `dt` with different `trip_start_date`, see https://github.com/cal-itp/data-infra/issues/2347#issuecomment-1464500392. I think we might need more refinement on this as #2347 is addressed, but I think the switch to `COUNT(DISTINCT trip_id))` for observed trips should work. 
* I also think that some issues here were being caused by grouping by both schedule and trip updates dataset key as opposed to just schedule in the `compare_trips` CTE, so I simplified that.
* I also did more spot checks and some of the behavior changes here are caused by the fact that the original version filtered to only `guidelines_assessed` quartets, whereas this preserves basically as much as possible. So in the original version a service would have one determination (based on its guidelines assessed quartet), here you can have multiple outcomes for the same service depending on its different quartets. 

Resolves #2373

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

Run locally & inspected results. Confirmed the specific issue identified with Commerce Municipal Bus Lines is fixed.

### DIFF
--- a/warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__scheduled_trips_in_tu_feed.sql
+++ b/warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__scheduled_trips_in_tu_feed.sql
@@ -25,17 +25,15 @@ compare_trips AS (
         scheduled_trips.service_date AS date,
         scheduled_trips.gtfs_dataset_key AS schedule_gtfs_dataset_key,
         scheduled_trips.feed_key AS schedule_feed_key,
-        observed_trips.tu_gtfs_dataset_key,
-        observed_trips.tu_base64_url,
         COUNT(DISTINCT scheduled_trips.trip_id) AS scheduled_trips,
-        COUNTIF(observed_trips.trip_id IS NOT NULL) AS observed_trips,
+        COUNT(DISTINCT observed_trips.trip_id) AS observed_trips,
     FROM fct_daily_scheduled_trips AS scheduled_trips
     LEFT JOIN fct_observed_trips AS observed_trips
       -- should this be activity date or service date? will depend on fix for https://github.com/cal-itp/data-infra/issues/2347
       ON scheduled_trips.service_date = observed_trips.dt
       AND scheduled_trips.gtfs_dataset_key = observed_trips.schedule_to_use_for_rt_validation_gtfs_dataset_key
       AND scheduled_trips.trip_id = observed_trips.trip_id
-    GROUP BY 1, 2, 3, 4, 5
+    GROUP BY 1, 2, 3
 ),
 
 check_start AS (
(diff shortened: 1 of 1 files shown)


owades:
Thank you!!

owades (approved):

atvaccaro (approved):

PR #2371, opened 2023-03-09
Actually migrate guideline indices / merge to main

lauriemerrell:
# Description
Constituent PRs: 
* #2340 
* #2375
* #2370
* #2379 
* #2402
* #2428

Post-merge follow-up items:
- #2401
- #2439

Resolves #2263 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

In final sub-PR (#2428) tested the end to end flow and confirmed things look reasonable.

PR #2379, opened 2023-03-14
Batch 3: Migrate guideline check indices

lauriemerrell:
# Description

See [Google Doc](https://docs.google.com/spreadsheets/d/15eYntYFoYyHuXYyp4dmqM3g5N18tw2r8dU2S7wDDM4E/edit#gid=1437513597) for list of models migrated here, those with #2379 in the PR column. Once again merging into staging branch, which will merge in #2371.

Notes:
* For RT feed checks on dates before RT files history are available, we might end up with `N/A - NO APPLICABLE ENTITY` instead of `BEFORE CHECK ASSESSED` because they will show has not having RT files (and therefore having no assessed entity). Not sure if this is a problem. 🤔 
* For the vehicle positions & trip updates freshness checks (updates every 20 sec/no updates older than 90 sec), we are creating new incremental fact models so that the check models themselves are not incremental. This should give more flexibility in future and keep the check models consistent across the board.
* For the updates every 20 seconds, we have discussed extensively offline. We have changed the measure from comparing one header timestamp to the subsequent header timestamp (which might be subject to flakiness if our scraper went down) to comparing a given header timestamp to the timestamp at which it was requested, showing the freshness of the data at the time it was received (this is already closer to the metric we were using for the `no stale` checks). Instead of checking that the maximum such value is < 20sec, we are checking that the 90th percentile is < 20sec, to account for the fact that this check seemed fairly noisy and we assume the intention is to capture the capacity to update every 20sec rather than to penalize a small number of lags during the day. 

Progress on #2263

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Tested each model individually to confirm that new statuses seem reasonable vs. old statuses. Also checked 
```
SELECT check, COUNT(*)
FROM `cal-itp-data-infra-staging.laurie_staging.int_gtfs_quality__guideline_checks_long_new_index`
GROUP BY check
```
to confirm that all checks have the same number of results (no fanout).

### DIFF
--- a/warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__no_rt_validation_errors.sql
+++ b/warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__no_rt_validation_errors.sql
@@ -1,57 +1,54 @@
-WITH feed_guideline_index AS (
-    SELECT * FROM {{ ref('int_gtfs_quality__rt_feed_guideline_index') }}
-),
-
-fct_daily_rt_feed_files AS (
-    SELECT * FROM {{ ref('fct_daily_rt_feed_files') }}
+WITH guideline_index AS (
+    SELECT
+        *
+    FROM {{ ref('int_gtfs_quality__guideline_checks_index') }}
+    WHERE check IN ({{ no_rt_validation_errors_vp() }},
+        {{ no_rt_validation_errors_tu() }},
+        {{ no_rt_validation_errors_sa() }})
 ),
 
 fct_daily_rt_feed_validation_notices AS (
     SELECT * FROM {{ ref('fct_daily_rt_feed_validation_notices') }}
 ),
 
-count_files AS (
-    SELECT
-        base64_url,
-        date,
-        SUM(parse_success_file_count) AS rt_files
-    FROM fct_daily_rt_feed_files
-    GROUP BY 1, 2
-),
-
 errors AS (
     SELECT
         date,
         base64_url,
-        SUM(total_notices) AS total_errors,
+        -- All error codes start with the letter E (warnings start with W)
+        COUNTIF(code LIKE "E%" and total_notices > 0) > 0 AS had_errors
     FROM fct_daily_rt_feed_validation_notices
-    -- All error codes start with the letter E (warnings start with W)
-    WHERE code LIKE "E%"
     GROUP BY 1, 2
+),
+
+check_start AS (
+    SELECT MIN(date) AS first_check_date
+    FROM errors
 ),
 
 int_gtfs_quality__no_rt_validation_errors AS (
     SELECT
-        idx.date,
-        idx.base64_url,
-        idx.feed_type,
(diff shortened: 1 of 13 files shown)


owades (approved):
tres elegant!

warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__no_rt_validation_errors.sql
```diff
@@ -1,57 +1,54 @@
...
-    GROUP BY 1, 2
-),
-
 errors AS (
     SELECT
         date,
         base64_url,
-        SUM(total_notices) AS total_errors,
+        -- All error codes start with the letter E (warnings start with W)
+        COUNTIF(code LIKE "E%" and total_notices > 0) > 0 AS had_errors
```
atvaccaro, line 19: nit: case insensitive?

warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__rt_https.sql
```diff
@@ -1,30 +1,48 @@
...
         CASE
-            WHEN string_url LIKE 'https%' THEN {{ guidelines_pass_status() }}
-            WHEN string_url IS NOT null AND string_url NOT LIKE 'https%' THEN {{ guidelines_fail_status() }}
+        -- check that the row has the right entity + check combo, then assign statuses
+            WHEN (idx.has_rt_url_tu AND check = {{ rt_https_trip_updates() }})
+                OR (idx.has_rt_url_vp AND check = {{ rt_https_vehicle_positions() }})
+                OR (idx.has_rt_url_sa AND check = {{ rt_https_service_alerts() }})
+                   THEN
+                    CASE
+                        WHEN string_url LIKE 'https%' THEN {{ guidelines_pass_status() }}
```
atvaccaro, line 34: nit: case sensitive, same below

technically our URLs could be uppercase or smth

atvaccaro (approved):

warehouse/models/mart/gtfs_quality/_mart_gtfs_quality.yml
```diff
@@ -628,6 +628,29 @@ models:
       - name: change_status
       - name: n
 
+  - name: fct_daily_vehicle_positions_message_age_summary
```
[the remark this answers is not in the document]
lauriemerrell: https://github.com/cal-itp/data-infra/pull/2379/commits/fc81a4f91e4163c70dabba31028197899270d0fe

warehouse/models/intermediate/gtfs_quality/guidelines_checks/int_gtfs_quality__no_rt_validation_errors.sql
lauriemerrell: https://github.com/cal-itp/data-infra/pull/2379/commits/b338b5e87870b7d2f5fee229de86b07db5e77970

PR #2383, opened 2023-03-14
Airtable columns requested in #2372

### LINKED ISSUE #2372: User Story: Include new table columns
## User stories

A user story is implemented as well as it is communicated.
If the context and the goals are made clear, it will be easier for everyone to implement it, test it, refer to it.

---

### Summary

A user story should typically have a summary structured this way:

1. **As a** [airtable data analyst]
2. **I want** [`services.Deprecated Date`, `services.Operating County Geographies`, and `organizations.HQ County Geography` to be added]
3. **So that** [will be ingested by the data warehouse.]

### Acceptance Criteria

**Example:**

1. [The columns referenced in this issue should be able to be queried in Metabase.]

### Tester [Stakeholder]

1. @evansiroky 

# Sprint Ready Checklist

charlie-costanzo:
# Description

This PR brings in the columns to the warehouse requested in #2372 
* services.deprecated_date
* services.operating_county_geographies
* organizations.hq_county_geography

Resolves #2372 

## Type of change

- [x] New feature


lauriemerrell (changes requested):
Thank you Charlie for tackling this!

(Responding to some discussion offline)

* The handling of the arrays is perfect up through the intermediate table. 
* In the mart itself, you need to create a bridge table, because these are versioned tables. Here's [an example bridge table](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/transit_database/bridge_organizations_x_ntd_agency_info.sql) between a versioned (orgs) and fake-versioned (NTD agency info) table; that's what you'll probably want to do here.
* Unfortunately it looks like we haven't previously created a mart table for the `county_geography` table that these are referencing, so you will need to bring that through all the way from staging. [Here's the external table](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_county_geography.yml).

lauriemerrell:
Ope, there is a staging table: https://github.com/cal-itp/data-infra/blob/main/warehouse/models/staging/transit_database/stg_transit_database__county_geography.sql

warehouse/models/staging/transit_database/stg_transit_database__county_geography.sql
```diff
@@ -18,8 +18,12 @@ stg_transit_database__county_geography AS (
         rtpa,
         mpo,
         place_geography,
+        unnested_organizations AS organizations,
+        unnested_services AS services,
         dt
     FROM once_daily_county_geography
+    LEFT JOIN UNNEST(once_daily_county_geography.organizations) AS unnested_organizations
```
lauriemerrell, line 25: I actually think we probably want to remove these UNNESTS -- we only `UNNEST` in the staging table if there's only one element per array in this table, otherwise just leave it as an array and handle later (see for example how we don't unnest `mobility_services_managed` in the `organizations` staging table, just leave it as an array) -- I would just add `organizations` and `services` to the column selection list, and then add them as array columns in the intermediate table when you make the dimension

warehouse/models/intermediate/transit_database/dimensions/int_transit_database__county_geography_dim.sql
```diff
@@ -0,0 +1,41 @@
+{{ config(materialized='table') }}
+
+WITH latest_county_geography AS (
+    {{ get_latest_dense_rank(
+        external_table = ref('stg_transit_database__county_geography'),
+        order_by = 'dt DESC'
```
lauriemerrell, line 6: Would need to add array columns for orgs and services here

warehouse/models/mart/transit_database/dim_organizations.sql
```diff
@@ -44,6 +55,8 @@ dim_organizations AS (
         ON ntd_to_org.ntd_id = ntd.ntd_id
         AND ntd._is_current
         AND ntd.year = 2021
+    LEFT JOIN bridge_organizations_x_county_geography AS geography_bridge
```
lauriemerrell, line 58: (reacting to offline convo) sorry I bet the NTD example was confusing here. We don't join the bridge table back onto the dimension because yes in that case it would cause fanout, that's why it lives in the separate bridge table. So we should just remove these left joins of the bridge tables onto the org and service dimensions

warehouse/models/mart/transit_database/bridge_organizations_x_county_geography.sql
```diff
@@ -0,0 +1,29 @@
...
+),
+
+geography AS ( -- noqa
+    SELECT *
+    FROM {{ ref('int_transit_database__county_geography_dim') }}
+),
+
+bridge_organizations_x_county_geography AS (
+    SELECT
+        orgs.source_record_id AS source_record_id,
```
lauriemerrell, line 15: I think we'd want `orgs.key AS organization_key` kind of following this: https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/transit_database/bridge_organizations_x_ntd_agency_info.sql.

warehouse/models/mart/transit_database/bridge_services_x_county_geography.sql
```diff
@@ -0,0 +1,29 @@
...
+),
+
+geography AS ( -- noqa
+    SELECT *
+    FROM {{ ref('int_transit_database__county_geography_dim') }}
+),
+
+bridge_services_x_county_geography AS (
+    SELECT
+        services.source_record_id AS source_record_id,
```
lauriemerrell, line 15: Same comment as orgs, we'd want `services.key AS service_key` -- following: https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/transit_database/bridge_organizations_x_ntd_agency_info.sql

warehouse/models/mart/transit_database/dim_services.sql
```diff
@@ -5,24 +5,36 @@ WITH dim AS (
     FROM {{ ref('int_transit_database__services_dim') }}
 ),
 
+bridge_services_x_county_geography AS (
```
lauriemerrell, line 8: Same comments as orgs, can remove this import and left join

lauriemerrell (changes requested):

warehouse/models/intermediate/transit_database/dimensions/int_transit_database__county_geography_dim.sql
lauriemerrell: Chatted offline, this was incorrect

warehouse/models/mart/transit_database/dim_county_geography.sql
```diff
@@ -0,0 +1,31 @@
...
+    SELECT
+
+        key,
+        source_record_id,
+        name,
+        fips,
+        msa,
+        caltrans_district,
+        caltrans_district_name,
+        -- rtpa is not a one-to-one relationship
```
lauriemerrell, line 18: If this is another array of organization keys, I would remove it from the final dim, and same for MPO... Generally I think we don't want unhandled arrays of keys in these tables

warehouse/models/staging/transit_database/stg_transit_database__county_geography.sql
lauriemerrell: Looked at this again, I still don't think we want to unnest this here -- on the county geography side this is a 1:many relationship so this makes many rows with the same source record ID

warehouse/models/mart/transit_database/_mart_transit_database.yml
```diff
@@ -508,6 +510,24 @@ models:
       - name: total_voms
       - name: volunteer_drivers
       - name: personal_vehicles
+  - name: dim_county_geography
+    description: '{{ doc("county_geography_table") }}'
+    columns:
+      - *key
```
lauriemerrell, line 516: Curious if tests on this pass re: comment about unnesting organizations in the staging table -- I think there would be dups?

warehouse/models/mart/transit_database/_mart_transit_database.yml
```diff
@@ -815,3 +835,5 @@ models:
           - relationships:
              to: ref('dim_modes')
              field: key
+  - name: bridge_organizations_x_county_geography
```
lauriemerrell, line 838: Would want to add descriptions kind of following this: https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/transit_database/_mart_transit_database.yml#L773-L779 (this is the reason I was focused on NTD agency info / org relationship as an example, because I knew it was one of these partially versioned ones.)

lauriemerrell (changes requested):
Just a few more tweaks:

* I think we want to not unnest in staging -- see for example:
```
SELECT dt, id, organizations 
FROM `cal-itp-data-infra-staging.charlie_staging.stg_transit_database__county_geography`
WHERE id = 'recuhtI5m5Y8J8zY1' AND dt = '2023-03-15'
```
See how there are multiple rows with the same ID, because the one to many relationship has been unnested.

* Need a bit more YAML.

lauriemerrell (approved):

PR #2384, opened 2023-03-14
Add table for reporting schedule compliance checks on reports site

owades:
# Description
The idea is this will allow us to begin reporting on the Schedule Compliance checks within the reports site framework.

Resolves #[234](https://github.com/cal-itp/reports/issues/234)

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml


owades:
@edasmalchi , let's discuss this tomorrow. We can take a look at `cal-itp-data-infra-staging.owades_mart_gtfs_quality.fct_monthly_reports_site_organization_guideline_checks` .

The main question in my mind right now is whether this would be enough, or if we want to add check results for each service within the organization.

warehouse/models/mart/gtfs_quality/fct_monthly_reports_site_organization_guideline_checks.sql
```diff
@@ -0,0 +1,61 @@
...
+        checks.feature,
+        checks.check,
+        checks.status
+    FROM idx_monthly_reports_site AS idx
+    LEFT JOIN generate_biweekly_dates AS dates
+        USING (publish_date)
+    LEFT JOIN checks
+        ON idx.organization_key = checks.organization_key
+        AND dates.sample_dates = checks.date
+    GROUP BY 1, 2, 3, 4, 5, 6, 7, 8, 9
```
lauriemerrell, line 58: I am a bit confused by the group-by here -- I think that if done carefully this join can be done without causing any fanout, so a group by / select distinct should not be necessary. Were you seeing extra rows? 👀

lauriemerrell (changes requested):

warehouse/models/mart/gtfs_quality/_mart_gtfs_quality.yml
lauriemerrell, line 1: Also meant to note -- when adding a new table that will be consumed by the reports site, we need to add it in the `exposures` section at the bottom of the YAML to show that it is used by the site

warehouse/models/mart/gtfs_quality/fct_monthly_reports_site_organization_guideline_checks.sql
owades: thank you! i was acting a bit too transactionally, copying over logic from another table. i checked, there are no extra/duplicate rows. removed the group by.

warehouse/models/mart/gtfs_quality/fct_monthly_reports_site_organization_guideline_checks.sql
```diff
@@ -0,0 +1,54 @@
...
+    FROM idx_monthly_reports_site
+    LEFT JOIN
+        UNNEST(GENERATE_DATE_ARRAY(
+            -- add a few days because otherwise it will always pick the first of the month,
+            -- which may be disproportionately likely to have new feed published and thus issues
+            DATE_ADD(CAST(date_start AS DATE), INTERVAL 3 DAY),
+            CAST(date_end AS DATE), INTERVAL 2 WEEK)) AS sample_dates
+),
+
+checks AS (
```
lauriemerrell, line 21: For stylistic consistency I would move this above the prior CTE because it has a `ref`

lauriemerrell (approved):
Just one style note, would appreciate if you reorder, but functionally I think this looks good to me!

PR #2390, opened 2023-03-15
split dbt build and visualize CI jobs

atvaccaro:
# Description

I only realized on a new PR that the visualize code would not run on warehouse model changes; it needs to be split from the image/script checks.

Also I changed the project profiles to use 8 threads instead of 4, which I'd been meaning to do for awhile.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Both should run on this PR; the DAG visualization does not gracefully handle a lack of changed models but in the future it'll only really run when models _have_ changed.

### DIFF
--- a/.github/workflows/build-warehouse-image.yml
+++ b/.github/workflows/build-warehouse-image.yml
@@ -54,7 +54,7 @@ jobs:
           poetry install
           poetry run dbt deps
           poetry run dbt compile --target prod
-          poetry run dbt docs generate
+          poetry run dbt docs generate --target prod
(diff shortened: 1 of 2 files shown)


atvaccaro:
Unsure why these are still failing, I ran a full refresh.

atvaccaro:
> Unsure why these are still failing, I ran a full refresh.

I'd missed the docs generate

atvaccaro:
Actually I'm going to use https://github.com/tj-actions/changed-files, it'll be simpler

PR #2393, opened 2023-03-15
only run parts of dbt action depending on files

atvaccaro:
# Description

Resolves # [issue]

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

### DIFF
--- a/.github/workflows/build-dbt.yml
+++ b/.github/workflows/build-dbt.yml
@@ -1,21 +1,16 @@
-name: Build and push dbt image
+name: build dbt
 
 on:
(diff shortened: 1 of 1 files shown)


atvaccaro:
> We'll have to keep an eye on third party tj-actions to ensure that the stuff we're relying on them for is kept up to date, but the stuff we're using here looks actively maintained and the changes make sense.

most actions are pinned to specific versions so we should be fine as long as we don't hit breaking GitHub action changes which aren't _super_ common

PR #2404, opened 2023-03-21
fix: add a few corrections to NTD seed data

evansiroky:
# Description

I went through the 2021 NTD data, the data we have in airtable, the existing seed table and came up with the following modifications to improve matching between Airtable organizations and NTD data.

I am a little unsure about my proposed changes to Los Angeles County. For some reason, they report all those little shuttles separately despite them all being Los Angeles County. Please advise on whether linking multiple NTD records to a single organization seems acceptable.

## Type of change

- [ ] New feature
- [ ] Documentation
- [x] NTD data

## How has this been tested?

N/A

### DIFF
--- a/warehouse/seeds/ntd_agency_to_organization.csv
+++ b/warehouse/seeds/ntd_agency_to_organization.csv
@@ -15,7 +15,7 @@ ntd_id,legacy_ntd_id,organization_record_id,organization_name
 90260,,recwTLLx2Qq4yzwWH,City of Compton
 99370,9T22,recYBRDOFAYGCfYmO,Tule River Tribe
 90062,9062,receZJ9sEnP9vy3g0,Monterey-Salinas Transit
-9R02-91082,9R02-094,,
+9R02-91082,9R02-094,recGqTUeT6ZZZZLEh,Mariposa County
 90291,,recSOvT9EjtcLUUeE,City of South Gate
 9R02-91110,9R02-024,recJgBQp06uDrbSiq,City of McFarland
 9R02-99438,,recx4ZG2lvZb7kGAL,Shasta County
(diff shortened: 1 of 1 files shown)


atvaccaro:
> Please advise on whether linking multiple NTD records to a single organization seems acceptable.

I think this will cause fanout; we've been assuming that a single organization can key to a single NTD record.
```sql
    LEFT JOIN dim_annual_database_agency_information AS ntd
        ON ntd_to_org.ntd_id = ntd.ntd_id
        AND ntd._is_current
        AND ntd.year = 2021
```

We have a TODO to use a bridge table for historical joins (instead of just 2021); that would also help with 1:many situations.

warehouse/seeds/ntd_agency_to_organization.csv
```diff
@@ -229,3 +229,25 @@ A0013,,,
...
+90272,,rec860cgWfE3MaiHw,Los Angeles County
+90276,,rec860cgWfE3MaiHw,Los Angeles County
+90277,,rec860cgWfE3MaiHw,Los Angeles County
+90269,,rec860cgWfE3MaiHw,Los Angeles County
+90273,,rec860cgWfE3MaiHw,Los Angeles County
+90274,,rec860cgWfE3MaiHw,Los Angeles County
+90275,,rec860cgWfE3MaiHw,Los Angeles County
+90278,,rec860cgWfE3MaiHw,Los Angeles County
+90279,,rec860cgWfE3MaiHw,Los Angeles County
+9R02-99442,,recFqVabH8109u70q,Calaveras Transit Agency
```
lauriemerrell, line 243: If more mappings have been added manually, we may need to update here: https://github.com/cal-itp/data-infra/blob/main/warehouse/seeds/_seeds.yml#L86 to describe what versions of NTD data are being used to make these mappings.

lauriemerrell (changes requested):
Thanks for looking at this Evan!

Just a few comments below. Agree with Andrew that making this mapping many to one would be disruptive to the existing data model, options would be:

1. Create new organization records for each of the shuttles and add LA County as a parent organization to those organizations? Part of me feels like separate NTD reporting is a pretty good reason to have separate organization records.
2. Make a separate ticket to support many to one and make a list of the affected IDs, update the seed at that time after adding downstream support.

warehouse/seeds/ntd_agency_to_organization.csv
```diff
@@ -229,3 +229,25 @@ A0013,,,
 9R02-91032,9R02-143,recbW86Xrtuw8PhiU,City of Auburn
 90026,9026,recZALk4vysuoTVjF,San Diego Metropolitan Transit System
 90278,,,
+90270,,rec860cgWfE3MaiHw,Los Angeles County
```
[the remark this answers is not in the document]
lauriemerrell: (Pending conversation above about the many to one)

warehouse/seeds/ntd_agency_to_organization.csv
evansiroky: I don't know how the initial seeds were created, so I don't know what to add here.

evansiroky: I added a note in [e52aa76](https://github.com/cal-itp/data-infra/pull/2404/commits/e52aa76163db893d4d374fdced1715beafc45ec5)

lauriemerrell (approved):

PR #2418, opened 2023-03-23
Jupyter image tweaks

atvaccaro:
# Description

Removing shared_utils install, and adding portfolio requirements based on feedback from @tiffanychu90.

Also took the opportunity to switch back to poetry-based installs, and using npm for vega instead of conda (much faster).

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Images will build in CI.


tiffanychu90:
@atvaccaro: Did you mean to remove the `requirements.txt`? Looking for [these](https://github.com/cal-itp/data-analyses/blob/remove-shared-utils-pkgs/_shared_utils/requirements.txt) also to be in the poetry file

tiffanychu90 (changes requested):
@atvaccaro: Did you mean to remove the requirements.txt? Looking for [these](https://github.com/cal-itp/data-analyses/blob/remove-shared-utils-pkgs/_shared_utils/requirements.txt) also to be in the poetry file

atvaccaro:
> @atvaccaro: Did you mean to remove the requirements.txt? Looking for [these](https://github.com/cal-itp/data-analyses/blob/remove-shared-utils-pkgs/_shared_utils/requirements.txt) also to be in the poetry file

Ah yes I meant to remove the file but good catch; will add the shared_utils requirements now that we're not installing the library itself anymore.

tiffanychu90 (approved):

PR #2419, opened 2023-03-24
Upgrade dbt_utils to 1.0

atvaccaro:
# Description

Get us on [dbt_utils 1.0](https://docs.getdbt.com/guides/migration/versions/upgrading-to-dbt-utils-v1.0) to remove some constant warnings from the output. It also means the developers have committed to a more stable and backwards-compatible API going forward.

The major change affecting us is the [surrogate key change](https://docs.getdbt.com/guides/migration/versions/upgrading-to-dbt-utils-v1.0#changes-to-surrogate_key), but we're opting into the legacy behavior (which is recommended) by setting `surrogate_key_treat_nulls_as_empty_strings: true` as a var in the project.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Testing some models; I've confirmed our compiled models coalesce with empty strings for surrogate key generation.

### DIFF
--- a/warehouse/dbt_project.yml
+++ b/warehouse/dbt_project.yml
@@ -22,6 +22,7 @@ clean-targets:         # directories to be removed by `dbt clean`
   - "dbt_packages"
 
 vars:
+  surrogate_key_treat_nulls_as_empty_strings: true # enable legacy behavior for dbt_utils surrogate keys, i.e. coalesce nulls to empty strings
   GTFS_SCHEDULE_START: '2021-04-16'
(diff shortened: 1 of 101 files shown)


lauriemerrell (approved):
I can rebase my own stuff in #2371 and #2417 once this merges

atvaccaro:
Oh I forgot to paste in, but I confirmed the keys match up with existing prod keys

```sql
SELECT count(0)
, count(prod.key)
, count(me.key)
FROM `cal-itp-data-infra`.mart_gtfs.dim_schedule_feeds prod
    full outer join `cal-itp-data-infra-staging`.andrew_mart_gtfs.dim_schedule_feeds me
using (key)
;
```

Issue #2382, opened 2023-03-14
User Story: monthly aggregation for mart_gtfs.fct_daily_scheduled_shapes


tiffanychu90:
## User stories

As an analyst, if I start with a monthly aggregated `service_hours` for a route, I want to be able to represent that on a map. I want to connect that monthly route statistic to a certain `shape_id` for the line geometry. To choose, I want to see which `shape_id` for that route was valid for the most days or for the most trips and select that for the visualization. Or, if I want to spatially join the route to a much larger geography (district, county, etc), I don't want to spatially join 30 `shape_ids` for the route, I want one selected to do the spatial join.

### Summary

Table Schema
* Grain: `source_record_id-name-shape_id-route_id-month-year`
* Timeframe for table: Dec 2022
* Identifier columns: `source_record_id, name, shape_id, route_id, route_short_name`
* Date columns: `month`, `year`
* Metric columns: `n_trips`, `n_days`

### Tester [Stakeholder]

1. @tiffanychu90 

### Notes
* Came out of exploratory work related to this ad hoc table: #2234
* Exploratory work showed challenges using `route-shape` aggregation for service hours. Instead, aggregate service hours to route (not route-shape) and figure out the shape piece separately. https://github.com/cal-itp/data-analyses/issues/673

PR #2426, opened 2023-03-27
Analyst POC tables (from issues #2381 and #2382)

### LINKED ISSUE #2381: User Story: add summary stats for to mart_gtfs.fct_observed_trips
### Summary
As an analyst, I want to have a schedule vs RT trip table that provide additional summary statistics for the RT trips observed.  

* This is follow-up for related work to #2236, but can be built off of another table of the right grain

### Table Schema
* build out the existing `mart_gtfs.gtfs.fct_observed_trips` table
* Grain: `gtfs_dataset_key-name-trip_id-activity_date`
* Metric columns: 
   * `tu_num_distinct_message_ids` (existing)
   * `tu_min_extract_ts` and `tu_max_extract_ts`. 
   Analyst could easily calculate this difference (`rt_trip_elapsed = max_extract_ts-min_extract_ts`). 
   Use that calculation to get `rt_trip_elapsed / num_distinct_message_ids` = number of distinct messages per minute. If we're hitting close to 3, that's close to capturing every 20 sec.
   * `number_of_distinct_minutes_with_vp` as a measure of RT availability for that trip. Analyst would compare against scheduled trip's `service_hours` --> converted to `service_minutes` and see how much coverage had RT. This could easily surpass a proportion of 1.0, esp when a trip takes longer than the scheduled time and there are vehicle positions keep on coming in. 
   * `time_of_day`: this could be where the heuristic is applied. `trip_id` appears in schedule and RT, use schedule to determine `time_of_day`, otherwise use `tu_min_extract_ts`
   
### Tester [Stakeholder]
* @tiffanychu90 

### Notes
WIP @amandaha8 providing exploratory context for @edasmalchi's idea: https://github.com/cal-itp/data-analyses/issues/668

charlie-costanzo:
# Description
Analyst POC tables as requested in #2381 and #2382
* `fct_observed_trips_summaries`
* `fct_daily_scheduled_shapes_monthly_aggregations`

Resolves #2381 and #2382

## Type of change

- [x] New feature

## How has this been tested?
local `dbt run` and `dbt test`


warehouse/models/mart/ad_hoc/fct_observed_trips_summaries.sql
charlie-costanzo, line 1: To-do:
* These should be added columns to `fct_observed_trips`
* Select only one week of data
* Make a fake incremental - this is large data and we don't want it run every day

charlie-costanzo: Questions:
* What do you want the number of distinct minutes to be
* Put it on fct_observed_trips - derive daypart - if there is a trip_start_time use that, if not, then use min_extract_ts
  * The, when we join schedule and RT in a more combined model, then we can have some logic to reconcile if the two have a different time of day for the same trip

warehouse/models/mart/ad_hoc/fct_daily_scheduled_shapes_monthly_aggregations.sql
charlie-costanzo, line 1: Actual question:
* For every month, for every route, what is the most common shape
* "`fct_monthly_common_shape_by_route`"

lauriemerrell (changes requested):
Overall looks good, I think `INTERVAL` is actually super clutch and we need to figure out how to use it more broadly for this, but I just have some suggested tweaks -- happy to discuss on a short call on Monday if you want

warehouse/models/mart/ad_hoc/fct_observed_trips_summaries.sql
```diff
@@ -0,0 +1,82 @@
...
+        dt,
+        tu_num_distinct_message_ids,
+
+        -- are these time zones correct??
+        tu_min_extract_ts,
+        tu_max_extract_ts,
+        trip_start_time,
+        vp_num_distinct_minutes_of_trip_with_vehicle_positions,
+
+        -- Cast to interval because it's a string, with <24hr values, so cast to time breaks
```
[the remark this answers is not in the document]
charlie-costanzo: yeah my bad this is a typo

mjumbewu:
The issues that spurred this PR have since been closed with other solutions. This PR doesn't seem necessary anymore, so closing (/cc @tiffanychu90).

PR #2428, opened 2023-03-27
Batch 6: Migrate guidelines checks to new index

lauriemerrell:
# Description

_will merge to main via #2371_

Migrates final checks, see [Google Doc](https://docs.google.com/spreadsheets/d/15eYntYFoYyHuXYyp4dmqM3g5N18tw2r8dU2S7wDDM4E/edit#gid=1437513597) for details. This PR also does the remaining cleanup from #2263, i.e., adds tests and does some assorted cleanup (ex. unused macros).

* I think that `all_tu_in_vp` may need some additional work post migration to become more robust, see the Google Doc for notes
* Start date for grading scheme is a bit weird because there is nothing that's not unknown or N/A (final row where it is no entity vs. manual check needed is historical/deprecated datasets that are treated as irrelevant for grading purposes):
image

Resolves # [issue]

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

Tested that the individual check statuses look reasonable against their prior versions. Tested that the guidelines checks long table passes its new tests (i.e., no fanout). Tested the mart organization table and verified it looks reasonable.


warehouse/tests/validate_guidelines_checks_rows_per_check_matches_index.sql
```diff
@@ -0,0 +1,20 @@
...
+)
+
+SELECT
+    check,
+    actual_checks,
+    idx_checks
+FROM check_cts
+LEFT JOIN idx_check_cts
+USING (check)
+WHERE actual_checks != idx_checks
```
atvaccaro, line 20: should this be a full outer?

warehouse/models/intermediate/gtfs_quality/int_gtfs_quality__scheduled_trip_version_summary.sql
```diff
@@ -0,0 +1,80 @@
...
+    WHERE _feed_valid_from >= '{{ max_ts }}'
+    {% else %}
+    WHERE _feed_valid_from >= CAST('{{ var('GTFS_SCHEDULE_START') }}' AS TIMESTAMP)
+    {% endif %}
+),
+
+dim_stops AS (
+    SELECT
+        *,
+        CONCAT(stop_lat, stop_lon) AS stop_location
```
atvaccaro, line 31: nit: this should probably have a delimiter

warehouse/models/intermediate/gtfs_quality/int_gtfs_quality__scheduled_trip_version_summary.sql
```diff
@@ -0,0 +1,80 @@
...
+        t1.service_id,
+        -- Creates a hash for a single field summarizing all stop_times this trip
+        MD5(
+            STRING_AGG(
+                t2.time_pair ORDER BY t2.stop_sequence ASC
+            )
+        ) AS trip_stop_times_hash,
+        -- Creates a hash for a single field summarizing all stop locations for this trip
+        MD5(
+            STRING_AGG(
```
atvaccaro, line 65: nit: these should probably have a delimiter

atvaccaro (approved):
approving and merging so we can get this deployed

PR #2435, opened 2023-03-28
update syvt url, per vendor's instructions

owades:
their vendor, Transnovation, says this is the new URL (as per Ali Attari).

As a reminder, GRaaS still uses agencies.yml as a source of truth, which is why I'm updating it here. 

I believe @evansiroky is making the corresponding update in Airtable, and will look for his confirmation.


evansiroky:
This has been updated in Airtable.

atvaccaro (approved):

Issue #2438, opened 2023-03-29
Set up dbt in JupyterHub


lauriemerrell:
We have some potential dbt developers (specifically analysts) who cannot install dbt locally and need to be able to contribute to the dbt project via JupyterHub.

Acceptance criteria:
- [ ] dbt and necessary dependencies are installed in the JH image 
- [ ] Investigate whether there are any low hanging fruit wins for developer experience to support SQL development in JupyterHub
- [ ] Extremely specific step by step documentation created for how to make a dbt contribution via JH -- we can check with analysts for where this should live, perhaps it can just get added to the warehouse README

PR #2446, opened 2023-03-31
Better dbt support in jupyterhub

atvaccaro:
# Description

Resolves https://github.com/cal-itp/data-infra/issues/2438

Cleans up Jupyter and adds necessary support for initial dbt work using a poetry-managed virtual environment. Also updates documentation and changes the CI to edit existing docs preview comments instead of creating a new one each commit.

I'd like to investigate https://github.com/yuvipanda/jupyterhub-ssh as a follow-up.

## Type of change

- [x] New feature
- [x] Documentation
- [ ] agencies.yml

## How has this been tested?
Been testing locally; the real test will be deploying to production and having someone go through the dbt setup from scratch.


docs/analytics_tools/github_setup.md
```diff
@@ -3,46 +3,47 @@
...
+  exit status 3
+  Please try entering the URL in your browser manually
+✓ Authentication complete.
+- gh config set -h github.com git_protocol ssh
+✓ Configured git protocol
+✓ Uploaded the SSH key to your GitHub account: /home/jovyan/.ssh/id_ed25519.pub
+✓ Logged in as atvaccaro
+```
 
 After completing the steps above be sure to complete the section below to persist your SSH key between sessions and enable extensions.
```
lauriemerrell, line 48: Is this still required? (persist the SSH key etc.)

lauriemerrell (commented):
Just one question. I'd defer to @charlie-costanzo or @SorenSpicknall on the actual JupyterHub updates

docs/analytics_tools/github_setup.md
atvaccaro: Yes, `profile.sh` does configure the key with keychain but if a `.bashrc` file exists, it'll ignore `profile.sh` unless you specifically source it (as best I can tell)

atvaccaro:
Ok docs are broken because of the pandas 2.0 release; [python-gbd](https://github.com/googleapis/python-bigquery-pandas/blob/main/setup.py#L28) does not specify a maximum version, so we'll need to do it ourselves. siuba is not compatible yet but it looks like michael is [working on it](https://github.com/machow/siuba/pull/472).

.github/workflows/publish-docs.yml
```diff
@@ -51,28 +53,18 @@ jobs:
         NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }}
         PR_NUMBER: ${{ github.event.number }}
 
-    - name: Add netlify link PR comment
-      uses: actions/github-script@v3
+    - uses: peter-evans/find-comment@v2
```
SorenSpicknall, line 56: Not actionable on this PR, but I've mentioned something similar before - we should try to find a mechanism to identify if an individually authored action goes stale or falls out of maintenance, so we don't end up scrambling for a replacement in that case.

docs/analytics_examples/warehouse_tutorial.md
```diff
@@ -33,7 +33,7 @@ The queries represented in the following tutorial are as follows:
 The tools that we can use to answer them are:
 * [**Metabase**](metabase) - our business insights and dashboarding tool
 * [**SQL**](querying-sql-jupyterhub) - using JupyterHub cloud notebooks
-* [**Python**](jupyterhub) - using JupyterHub cloud notebooks
+* [**Python**](jupyterhub-intro) - using JupyterHub cloud notebooks
```
SorenSpicknall, line 36: Do we have externally maintained links (in Google docs, slide decks, etc.) that are going to need updating from this?

SorenSpicknall (approved):

docs/analytics_examples/warehouse_tutorial.md
atvaccaro: Hard to know for sure but unlikely, we tend to just link to the top-level page

atvaccaro:
I just added back the portfolio dependency group; I'd removed when trying to install dbt in the image directly, but we can keep it if we're using `poetry` for warehouse work.

Issue #2453, opened 2023-04-03
Run GTFS Schedule unzip/validate job hourly


lauriemerrell:
Per discussion with @atvaccaro, we think that the GTFS schedule unzip/validate Airflow DAG task should run hourly instead of daily to help mitigate this:

![image](https://user-images.githubusercontent.com/55149902/229624909-89596c77-0abc-4837-b7e5-765393c2836b.png)

Basically, GTFS schedule data doesn't make it into the BQ warehouse until 30+ hours after it is downloaded because of how we have the unzip/validate job structured to wait until the day is over. If we switched to hourly instead of daily processing, we could get schedule data into the warehouse within 12 hours, more in line with Airtable and GTFS RT and reduce the significant delay.

Steps:
- [ ] Change unzip outcomes partitioning to be hourly (no changes needed for raw data, just outcomes); this may require changing metadata attributes as well
- [ ] Backfill old outcomes into new partitions
   - [ ] This will require splitting up files for days that had multiple downloads
   - [ ] This may require backfilling new metadata attributes
- [ ] Change the schedule of the Airflow task

PR #2460, opened 2023-04-05
Parse schedule data hourly

atvaccaro:
# Description

Resolves https://github.com/cal-itp/data-infra/issues/2453

My deploy plan is as follows.
1. Create a new `_hourly` DAG that populates `_hourly` buckets.
2. Let that DAG backfill a couple of weeks.
3. Test that data, then do a full backfill.
4. Create a new PR to point the external tables at the new `_hourly` buckets.

Three new environment variables support this deployment.
* `CALITP_BUCKET__GTFS_SCHEDULE_VALIDATION_HOURLY`
* `CALITP_BUCKET__GTFS_SCHEDULE_UNZIPPED_HOURLY`
* `CALITP_BUCKET__GTFS_SCHEDULE_PARSED_HOURLY`

The corresponding buckets have been created.

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Tested the new jobs locally.


airflow/dags/unzip_and_validate_gtfs_schedule_hourly/unzip_gtfs_schedule.py
```diff
@@ -0,0 +1,282 @@
...
+    pbar=None,
+) -> Tuple[List[GTFSScheduleFeedFile], str]:
+    zipfile_files = []
+    md5hash = hashlib.md5()
+    if not is_valid:
+        raise ValueError(
+            "Unparseable zip: File/directory structure within zipfile cannot be unpacked"
+        )
+
+    # sorting is new here, to make the hash deterministic
```
lauriemerrell, line 126: did you add this comment? can you just say "sorting makes the hash deterministic"? I don't like comments that are specific to the time context of the PR that added them, they will not make sense later

SorenSpicknall (approved):

lauriemerrell (commented):
Why does this require new unzipped and parsed buckets long term? Can we not backfill into the existing buckets, just with one partition layer added for the outcomes? Asking because I thought our original goal was to only ever version the raw data bucket and that everything else was malleable in place. Sorry if I am misunderstanding something

atvaccaro:
> Why does this require new unzipped and parsed buckets long term? Can we not backfill into the existing buckets, just with one partition layer added for the outcomes? Asking because I thought our original goal was to only ever version the raw data bucket and that everything else was malleable in place. Sorry if I am misunderstanding something

My worry was that using existing buckets means there's some amount of downtime, but if we use new buckets we can just swap the external tables once we've tested things.

It'd be different as a backfill of metadata or processing changes, where we can actually replace existing files. But here, we have to create new files and drop old ones.

lauriemerrell:
>My worry was that using existing buckets means there's some amount of downtime, but if we use new buckets we can just swap the external tables once we've tested things.
> It'd be different as a backfill of metadata or processing changes, where we can actually replace existing files. But here, we have to create new files and drop old ones.

Can we use the new buckets for the backfill but swap them into place when the backfill is done so that "prod" ultimately stays the same? Similar to the concern about deprecating tables, I think we should not be keeping a lot of buckets around with `_vX` because I think it does get confusing and makes mistakes more likely. If we're doing this I'd want to add `deprecated` to the old buckets so it's more obvious and they're alphabetically distinct.

atvaccaro:
Sure, it'll be copies since we can't rename buckets, but I can swap them back when done.

lauriemerrell (approved):

PR #2473, opened 2023-04-12
SAFE_CAST of lat/long/money values to NUMERIC in warehouse

charlie-costanzo:
# Description

The payments v2 pipeline has the following values in `STRING` format. This PR casts them as `NUMERIC`

In `stg_littlepay__micropayments`
* charge_amount (and thus `refund_amount` downstream)
* `nominal_amount`

In `fct_payments_rides_v2` (couldn't perform this cast in the staging table `stg_littlepay__micropayment_adjustments` because it needs to be a `STRING` for partition)
* `adjustment_amount`

## Type of change

- [x] New feature

## How has this been tested?
locally with dbt


atvaccaro (changes requested):
can you use numeric instead of float? https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#decimal_types

atvaccaro (approved):

atvaccaro (approved):

PR #2477, opened 2023-04-13
V2 payments reliability tables

charlie-costanzo:
# Description

This PR replicates the tables that power the payments reliability dashboard in the v1 pipeline for the v2 pipeline

Question @atvaccaro - is there a better place that this should live?

## Type of change

- [x] New feature


atvaccaro (approved):

atvaccaro:
> Question @atvaccaro - is there a better place that this should live?

I think this is OK; we could potentially have `mart_payments_reliability` or something but `mart_payments` is probably fine

PR #2492, opened 2023-04-21
Handle frequency-based trips in GTFS schedule

### LINKED ISSUE #2471: Handle frequency based trips in the warehouse
Right now, we don't handle frequency-based trips (trips defined in `frequencies.txt`) in `fct_daily_scheduled_trips` or in `fct_observed_trips`. 

We should:

- [x] Add an intermediate table that expands out frequency based trips using stop times
- [x] Add a second intermediate table that groups that first table analogous to `int_gtfs_schedule__stop_times_grouped`
- [x] Incorporate the new intermediate tables in `fct_daily_scheduled_trips`, `_stops`, and `_shapes`
- [ ] ~Add better handling for frequency-based trips in any scheduled/RT comparisons (this may require a follow-up ticket depending on when this is undertaken relative to #2236) -- this could potentially involve fixing #2413 by standardizing trip identifiers based on whether or not the trip is a frequency-based trip~

lauriemerrell:
# Description

This PR adds support for trips defined in `frequencies.txt` in the GTFS schedule portion of our data warehouse. Previously these trips were not reflected correctly; we would only list them as they appeared in `stop_times`, even though the spec best practices say that the times in `stop_times` should only be used for stop-to-stop durations and they do not have to correspond to the times of actual trip activity for frequencies-defined trips: 
image

Resolves #2471 

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

Ran & tested all affected tables. Specifically checked that we aren't losing any rows that currently exist in the `_stops` mart table.


warehouse/macros/gtfs_time_string_to_interval.sql
```diff
@@ -0,0 +1,10 @@
+{% macro gtfs_time_string_to_interval(gtfs_time_field) %}
+-- convert a GTFS "Time" string to a BQ interval type
+-- INTERVAL type allows us to handle times past midnight (ex. 26:30:30)
+-- see: https://gtfs.org/schedule/reference/#field-types for how GTFS defines a "Time"
+
+CASE
+    WHEN REGEXP_CONTAINS({{ gtfs_time_field }}, "^[0-9]+:[0-5][0-9]:[0-5][0-9]$") THEN CAST({{ gtfs_time_field }} AS INTERVAL)
```
SorenSpicknall, line 7: Can you think of any cases where a GTFS time field is potentially negative? And if negatives are possible source values here, does this properly cast to the INTERVAL in a way that distributes the negative across hour/minute/second values?

SorenSpicknall (approved):
Generally looks good to me, but I had one substantive question on interval calculation that I want to make sure we've covered cases for. Tentatively approving.

warehouse/macros/gtfs_time_string_to_interval.sql
lauriemerrell: Times are not supposed to be negative. GTFS doesn't officially call these intervals; they portray it as like an extension of `HH:MM:SS` rather than an interval, we are just using the BigQuery interval type because it natively handles `HH > 24`. 
image

Issue #2362, opened 2023-03-08
Log individual dbt-metabase failures to Sentry at the model level


atvaccaro:
**Is your feature request related to a problem? Please describe.**
We've been just failing the Airflow job if dbt-metabase throws an exception, when only 1 model might have failed and the consequences are low.

**Describe the solution you'd like**
I've stopped failing the job in https://github.com/cal-itp/data-infra/pull/2361 but now we won't see any failures except in logs. We should parse the output of the sync and log a custom exception to Sentry for each model that needs fixing.

Also, @lauriemerrell requested adding the model path to dbt test failures/warns, and I think we should try to add those here too.

**Describe alternatives you've considered**
I don't think we want to be louder and quieter than 1 sentry issue per model?

**Additional context**
n/a

PR #2501, opened 2023-04-24
log dbt-metabase errors and add enrichment/filepaths

atvaccaro:
# Description

~I actually have mixed feelings about https://github.com/cal-itp/data-infra/issues/2362; my worry is that we don't actually know when Metabase has synced a schema. At best, we could trigger a manual sync and wait for a period of time. I don't think we can actually monitor for sync completion.~

~Actually it looks like we can use https://www.metabase.com/docs/latest/api/notify potentially~

Even better, dbt-metabase already triggers a sync and waits before attempting to update metadata.

Resolves https://github.com/cal-itp/data-infra/issues/2362

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
See [Sentry](https://sentry.calitp.org/organizations/sentry/issues/70462/?environment=development&project=2&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h) for examples

### DIFF
--- a/warehouse/scripts/run_and_upload.py
+++ b/warehouse/scripts/run_and_upload.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 import json
 import os
+import re
 import shutil
 import subprocess
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


warehouse/scripts/run_and_upload.py
```diff
@@ -284,12 +291,59 @@ def get_command(*args) -> List[str]:
...
+                capture_output=True,
             )
+
+            with open("./target/manifest.json") as f:
+                manifest = Manifest(**json.load(f))
+            matches = {}
+            for line in p.stdout.decode().splitlines():
+                print(line)
+                for pattern in (
+                    r"WARNING\s+Model\s(?P<dbt_schema>\w+)\.(?P<model>\w+)\snot\sfound\sin\s(?P<db_schema>\w+)",
```
SorenSpicknall, line 316: Just noting we'll have to be sure to come back to this later if we change the strings underlying this pattern we're looking for - since we don't have a direct linkage here to trace back to the strings themselves in code, just a regex, it's potentially brittle/easily missed in future changes.

SorenSpicknall (approved):

PR #2502, opened 2023-04-24
notify sentry of releases

atvaccaro:
# Description

Resolves # [issue]

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

### DIFF
--- a/.github/workflows/sentry-release.yml
+++ b/.github/workflows/sentry-release.yml
@@ -1,0 +1,29 @@
+name: Notify Sentry of releases
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


atvaccaro:
I think squash-merges may make releases weird...

atvaccaro:
> I think squash-merges may make releases weird...

Ok this will be weird in test since the release/commit won't exist in production, but I think it's probably OK within each environment. (We don't really try for 1:1 parity between test and prod anyways.)

SorenSpicknall (approved):

Issue #2485, opened 2023-04-17
Mechanism for partial re-run of incremental tables


lauriemerrell:
Currently in our GTFS data models we tend to have incremental models that either full-refresh (re-run all of history since a set start date) or only update data since the last time the table was run. It would be useful to have a mechanism to re-run in an incremental fashion since some intermediate date so that if a new data issue is encountered (ex. #2484) we can re-run the affected period without re-running all of history. 

Likely this would be a new variable that can be optionally invoked instead of `ts > select max(timestamp from {{ this }})` in `is_incremental` clauses.

PR #2504, opened 2023-04-25
create new incremental macro to support partial re-processing of historical GTFS data

atvaccaro:
# Description

~Blocked by fully backfilling https://github.com/cal-itp/data-infra/pull/2121 as we need line numbers on Schedule rows to merge in the incremental dim models~

Resolves https://github.com/cal-itp/data-infra/issues/2485
Tabling https://github.com/cal-itp/data-infra/issues/2541 as we don't have a reliable way of avoiding attempting to re-read empty files constantly

Request from @lauriemerrell: allow passing `vars` to warehouse/warehouse_full_refresh jobs so a user can specify the new lookback (this is done)

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Testing specific table (`dim_stop_times` with the incremental change, and then testing all GTFS Schedule mart tables for the test changes.
`poetry run dbt build -s dim_stop_times --vars 'INCREMENTAL_MAX_DT: 2023-05-01'`
`poetry run dbt run -m +mart.gtfs_schedule_latest`
`poetry run dbt test -s dim_schedule_feeds+1,mart.gtfs --indirect-selection=cautious`
`poetry run dbt run -m +fct_daily_scheduled_trips --full-refresh`
`poetry run dbt test -m fct_daily_scheduled_trips`

I've also confirmed the vars are passable in a configured Airflow run.


lauriemerrell (commented):
Can you document the new `_dt` and `_line_number` columns in the relevant YAML? Can just make a doc template and use it everywhere. 

Specifically though you need to add them to the `_latest` YAML and suppress them from publishing otherwise they will be published by default (perhaps we should flip this to be publishing is suppressed by default?) 

Also I think I am missing something but I am not really understanding some the implementation here wrt `_dt` (see inline comments).

warehouse/models/mart/gtfs/dim_agency.sql
```diff
@@ -18,7 +18,9 @@ dim_agency AS (
         agency_fare_url,
         agency_email,
         base64_url,
+        _dt,
```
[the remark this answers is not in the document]
atvaccaro: `_dt` is the date rather than timestamp, since we have to eliminate date partitions first if we want the read savings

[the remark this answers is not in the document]
atvaccaro: see above

warehouse/models/intermediate/gtfs/int_gtfs_schedule__frequencies_stop_times.sql
```diff
@@ -60,7 +60,9 @@ int_gtfs_schedule__frequencies_stop_times AS (
         warning_duplicate_primary_key AS warning_duplicate_stop_times_primary_key,
         warning_missing_foreign_key_stop_id,
         base64_url,
+        _dt,
         _feed_valid_from,
+        _line_number,
```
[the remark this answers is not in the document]
atvaccaro: ah yeah bad find-replace, will fix this plus below

warehouse/models/intermediate/gtfs/int_gtfs_schedule__all_scheduled_service.sql
```diff
@@ -16,7 +16,9 @@ boolean_calendar_dates AS (
         {{ dbt_utils.generate_surrogate_key(['feed_key', 'service_id', 'date']) }} AS key,
         date AS service_date,
         feed_key,
+        _dt,
```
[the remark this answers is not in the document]
lauriemerrell: But these intermediate tables are on top of materialized tables, they aren't reading from the external tables.

lauriemerrell:
>  I'd prefer to keep any default changes out of this PR.

Wrote #2651

warehouse/models/intermediate/gtfs/int_gtfs_schedule__all_scheduled_service.sql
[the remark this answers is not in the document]
atvaccaro: Ah yeah sorry I'm removing these; same bad find-replace

warehouse/models/mart/gtfs/_mart_gtfs_dims.yml
```diff
@@ -51,14 +51,27 @@ models:
...
+          Synthetic primary key constructed from `feed_key` and `_line_number`.
+        tests:
+          - not_null
+          - unique
+      - &schedule_dim_dt
+        name: _dt
+        description: Date of the feed version; used for upstream partition elimination.
+        tests:
           - not_null
+        meta:
```
lauriemerrell, line 66: You don't need `publish.ignore` here

warehouse/models/mart/gtfs/dim_shapes_arrays.sql
```diff
@@ -50,7 +48,7 @@ initial_pt_array AS (
         -- count number of rows so we can check for nulls (drops) later
         COUNT(*) AS ct
     FROM lat_long
-    GROUP BY feed_key, base64_url, shape_id, _feed_valid_from, feed_timezone
+    GROUP BY feed_key, base64_url, shape_id, _feed_valid_from, _line_number, feed_timezone
```
lauriemerrell, line 51: cannot group by line number here! each input point has its own line

warehouse/models/mart/gtfs_schedule_latest/_gtfs_schedule_latest.yml
```diff
@@ -11,14 +11,24 @@ models:
     meta:
       ckan.authority: https://gtfs.org/schedule/reference/#agencytxt
     columns:
-      - &primary_key
+      - &schedule_dim_pk
         name: key
         description: |
           Synthetic primary key constructed from `feed_key` and `_line_number`.
         tests:
-          - unique:
```
lauriemerrell, line 18: Have we actually dealt with this? I thought we reintroduced this issue, I think without this filter the test will fail

lauriemerrell (changes requested):
I think you need to run a full run locally including tests before we merge this, I caught a few things but worried I might be missing others. Also as discussed offline I am disinclined to merge this on a Friday before a long weekend, we can aim to merge Tuesday once changes addressed

warehouse/models/mart/gtfs/dim_shapes_arrays.sql
atvaccaro: yeah I realized right after I pushed 🙃 fixing

warehouse/models/mart/gtfs_schedule_latest/_gtfs_schedule_latest.yml
atvaccaro: good question, it's fine now because line_number will always exist; the issue was null values being equal

atvaccaro:
> I think you need to run a full run locally including tests before we merge this, I caught a few things but worried I might be missing others. Also as discussed offline I am disinclined to merge this on a Friday before a long weekend, we can aim to merge Tuesday once changes addressed

yep sounds good to me

atvaccaro:
discussed with @lauriemerrell a bit; I'm going to re-add those unique-unless-flag tests as well as add threshold tests on those flags (e.g. test than the flag is present on no more then 50% of rows)

warehouse/models/mart/gtfs_schedule_latest/_gtfs_schedule_latest.yml
lauriemerrell: https://github.com/cal-itp/data-infra/pull/2504#issuecomment-1564919117

warehouse/models/mart/gtfs/_mart_gtfs_dims.yml
```diff
@@ -252,10 +294,14 @@ models:
       Definitions for the original GTFS fields are available at:
       https://gtfs.org/schedule/reference/#fare_productstxt.
     columns:
-      - name: key
-        description: |
-          Synthetic primary key constructed from `feed_key` and `fare_product_id`.
-        tests: *pk_tests_with_dups
+      - *schedule_dim_pk
+      - *schedule_dim_gtfs_key_with_dupe_flag
+      # some very large feeds have many duplicates, so row-level counts on this violate our test
```
lauriemerrell, line 299: Should this comment be moved up to where the anchor is defined? Specifically to define why you're excluding the MTC regional feed? And also, why exclude that feed from *all* such tests? Seems a bit odd?

warehouse/models/mart/gtfs/dim_calendar_dates.sql
```diff
@@ -6,17 +6,23 @@ WITH make_dim AS (
...
         date,
         exception_type,
         base64_url,
+        _dt,
         _feed_valid_from,
+        _line_number,
         feed_timezone,
     FROM make_dim
+    -- filter rather than flag; lots of downstream models expect uniqueness
+    -- we could remove if we handled it everywhere but these are mostly full duplicates
```
lauriemerrell, line 23: not "mostly"; these are all full duplicates

warehouse/models/mart/gtfs/dim_calendar_dates.sql
```diff
@@ -6,17 +6,23 @@ WITH make_dim AS (
...
+    -- and they occur on many feeds
+    QUALIFY ROW_NUMBER() OVER (PARTITION BY feed_key, service_id, date ORDER BY _line_number) = 1
```
lauriemerrell, line 25: should have exception type too; if we have dups with conflicting exception types we have to figure something else out

warehouse/models/mart/gtfs/dim_attributions.sql
```diff
@@ -32,15 +22,16 @@ dim_attributions AS (
         attribution_email,
         attribution_phone,
         make_dim.base64_url,
-        COALESCE(warning_duplicate_primary_key, FALSE) AS warning_duplicate_primary_key,
+        COUNT(
```
lauriemerrell, line 25: Why did you refactor this one away from `bad_rows` but not others

warehouse/models/mart/gtfs_schedule_latest/_gtfs_schedule_latest.yml
lauriemerrell, line 1: I think you have to add a publish ignore for your new `_gtfs_key` column to all these too

warehouse/models/mart/gtfs/_mart_gtfs_dims.yml
```diff
@@ -51,13 +51,35 @@ models:
...
+        tests:
+          - not_null
           - unique:
               # TODO: remove this once we deal with Auburn feed that has tab-delimited files
+              # and at that point can move the &primary_key_tests anchor up to here
+              # (until exclusion is dropped, this exclusion condition is different because it's for key instead of feed_key)
               where: "feed_key != '6368fe701bdd68c4f521751a9a222a10'"
+      - &schedule_dim_dt
+        name: _dt
+        description: Date of the feed version; used for upstream partition elimination.
```
lauriemerrell, line 76: Could add "corresponds to date extracted from `_feed_valid_from`" or note that it's a UTC date not a Pacific one

lauriemerrell (commented):
just a few more things

warehouse/models/mart/gtfs/dim_attributions.sql
atvaccaro: Only this one had the null condition in the join and it was causing bad fanout

warehouse/models/mart/gtfs/_mart_gtfs_dims.yml
atvaccaro: I will clarify in the YAML; I found this was causing multiple natural key tests to fail the 50% check since it's a very large feed that can dominate record-level tests that check for count or proportion

atvaccaro: Oh wait sorry this specific comment is specific to `dim_fare_products`; I will still clarify in both places

warehouse/models/mart/gtfs/_mart_gtfs_dims.yml
```diff
@@ -143,6 +143,8 @@ models:
         description: |
           Flags rows with a duplicate _gtfs_key; e.g. area_id duplicates would result in _gtfs_key duplicates within a feed version.
           Treat these rows with caution as they can cause fanout when joining on natural (i.e. GTFS) keys.
+          This test excludes the regional MTC feed because it's a large feed that often violates GTFS uniqueness
```
lauriemerrell, line 146: massive nit: should this be a comment not part of the column description? Don't think that including it in description will make sense once it's displayed in dbt docs

lauriemerrell (approved):
just one nit... I am still a bit confused, I thought that the MTC feed is usually pretty good about prefixing IDs with agency identifiers so I would NOT think they would introduce big issues with duplicates, especially for fields like `*_id` but it's probably fine

warehouse/models/mart/gtfs/_mart_gtfs_dims.yml
```diff
@@ -330,7 +367,8 @@ models:
       https://gtfs.org/schedule/reference/#fare_transfer_rulestxt.
     columns:
       - *schedule_dim_pk
-      - *schedule_dim_gtfs_key
+
```
lauriemerrell, line 370: shouldn't this still have `*schedule_dim_gtfs_key_with_dupe_flag`?

lauriemerrell (commented):
ok sorry we are really close, just a few tweaks

lauriemerrell (approved):

PR #2511, opened 2023-04-27
point external Schedule tables at new hourly buckets with new hive schema

atvaccaro:
# Description

Should only merge after https://github.com/cal-itp/data-infra/pull/2523 is merged and the backfill is done.

This will address some Sentry issues (dbt errors / failures) that are caused by timing issues:
Fixes CAL-ITP-DATA-INFRA-24AT
Fixes CAL-ITP-DATA-INFRA-25CC

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml


lauriemerrell (commented):
Where is `CALITP_DBT_SOURCE_DATABASE` set? The name is changed here but I don't see it actually being set (is it in a Docker image somewhere)? Otherwise I think this looks good

PR #2512, opened 2023-04-27
Bring Elavon batch type B records into staging table

SorenSpicknall:
# Description

Partway through the operation of our Elavon data pipeline, we received what appears to be a monthly drop of batch type B (billing) data. These records don't have a `trn_ref_num`, which we previously were able to treat as a unique ID. We are still awaiting more detailed information about the dynamics of these data, but because this pipeline is currently doing a full file download + full reload into an external table every day, we can modify the dbt job that generates the staging table to give us the correct rows for B-batch data while not interrupting the logic that works for other batch types.

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?

Staging table generated from scratch in the staging warehouse

### DIFF
--- a/warehouse/models/staging/payments/elavon/stg_elavon__transactions.sql
+++ b/warehouse/models/staging/payments/elavon/stg_elavon__transactions.sql
@@ -1,4 +1,4 @@
-WITH source AS (
+WITH primary_batch_transactions AS (
     SELECT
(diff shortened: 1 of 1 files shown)


charlie-costanzo (approved):
This LGTM! Thank you for taking care of this

PR #2518, opened 2023-04-27
Post demo updates

atvaccaro:
# Description

Updates docs. Adds max bytes billed as an override-able profile template default.

Still looking into checkpoint issue.

## Type of change

- [x] New feature
- [x] Documentation
- [ ] agencies.yml

## How has this been tested?
Tested `dbt init`.


warehouse/README.md
```diff
@@ -43,6 +46,7 @@ are already configured/installed.
          schema: <yourname>
          threads: 8
          timeout_seconds: 3000
+         maximum_bytes_billed: 100000000000
```
lauriemerrell, line 49: What happens if you run initial setup with this set? This is not enough for initial build, right? Do only the large tables fail?

atvaccaro: Yes, I'll bump to 2 TB so most things (except trip updates) will fail. A full `dim_stop_times` is ~1.1 TB right now.

atvaccaro: Our lookback for trip updates is about 1 TB, so 2 will work for awhile.

lauriemerrell (approved):

PR #2523, opened 2023-04-27
backfill hourly schedule parsing to 2021

atvaccaro:
# Description

Follow-up to https://github.com/cal-itp/data-infra/pull/2460

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml


lauriemerrell (approved):

atvaccaro:
I've tested in staging by pointing at the production buckets with 2 weeks of data; I was able to produce a proper `dim_schedule_feeds` table (i.e. a few new feeds per day) so I think we can perform the full backfill.

atvaccaro:
I've also tested that row counts match on a couple outcome types (cc @lauriemerrell)

Issue #2478, opened 2023-04-14
Bug: need to adjust RT fetch exception fingerprinting and exception secret cleaning


atvaccaro:
**Describe the bug**
We should adjust the RT exception fingerprinting to be based on just the three following fields.

1. Top-level exception _type_ (not body/string representation, which can contain underlying exceptions)
2. URL with query param stripped for 511 feeds
3. (If relevant) HTTP error code

In addition, it looks like we occasionally have [API keys](https://sentry.calitp.org/organizations/sentry/issues/69491/?environment=cal-itp-data-infra&project=2&query=is%3Aunresolved+ConnectTimeoutError&referrer=issue-stream&statsPeriod=7d) showing up as query params; we should remove those if Sentry isn't catching them already.

**To Reproduce**
See these two sets of issues that should be combined.

Separate issues because the underlying exception/object reference differs:
[One](https://sentry.calitp.org/organizations/sentry/issues/70019/?environment=cal-itp-data-infra&project=2&query=is%3Aunresolved+0x&referrer=issue-stream&statsPeriod=14d)
[Two](https://sentry.calitp.org/organizations/sentry/issues/69530/?environment=cal-itp-data-infra&project=2&query=is%3Aunresolved+0x&referrer=issue-stream&statsPeriod=14d)

Separate 511 fetch errors:
[One](https://sentry.calitp.org/organizations/sentry/issues/39535/?environment=cal-itp-data-infra&project=2&query=is%3Aunresolved+rtfetchexception&referrer=issue-stream&sort=date&statsPeriod=24h)
[Two](https://sentry.calitp.org/organizations/sentry/issues/39534/?environment=cal-itp-data-infra&project=2&query=is%3Aunresolved+rtfetchexception&referrer=issue-stream&sort=date&statsPeriod=24h)

**Expected behavior**
These issues should be grouped going forward, and we should merge the old issues together.

**Additional context**
_Please enter any other context about the problem here._

PR #2528, opened 2023-04-28
RT fingerprint adjustments

atvaccaro:
# Description

Resolves https://github.com/cal-itp/data-infra/issues/2478

1. Get the top-level exception type for the fingerprint; avoids object hashes in (potentially very nested) exception bodies
2. Strips agency query params from end of URL (to combine 511 issues)

I've also added masking in the [Sentry UI](https://sentry.calitp.org/settings/sentry/security-and-privacy/) but it doesn't seem to be entirely working yet.

fixes CAL-ITP-DATA-INFRA-24S9

(testing that)

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Deployed to test now.

See [this issue](https://sentry.calitp.org/organizations/sentry/issues/70957/?environment=development&project=2&query=is%3Aunresolved&referrer=issue-stream) as an example of a new issue with the updated exception body and the 511 change.

### DIFF
--- a/services/gtfs-rt-archiver-v3/gtfs_rt_archiver_v3/consumer.py
+++ b/services/gtfs-rt-archiver-v3/gtfs_rt_archiver_v3/consumer.py
@@ -25,11 +25,14 @@ def set_exception_fingerprint(event, hint):
 
     exception = hint["exc_info"][1]
     if isinstance(exception, RTFetchException):
+        # strip out 511 agency ids, since the same outage often affects all of them
         cleaned_url = re.sub(
-            r"agency=\w+$", str(exception.url), "agency=XYZ"
-        )  # strip out 511 agency ids, since the same outage often affects all of them
+            pattern=r"agency=\w+$",
+            repl="agency=XYZ",
+            string=str(exception.url),
+        )
+        # use just the type to avoid differentiating based on underlying exceptions as well as object hashes in exception strings
         event["fingerprint"] = [
-            # use just the type to avoid differentiating based on underlying exceptions as well as object hashes in exception strings
             type(exception.cause).__name__,
             cleaned_url,
         ]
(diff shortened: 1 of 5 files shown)


services/gtfs-rt-archiver-v3/gtfs_rt_archiver_v3/consumer.py
```diff
@@ -25,12 +25,16 @@ def set_exception_fingerprint(event, hint):
 
     exception = hint["exc_info"][1]
     if isinstance(exception, RTFetchException):
+        # strip out 511 agency ids, since the same outage often affects all of them
+        cleaned_url = re.sub(
```
lauriemerrell, line 29: Do we have access here to the original URL from Airtable? Can we just use that for the API key scrubbing?

lauriemerrell (commented):
Wondering if we should tweak this at all:
1. Should the MTC agency placeholder be more obviously a placeholder? 
2. Can we not access the original input URL and its components instead of just passing through the compiled URL? (Could have one attribute for `pipeline_url` and another for API key name if you want to be able to identify issues with specific keys)

services/gtfs-rt-archiver-v3/gtfs_rt_archiver_v3/consumer.py
```diff
@@ -25,12 +25,16 @@ def set_exception_fingerprint(event, hint):
...
+            pattern=r"agency=\w+$",
+            repl="agency=XYZ",
```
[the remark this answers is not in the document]
atvaccaro: I can make it more explicit

services/gtfs-rt-archiver-v3/gtfs_rt_archiver_v3/consumer.py
atvaccaro: ooh that's a great idea

atvaccaro: Oh wait no this is separate; the API key issue exists within the body of the nested exception message (i.e. `requests` or `urllib` or something downstream formatting their own exception with the full URL path)

lauriemerrell (approved):

Issue #2537, opened 2023-05-01
Bug: the new dbt run config/selector only applies to runs


atvaccaro:
**Describe the bug**
see title

**To Reproduce**
Execute either of the warehouse DAGs with JSON config.

**Expected behavior**
It should also apply to `dbt test` calls.

**Additional context**
n/a

PR #2539, opened 2023-05-01
also apply config selector to test

atvaccaro:
# Description

Resolves https://github.com/cal-itp/data-infra/issues/2537

## Type of change

- [ ] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Testing locally via docker-compose, it did not work yet.

### DIFF
--- a/airflow/dags/transform_warehouse/dbt_test.yml
+++ b/airflow/dags/transform_warehouse/dbt_test.yml
@@ -10,6 +10,8 @@ arguments:
   - '--no-dbt-seed'
   - '--no-dbt-run'
   - '--dbt-test'
+  - '--select'
+  - "{{ dag_run.conf.get('dbt_select_statement', '') }}"
 
 dependencies:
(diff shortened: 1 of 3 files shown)


airflow/dags/transform_warehouse/dbt_test.yml
```diff
@@ -7,8 +7,11 @@ cmds:
 arguments:
   - '/app/scripts/run_and_upload.py'
   - 'run'
```
lauriemerrell, line 9: just noting that it is kind of confusing that the command here is `run` but 🤷‍♀️

lauriemerrell (approved):

airflow/dags/transform_warehouse/dbt_test.yml
atvaccaro: Yeah, I'll think of a new name if we ever revisit this

Issue #2535, opened 2023-05-01
See if we can avoid truncating the long process stderr in Sentry context (for RT validation)


atvaccaro:
The RT validator output is verbose, and it gets truncated when [logged as context](https://sentry.calitp.org/organizations/sentry/issues/71024/?environment=cal-itp-data-infra&project=2&query=is%3Aunresolved+firstSeen%3A-72h&referrer=issue-stream&statsPeriod=24h#context-process); we should see if we can control that so we can see the whole stacktrace.

See https://stackoverflow.com/questions/53699110/how-do-i-increase-the-max-length-of-captured-python-parameters-in-sentry for a potential solution

Issue #2534, opened 2023-05-01
Add single example RT extract to CalledProcessError context


atvaccaro:
Currently, we set the [Schedule extract](https://sentry.calitp.org/organizations/sentry/issues/70328/?environment=cal-itp-data-infra&project=2&query=is%3Aunresolved+firstSeen%3A-72h&referrer=issue-stream&statsPeriod=24h#context-schedule_extract) as context when validating RT feeds, but only the [hourly aggregation](https://sentry.calitp.org/organizations/sentry/issues/70328/?environment=cal-itp-data-infra&project=2&query=is%3Aunresolved+firstSeen%3A-72h&referrer=issue-stream&statsPeriod=24h#context-hour) of RT feeds being validated. We should also set a single example RT extract in the Sentry context, so it's easier to debug.

PR #2544, opened 2023-05-01
improve RT sentry context

atvaccaro:
# Description

Resolves https://github.com/cal-itp/data-infra/issues/2534
Resolves https://github.com/cal-itp/data-infra/issues/2535

## Type of change

- [x] New feature
- [ ] Documentation
- [ ] agencies.yml

## How has this been tested?
Tested locally; [example issue](https://sentry.calitp.org/organizations/sentry/issues/70317/?environment=local&project=2&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=1h).

### DIFF
--- a/jobs/gtfs-rt-parser-v2/gtfs_rt_parser.py
+++ b/jobs/gtfs-rt-parser-v2/gtfs_rt_parser.py
@@ -62,7 +62,7 @@ GTFS_RT_VALIDATOR_VERSION = os.environ["GTFS_RT_VALIDATOR_VERSION"]
 
 app = typer.Typer(pretty_exceptions_enable=False)
 
-sentry_sdk.utils.MAX_STRING_LENGTH = 2048
+sentry_sdk.utils.MAX_STRING_LENGTH = 2048  # default is 512 which will cut off validator stderr stacktrace; see https://stackoverflow.com/a/58124859
 sentry_sdk.init()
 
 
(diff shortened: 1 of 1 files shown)


jobs/gtfs-rt-parser-v2/gtfs_rt_parser.py
```diff
@@ -640,20 +643,23 @@ def parse_and_validate(
...
-                        "schedule_extract", json.loads(schedule_extract.json())
+                        "First RT Extract in Hour", json.loads(hour.extracts[0].json())
+                    )
+                    scope.set_context(
+                        "Schedule Extract", json.loads(schedule_extract.json())
                     )
+                    # get the end of stderr, just enough to fit in MAX_STRING_LENGTH defined above
                     scope.set_context(
-                        "process", {"stderr": e.stderr.decode("utf-8")[-1800:]}
+                        "Process", {"stderr": e.stderr.decode("utf-8")[-2000:]}
```
lauriemerrell, line 662: sorry but why is this `2000` instead of exactly the same as the max string length?

atvaccaro: No reason; I just stuck with a power of 2 for the library setting.

lauriemerrell (approved):
Ok can you maybe just add a bit of a comment about why you set the max string length / what it actually controls because you don't actually invoke it later?

PR #2547, opened 2023-05-01
New column names for quality tables

### LINKED ISSUE #2464: Rename guidelines_assessed / reports_site_assessed columns to be more explicit

_Please describe your need, outlining the key users, the feature being requested, and the goal that that the feature will facilitate. For example: **As a [user or stakeholder type], I want [software feature] so that [some business value]**_

As data quality-interested parties, Evan, Scott, Eric, and myself (reviewers of [this doc](https://docs.google.com/document/d/12klLxaKaGCz0_dYwvwNY4T1XXgOt57twFxxr0g2THNE/edit#)) want to change the names of the `guidelines_assessed` and `reports_site_assessed` columns in the `gtfs_quality` warehouse to something that actually reflects that they represent, for example `public_customer_facing` and `public_customer_facing_or_regional_subfeed` (respectively) or similar.

### Acceptance Criteria

Columns (and references) are renamed in all locations to something that reflects their semantic meaning (public, customer facing, subfeed) vs. their intended usage. 

### Notes

SorenSpicknall:
# Description
Responsive to #2464, renames two columns used in quality tables to reflect a more commonsense understanding of those columns' meanings.

Resolves #2464

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?
Working through dbt tests at the time of PR creation.

## Post-merge follow-ups
- [ ] No action required

Downstream references to the old columns (e.g. in Metabase) will need to be switched out.

### DIFF
--- a/docs/analytics_tools/rt_analysis.md
+++ b/docs/analytics_tools/rt_analysis.md
@@ -28,7 +28,7 @@ This section includes detailed information about the data model and processing s
 
 All of the above are sourced from the v2 warehouse. Note that all components must be present and consistently keyed in order to successfully analyze. This module works at the organization level in order to match the reports site and maintain the structure of the speedmap site.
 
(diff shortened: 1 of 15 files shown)


lauriemerrell (commented):
I think this looks good, just two questions (neither of them blocking necessarily):

1. We should double check the reports repo and make sure there's nothing in the reports generation code that references the `reports_site_assessed` column by name. I just did a quick search and I think it's ok but just something to consider whenever touching reports code.

2. This is a question primarily for @evansiroky / @owades / @edasmalchi which maybe we can take to Slack: This PR renames the `guidelines_assessed` column to `public_customer_facing` but technically the logic here does also include the `fixed_route` check because that's how the logic for `organization_assessed` is currently set in Airtable. Should we include fixed route in the name and make a new column when the [proposed changes](https://docs.google.com/document/d/1pLjxYGiOCC6ctLUuKg0i4MqAKWgzThqkLrrAAHDfvww/edit) take effect? And should `currently_operating` be part of the name?

Issue #2484, opened 2023-04-17
Handle tab-delimited data


lauriemerrell:
In #2465 and #2468 I suppressed tests on an Auburn feed that had some tab- (rather than comma-) delimited files. Ideally we should implement a more substantive fix for these files. Per [this Google doc](https://docs.google.com/document/d/15VwaDbp72hP69TWKM7LL3b31vxtrD5S_wsf_2-_PL34/edit#) we are evenly divided on whether it's preferable to try to just parse the tab-delimited files in the pipeline or whether we should suppress the null rows that result. 

We should assess the level of effort associated with each approach and pick one based on what is feasible right now.

lauriemerrell:
Reopening because of #2567; initial fix approach did not work

atvaccaro:
Re-opening until the backfill is performed.

lauriemerrell:
This has been backfilled

PR #2567, opened 2023-05-08
Revert csv sniffing

atvaccaro:
# Description

It turns out that the CSV sniffing regex blows up on specific files, but we can't really just sniff a sample in case of doublequote issues. I'm going to revert this, and just specifically handle commas and tabs within the `'excel'` dialect (the default).

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

## Post-merge follow-ups

- [ ] No action required

Need to clear the stalled stop_times and re-run external tables.


airflow/plugins/operators/gtfs_csv_to_jsonl_hourly.py
```diff
@@ -90,8 +90,7 @@ def parse_individual_file(
     lines = []
     try:
         with fs.open(input_file.path, newline="", mode="r", encoding="utf-8-sig") as f:
-            # read the whole file so we see any doublequote usage, mainly
-            dialect = csv.Sniffer().sniff(f.read())
+            dialect = csv.Sniffer().sniff(f.readline())
```
lauriemerrell, line 93: Does this also need to be deleted?

atvaccaro: yes I just realized it was a separate PR and was about to push

lauriemerrell (approved):
Do we want a follow up ticket to handle tabs through some other manner? You reference Excel dialect, I thought that by reverting this all as-is we will continue to fail on the tab-delimited feed

lauriemerrell:
I will reopen #2484?

PR #2582, opened 2023-05-10
NTD source cleanup

### LINKED ISSUE #2424: Clean up mart NTD data
Per [discussion on Slack](https://cal-itp.slack.com/archives/C02KH3DGZL7/p1669161731863779) we want to start adjusting how NTD data is maintained in the warehouse.

Immediate actions to take:

* Anywhere that `organization_ntd_id` is present in the warehouse, it should be sourced from the seed file: https://github.com/cal-itp/data-infra/blob/main/warehouse/seeds/ntd_agency_to_organization.csv, NOT from the Airtable `ntd_agency_info` table.
* `dim_ntd_agency_info` should be deprecated and removed from the mart (see also #1355) to avoid confusion. `dim_annual_database_agency_information` is canonical now.
* Perhaps we should rename `dim_annual_database_agency_information` to have `ntd` in the name to make it more discoverable.

SorenSpicknall:
# Description
This PR will switch all existing tables that use `organization_ntd_id` to rely upon the NTD seed, rather than the current lineage that points back to Airtable and by passes the seed. We accomplish this by adding `ntd_id` sourced from the seed into the `dim_organizations` model. Several other models, like `dim_ntd_agency_info`, are deprecated in the process.

Resolves #2424

## Type of change

- [x] New feature
- [x] Documentation

## How has this been tested?
Changes tables and several downstream tables tested via local dbt

## Post-merge follow-ups
The availability of `ntd_id` directly on `dim_organizations` should be announced to users, and we should take our usual table deprecation/deletion steps on the models deleted via this PR.

- [ ] No action required


warehouse/models/mart/transit_database/dim_organizations.sql
```diff
@@ -44,7 +45,7 @@ dim_organizations AS (
     FROM dim
     LEFT JOIN ntd_agency_to_organization ntd_to_org
         ON dim.source_record_id = ntd_to_org.organization_record_id
-    LEFT JOIN dim_annual_database_agency_information AS ntd
+    LEFT JOIN dim_annual_ntd_agency_information AS ntd
         -- TODO: make this historical and use a bridge table
```
lauriemerrell, line 49: This `TODO` can be removed. Is there an `_is_current` column in `dim_annual_ntd_agency_information`? Also see above I think we may not even need this second join.

warehouse/models/mart/transit_database/dim_organizations.sql
```diff
@@ -20,6 +20,7 @@ dim_organizations AS (
         dim.source_record_id,
 
         ntd.key AS annual_database_agency_information_2021_key,
+        ntd.ntd_id,
```
lauriemerrell, line 23: Can you just use `ntd_to_org.ntd_id` here? Do we even need the second join?

lauriemerrell (commented):
Just one suggestion. Also, I'd want to double check that `dim_organizations` still passes all its tests just to confirm there's no fanout (no organization with multiple NTD IDs). If there is fanout we may have to convert this to an array or something 🤔

warehouse/models/mart/transit_database/dim_organizations.sql
SorenSpicknall: The join was redundant, so was removed in [84691c6](https://github.com/cal-itp/data-infra/pull/2582/commits/84691c609797dde00a57a56a90e3b0e939a7442d).

warehouse/models/mart/transit_database/dim_organizations.sql
SorenSpicknall: Resolved in [84691c6](https://github.com/cal-itp/data-infra/pull/2582/commits/84691c609797dde00a57a56a90e3b0e939a7442d).

lauriemerrell (approved):

PR #2593, opened 2023-05-12
GTFS Schedule Performance Improvements

### LINKED ISSUE #2488: GTFS schedule performance improvements
In the course of #2457 I discovered a few tables that are expensive to build / query:

* `int_gtts_schedule__stop_times_grouped`: At time of writing this is about 155 GB to build. We should make this table incremental, possibly based on `feed_key` and `_feed_valid_from` (because this table is append-only, making it incremental is pretty trivial; there should be no need to update existing rows)
* `fct_daily_scheduled_stops`: Currently ~127 GB to build, could probably be made incremental similar to the above, as long as we suppress future service where `date > current_date()` (which I mostly think we should do anyway)
* `fct_daily_scheduled_trips`:
   * I actually think that we should rename this to `fct_scheduled_trips` (though that would potentially be disruptive, need to check with consumers) because it is not actually grouped to a day level -- it should be grouped to an individual trip instance level (see #2471)
   * This table is relatively expensive to query -- we probably should cluster on `base64_url` or something, may want to inspect query patterns to determine

SorenSpicknall:
# Description
Implements several changes to build behavior and clustering in order to clear up some cost-intensive operations on commonly used models/tables.

Resolves #2488

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?
Changes models run with local dbt

## Post-merge follow-ups
- [ ] No action required

We will want to benchmark performance following these changes against past behavior to measure performance impacts.


SorenSpicknall:
NOTE: this PR doesn't address the table rename requested in #2488, which has been deferred to #2639 in order to decouple the performance improvements from the rename process and roll out each as they're ready.

lauriemerrell (changes requested):
I think that as written these changes will not actually lead to performance improvements, because the primary data consumption in these builds is the data *reading*, so the performance wins would come from adding an efficient `where` clause on the tables being read in (particularly `dim_stop_times`, in the case of `stop_times_grouped`), not just adding keys to *build* incrementally (i.e., as written these changes will still *read* all the same data and therefore will still be just as expensive to build; we need to limit how much data is read on a daily incremental build). See [this section of dbt's incremental docs](https://docs.getdbt.com/docs/build/incremental-models#filtering-rows-on-an-incremental-run). 

The way to test this would be to run the newly-incremental model *twice* locally. The second run should read significantly less data because it can build on top of the first run and you'll see a much lower bytes billed in the terminal output. 

Happy to chat more offline if you want on how to approach this... I think the most efficient might be identify `feed_key`s that exist in `dim_stop_times` but not in `stop_times_grouped`, because dim stop times is clustered on feed key.

atvaccaro:
Sorry for butting in, but I think some of the new bigquery performance insights are relevant here
image

SorenSpicknall:
Closing in favor of #2711

PR #2594, opened 2023-05-15
Update NTD scraper with new calitp-data-infra import

SorenSpicknall:
# Description
While working through old import references in our docs, I found this old import still active in our NTD scraper. Best to patch it up before we need to run it next.

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?
Requirements.txt imports checked with new calitp-data-infra version reference, and a modified version of the NTD import (with no real effect) was tested to ensure the import made it in okay.

## Post-merge follow-ups
- [x] No action required


atvaccaro (approved):
this might break some of the old scripts but we should probably delete them anyways

Issue #2569, opened 2023-05-08
Change dbt CI Action to highlight affected incremental tables 


lauriemerrell:
We've been test-driving the CI Action that generates a list of affected dbt models for a little while and personally I have not found it particularly actionable. Per discussion with @atvaccaro, I think it might be more useful if the action specifically called out affected *incremental* tables and noted what follow-up actions may be needed. 

This would be in addition to #2568.

PR #2489, opened 2023-04-20
RT trip labeling / matching updates

lauriemerrell:
# Description
_Describe your changes and why you're making them. Please include the context, motivation, and relevant dependencies._

Progress towards (resolution of?) #2413 and #2347 

TODO:
- [ ] handle times past midnight: add a test 
- [x] Put schedule feed key (join via base64 URL), URL, time zone on the RT messages models -- can add calculated service date; use `_config_extract_ts` for message joins
- [ ] Add test that time zone doesn't change over time
- [x] Revert changes to `fct daily rt feed files`~
- [ ] Make inter- and intra-feed RT trip key (inter: schedule base64 URL, trip id, calculated service date, and trip start time; intra: what it already is) 
- [ ] Vehicle positions
- [ ] Document intermediate models 
- [ ] Add imputed `iteration_num` to these tables to prepare for join with schedule
- [ ] Re-do `fct_observed_trips`
- [ ] Do we need to check for whether the calculated service date is between the active dates for service alerts 🤔 

Future:
* Add some test to try to validate scheduled time vs. observed time and warn if they seem really out of whack 

Resolves #[issue]

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

## How has this been tested?
_Include commands/logs/screenshots as relevant._

## Post-merge follow-ups
_Document any actions that must be taken post-merge to deploy or otherwise implement the changes in this PR (for example, running a full refresh of some incremental model in dbt). If these actions will take more than a few hours after the merge or if they will be completed by someone other than the PR author, please create a dedicated follow-up issue and link it here to track resolution._

- [ ] No action required
- [x] Actions required (specified below)

* Full refresh... Must include:
   * `fct_daily_rt_feed_files` 
* Post in Slack explaining changes
* Update [RT files Metabase dashboard](https://dashboards.calitp.org/dashboard/98) to no longer need a join to get dataset name

**TODO: add literal selectors for what needs full refresh**

PR #2605, opened 2023-05-16
Bring back cml CI and change what we print

atvaccaro:
# Description

Resolves https://github.com/cal-itp/data-infra/issues/2569

Note: I've rebased off handle-stop-times-enums to test the visualization; need to undo before merging.

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

CI should run and display.

## Post-merge follow-ups

- [x] No action required

### DIFF
--- a/warehouse/scripts/dbt_artifacts/__init__.py
+++ b/warehouse/scripts/dbt_artifacts/__init__.py
@@ -71,6 +71,7 @@ DependsOn.resolved_nodes = property(  # type: ignore[attr-defined]
 )
 ColumnInfo.publish = property(lambda self: not self.meta.get("publish.ignore", False))  # type: ignore[attr-defined]
 
+
 # End monkey patches
 
 
@@ -111,7 +112,10 @@ class NodeModelMixin(BaseModel):
         columns = [
             c
             for c in self.sqlalchemy_table(engine).columns
-            if not self.columns or c.name not in self.columns or self.columns[c.name].publish  # type: ignore[attr-defined]
+            if not self.columns
+            or c.name not in self.columns
+            or self.columns[c.name].publish
+            # type: ignore[attr-defined]
         ]
         return select(columns=columns)
 
@@ -120,12 +124,7 @@ class NodeModelMixin(BaseModel):
         """
         Returns a string representation intended for graphviz labels
         """
-        return "\n".join(
-            [
-                self.config.materialized or self.resource_type.value,  # type: ignore[attr-defined]
-                self.name,
-            ]
-        )
+        return self.unique_id
 
     @property
     def gvattrs(self) -> Dict[str, Any]:
@@ -133,7 +132,13 @@ class NodeModelMixin(BaseModel):
         Return a dictionary of graphviz attrs for DAG visualization
         """
         return {
-            "fillcolor": "black",
+            "fillcolor": "white",  # this is already the default but make explicit
+            "label": "\n".join(
+                [
+                    self.config.materialized or self.resource_type.value,  # type: ignore[attr-defined]
+                    self.name,
+                ]
+            ),
         }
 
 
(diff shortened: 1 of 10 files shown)


lauriemerrell:
We discussed offline: 
* Add a key to explain what the visualization means
* Add more text / checklists on what to do with the information, ex: if there are any incremental models in the diagram, did the author identify whether or not they need to be fully refreshed after merge?
* Confirm that color scheme is colorblind accessible and that there is text on everything (no color-only semantics)

warehouse/scripts/dbt_artifacts/__init__.py
```diff
@@ -151,43 +154,54 @@ class HookNode(BaseHookNode, NodeModelMixin):
...
+        more_than_100gb = (
             self.catalog_entry
             and self.catalog_entry.num_bytes  # type: ignore[attr-defined]
             and self.catalog_entry.num_bytes > 100_000_000_000  # type: ignore[attr-defined]
+        )
+
+        if self.config.materialized in ("table", "incremental"):
+            fillcolor = "aquamarine"
+
+        if self.config.materialized == "table" and more_than_100gb:
```
lauriemerrell, line 184: I am confused, don't we want to override this if it *is* clustered/partitioned already?

warehouse/scripts/templates/ci_report.md
```diff
@@ -0,0 +1,39 @@
...
+{% if changed_incremental_models %}
+### Changed incremental models 🔀
+
+{% for model in changed_incremental_models %}
+{{ model }}
+{% endfor %}
+{% endif %}
+
+### DAG
+Legend (in order of precedence)
```
lauriemerrell, line 20: I think this should go below the viz to avoid taking a focus area

warehouse/scripts/templates/ci_report.md
```diff
@@ -0,0 +1,39 @@
...
+
+| Resource type                                  | Indicator | Resolution                            |
+|------------------------------------------------|-----------|---------------------------------------|
+| Large table-materialized model                 | Red       | Make the model incremental            |
+| Large model without partitioning or clustering | Orange    | Add partitioning and/or clustering    |
+| View with more than one child                  | Yellow    | Materialize as a table or incremental |
+| Incremental                                    | Blue      |                                       |
```
lauriemerrell, line 27: I am confused about the blue vs. green when the main color I've actually seen is "aquamarine"

warehouse/scripts/templates/ci_report.md
```diff
@@ -0,0 +1,39 @@
...
+| Table                                          | Green     |                                       |
+| View                                           | White     |                                       |
+
+![](./dag.png "New and changed models")
+
+### Checks/potential follow-ups
```
lauriemerrell, line 33: I would move this to the very top of the message

warehouse/scripts/visualize.py
```diff
@@ -216,5 +238,72 @@ def viz(
...
+            "state:modified",
+            "--exclude",
+            "state:new",
+            "--state",
+            latest_dir,
+        ]
+    ).result
+    assert isinstance(modified_models, list)
+    viz(
+        "man",
```
lauriemerrell, line 288: `"man"` lol

lauriemerrell (commented):
I am a bit confused about some of the color setting

warehouse/scripts/dbt_artifacts/__init__.py
atvaccaro: Ok, I'm actually going to combine these conditions. Technically a large table could be cheap to build but end up with many rows (e.g. huge cross join) which isn't _necessarily_ a problem.

warehouse/scripts/templates/ci_report.md
```diff
@@ -0,0 +1,47 @@
...
+
+{% if new_models or changed_incremental_models %}
+### Checks/potential follow-ups
+
+Checks indicate the following action items may be necessary.
+{% if new_models -%}
+- [ ] For new models, do they all have a surrogate primary key that is tested to be not-null and unique?
+{%- endif %}
+{% if changed_incremental_models -%}
+- [ ] For changed incremental models, does the PR description identify whether a full refresh is needed for these tables?
```
lauriemerrell, line 11: `changed or downstream`

atvaccaro: good catch, I need to change the code for this

atvaccaro:
Actually I need to change this to add the models _between_ the selected ones, too.

atvaccaro:
I've tested this locally; it just won't render here without model changes, but there's no convenient PR to use to test right now.

warehouse/scripts/dbt_artifacts/__init__.py
```diff
@@ -416,21 +429,17 @@ def gvattrs(self) -> Dict[str, Any]:
         """
         Returns a string representation intended for graphviz labels
         """
+        color = "white"
         assert self.node is not None
-        # TODO: do an actual linear transform on this
-        # the top colors are too dark to use as a background
-        white, yellow, orange, red, _, _ = LaJolla_6.hex_colors
+        # TODO: use palettable
         if self.bytes_processed > 500_000_000_000:
```
lauriemerrell, line 435: These colors are not mentioned in the legend?

lauriemerrell (commented):
>  there's no convenient PR to use to test right now.

Could use #2489? 

One other question: Are we actually checking different things for `Large table-materialized model` vs. `Large model without partitioning or clustering | Orange | Add partitioning and/or clustering`? Specifically for the second one, don't we care about size to query vs. size to build? I am just wondering if you can actually assess need to partition/cluster based on bytes billed when building that table or if it's more like assessing the *parents* of a table that's expensive to build, if that makes sense?

warehouse/scripts/dbt_artifacts/__init__.py
atvaccaro: these are for run results, not manifest

atvaccaro:
> Could use #2489?

Will do, good idea.

> One other question: Are we actually checking different things for `Large table-materialized model` vs. `Large model without partitioning or clustering | Orange | Add partitioning and/or clustering`? Specifically for the second one, don't we care about size to query vs. size to build? I am just wondering if you can actually assess need to partition/cluster based on bytes billed when building that table or if it's more like assessing the _parents_ of a table that's expensive to build, if that makes sense?

The CI visualization is based on the manifest rather than run results; so it's just the storage size of the table, though theoretically that affects users as well as downstream tables. We could add latest run result in the future to recommend build changes but right now I think that's better handled by checking the Metabase dashboard.

lauriemerrell (approved):

PR #2610, opened 2023-05-16
Reduce not-null test threshold for fct_daily_rt_feed_files

### LINKED ISSUE #2571: Update not null proportion for fct_daily_rt_feed_files.gtfs_dataset_key
In #2555 we changed this to a not-null proportion test but this table actually has about 98% nulls in the gtfs_dataset_key column as of today 5/9/23 (the data was more problematic in the past, so more recent dates would pass a stricter percentage but overall this fails at the .999 threshold)

Sentry Issue: [CAL-ITP-DATA-INFRA-25BJ](https://sentry.calitp.org/organizations/sentry/issues/71304/?referrer=github_integration)

```
DbtTestFail: test.calitp_warehouse.dbt_utils_not_null_proportion_fct_daily_rt_feed_files_0_999__gtfs_dataset_key.543d04a4d2 - Got 1 result, configured to fail if != 0
```

SorenSpicknall:
# Description
Due to missing older data before our ingestion and modeling were evolved, fct_daily_rt_feed_files doesn't currently test a .999-threshold not null test for the `gtfs_dataset_key`. This lowers the threshold to .998 in order to pass, while keeping it high enough that newly introduced issues should still trip up alarms.

Resolves #2571

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?
dbt tests run locally, using the production table as a reference.

## Post-merge follow-ups
- [ ] No action required

Eventually, as this table grows, we should likely move this back to a .999-threshold test.

### DIFF
--- a/warehouse/models/mart/gtfs/_mart_gtfs_fcts.yml
+++ b/warehouse/models/mart/gtfs/_mart_gtfs_fcts.yml
@@ -630,7 +630,7 @@ models:
         historical GTFS data with their associated transit database records).
       tests:
         - dbt_utils.not_null_proportion:
-            at_least: 0.999
+            at_least: 0.998
         - relationships:
             to: ref('dim_gtfs_datasets')
             field: key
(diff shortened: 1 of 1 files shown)


lauriemerrell (approved):
If you wanted to add a TODO inline to raise this back up once enough time has passed, we could, but not a big deal

PR #2626, opened 2023-05-22
create DAGs for Sunday warehouse `--full-refresh`

### LINKED ISSUE #2568: Schedule weekly full refresh for non-RT data
In the course of re-running recent data post #2562, I encountered a case where a few DAG parsing tasks had been re-run in early April with code changes that changed the outputs of the parse job but a full refresh had not been run, so the parsed data in GCS was broken and had diverged from the associated mart table.

This led me and @atvaccaro to discuss whether we should be running full refreshes of schedule data more often so that it's less possible for these kinds of gaps to emerge and fester silently. We make frequent enough code changes that affect incremental schedule tables that a weekly (perhaps Sunday morning) full refresh of schedule data only (*not* RT for cost) would probably be beneficial.

AC:
- Sunday morning dbt run is a full refresh (excluding GTFS RT data) instead of just a normal run

charlie-costanzo:
# Description
This PR creates two new `transform_warehosue` DAGs that run on Sundays, and changes the configuration of the regular `transform_warehouse` DAG to not run on Sundays. These new DAGS compliment each other: one runs a `dbt run --full-refresh` of smaller (non-RT) data, and the other runs a normal `dbt run` for the remaining (larger) RT data.

The purpose of the two new DAGs are:
* `transform_warehouse_full_refresh_sunday_exclude_rt`
  * This DAG runs a `dbt run --full-refresh` every week on Sunday of *non-RT data only*, so as to capture any changes in the code that may not yet be represented in incremental tables (`--exclude gtfs_rt_external_tables+`)
* `transform_warehouse_sunday_select_rt`
  * This DAG runs a normal `dbt run` every Sunday of *only RT data* after the DAG run above,  so as to make sure that the RT data is also fresh, without running the expensive full-refresh on RT data (`--select gtfs_rt_external_tables+`)

Resolves #2568 

## Type of change

- [x] New feature

## How has this been tested?
successful local build of docker image 

## Post-merge follow-ups
Ensure this dag is running on schedule as intended


charlie-costanzo:
@atvaccaro a few questions about this:
* I broke this out to a new dag called `transform_warehouse_full_refresh_exclude_rt` and preserved the original dag `transform_warehouse_full_refresh` so that it can be used in the future. is that best?
* is my syntax for excluding rt  `-  '--exclude'` and `- 'gtfs_rt_external_tables+'` correct?
* Is my cron scheduling syntax correct?
* Is there a best way to test this? as it is mostly a scheduling change

lauriemerrell:
TLDR -- @atvaccaro, @charlie-costanzo, and I chatted about this offline and realized that this task has a more complicated scope & implications than originally understood, so we need to pause and reset a bit.

-----

**Things to re-consider:**

* GTFS RT data is consumed by some `gtfs_quality` models, so if the GTFS RT data is entirely skipped on Sunday, those models will have a weird hybrid of up to date schedule data and out of date RT data.
* Therefore, we probably need to run a *regular* (non-full) refresh of RT data first and *then* run a full-refresh job that excludes the expensive RT data. (We need *two* DAG task runs on Sundays with different selectors.)
* The RT data is complicated enough that we probably need to use tags or something to identify which RT models are safe to full refresh regularly, rather than relying on schemas or DAG relationships exclusively (we want to make sure it's an affirmative tag to indicate safety so if someone adds an RT model and forgets to tag, the model would be excluded from full refreshes by default.) 
* @atvaccaro thinks that the added complexity here warrants a switch to the dbt Python API rather than continuing to use `subprocess` calls so that we can do some assert statements to systematically check our selectors. The regular Sunday task model selectors plus the full refresh Sunday task selectors should be MECE.

So, TODO:

* Convert the dbt Airflow tasks to use the dbt Python API.
* Apply safety tags to cheaper RT models.
* Create *two* new Sunday jobs: one regular job that does a daily refresh of expensive RT data, and a full refresh for everything else. 
* Perform checks on the selectors for the two jobs to ensure that they are MECE. 

----
**Answers to Charlie's original questions:**

> I broke this out to a new dag called transform_warehouse_full_refresh_exclude_rt and preserved the original dag transform_warehouse_full_refresh so that it can be used in the future. is that best?

I think this is ok, to preserve the other one for primarily ad-hoc use. We may want to clarify that in the respective descriptions though. I don't know if there's an elegant way to use JSON config within the scheduled task to reuse them cleanly (i.e., to allow ad-hoc runs that include the RT data when needed.)

> is my syntax for excluding rt -  '--exclude' and - 'gtfs_rt_external_tables+' correct?

General syntax looks good, but there are a few specifics to change. The best way to test this is `poetry run dbt ls -s gtfs_rt_external_tables+ --resource-type model` to see what your select statement is grabbing (`dbt run` includes that `--resource-type model` filter so you don't need to include that on `run` select statements). Running `ls` you'll see that it's not getting any nodes:
image

There are a few reasons: 

1. For sources you need to prepend with `source`, like `poetry run dbt ls -s source:gtfs_rt_external_tables+ --resource-type model`.
2. The source name is actually `external_gtfs_rt` (defined [here](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/staging/gtfs/_src_gtfs_rt_external_tables.yml#LL4C11-L4C27)).

So `poetry run dbt ls -s source:external_gtfs_rt+ --resource-type model` will show what I think you were targeting with this exclude statement (and you'd convert it to `--exclude source:external_gtfs_rt+`). However, looking at the output, you'll see that this suppresses a ton of stuff, including things like many of the GTFS quality guideline checks. See above for the longer outcomes of the discussion on this point. 

> Is my cron scheduling syntax correct?

I think that `1-6` should be `1,2,3,4,5,6`, per [this](https://en.wikipedia.org/wiki/Cron#CRON_expression), which is linked from the [Airflow cron docs](https://airflow.apache.org/docs/apache-airflow/1.10.1/scheduler.html):
image


> Is there a best way to test this? as it is mostly a scheduling change

If it's possible to call `dbt ls` in the Airflow DAG tasks for testing and then run that through local Airflow, you could check that it's grabbing what you expect.

mjumbewu:
@charlie-costanzo and I had a discussion about this, and (unfortunately without the _complete_ context from before) it seems like a good enough compromise is to modify the new Sunday dag to do the following:

1. Run a regular refresh of only `gtfs_rt_external_tables+` (i.e., everything that is a direct descendent of the RT external tables)
2. Run a full refresh of everything excluding `gtfs_rt_external_tables+`

We _think_ this is basically in line with the recommendations/reconsiderations in [Laurie's comment](https://github.com/cal-itp/data-infra/pull/2626#issuecomment-1558022761), except making the simplifying assumption that _all_ RT data is expensive to do a full refresh on, or at least doesn't need to be trued-up as frequently as the static data.

airflow/dags/transform_warehouse_sunday_select_rt/METADATA.yml
```diff
@@ -0,0 +1,18 @@
...
+schedule_interval: "0 12 * * 0"
+tags:
+  - all_gusty_features
+default_args:
+    owner: airflow
+    depends_on_past: False
+    start_date: !days_ago 1
+    email:
+      - "andrew.v@jarv.us"
+      - "jameelah.y@jarv.us"
```
mjumbewu, line 11: **thought (non-blocking):** I wonder whether there are any Caltrans addresses we can start changing these to. Maybe a conversation for the next 12L check-in?

mjumbewu (commented):
Just a few comments, the most substantial asking whether the two new DAGs should be one.

charlie-costanzo:
to do:
* make sure we change the select/exclude statements to use new RT source tables
  * also use `source:`
* Also exclude old RT source tables in exclude statements
* manage order of DAG runs through dependencies

mjumbewu:
Also @charlie-costanzo, be sure to read through Laurie's comment for the testing tips as well. I think the cron syntax is fine though: https://crontab.guru/ is my go-to checker.

airflow/dags/transform_warehouse_full_refresh/METADATA.yml
```diff
@@ -1,5 +1,5 @@
 description: "Runs dbt with --full-refresh; use this to refresh/rebuild/backfill incremental models."
-schedule_interval: null
+schedule_interval: Null
```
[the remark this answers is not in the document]
charlie-costanzo: hm yeah this was in the original commits from ~1 year ago, not sure why. just changed it back

airflow/dags/transform_warehouse_sunday_select_rt/METADATA.yml
charlie-costanzo: I just changed this to Hunter and Evan, which is how some of our other, newer dags are configured

airflow/dags/transform_warehouse_full_refresh_sunday_exclude_rt/dbt_run_and_upload_artifacts.yml
```diff
@@ -0,0 +1,77 @@
...
+  - 'run'
+  - '--full-refresh'
+  - '--dbt-docs'
+  - '--save-artifacts'
+  - '--deploy-docs'
+  - '--sync-metabase'
+  - '--select'
+  - "{{ dag_run.conf.get('dbt_select_statement', '') }}"
+  - '--exclude'
+  - 'gtfs_rt_external_tables+'
```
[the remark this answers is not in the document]
charlie-costanzo: changed

charlie-costanzo:
> Also @charlie-costanzo, be sure to read through Laurie's comment for the testing tips as well. I think the cron syntax is fine though: https://crontab.guru/ is my go-to checker.

I just did testing with Laurie's recommended `ls` command on the command line and everything works as expected. I didn't run the `dbt ls` from within local airflow though, let me know if you think I should figure out a way to do that.

airflow/dags/transform_warehouse_full_refresh_sunday/dbt_run_and_upload_artifacts_full_refresh_exclude_rt.yml
```diff
@@ -0,0 +1,77 @@
+operator: 'operators.PodOperator'
+name: 'dbt-run-and-upload-artifacts'
```
mjumbewu, line 2: **suggestion:** I'm not sure whether it's necessary, but it seems that every other `PodOperator` in the repo has the `name` attribute match the file name. Should probably update for the tasks in this DAG.

mjumbewu (approved):
Just one small change suggested below (the `name` attributes on the operators). Other than that, I _do_ suggest spinning this up in a local airflow and replacing the `dbt run`s with `dbt ls`, just to sanity check.

Assuming everything checks out after you do that, then this looks good to me!

PR #2655, opened 2023-05-26
dockerfile update to get dbt running on m1

owades:
# Description
Worked through my open issues with @atvaccaro, got it running with this tiny change.

If you're experiencing this issue too (memray install issues), merge this change and then:

1. `build -t local-dbt -f Dockerfile.local --no-cache .`
2. Delete warehouse/dbt_packages dir
3. run `dbt deps`

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?
I was able to run dbt build

## Post-merge follow-ups
- [x] No action required


atvaccaro (approved):

atvaccaro:
For context, we had to adds these because the docker build on M1 did not pull in wheels for memray, and these libraries are needed to build it from source; we never encountered this on Intel macbooks or in CI

PR #2657, opened 2023-05-29
Updates to support adding gtfs schedule compliance to reports site

### LINKED ISSUE #2656: Add "reports site status" as additional guideline check column
@owades commented on [Thu May 25 2023](https://github.com/cal-itp/reports/issues/258)

**Background**

For the reports site, we are simplifying guideline check reporting slightly, so that the only possible guideline check statuses are:
- Pass
- Fail
- Not applicable/data unavailable

In order to maintain clarity and consistency, we'd like to update existing guideline check reporting so that this simplification logic occurs in the source table. Specifically, let's add a column called **reports_site_status** to `fct_daily_organization_combined_guideline_checks` (or probably the long table upstream of that), with logic roughly as follows:
```
CASE status
    WHEN "PASS" THEN "PASS"
    WHEN "FAIL" THEN "FAIL"
ELSE "NOT APPLICABLE/DATA UNAVAILABLE"
END AS reports_site_status
```
Along with this update, lets update `fct_monthly_reports_site_organization_guideline_checks` to refer to this new status column.

**Acceptance criteria**

- fct_daily_organization_combined_guideline_checks has a new column called "reports_site_status" or similar
- fct_monthly_reports_site_organization_guideline_checks refer to this column, preferably continuing to output the column with the name "status", so that downstream logic won't need to change.

owades:
# Description
Some updates in support of https://github.com/cal-itp/reports/pull/256, mostly based on discussion on that PR.

Namely:
- create `reports_status` column
- create is_manual column in reports support table, based on a new seed file of manual checks

Resolves #2656

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?
Confirmed via spot-checking that is_manual column works as expected.

## Post-merge follow-ups
- [ ] No action required

Update https://github.com/cal-itp/reports/pull/256 to refer to new "reports_status" field


warehouse/models/mart/gtfs_quality/fct_daily_service_combined_guideline_checks.sql
```diff
@@ -17,6 +17,7 @@ fct_daily_service_combined_guideline_checks AS (
         feature,
         check,
         {{ guidelines_aggregation_logic() }} as status,
+        {{ guidelines_aggregation_logic_reports }} as reports_status,
```
lauriemerrell, line 20: I think you need parentheses here, mirroring the call on the line above?

warehouse/models/mart/gtfs_quality/fct_monthly_reports_site_organization_guideline_checks.sql
```diff
@@ -42,13 +46,19 @@ fct_monthly_reports_site_organization_guideline_checks AS (
...
+        checks.reports_status,
+        CASE
+            WHEN guideline_manual_checks.check IS NOT null THEN 1
+            ELSE 0
+        END AS is_manual 
     FROM idx_monthly_reports_site AS idx
     LEFT JOIN generate_biweekly_dates AS dates
         USING (publish_date)
-    LEFT JOIN checks
+    JOIN checks
```
lauriemerrell, line 57: Just curious as to the rationale for changing this from a left to an inner join? We want to drop rows that won't have checks? Either way would prefer to specifically say `INNER JOIN` even though it is redundant but I think we do that for consistency (and maybe SQLFluff yells about it somewhere?)

warehouse/seeds/guideline_manual_checks.csv
lauriemerrell, line 1: Ok this is my bad, I suggested a seed, but in retrospect I don't think this is the best solution because hard-coding these strings is going to be fragile.

Talked to @atvaccaro and he suggested a more robust solution of adding a `manual` column in [`stg_gtfs_quality__intended_checks`](https://dbt-docs.calitp.org/#!/model/model.calitp_warehouse.stg_gtfs_quality__intended_checks) because that's a place we already enumerate all the checks and then you can use that instead of this seed. Does that make sense/seem ok?

lauriemerrell (changes requested):
One substantial change (my bad), one suspected typo, one question.

warehouse/models/mart/gtfs_quality/fct_daily_service_combined_guideline_checks.sql
owades: yes! darnit, I already caught this but I guess I didn't push it

warehouse/models/mart/gtfs_quality/fct_monthly_reports_site_organization_guideline_checks.sql
owades: Yes exactly that - there were many rows with null check values, which I ended up filtering out in python within reports site infra. I figured it was cleaner to filter further upstream - i will leave a comment about that and switch to INNER JOIN

warehouse/seeds/guideline_manual_checks.csv
owades: Yes, this is definitely a better solution!

warehouse/models/mart/gtfs_quality/fct_daily_service_combined_guideline_checks.sql
owades: Weird, this commit did go thorugh yesterday, yet you still saw the old version.  🤷

warehouse/models/mart/gtfs_quality/_mart_gtfs_quality.yml
```diff
@@ -185,6 +185,10 @@ models:
         tests:
           - not_null
       - name: feature
+      - &is_manual
+        name: is_manual
```
lauriemerrell, line 189: Can we add just a short description of what this flag means?

lauriemerrell (commented):
Two more comments

lauriemerrell (approved):

PR #2664, opened 2023-05-30
Switch all get_fs references to data-infra package

SorenSpicknall:
# Description

Related to [this change in calitp-py](https://github.com/cal-itp/calitp-py/pull/105), this updates all outstanding get_fs function imports and references to use the version sourced from calitp-data (rather than calitp-data-infra), and updates the referenced version number for calitp-data-infra throughout. This should not be merged until the new version of calitp-data-infra has been published on PyPI.

## Type of change

- [x] New feature
- [x] Documentation

## How has this been tested?
Reference changes tested ad hoc

## Post-merge follow-ups
- [x] No action required


atvaccaro (changes requested):
Did you manually edit these or update with `poetry add ...`? `poetry add` will also updated the lockfile with the actual versions installed.

Also, you shouldn't need to include `calitp-data` since it's already required by `calitp-data-infra`.

Are you planning to also update and test new versions of the images that are changing?

atvaccaro (changes requested):
Did you manually edit these or update with `poetry add ...`? `poetry add` will also updated the lockfile with the actual versions installed.

Also, you shouldn't need to include `calitp-data` since it's already required by `calitp-data-infra`.

Are you planning to also update and test new versions of the images that are changing?

SorenSpicknall:
> Did you manually edit these or update with `poetry add ...`? `poetry add` will also updated the lockfile with the actual versions installed.
> 
> Also, you shouldn't need to include `calitp-data` since it's already required by `calitp-data-infra`.
> 
> Are you planning to also update and test new versions of the images that are changing?

Now that we've got the new calitp-data-infra package version released, I was able to get all our automated builds updated and passing. Are there any additional steps you'd advise taking before merge?

atvaccaro:
> Now that we've got the new calitp-data-infra package version released, I was able to get all our automated builds updated and passing. Are there any additional steps you'd advise taking before merge?

Ok cool I think that's one step, but can you test the jobs themselves that rely on the imports? Most should have mypy but I'm not sure that everything does.

atvaccaro (approved):
approved pending test archiver deploy

Issue #2628, opened 2023-05-22
User stories: Include organizations.raw_ntd_id column


SLewis1120:
## User story / feature request

_Please describe your need, outlining the key users, the feature being requested, and the goal that that the feature will facilitate. For example: 

As a airtable data analyst, I want `organizations.raw_ntd_id` to be added and ingested by the data warehouse.

### Acceptance Criteria
_Please enter something that can be verified to show that this user story is satisfied. For example: 
1. The columns referenced in this issue should be able to be queried in Metabase.

### Tester [Stakeholder]
1. @evansiroky

Issue #2627, opened 2023-05-22
User Story: Include new Airtable services table columns


SLewis1120:
## User story / feature request

Please describe your need, outlining the key users, the feature being requested, and the goal that that the feature will facilitate. For example: 

As a airtable data analyst, I want the following to be added and ingested by the data warehouse:
1. `services.fixed-route`
2. `services.Is Public`
3. `services.Start Date`
4. `services.Operational Status`
5. `services.Public Currently Operating`
6. `services.Public Currently Operating Fixed Route`
7. `organizations.Public Currently Operating`
8. `organizations.Currently Operating Fixed Route`
9. `organizations.Is Public Entity`

### Acceptance Criteria
Please enter something that can be verified to show that this user story is satisfied. For example:  
1. The columns referenced in this issue should be able to be queried in Metabase.

### Tester [Stakeholder]
1. @evansiroky

Issue #2584, opened 2023-05-11
Determine and Implement Updated Source of Truth for NTD ID


SorenSpicknall:
## User story / feature request

In the process of changing sourcing for NTD ID to the seed file in #2582, we discussed how evolving columns in Airtable [might allow for us to determine a new source](https://cal-itp.slack.com/archives/C02KH3DGZL7/p1683837993314519?thread_ts=1669161731.863779&cid=C02KH3DGZL7) of truth in the future, and possibly even eliminate the seed file. As a follow-up to that PR, we should set in place a new ruleset for bringing NTD ID into warehouse tables based on the current state of Airtable.

### Acceptance Criteria

Anybody interacting regularly with NTD ID (analysts, pipeline maintainers, etc.) should have a shared understanding of where that ID comes from, and how it changes year over year.

PR #2675, opened 2023-06-01
Update services & organizations tables with new columns & new assessment logic

lauriemerrell:
# Description

This PR: 
* Adds the new columns from Airtable requested in #2627 and #2628
* Updates the logic for determining assessed entities according to those changes, per [discussion](https://docs.google.com/document/d/1F4METWYNip5nobcPZSUg-5XGtC1XX1hWMQfEmERPPd4/edit) & [confirmation on Slack](https://cal-itp.slack.com/archives/C040XD9UB33/p1685734771519669?thread_ts=1684873084.739219&cid=C040XD9UB33)
* Also handles #2584 by adding a May 23, 2023 (same date as assessment logic change) cutover date from the NTD data seed to the new Airtable column `organizations.raw_ntd_id`

May 23, 2023 selected as cutover date because that appeared to be the date that the data entry for the new columns was completed.

Resolves:
* #2627 
* #2628 
* #2584 

## Type of change

- [x] New feature
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected): see above links for confirmations that we are ok with the assessment logic changes
- [ ] Documentation

## How has this been tested?
`poetry run dbt run -s stg_transit_database__services+ stg_transit_database__organizations+ --full-refresh`
`poetry run dbt test -s stg_transit_database__services+ stg_transit_database__organizations+`

## Post-merge follow-ups

- [ ] No action required

* Full refresh everything downstream of services & organizations staging tables because we will have version changes based on these new columns


SorenSpicknall (approved):
This looks good to me - we should just make sure as a matter of process that future maintainers of Airtable know now to delete the pre-`raw_ntd_id` column in case we need to backfill.

lauriemerrell:
Thanks Soren!

> This looks good to me - we should just make sure as a matter of process that future maintainers of Airtable know now to delete the pre-raw_ntd_id column in case we need to backfill.

I think the policy right now is not to delete columns in Airtable, but even if we don't, I think that this will always fall back to the seed so changes to the prior data in Airtable won't matter from this point forward. (And the pre-`raw_ntd_id` column is a foreign key to the NTD agency info table in Airtable, so it's not a raw string value, it requires a lookup to populate).

PR #2682, opened 2023-06-02
Source NTD ID source to Airtable Instead of Seed

### LINKED ISSUE #2584: Determine and Implement Updated Source of Truth for NTD ID

SorenSpicknall:
# Description
Following the work performed in #2582, we determined that we wanted to deprecate the NTD information seed and rely on a regularly maintained column in Airtable instead. This PR makes that switch, but at the time of creation it leaves several additional pieces alone - for instance, it does not delete the seed, nor does it get rid of the logic that creates the `unnested_ntd_records` field in `stg_transit_database__organizations`. I'd like advice from other stakeholders about how to proceed with those - perhaps they can be cleaved away in a follow-up PR to keep this one focused on the narrow scope of #2584?

Resolves #2584

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?
The `dim_organizations` model was run from scratch in a staging environment, available to staging warehouse users at `soren_mart_transit_database.dim_organizations`. The total number of rows with `ntd_id` values differs from the production table, but the total number of rows with `ntd_id` values for which `_is_current` is true does not differ. There may be implications for historic data here - I welcome input.

## Post-merge follow-ups
- [ ] No action required

If we take the minimal route on this PR, we will likely need to follow up with some cleanup tasks related to the current seed file and `stg_transit_database__organizations`.

### DIFF
--- a/warehouse/models/intermediate/transit_database/dimensions/int_transit_database__organizations_dim.sql
+++ b/warehouse/models/intermediate/transit_database/dimensions/int_transit_database__organizations_dim.sql
@@ -34,6 +34,7 @@ int_transit_database__organizations_dim AS (
         assessment_status,
         manual_check__contact_on_website,
         hq_county_geography,
+        raw_ntd_id,
         _is_current,
         _valid_from,
         _valid_to
--- a/warehouse/models/mart/transit_database/dim_organizations.sql
+++ b/warehouse/models/mart/transit_database/dim_organizations.sql
@@ -4,18 +4,12 @@ WITH dim AS (
     SELECT * FROM {{ ref('int_transit_database__organizations_dim') }}
 ),
 
-ntd_agency_to_organization AS (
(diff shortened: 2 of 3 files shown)


SorenSpicknall:
Closing because Laurie is replicating these changes on another PR that includes other tweaks.

PR #2689, opened 2023-06-06
handle tab-separated schedule files

atvaccaro:
# Description

Resolves https://github.com/cal-itp/data-infra/issues/2484

Third time's the charm. 😅 

**Question: I'm adding the dialect and number of lines to the outcome metadata; are there any other attributes that might be useful to add?**

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

Tested locally, confirming that a tab-delimited stop_times file changes appropriately.

```json
"fields": [
    "trip_id\tarrival_time\tdeparture_time\tstop_id\tstop_sequence\tstop_headsign\ttimepoint"
  ]
... make change ...
"fields": [
    "trip_id",
    "arrival_time",
    "departure_time",
    "stop_id",
    "stop_sequence",
    "stop_headsign",
    "timepoint"
  ],
"dialect": "excel-tab",
"num_lines": 23617
```

```sql
select parsed_file.dialect, sum(parsed_file.num_lines)
from cal-itp-data-infra-staging.external_gtfs_schedule.stops_txt_parse_outcomes
group by 1
;
```

```bash
env DBT_SOURCE_DATABASE=cal-itp-data-infra-staging poetry run dbt run -m +fct_schedule_feed_files --full-refresh
```

## Post-merge follow-ups

- [ ] No action required
Reprocess the days that we know had TSVs and backfill warehouse models.


lauriemerrell (commented):
Questions about post-merge actions:

1. Will this be the first time we change the outcomes schema (adding `dialect` and `num_lines`) without a full backfill? 
2. Can you add these new fields to the external table schema in this PR? And then would you do a follow-up PR to get them into the staging model? 
3. Did you check whether there are any files with just one column & confirm whether they were handled correctly? Best candidate might be `areas.txt` which only has one required column but I think most cases where `area_name` is null they have a `wkt` field populated that we don't ingest.
4. Related: Any chance it's worth adding a list of column names to the outcomes in this PR, since you're already inspecting the top row? And then perhaps noting that a longer backfill would be good at some point?

atvaccaro:
> * Will this be the first time we change the outcomes schema (adding `dialect` and `num_lines`) without a full backfill?

I think we've added fields before without a backfill; main thing is marking them optional.

> * Can you add these new fields to the external table schema in this PR? And then would you do a follow-up PR to get them into the staging model?

Yes let me do that.

> * Did you check whether there are any files with just one column & confirm whether they were handled correctly? Best candidate might be `areas.txt` which only has one required column but I think most cases where `area_name` is null they have a `wkt` field populated that we don't ingest.

I can write a unit test for this; we should only be using `excel-tab` when we ended up with more than 1 column from splitting on tabs.

> * Related: Any chance it's worth adding a list of column names to the outcomes in this PR, since you're already inspecting the top row? And then perhaps noting that a longer backfill would be good at some point?

Yes, can do.

atvaccaro:
> > * Related: Any chance it's worth adding a list of column names to the outcomes in this PR, since you're already inspecting the top row? And then perhaps noting that a longer backfill would be good at some point?
> 
> Yes, can do.

Actually we already have `fields`, is that what you're referring to?

atvaccaro:
> TIL... This doesn't come into the warehouse anywhere IDT? Ideally these would make it onto `fct_schedule_feed_files` but it doesn't.

It's in the external tables but that's it.

atvaccaro:
known dbt CI failure

atvaccaro:
> 1. What specific days do we need to reprocess? Is that going to lead to the new dialect and line num attributes being available for like 2 random days in April 2023 and then consistently from June?

I think we can backfill from those chosen days forward (schedule is pretty cheap to backfill so it's fine). Maybe we can just pick April 1 forward?

> 2. Backfill which exact warehouse models? I think there was only one affected file in the feed (maybe calendar? I'd have to double check), did you want to only backfill that?

I think we wanted a full refresh at some point, right? Maybe we should just do one.

lauriemerrell:
> I think we wanted a full refresh at some point, right? Maybe we should just do one.

We just full-refreshed `dim_schedule_feeds+` (i.e., all the schedule dims) on 6/2/23. 

Maybe this is an opportunity to test your new date param and try setting that as the GTFS schedule start date and do it again?

PR #2700, opened 2023-06-08
Default to non-publication for new columns on open data tables

### LINKED ISSUE #2651: Open data publishing: Switch default so that columns are not published unless flagged
Per [recent discussion](https://docs.google.com/document/d/1PlVDdt9PtazBfbfWLoZH9UVa0VFq4AMtMTUKboOdL1A/edit#heading=h.xr26c3ku06p9), we would like to make the default that new columns added to published tables be suppressed from open data publishing until we make an affirmative decision to add them. This ensures that all documentation etc. is updated appropriately.

However, the [default behavior currently](https://github.com/cal-itp/data-infra/blob/main/warehouse/scripts/publish.py#L347) is to publish columns, and only ignore columns if an ignore flag is set. This means that every time someone adds a column to a published table, they have to remember to specifically set that ignore flag in the `_latest` tables. This risks us unintentionally publishing new columns. We should swap the behavior so that publishing must be explicitly enabled.

Acceptance criteria:
- [ ] Default behavior in `publish.py` looks for an explicit `meta.publish = True` type flag to include columns in data publishes
- [ ] This new flag is set on all currently-published columns

SorenSpicknall:
# Description
In order to facilitate smoother addition of new columns to tables that are present in open data publishing, this PR defaults to non-publication and adds a mechanism by which we can manually designate columns for publication. This flips the previous logic, which attempted to publish all columns which weren't explicitly excluded via a `publish.ignore` meta entry.

Resolves #2651

## Type of change
- [x] New feature
- [ ] Documentation

## How has this been tested?
Tested for column parity against currently published tables by generating a manifest locally and using that to generate local dictionary and metadata files.

## Post-merge follow-ups
- [ ] No action required

On the original Issue, [Andrew mentioned](https://github.com/cal-itp/data-infra/issues/2651#issuecomment-1572794033) eschewing the `_latest` models altogether and making the qualifying logic for publication into a `where` clause on the models from which our current `_latest` models are derived. I'm generally supportive of that, but would like to tackle it in the next sprint instead of in this PR.

### DIFF
--- a/warehouse/models/mart/gtfs_schedule_latest/_gtfs_schedule_latest.yml
+++ b/warehouse/models/mart/gtfs_schedule_latest/_gtfs_schedule_latest.yml
@@ -18,8 +18,6 @@ models:
         tests:
           - not_null
           - unique
-        meta:
-          publish.ignore: true
       - &schedule_dim_gtfs_key
         name: _gtfs_key
         description:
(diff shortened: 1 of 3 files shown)


lauriemerrell (approved):
LGTM!

PR #2704, opened 2023-06-12
Patch GTFS datasets model for new publication rules

SorenSpicknall:
# Description
The `dim_gtfs_datasets_latest` was inadvertently left out of #2700 because it sits in a different dataset than the other `_latest` models. This setup can likely be revisited when we consider ridding ourselves of the `_latest` models altogether in an upcoming sprint. But for now, this will align the model with the changes we made last week to default to non-publication on new columns.

## Type of change
- [ ] New feature
- [ ] Documentation

## How has this been tested?
Table documentation generated and compared to the most recent version we sent to Chad.

## Post-merge follow-ups
- [ ] No action required

Rerun open data publishing for this week


lauriemerrell (approved):
✅ , sorry I didn't think of this when reviewing the original PR. Definitely another reason to migrate to the `where` based publication logic

PR #2711, opened 2023-06-14
Heavy schedule model/table performance improvements

### LINKED ISSUE #2488: GTFS schedule performance improvements

SorenSpicknall:
# Description
Following up on work from #2593 that's now outdated, this makes changes to build and storage behavior for several GTFS schedule tables to save us significant computing costs in the long run.

Resolves #2488

## Type of change
- [x] New feature
- [ ] Documentation

## How has this been tested?
Benchmarked against current (full refresh) behavior

## Post-merge follow-ups
- [ ] No action required
The models with new columns will need a full refresh after merging


warehouse/models/intermediate/gtfs/int_gtfs_schedule__stop_times_grouped.sql
```diff
@@ -1,9 +1,16 @@
-{{ config(materialized='table') }}
+{{
+    config(
+        materialized='incremental',
+        unique_key = 'key',
```
lauriemerrell, line 4: I am slightly curious as to why these use a key-based strategy as opposed to just date partitions (@atvaccaro?), but I see this is the same as the parent so I assume it's fine.

warehouse/models/intermediate/gtfs/int_gtfs_schedule__stop_times_grouped.sql
```diff
@@ -59,6 +66,7 @@ grouped AS (
         trip_end_timezone,
         iteration_num,
         exact_times,
+        MAX(_feed_valid_from) AS _feed_valid_from,
```
lauriemerrell, line 69: Is there a specific reason you think it's preferable to group this as opposed to including it as part of the grouping? I probably would have just added it to the grouping logic; it won't change the grain and it will be unique with `feed_key` (like, a given `feed_key` necessarily comes with one and only one non-null `_feed_valid_from`)

warehouse/models/mart/gtfs/fct_daily_scheduled_shapes.sql
```diff
@@ -1,4 +1,11 @@
-{{ config(materialized='table') }}
+{{ config(
+    materialized='table',
```
lauriemerrell, line 2: Can I ask why we're partitioning this without making it incremental? Is this for the performance of something downstream?

lauriemerrell (commented):
Blocking question for me is the partitioning of shapes, the rest are optional / opinions

warehouse/models/mart/gtfs/fct_daily_scheduled_shapes.sql
SorenSpicknall: This was one of several automatically surfaced BQ recommendations responsive to query patterns, rather than anything meant to address built efficiency.

warehouse/models/intermediate/gtfs/int_gtfs_schedule__stop_times_grouped.sql
atvaccaro: Originally I wanted to eventually support late-arriving data, and I don't think we'd benefit from using `insert_overwrite` so might as well stick with key.

warehouse/models/intermediate/gtfs/int_gtfs_schedule__stop_times_grouped.sql
SorenSpicknall: Fixed in [fb59602](https://github.com/cal-itp/data-infra/pull/2711/commits/fb596023e874459853c5f2b6bfdef501267cdd0c) - I wrote that before fully understanding the function of `_feed_valid_from`.

Issue #2670, opened 2023-06-01
Set up doit-based deploy CI for one application


atvaccaro:
One of our follow-ups from our infra handoff meeting is some initial deploy CI using something like https://github.com/pydoit/doit and Google Secret Manager as our basic tools for automating deploys. @lottspot recommended just fetching secrets in Actions rather than something like sops or sealed secrets, and a build tool like doit can streamline that pattern.

Some possible options:
* Archiver (kustomize)
* Metabase (helm)
* JupyterHub (helm)

atvaccaro:
https://github.com/pyinvoke/invoke is another interesting option

PR #2736, opened 2023-06-20
migrate CI to use python invoke

atvaccaro:
# Description

Resolves https://github.com/cal-itp/data-infra/issues/2670

I'm finding myself having to resolve conflicts between candidate and releases branches occasionally, but I haven't dug into the git commands yet (that's next, at minimum just to understand them).

~Marking as do-not-merge while I update the docs.~ I'm also working on an architecture diagram, but that doesn't block this.

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

I've deployed to test.

## Post-merge follow-ups

- [ ] No action required
Deploy to prod.


ci/invoke.yaml
```diff
@@ -0,0 +1,83 @@
...
+          namespace: metabase-test
+          helm_name: metabase-test
+          helm_chart: kubernetes/apps/charts/metabase
+          helm_values: kubernetes/apps/values/metabase-test.yaml
+        - name: postgresql-backup-metabase
+          driver: helm
+          namespace: metabase-test
+          helm_name: postgresql-backup
+          helm_chart: kubernetes/apps/charts/postgresql-backup
+          helm_values: kubernetes/apps/values/postgresql-backup-metabase.yaml:kubernetes/apps/values/postgresql-backup-metabase-test.yaml
```
lottspot, line 23: Would it be nicer to use yaml list syntax for multiple values files?

ci/tasks.py
```diff
@@ -0,0 +1,196 @@
...
+                full_diff += result.stdout
+    c.update({"hdiff": full_diff})
+    if outfile:
+        msg = f"```{full_diff}```" if full_diff else "No helm changes found.\n"
+        print(f"writing {len(msg)=} to {outfile}", flush=True)
+        with open(outfile, "w") as f:
+            f.write(msg)
+
+
+# TODO: we may want to split up channels into separate files so channel is not an argument but a config file
```
lottspot, line 159: +1

ci/tasks.py
```diff
@@ -0,0 +1,196 @@
...
+    c.update({"kdiff": full_diff})
+    if outfile:
+        msg = f"```{full_diff}```" if full_diff else "No kustomize changes found.\n"
+        print(f"writing {len(msg)=} to {outfile}", flush=True)
+        with open(outfile, "w") as f:
+            f.write(msg)
+
+
+@task(parse_calitp_config)
+def hdiff(
```
lottspot: Isn't dealing with helm so much fun 😄

.github/workflows/service-release-diff.yml
```diff
@@ -0,0 +1,68 @@
...
+        name: Run poetry invoke
+        shell: bash
+        working-directory: ci
+        run: |
+          # GITHUB_JOB is not populated until the job is running
+          git config user.name "Github Action $GITHUB_JOB"
+          git config user.email "$(whoami)@$(uname -n)"
+          export RELEASE_CHANNEL=${GITHUB_BASE_REF#releases/}
+          printf 'WORKFLOW: service-release; RELEASE_CHANNEL=%s\n' "$RELEASE_CHANNEL"
+          set -- "$GITHUB_WORKSPACE/ci/vars/project.env"
```
lottspot, line 51: shouldn't need this anymore now that `*.env` inputs are going away

.github/workflows/service-release-diff.yml
```diff
@@ -0,0 +1,68 @@
...
+          poetry install
+          poetry run invoke kdiff $RELEASE_CHANNEL --outfile=kdiff.txt
+          poetry run invoke hdiff $RELEASE_CHANNEL --outfile=hdiff.txt
+          cat kdiff.txt hdiff.txt > diff.txt
```
lottspot: I like this method of diff reporting 👏

.github/workflows/service-release-diff.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      # Diff and write back to PR
+      - id: diff
+        name: Run poetry invoke
+        shell: bash
+        working-directory: ci
+        run: |
+          # GITHUB_JOB is not populated until the job is running
+          git config user.name "Github Action $GITHUB_JOB"
+          git config user.email "$(whoami)@$(uname -n)"
```
lottspot, line 48: A git identity is probably not necessary since it doesn't look like you are authoring any commits or tags in this iteration of CI

lottspot (approved):
This is a great leap forward!

ci/invoke.yaml
atvaccaro: yep, just pushed a commit with that!

ci/tasks.py
atvaccaro: I'll do that; I'd gone back-and-forth on everything in `invoke.yaml` vs providing configs, and I'm realizing that repetitive arguments in task definitions are kinda annoying. [pre/post-task data passing](https://github.com/pyinvoke/invoke/issues/261) is an open issue, so I think relying on config + context is the best pattern for now anyways.

ci/tasks.py
atvaccaro: luckily I get to just copy-paste what you've already figured out 🤣

docs/infrastructure/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+[invoke](https://docs.pyinvoke.org/en/stable/) is a Python framework for executing subprocess and building a CLI application.
+The tasks are defined in `tasks.py` and configuration in `invoke.yaml`; config values under the top-level `calitp`
+are specific to our defined tasks.
+
+Run `poetry run invoke -l` to list the available commands, and `poetry run invoke -h <command>` to get more detailed help for each individual command.
+
+## CI/CD
+
+All CI/CD automation in this project is executed via GitHub Actions, whose workflow files .
```
lauriemerrell, line 13: sentence missing end?

ci/tasks.py
```diff
@@ -0,0 +1,259 @@
...
+    namespace: Optional[str]
+    helm_name: Optional[str]
+    helm_chart: Optional[Path]
+    helm_values: List[Path] = []
+
+    # for kustomize
+    kustomize_dir: Optional[Path]
+
+
+# TODO: rename this
```
lauriemerrell, line 50: TODO here? or later?

ci/tasks.py
```diff
@@ -0,0 +1,259 @@
...
+    ret = []
+    for release in c.calitp_config.releases:
+        if (not driver or release.driver == driver) and (
+            not app or app == release.name
+        ):
+            ret.append(release)
+    return ret
+
+
+# TODO: kubectl diff now supports masking secrets, so we shoulds be able to render secret diffs in PRs
```
lauriemerrell, line 92: do you want a follow up ticket for this TODO? also typo `shoulds`

ci/tasks.py
```diff
@@ -0,0 +1,259 @@
...
+                    name = f"projects/1005246706141/secrets/{release_secret}/versions/latest"
+                    secret_contents = client.access_secret_version(
+                        request={"name": name}
+                    ).payload.data.decode("UTF-8")
+
+                    if release.namespace:
+                        ns_str = f"--namespace {release.namespace}"
+                    else:
+                        ns_str = ""
+                        assert (
```
lauriemerrell, line 128: not sure if this is off base, but this kind of `assert` with a yaml safe load makes me think that something here could be turned into a pydantic class instead?

ci/tasks.py
```diff
@@ -0,0 +1,259 @@
...
+    Applies kubectl diff to manifests for a given channel.
+    """
+    assert c.calitp_config.git_root is not None
+    actual_driver = ReleaseDriver[driver] if driver else None
+    full_diff = ""
+    result: Result
+
+    for release in get_releases(c, driver=actual_driver, app=app):
+        if release.driver == ReleaseDriver.kustomize:
+            assert release.kustomize_dir is not None
```
lauriemerrell, line 174: similarly, something here seems a bit weird re: having asserts on things that you defined above... should these attributes just be required in the relevant pydantic classes?

ci/tasks.py
```diff
@@ -0,0 +1,259 @@
...
+    )
+    if outfile:
+        print(f"writing {len(msg)=} to {outfile}", flush=True)
+        with open(outfile, "w") as f:
+            f.write(msg)
+    else:
+        print(msg, flush=True)
+
+
+# TODO: we may want to split up channels into separate files so channel is not an argument but a config file
```
lauriemerrell, line 218: in comment you said you'll do this but comment is still here, should this be made into a ticket?

ci/tasks.py
```diff
@@ -0,0 +1,259 @@
...
+):
+    """
+    Releases (i.e. deploys) apps into a specific channel.
+    """
+    assert c.calitp_config.git_root is not None
+    actual_driver = ReleaseDriver[driver] if driver else None
+
+    for release in get_releases(c, driver=actual_driver, app=app):
+        if release.driver == ReleaseDriver.kustomize:
+            assert release.kustomize_dir is not None
```
lauriemerrell, line 233: just noting that this exact pattern (to build `kustomize_dir`) appears twice right? so yeah wondering if that can be simplified. and same for `chart_path`

docs/infrastructure/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## deploy-airflow.yml
+
+While we're using GCP Composer, "deployment" of Airflow consists of two parts:
+
+1. Calling `gcloud composer environments update ...` to update the Composer environment with new (or specific versions of) packages
+2. Copying the `dags` and `plugins` folders to a GCS bucket that Composer reads
```
lauriemerrell, line 20: any guidance on how to identify the correct GCS bucket?

lauriemerrell (approved):
Pretty superficial comments, feel free to take or leave it. Except a few typos that should just be cleaned up

ci/tasks.py
atvaccaro: Not necessarily in this case; a Secret without a namespace is totally valid, but in our specific case it's fine if it is specified in the Release

ci/tasks.py
atvaccaro: An alternative would be to use a [discriminated union](https://docs.pydantic.dev/dev-v2/usage/types/unions/#discriminated-unions-aka-tagged-unions) but I think it's overkill here; most of these asserts are primarily to make mypy happy. I think if we were to implemented a third driver (e.g. just vanilla `kubectl`) then it'd be worth creating a union of 3 types.

ci/tasks.py
atvaccaro: oh yep I did this, will remove

lauriemerrell (approved):

PR #2743, opened 2023-06-21
analyst ad-hoc: `fct_daily_scheduled_shapes_monthly_aggregations`

### LINKED ISSUE #2382: User Story: monthly aggregation for mart_gtfs.fct_daily_scheduled_shapes

tiffanychu90:
# Description

Build off of @charlie-costanzo's [branch](https://github.com/cal-itp/data-infra/tree/analyst-ad-hoc-tables) in aggregating shapes to the month. 
Test this with just a couple days before expanding time horizon.

Resolves #2382 

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?
fct_monthly_routes

## Post-merge follow-ups

- [ ] No action required


tiffanychu90:
@lauriemerrell: I tested this on a very small band of dates, but my main question is...are you worried about the logic in setting up these CTEs....is it running too many large tables? Not sure where I should go from my first attempt.

lauriemerrell (commented):
Thank you for tackling this and sorry for any confusion caused by picking up the in-progress work, which I am now realizing may have set us on a slightly-unhelpful path. 

I have a few suggestions here; sorry in advance for long comment 😅 but figured this was an opportunity to describe how I would approach this feature, in case that's helpful. I'd be very happy to meet for 30 minutes and talk through this if that would be useful. 

**To the specific question about performance / querying large tables:** The main way that we manage the need to read in large tables is by using [incremental tables](https://docs.getdbt.com/docs/build/materializations#incremental). That way each time the table is built, only new data is read in. I think you could calculate the last day of the given month and then use that as the date column for our `incremental_where` macro, similar to the work that Soren did in this PR: https://github.com/cal-itp/data-infra/pull/2711/. I assume we don't want to add a given month until it's over, so you could also include a filter like we do for the reports site ([here](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/gtfs_quality/idx_monthly_reports_site.sql#L159)).

Overall conceptual comments:
* It seems like the intended grain for this table is actually more of a `fct_monthly_routes`, because the goal is that there would be one row per route per month, and the shape is basically an attribute of the route for that month. If there were other aggregations we wanted by route by month (for example total trips), it seems like they could go on this table too. So I would consider renaming this table in that direction and kind of reconceptualizing the approach around that perspective.
* Similarly, I think it would be fine for this table to just live in `mart_gtfs`. For tables that we intend to have people use across multiple analyses (with no specific limitations), it seems appropriate for them to live in a non-ad hoc location. 

Implementation comments:
Based on the first comment above, it might be helpful to think about the target grain as route per month and then kind of work backwards from there. With that in mind, the way I would approach this generally is:

* The final key for this row will actually be `base64_url` + `month/year` + `route_id`. The other data is attached/associated with this but it's not part of the key. If we want foreign keys to other tables, we need a way to pick the right version (for example, `shape_key` could be the most recent version within the relevant month of the most common `shape_id`, so you would need a way to identify "latest version" of the shape within the month).
*  So, starting with `fct_scheduled_trips`, you can get the route/month/year/shape grain, and pick the most common shape id by month. (This would be a groupby and then a window function, similar to what is in here now.)
*  You can then group a second time, **dropping shape id from the groupby** to get the desired month/route grain. You would then join in the correct shape id as an attribute on that route grain.
* Separately, there is the decision on how to join that shape id to get its point array and pick what version of the shape you're using. I think that could be accomplished either by using `dim_shapes_arrays` and `dim_schedule_feeds` to get the latest version of a shape id per month, or by using `fct_daily_scheduled_shapes` to get that latest version. So joining on the pt array is pretty separate from main aggregation/grain logic, and is treated as more of just a "lookup". 

I hope that makes sense... Again, happy to meet to discuss.

tiffanychu90:
@lauriemerrell The comment is really helpful for me. I'll work on those changes and start tackling it with the new approach

tiffanychu90:
@lauriemerrell: made some changes and followed your conceptual approach, and now it's moved to `mart_gtfs.fct_monthly_shapes`. I have a couple of items I need help with:
1. incremental table -- do I need to create a key, what should it be clustered by (year/month or `source_record_id`)?
2. how to subset the service dates from `fct_scheduled_service` for only dates where a full month has elapsed
3.

warehouse/models/mart/gtfs/fct_monthly_shapes.sql
```diff
@@ -0,0 +1,70 @@
...
+
+trips_by_shape AS (
+
+    SELECT
+        base64_url,
+        route_id,
+        shape_id,
+        EXTRACT(month FROM service_date) AS month,
+        EXTRACT(year FROM service_date) AS year,
+        COUNT(*) AS n_trips,
```
lauriemerrell, line 38: I would use something like this: https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/gtfs_quality/idx_monthly_reports_site.sql#L15 to get the last date of the given month as well (`LAST_DAY(service_date, MONTH) AS month_last_day`), so you can use that individual date for the join with the shapes and use that to filter out incomplete months

warehouse/models/mart/gtfs/fct_monthly_shapes.sql
```diff
@@ -0,0 +1,70 @@
...
+        shapes.key AS shape_array_key,
+        shapes.pt_array
+
+    FROM trips_by_shape AS trips
+    INNER JOIN dim_gtfs_datasets
+        ON trips.base64_url = dim_gtfs_datasets.base64_url
+    INNER JOIN dim_shapes_arrays AS shapes
+        ON trips.base64_url = shapes.base64_url
+        AND trips.shape_id = shapes.shape_id
+)
```
lauriemerrell, line 68: Here you can add `WHERE month_last_day <= CURRENT_DATE()` if you calculate the month last date as described on line 38

warehouse/models/mart/gtfs/fct_monthly_shapes.sql
```diff
@@ -0,0 +1,70 @@
...
+        trips.shape_id,
+        trips.month,
+        trips.year,
+
+        shapes.key AS shape_array_key,
+        shapes.pt_array
+
+    FROM trips_by_shape AS trips
+    INNER JOIN dim_gtfs_datasets
+        ON trips.base64_url = dim_gtfs_datasets.base64_url
```
lauriemerrell, line 64: This is going to result in fanout because `dim_gtfs_datasets` is versioned (i.e., there are a lot of rows with the same `base64_url`: one row per dataset record version). See line 41 for an alternate suggestion

warehouse/models/mart/gtfs/fct_monthly_shapes.sql
```diff
@@ -0,0 +1,70 @@
...
+    INNER JOIN dim_shapes_arrays AS shapes
+        ON trips.base64_url = shapes.base64_url
+        AND trips.shape_id = shapes.shape_id
```
lauriemerrell, line 67: If you add `month_last_day` as described on line 38, that can be used as the date for this join (can just add another join condition: `AND month_last_day BETWEEN shapes._valid_from AND shapes._valid_to`). Because `dim_shapes_array` is versioned, this join as written would result in fanout; there are multiple rows in `dim_shapes_array` with the same `base64_url` and `shape_id` values (one for each version of that record). Note that if there is a case where the most common shape for the route was deleted and not active at the end of the month, this would be null. Would want to check if/how often that happens.

warehouse/models/mart/gtfs/fct_monthly_shapes.sql
```diff
@@ -0,0 +1,70 @@
...
+
+    SELECT
+        key,
+        base64_url,
+        shape_id,
+        pt_array
+
+    FROM {{ ref('dim_shapes_arrays') }}
+    WHERE {{ incremental_where(default_start_var='GTFS_SCHEDULE_START',
+                               this_dt_column='_feed_valid_from',
```
lauriemerrell, line 24: `this_dt_column` needs to refer to a column that is present on the final table being defined here... Could potentially use `month_last_day`

warehouse/models/mart/gtfs/_mart_gtfs_fcts.yml
```diff
@@ -2098,3 +2098,20 @@ models:
...
+      An aggregation of GTFS schedule shapes by month, filtered to the shape_id associated with each route_id with the most trips.
+    columns:
+      - name: source_record_id
+      - name: name
+      - name: base64_url
+      - name: route_id
+      - name: shape_id
+      - name: month
+        description: |
+          Actual calendar month (Pacific Time dates) in which this service was scheduled to occur.
```
lauriemerrell, line 2112: If you are using `service_date`, then this is not necessarily an actual Pacific Time calendar month, this is the dates within that month as defined by the agency's definition of a service date. I still think `service_date` is probably the best way to do this, but would just update this to say `service month`

lauriemerrell (commented):
> incremental table -- do I need to create a key, what should it be clustered by (year/month or source_record_id)?

The keying, incremental materialization, and clustering are actually a bit separate. 

Keying just means figuring out what the unique key for an individual row would be.  If the table is called `fct_monthly_shapes`, that implies the grain is one row per shape (per feed) per month, so the key should identify one shape/feed/month, so it seems like the key would be `base64_url, month + year (or month_last_day, see below), and shape_id`. Then ideally that should be tested for uniqueness in the yaml. 

Incremental materializations just mean only adding/updating a certain subset of rows instead of reloading the entire table every day. This can be based on the keys for an individual row (where you use the CTEs to define the set of rows that need to be updated) or for BigQuery we can [use the `insert_overwrite` strategy to update an entire partition at a time](https://docs.getdbt.com/reference/resource-configs/bigquery-configs#the-insert_overwrite-strategy). For a monthly table, it might make sense to use an `insert_overwrite` strategy where you update a whole month at a time, rather than a row-based update based on a row's unique key. In that case, you'd probably add the incremental where clause on `fct_scheduled_trips` and filter just to trips where the service date is in a month that:
1. Isn't already present in the current table
2. Is over

Clustering is a way to make queries later more efficient, so you'd want to cluster by whatever the query pattern is most likely to be. Will people query by a full month batch at a time, or by URL? You could potentially leave this for later if you're not sure. If we use the `insert_overwrite` strategy, we'd have to partition this table, so we may not need to cluster it in addition.

I'm also not entirely sure that this needs an incremental materialization... Once you remove the July-only date filters, it's worth just checking how much data this actually reads. If it doesn't read more than 50 GB, it probably can be left as just a regular table.

Another, non-incremental, way to make this more performant would be by using `fct_daily_schedule_feeds` to get the `feed_key` of the last day of the month for a given shape and then including that `feed_key` in the join with `dim_shapes_array` because `dim_shapes_array` is clustered on feed key so joins that use that column will be more efficient.

> how to subset the service dates from fct_scheduled_service for only dates where a full month has elapsed

Added some inline comments, I'd basically just calculate the last day of every month and check that the last day of a month is before today.

tiffanychu90:
@lauriemerrell: 
> Keying just means figuring out what the unique key for an individual row would be.  If the table is called `fct_monthly_shapes`, that implies the grain is one row per shape (per feed) per month, so the key should identify one shape/feed/month, so it seems like the key would be `base64_url, month + year (or month_last_day, see below), and shape_id`. Then ideally that should be tested for uniqueness in the yaml. 

Maybe this should be called `fct_monthly_route_shapes` or `fct_monthly_routes`? I added the `primary_key_tests` for operator/route. I'd want the table to show me the shape to use for that route, but this would probably fail a shape uniqueness test.

> I'm also not entirely sure that this needs an incremental materialization... Once you remove the July-only date filters, it's worth just checking how much data this actually reads. If it doesn't read more than 50 GB, it probably can be left as just a regular table.

For Jun/Jul, 2 months worth, this is print out from `poetry run dbt run` `20:09:01  1 of 1 OK created sql table model tiffany_mart_gtfs.fct_monthly_shapes ......... [CREATE TABLE (7.6k rows, 57.2 GB processed) in 23.16s]` . The final storage is
fct_monthly_shapes_storage

Is "reading more than 50 GB" based on the output table....and in this case, it's small enough to let it be a table?

> Another, non-incremental, way to make this more performant would be by using `fct_daily_schedule_feeds` to get the `feed_key` of the last day of the month for a given shape and then including that `feed_key` in the join with `dim_shapes_array` because `dim_shapes_array` is clustered on feed key so joins that use that column will be more efficient. 

I couldn't find the `_valid_to` column, just `_feed_valid_from` in `dim_shapes_arrays`, so I [did this](https://github.com/cal-itp/data-infra/blob/3781db5488fdec613c864f75219202f7c61e6ab7/warehouse/models/mart/gtfs/fct_monthly_shapes.sql#L28-L32)...which I'm guessing is not performant, so wanting to call this out specifically for feedback

warehouse/models/mart/gtfs/fct_monthly_shapes.sql
```diff
@@ -0,0 +1,82 @@
...
+    SELECT *
+    FROM {{ ref('dim_gtfs_datasets') }}
+),
+
+fct_scheduled_trips AS (
+    SELECT *
+    FROM {{ ref('fct_scheduled_trips') }}
+    WHERE service_date >= '2023-06-01'
+),
+
```
lauriemerrell, line 13: I'd add `fct_daily_schedule_feeds` here to help with the join with dim shapes array below

warehouse/models/mart/gtfs/fct_monthly_shapes.sql
lauriemerrell: If you first join with `fct_daily_schedule_feeds`, you can use `AND feeds.feed_key = shapes.feed_key` and that will make the join more performant

warehouse/models/mart/gtfs/fct_monthly_shapes.sql
```diff
@@ -0,0 +1,82 @@
...
+dim_shapes_arrays AS (
+
+    SELECT
+        base64_url,
+        shape_id,
+        pt_array,
+        _feed_valid_from
+
```
lauriemerrell, line 21: would need to add `feed_key` here to use that in the join

lauriemerrell (commented):
> Maybe this should be called fct_monthly_route_shapes or fct_monthly_routes? I added the primary_key_tests for operator/route. I'd want the table to show me the shape to use for that route, but this would probably fail a shape uniqueness test.

Yes, I think `fct_monthly_routes` would make sense! And then for example if you later want a count of overall trips by route by month, that would be added to this table. Anything that is summarized at the route/month level could be added. 

> Is "reading more than 50 GB" based on the output table....and in this case, it's small enough to let it be a table?

Ah yeah sorry I forgot about all the things it shows -- the 50 GB would be based on the bytes billed in the dbt output, so I think this is probably too big. 

That being said, I am not sure that making this incremental is the most important performance improvement; I think that the join with `dim_shapes_arrays` is the most expensive part, and I'm not sure that can be solved just by loading it incrementally.  I would use `fct_daily_schedule_feeds` to get a feed key to use in the join (that would also address the issue of no `_feed_valid_to` being present on the shapes table). I tested that locally and it was 70 GB to build for all of time (dropping the 2 month filter). I think that's probably fine for now.

tiffanychu90:
@lauriemerrell Ready for review....hoping I did the selecting month end's `feed_key` and joining correctly. The last thing that I haven't done is how to update the table with new rows for a new month with `insert_overwrite`. I could see this partitioned by month as a use case, but I also want to leave this open-ended for now until we have more use cases.

lauriemerrell (commented):
🎉 I think this looks good to me! Not approving just because you had questions, happy to look one final time.

>The last thing that I haven't done is how to update the table with new rows for a new month with insert_overwrite. I could see this partitioned by month as a use case, but I also want to leave this open-ended for now until we have more use cases.

Sorry if I was confusing before, `insert_overwrite` is only for incremental tables. Since you removed the incremental materialization (correctly, per our prior convo), I think you're ok! 

A few small caveats (optional changes, up to you on whether to make them; if not I might update the documentation either in the YAML or in SQL comments just to record them):
* I think that by inner joining with `dim_shapes_arrays`, you will drop routes that don't have a shape defined. Since shapes are technically optional, this might drop valid routes that had scheduled activity in that month. You could consider using a left join there and then filtering out routes without a shape in downstream use cases. 
* I think that by inner joining with feeds on the last day of the month, it is technically possible that you could get a weird situation where a route used one shape for most of the month and then that shape was deleted and was no longer present in the final feed of the month, so that could be another source of possibly dropping rows.  This shouldn't come up too often (I don't think agencies delete shapes very often) but I think it is theoretically possible.

tiffanychu90:
@lauriemerrell All that makes sense, so I made all those changes!

lauriemerrell (approved):
🎉 Cool cool! The corresponding caveat is now you can technically have a route with no shape, but hopefully that is a little clearer (it would happen in that deletion case; with the new joins the route would still be there but there would be no shape. Again I think that case is rare.)

PR #2772, opened 2023-06-27
add generic loading icon

atvaccaro:
# Description

Resolves #[issue]

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

## Post-merge follow-ups

- [ ] No action required


atvaccaro:
@edasmalchi it looks like https://github.com/ebrelsford/Leaflet.loading/issues/26 prevents us from the cleaner spinner handling we'd prefer; theoretically we could fork leaflet-loading but I'm worried about maintainability.

atvaccaro:
I did just add event fires, but it's messy because of the baselayer's own events.

apps/maps/src/routes/+page.svelte
```diff
@@ -149,7 +149,7 @@
         return {
           html: `
             <h2 class="has-text-weight-bold has-text-teal-bold">
-              ${stop_name ?? '(Stop name unavailable)'}
+              ${stop_name ?? 'Non-stop segment'}
```
lauriemerrell, line 152: Just noting that `non-stop` may be confusing like the adjective `nonstop`

lauriemerrell (approved):
One language nit

apps/maps/src/routes/+page.svelte
edasmalchi: Right, though in this case I think either meaning is roughly correct (they're both not stops and the bus does not stop). These segments are created to provide more granular detail when there aren't any stops for long distances. "Stop name unavailable" implies a data issue which isn't the case.

Certainly open to other ideas if you have any though!

lauriemerrell: I guess that makes sense, and no great alternatives come to mind.

PR #2778, opened 2023-07-03
Surface code descriptions in fct_daily_rt_feed_validation_notices

### LINKED ISSUE #2761: Add RT validation code descriptions to the mart
Currently, there are no models in the mart (and therefore accessible in Metabase) that surface the actual descriptions for RT validation notices (as opposed to just the codes). 

There is the staging table `stg_gtfs_quality__rt_validation_code_descriptions` but that is not available in the mart.

Two options are available:

1. Just move `stg_gtfs_quality__rt_validation_code_descriptions` to the mart so it's accessible in Metabase
2. Propagate the `description` column onto `fct_daily_rt_feed_validation_notices` so that the description is available with the notice count by feed

Can do both or just one of these, [waiting for input in Slack on which is more useful](https://cal-itp.slack.com/archives/C02KH3DGZL7/p1687554701794959).

SorenSpicknall:
# Description
Surfaces the `description` field from `stg_gtfs_quality__rt_validation_code_descriptions` in `fct_daily_rt_feed_validation_notices` to make it easier to interpret validation notice codes.

Resolves #2761

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?
The existing `fct_daily_rt_feed_validation_notices` model was recreated from scratch in a staging environment with the new column present, simulating a full refresh/backfill of the table.

## Post-merge follow-ups
- [ ] No action required

### DIFF
--- a/warehouse/models/mart/gtfs_quality/fct_daily_rt_feed_validation_notices.sql
+++ b/warehouse/models/mart/gtfs_quality/fct_daily_rt_feed_validation_notices.sql
@@ -30,6 +30,7 @@ fct_daily_rt_feed_validation_notices AS (
         daily_feeds.date,
         daily_feeds.base64_url,
         codes.code,
+        codes.description,
         codes.is_critical,
         -- TODO: at some point, these codes will be versioned by validator version
         ARRAY_AGG(DISTINCT notices.gtfs_validator_version IGNORE NULLS) AS gtfs_validator_versions,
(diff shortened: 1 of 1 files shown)


warehouse/models/mart/gtfs_quality/_mart_gtfs_quality.yml
```diff
@@ -26,6 +26,9 @@ models:
       - name: code
         tests:
           - not_null
+      - name: description
```
lauriemerrell, line 29: kind of a nitpick but in the spirit of handoff, can we add a description here? I'm trying to be more aggressive about documenting everything at this point.

lauriemerrell (approved):
One nitpick and also I don't think any post-merge actions are required (indicated in PR description)? This is not an incremental table. Otherwise LGTM.

SorenSpicknall:
> One nitpick and also I don't think any post-merge actions are required (indicated in PR description)? This is not an incremental table. Otherwise LGTM.

Yeah, that was just a wrong checkbox situation. No action is needed after merge.

PR #2781, opened 2023-07-05
WIP: Add Transit Data Quality Issues Airtable base to warehouse

### LINKED ISSUE #2724: Download Transit Data Quality Issues Base just in case it's needed later

The Transit Data Quality Team is starting to use a new Airtable Base called `Transit Data Quality Issues`. It is requested that this base be downloaded in raw format if it is not already so that it could be used in metabase at a later time if needed.

### Acceptance Criteria

The Transit Data Quality Issues Base is downloaded nightly like the other bases are.

### Notes

There is some discussion about whether Hubspot should be used exclusively to track transit data quality issues, but for now this Airtable base is being used for the time being.

SorenSpicknall:
# Description
Brings table from the Transit Data Quality Issues Airtable base into BigQuery using much the same process as the two other existing bases (California Transit and Transit Tech Stacks)

Resolves #2724

## Type of change
- [x] New feature
- [x] Documentation

## How has this been tested?
Not yet tested at the time of WIP PR creation

## Post-merge follow-ups
- [x] No action required

### DIFF
--- a/airflow/dags/airtable_loader_v2/transit_data_quality_issues_gtfs_datasets.yml
+++ b/airflow/dags/airtable_loader_v2/transit_data_quality_issues_gtfs_datasets.yml
@@ -1,0 +1,6 @@
+operator: operators.AirtableToGCSOperator
+
+air_base_id: appmBGOFTvsDv4jdJ
+air_table_name: "GTFS Datasets"
+air_base_name: transit_data_quality_issues
+bucket: "{{ env_var('CALITP_BUCKET__AIRTABLE') }}"
--- a/airflow/dags/airtable_loader_v2/transit_data_quality_issues_issue_types.yml
+++ b/airflow/dags/airtable_loader_v2/transit_data_quality_issues_issue_types.yml
@@ -1,0 +1,6 @@
+operator: operators.AirtableToGCSOperator
+
+air_base_id: appmBGOFTvsDv4jdJ
+air_table_name: "Issue Types"
+air_base_name: transit_data_quality_issues
+bucket: "{{ env_var('CALITP_BUCKET__AIRTABLE') }}"
--- a/airflow/dags/airtable_loader_v2/transit_data_quality_issues_services.yml
+++ b/airflow/dags/airtable_loader_v2/transit_data_quality_issues_services.yml
@@ -1,0 +1,6 @@
+operator: operators.AirtableToGCSOperator
+
+air_base_id: appmBGOFTvsDv4jdJ
+air_table_name: "Services"
+air_base_name: transit_data_quality_issues
+bucket: "{{ env_var('CALITP_BUCKET__AIRTABLE') }}"
--- a/airflow/dags/airtable_loader_v2/transit_data_quality_issues_transit_data_quality_issues.yml
+++ b/airflow/dags/airtable_loader_v2/transit_data_quality_issues_transit_data_quality_issues.yml
@@ -1,0 +1,6 @@
+operator: operators.AirtableToGCSOperator
+
+air_base_id: appmBGOFTvsDv4jdJ
+air_table_name: "Transit Data Quality Issues"
+air_base_name: transit_data_quality_issues
+bucket: "{{ env_var('CALITP_BUCKET__AIRTABLE') }}"
--- a/airflow/dags/create_external_tables/airtable/external_airtable_transit_data_quality_issues_gtfs_datasets.yml
+++ b/airflow/dags/create_external_tables/airtable/external_airtable_transit_data_quality_issues_gtfs_datasets.yml
@@ -1,0 +1,15 @@
+operator: operators.ExternalTable
+bucket: gs://calitp-airtable
+prefix_bucket: true
+post_hook: |
+  SELECT *
+  FROM `{{ get_project_id() }}`.external_airtable.transit_data_quality_issues__gtfs_datasets
+  LIMIT 1;
+source_objects:
+  - "transit_data_quality_issues__gtfs_datasets/*.jsonl.gz"
+destination_project_dataset_table: "external_airtable.transit_data_quality_issues__gtfs_datasets"
+source_format: NEWLINE_DELIMITED_JSON
+use_bq_client: true
+hive_options:
+  mode: AUTO
+  source_uri_prefix: "transit_data_quality_issues__gtfs_datasets/"
--- a/airflow/dags/create_external_tables/airtable/external_airtable_transit_data_quality_issues_issue_types.yml
+++ b/airflow/dags/create_external_tables/airtable/external_airtable_transit_data_quality_issues_issue_types.yml
@@ -1,0 +1,15 @@
+operator: operators.ExternalTable
+bucket: gs://calitp-airtable
+prefix_bucket: true
+post_hook: |
+  SELECT *
+  FROM `{{ get_project_id() }}`.external_airtable.transit_data_quality_issues__issue_types
+  LIMIT 1;
+source_objects:
+  - "transit_data_quality_issues__issue_types/*.jsonl.gz"
+destination_project_dataset_table: "external_airtable.transit_data_quality_issues__issue_types"
+source_format: NEWLINE_DELIMITED_JSON
+use_bq_client: true
+hive_options:
+  mode: AUTO
+  source_uri_prefix: "transit_data_quality_issues__issue_types/"
--- a/airflow/dags/create_external_tables/airtable/external_airtable_transit_data_quality_issues_services.yml
+++ b/airflow/dags/create_external_tables/airtable/external_airtable_transit_data_quality_issues_services.yml
@@ -1,0 +1,15 @@
+operator: operators.ExternalTable
+bucket: gs://calitp-airtable
+prefix_bucket: true
+post_hook: |
+  SELECT *
+  FROM `{{ get_project_id() }}`.external_airtable.transit_data_quality_issues__services
+  LIMIT 1;
+source_objects:
+  - "transit_data_quality_issues__services/*.jsonl.gz"
+destination_project_dataset_table: "external_airtable.transit_data_quality_issues__services"
+source_format: NEWLINE_DELIMITED_JSON
+use_bq_client: true
+hive_options:
+  mode: AUTO
+  source_uri_prefix: "transit_data_quality_issues__services/"
--- a/airflow/dags/create_external_tables/airtable/external_airtable_transit_data_quality_issues_transit_data_quality_issues.yml
+++ b/airflow/dags/create_external_tables/airtable/external_airtable_transit_data_quality_issues_transit_data_quality_issues.yml
@@ -1,0 +1,15 @@
+operator: operators.ExternalTable
+bucket: gs://calitp-airtable
+prefix_bucket: true
+post_hook: |
+  SELECT *
+  FROM `{{ get_project_id() }}`.external_airtable.transit_data_quality_issues__transit_data_quality_issues
+  LIMIT 1;
+source_objects:
+  - "transit_data_quality_issues__transit_data_quality_issues/*.jsonl.gz"
+destination_project_dataset_table: "external_airtable.transit_data_quality_issues__transit_data_quality_issues"
+source_format: NEWLINE_DELIMITED_JSON
+use_bq_client: true
+hive_options:
+  mode: AUTO
+  source_uri_prefix: "transit_data_quality_issues__transit_data_quality_issues/"


warehouse/dbt_project.yml
```diff
@@ -92,3 +92,5 @@ models:
         schema: mart_payments
       benefits:
         schema: mart_benefits
+      transit_data_quality_issues:
```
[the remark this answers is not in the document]
lauriemerrell: Or maybe `mart_transit_database`... Since this will have foreign keys to other tables there and `transit database` is kind of just an alias for `Airtable`.

lauriemerrell (changes requested):
I didn't realize this when we originally discussed this request, but this is going to need some significant refactoring to account for the fact that `GTFS datasets` and `services` are **synced tables** in Airtable, meaning that they are just direct copies of the `California Transit.gtfs datasets` and `California Transit.services` tables.

This means we do *not* want to create new intermediate or mart tables for them (it should raise some red flags to have a table with the same name as another table but prefixed with a different domain, like `transit_data_quality_gtfs_datasets`). Instead, what we have to do is map these synced tables to their California Transit base equivalents, because we will want transit data quality issues to point to the relevant record in the canonical `dim_gtfs_datasets` and `dim_services` tables. 

Basically when there is a synced table in Airtable, all the actual data is copied from the source table, but Airtable assigns a new record ID. So when there is a foreign key to `gtfs datasets` from `transit data quality issues`, it's going to reference the synced table record ID rather than the record ID from the source `California Transit.gtfs datasets` table.

We already have handling for cases like this because this issue occurs for the Transit Tech Stacks base too. 

So what needs to happen is:
* Create a mapping table that maps the synced table ID to the canonical source table ID (see: https://github.com/cal-itp/data-infra/blob/main/warehouse/models/staging/transit_database/base/base_tts_services_ct_services_map.sql for example)
* Use that mapping table to replace all references to synced table record IDs with their canonical counterparts (following ex. https://github.com/cal-itp/data-infra/blob/main/warehouse/models/staging/transit_database/base/base_tts_products_idmap.sql)
* Then we don't reference these copied tables any more (i.e., they don't get staging, intermediate, or mart tables) and instead we just reference the existing `dim_services` and `dim_gtfs_datasets` tables. 

Also, it looks like one issue can be linked to multiple datasets/services, so I think we will need bridge tables for those? À la https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/transit_database/bridge_organizations_x_funding_programs.sql.

warehouse/dbt_project.yml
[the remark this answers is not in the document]
SorenSpicknall: Resolved in [d43974a](https://github.com/cal-itp/data-infra/pull/2781/commits/d43974a56418e258bfecc83cbff34011e70b5dbc)

warehouse/models/staging/transit_database/base/_base_transit_database.yml
```diff
@@ -1,6 +1,38 @@
...
+    description: |
+      Mapping between two instances of synced `gtfs_datasets` table in Airtable;
+      one copy in California Transit base and the other in Transit
+      Transit Data Quality Issues base. Different Airtable internal record IDs
+      are assigned in each instance so this table maps the IDs between
+      instances, using `name` (Airtable primary field) to join.
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - tts_key
```
lauriemerrell, line 14: I think that these should be `tdqi_key` and `tdqi_date` -- the prefix is the base abbreviation

warehouse/models/staging/transit_database/_stg_transit_database.yml
```diff
@@ -330,3 +330,45 @@ models:
...
       - name: dt
+  - name: stg_transit_database__issue_types
+    description: TODO
+    columns:
+      - name: id
+      - name: dataset_type
+      - name: transit_data_quality_issues
+      - name: name
+      - name: notes
+      - &dt
```
lauriemerrell, line 341: If `dt` and `ts` appear elsewhere above, can you use these anchors for those too? (I see `dt` on line 332)

warehouse/models/intermediate/transit_database/_int_transit_database.yml
lauriemerrell, line 1: Can these models be moved into the `dimensions` subfolder in this directory and renamed with the others? If you're using those as a template, seems appropriate to keep them all grouped together.

warehouse/models/mart/transit_database/dim_transit_data_quality_issues.sql
```diff
@@ -0,0 +1,41 @@
...
+    SELECT * FROM {{ ref('int_transit_database__transit_data_quality_issues') }}
+),
+
+dim_transit_data_quality_issues AS (
+    SELECT
+        key,
+        source_record_id,
+        description,
+        issue_type,
+        gtfs_dataset,
```
lauriemerrell, line 13: This is where I think we still may need a bridge table, because ideally this would be a `gtfs_dataset_key` value, rather than a `source_record_id` value (I think if you just unnest and keep the outcome, you're going to have the `source_record_id` for the associated GTFS dataset, which can map to multiple actual records during this time period if updates were being made to that foreign record.)

The purpose of the bridge table would be to provide a 1:1 **versioned** relationship for this issue to all associated GTFS dataset records. Even though this is current-only, because it has the fake `_valid` dates, it will join with a lot of versions of the associated records if you just join on `source_record_id`

warehouse/models/mart/transit_database/dim_transit_data_quality_issue_types.sql
```diff
@@ -0,0 +1,21 @@
...
+WITH int_transit_database__issue_types AS (
+    SELECT * FROM {{ ref('int_transit_database__issue_types') }}
+),
+
+dim_transit_data_quality_issue_types AS (
+    SELECT
+        key,
+        source_record_id,
+        dataset_type,
+        transit_data_quality_issue,
```
lauriemerrell, line 12: This looks like it's also a foreign key -- this is the `source_record_id` for an issue. In this case, it seems like this types table is literally just a table that stores a string value (`name`) so what I would probably do here is just stick this on the other table as an issue type. Per the below query it looks like this relationship is always 1:1 (they pick just one type for each issue), so you can probably just do this lookup and not have this as a dedicated table in the mart. (If we did want a mart table, we would want to drop this foreign key and dedupe to just the type names.... I think that this is basically just a way to have a single select field in Airtable though so there's no need for us to maintain this normalization in the mart.)

```
SELECT transit_data_quality_issue, COUNT(source_record_id) FROM `cal-itp-data-infra-staging.soren_mart_transit_database.dim_transit_data_quality_issue_types`
GROUP BY 1 
ORDER BY 2 DESC
```

lauriemerrell (changes requested):
I do think we actually still need bridge tables, there are a few instances where you have `source_record_id` values here as foreign keys which we try not to do in the mart... See inline for more details.

warehouse/models/staging/transit_database/base/_base_transit_database.yml
SorenSpicknall: Resolved in [ce2d2e4](https://github.com/cal-itp/data-infra/pull/2781/commits/ce2d2e4af50b81448381ff68ec3c28a31913732c)

warehouse/models/intermediate/transit_database/_int_transit_database.yml
SorenSpicknall: Resolved in [ce2d2e4](https://github.com/cal-itp/data-infra/pull/2781/commits/ce2d2e4af50b81448381ff68ec3c28a31913732c)

warehouse/models/staging/transit_database/_stg_transit_database.yml
SorenSpicknall: Resolved in [ce2d2e4](https://github.com/cal-itp/data-infra/pull/2781/commits/ce2d2e4af50b81448381ff68ec3c28a31913732c)

warehouse/models/mart/transit_database/dim_transit_data_quality_issue_types.sql
SorenSpicknall: Resolved in [b9b4c3f](https://github.com/cal-itp/data-infra/pull/2781/commits/b9b4c3fbe32bc8b0c4d65d4417a209a017ae62f3)

warehouse/models/mart/transit_database/_mart_transit_database.yml
```diff
@@ -848,3 +848,35 @@ models:
...
+      - name: last_update
+      - name: last_update_month
+      - name: last_update_year
+      - name: status_notes
+      - name: waiting_since
+      - name: outreach_status
+      - name: should_wait_until
+      - *valid_from_placeholder
+      - *valid_to_placeholder
+      - *is_current_placeholder
```
lauriemerrell, line 882: nit: Can we add notes about GTFS dataset key/service key at creation/resolution and add some documentation about what those are

warehouse/models/intermediate/transit_database/dimensions/int_transit_database__transit_data_quality_issues_dim.sql
```diff
@@ -0,0 +1,270 @@
...
+    FROM join_services_at_resolution
+    INNER JOIN (select distinct source_record_id, name, _is_current, _valid_from, _valid_to from dim_issue_types) as distinct_issue_types
+        ON join_services_at_resolution.issue_type_key = distinct_issue_types.source_record_id
+        AND join_services_at_resolution._valid_from < distinct_issue_types._valid_to
+        AND join_services_at_resolution._valid_to > distinct_issue_types._valid_from
+),
+
+int_transit_database__transit_data_quality_issues_dim AS (
+    SELECT
+        {{ dbt_utils.generate_surrogate_key(['source_record_id', '_valid_from', 'gtfs_dataset_key_at_creation', 'gtfs_dataset_key_at_resolution', 'service_key_at_creation', 'service_key_at_resolution', 'issue_type_key']) }} AS key,
```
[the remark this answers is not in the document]
SorenSpicknall: Resolved in [7601230](https://github.com/cal-itp/data-infra/pull/2781/commits/76012305ba1e92c6f846147053e86b6c02548bef)

warehouse/models/intermediate/transit_database/dimensions/int_transit_database__transit_data_quality_issues_dim.sql
```diff
@@ -0,0 +1,270 @@
...
+        waiting_since,
+        outreach_status,
+        should_wait_until,
+        historical._is_current,
+        historical._valid_from,
+        historical._valid_to
+    FROM historical
+    INNER JOIN dim_gtfs_datasets
+        ON historical.gtfs_dataset_key = dim_gtfs_datasets.source_record_id
+        AND historical.issue_creation_time < dim_gtfs_datasets._valid_to
```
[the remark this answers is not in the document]
SorenSpicknall: Resolved in [7601230](https://github.com/cal-itp/data-infra/pull/2781/commits/76012305ba1e92c6f846147053e86b6c02548bef)

warehouse/models/intermediate/transit_database/dimensions/int_transit_database__transit_data_quality_issues_dim.sql
```diff
@@ -0,0 +1,270 @@
...
+        is_open,
+        last_update,
+        last_update_month,
+        last_update_year,
+        status_notes,
+        waiting_since,
+        outreach_status,
+        should_wait_until,
+        historical._is_current,
+        historical._valid_from,
```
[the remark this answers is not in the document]
SorenSpicknall: Resolved in [cdc9c38](https://github.com/cal-itp/data-infra/pull/2781/commits/cdc9c3851fb69d34f8c7bcd3c9d25ccc28e0836f)

warehouse/models/mart/transit_database/_mart_transit_database.yml
SorenSpicknall: Resolved in [bfc7865](https://github.com/cal-itp/data-infra/pull/2781/commits/bfc7865fd236f569abe061fbd6ff35ac82c2d74a)

warehouse/models/intermediate/transit_database/int_transit_database__transit_data_quality_issues.sql
```diff
@@ -0,0 +1,278 @@
...
+        outreach_status,
+        should_wait_until
+    FROM join_services_at_resolution
+    INNER JOIN (select distinct source_record_id, name, _is_current, _valid_from, _valid_to from dim_issue_types) as distinct_issue_types
+        ON join_services_at_resolution.issue_type_key = distinct_issue_types.source_record_id
+),
+
+int_transit_database__transit_data_quality_issues AS (
+    SELECT
+        {{ dbt_utils.generate_surrogate_key(['source_record_id', 'gtfs_dataset_key_at_creation', 'gtfs_dataset_key_at_resolution', 'service_key_at_creation', 'service_key_at_resolution']) }} AS key,
```
lauriemerrell, line 246: I think that it would actually be safer to use `gtfs_dataset.source_record_id` and `service.source_record_id` in the composite key (without the `_key`s) just to confirm that you're only getting one combination here (make sure you didn't accidentally get any fanout in the other joins.... by testing on the versioned keys there's more room to end up with something unexpected)

warehouse/models/mart/transit_database/fct_transit_data_quality_issues.sql
```diff
@@ -0,0 +1,40 @@
...
+WITH int_transit_database__transit_data_quality_issues AS (
+    SELECT * FROM {{ ref('int_transit_database__transit_data_quality_issues') }}
+),
+
+fct_transit_data_quality_issues AS (
+    SELECT
+        key,
+        source_record_id,
+        description,
+        issue_type_key,
```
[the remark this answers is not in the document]
SorenSpicknall: Resolved in [1221aff](https://github.com/cal-itp/data-infra/pull/2781/commits/1221aff4a0fe4dd4f167bce441e42346d5fb406b)

lauriemerrell (approved):

PR #2795, opened 2023-07-12
Restrict qualified columns for publishing in node.select

### LINKED ISSUE #2792: Bug: CKAN publishing script reveals columns without YAML entries in dbt models
**Describe the bug**
When the YAML associated with a folder of dbt models is missing one or more columns from the model SQL, those columns are fed through `publish.py` as if they are set to publish, which contradicts [our current default behavior](https://github.com/cal-itp/data-infra/pull/2700). Additionally, due to weirdness in the logic pathways, when this happens it appears that the resulting tables are published in CKAN with the downloadable filenames `dim_[resource]_latest.csv`, rather than the properly stripped name of `[resource].csv`.

**To Reproduce**
Remove one of the column entries from the YAML associated with a CKAN-published dbt model and then run the publication script (target the staging environment if you actually do this, though, since running it in prod would break our open data).

**Expected behavior**
Even if somebody makes a YAML mistake when creating or updating a model used in open data publication and forgets to include a column in the associated YAML, that column should not publish. Columns should consistently adhere to our current default non-publication behavior, to be overridden only by explicit inclusion in the `california-open-data` exposure. 

**Additional context**
An error resulting ultimately from this bug was reported by a UCB researcher when it produced visible errors in the California open data portal. The surface-level error was patched via [this PR](https://github.com/cal-itp/data-infra/pull/2786) that corrected the YAML of the publishing models.

SorenSpicknall:
# Description
Turns out our publishing issue came from a layered loose definition of columns qualified for publishing within the `NodeModelMixIn` class, outside the publication script itself. This tightens up the column selection to adhere to our current practice of explicit inclusion for publishing, rather than explicit exclusion. This will not immediately impact the columns currently being published, since the initial bug only appeared where there was a mismatch between a model's YAML and its select statement, but it will protect us from future such issues and bring our selection logic within this class in line with the selection logic used by the publication script.

Resolves #2792

## Type of change
- [ ] New feature
- [ ] Documentation

## How has this been tested?
Column selection tested locally

## Post-merge follow-ups
- [x] No action required

### DIFF
--- a/warehouse/scripts/dbt_artifacts/__init__.py
+++ b/warehouse/scripts/dbt_artifacts/__init__.py
@@ -111,9 +111,7 @@ class NodeModelMixin(BaseModel):
         columns = [
             c
             for c in self.sqlalchemy_table(engine).columns
-            if not self.columns
-            or c.name not in self.columns
-            or self.columns[c.name].publish  # type: ignore[attr-defined]
+            if self.columns[c.name].publish  # type: ignore[attr-defined]
         ]
         return select(columns=columns)
 


atvaccaro (approved):
have you tested what happens if no columns are enumerated and/or marked as published?

SorenSpicknall:
> have you tested what happens if no columns are enumerated and/or marked as published?

Noting for posterity that this was discussed internally. Since `publish.py` only passes in tables that have `publish.include: true` on at least one column, this condition can't be triggered by our current use case. However, I am investigating post-merge with intentional violation of those constraints to identify how SQLAlchemy behaves under a no-column SELECT, and will create a follow-up issue to close the logic hole if it exists in order to future-proof functionality.

PR #2808, opened 2023-07-18
Reclassify gcs data as cold storage

### LINKED ISSUE #2773: Move gtfs-data and gtfs-data-test to cold storage
We store a ton of old data in gtfs-data (duplicated in gtfs-data-test) that isn't accessed anymore. We've confirmed we do not plan to backfill the RT data from H1 of 2022, so we can go ahead and move these to a colder storage class and save some storage money.

1. Confirm there is no traffic to these buckets
2. Consider limiting read access to this bucket
3. Change the default storage class
4. Write a script/notebook to convert all the existing objects to a colder storage class

This may replace https://github.com/cal-itp/data-infra/issues/859

charlie-costanzo:
# Description
As found in #2773 - this notebook allows us to reclassify data in gcs as cold storage

Resolves #2773

## Type of change

- [x] New feature

## How has this been tested?
Still needs to be run all the way through - the rt data is too big so it causes Jupyter to crash. Still need to decide if we will run it as an Airflow DAG or if I will add logic to only run what hasn't been run yet.

## Post-merge follow-ups
Once the notebook works we need to run it on both the `gtfs-data` and `gtfs-data-test` buckets


charlie-costanzo:
This work was replaced by more recent cost analysis work

PR #2809, opened 2023-07-18
remove reports check feature filter

owades:
# Description
This is a followup to https://github.com/cal-itp/reports/pull/265. By moving the feature-level filter from the data infra side to the reports side, I will limit future modifications on the data-infra side.

My next step for reports is to add RT Compliance checks to the site - by making this change, I won't have to manually add that as a filter.

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?
I tested this in testing https://github.com/cal-itp/reports/pull/265, by confirming that the feature-level filter continues to return the Schedule Compliance checks. As an extra sanity check, I'll rerun everything after this is merged, to ensure everything looks good before the July reports are created.

## Post-merge follow-ups

Generate sample reports, to double check that no extra checks are listed.


owades:
@lauriemerrell can I get your review of this when you have a chance?

lauriemerrell (approved):

PR #2844, opened 2023-07-26
reports_order for RT compliance to intended checks

owades:
# Description
One change I forgot to make before: adding reports_order for RT checks to intended checks. This is very straightforward. 

Also fixed a typo in a check name.

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?
I spot checked to confirm that there's only one of each number added (from 0 to 15). Do you think I ought to rerun the whole model to test this?

## Post-merge follow-ups

- [x] No action required


owades:
@lauriemerrell I'd love a quick review on this!

lauriemerrell (approved):
This looks good, re: testing you could add a few new tests [here](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/staging/gtfs_quality/_stg_gtfs_quality.yml#L20):

* `not_null` with a `where` clause and the `feature` is any of the features included in the reports site
* `unique_combination_of_columns` for `feature / reports_order` where `reports_order` is not null

And combined those tests would I think guarantee what you want.

owades:
Thanks @lauriemerrell, I added the simpler of the two checks you suggested for now (the other one required some workarounds as it would require using a custom macro in a place where dbt doesn't like macros. i'm happy with just this for now).

PR #2846, opened 2023-07-28
Update guidance to ID GCS buckets for deprecation process

SorenSpicknall:
# Description
Following up on #2845, this tightens up the metrics used to determine which GCS buckets are deprecation candidates during any given deprecation sweep.

## Type of change

- [ ] New feature
- [x] Documentation

## Post-merge follow-ups
- [ ] No action required

Follow the outlined process for a first round of GCS bucket deprecations


atvaccaro (approved):
lgtm! you could also mention a GitHub organization-wide string search, though that wouldn't capture issues like buckets referenced in Composer variables

Issue #1938, opened 2022-10-31
Automatically reload secrets in RT archiver every 5 minutes


atvaccaro:
**Is your feature request related to a problem? Please describe.**
Currently, loading new/changed secrets in the new RT archiver requires restarting the consumer pods.

**Describe the solution you'd like**
The ticker already reloads `gtfs_config` records every 5 minutes; the consumers should do the same with secrets.

**Describe alternatives you've considered**
Technically we could just restart the pods nightly or something but that seems silly.

**Additional context**
https://cal-itp.slack.com/archives/C01FNDG1ZPA/p1667240594043489

PR #2876, opened 2023-08-09
Fetch latest RT secrets every minute

atvaccaro:
# Description

Resolves https://github.com/cal-itp/data-infra/issues/1938
(see https://cal-itp.slack.com/archives/C01FNDG1ZPA/p1691601423249999)

Note: I'm choosing an easier/simpler implementation that uses the ticker to grab and pass auth keys in task calls. This means secrets are serialized into Redis, but it's not a huge deal since we don't persist/backup Redis.

~Also, fetching secrets when we fetch configs does add to the fetch delay every 5 minutes. The alternative would be a separate scheduled function call, which I am going to experiment with. (We could use this same scheduled call to update the configs and reduce delay even more).~

I implemented this; the same scheduler that calls `tick` now loads configs and secrets at the `XX:45` every minute.

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

Testing locally and will deploy to test first.

## Post-merge follow-ups

- [ ] No action required
Deploy to prod.

### DIFF
--- a/services/gtfs-rt-archiver-v3/gtfs_rt_archiver_v3/ticker.py
+++ b/services/gtfs-rt-archiver-v3/gtfs_rt_archiver_v3/ticker.py
@@ -5,7 +5,7 @@ import random
 import time
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import List, Tuple
+from typing import List, Mapping, Tuple
 
 import pendulum
 import schedule  # type: ignore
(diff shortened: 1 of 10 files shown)


services/gtfs-rt-archiver-v3/gtfs_rt_archiver_v3/ticker.py
```diff
@@ -90,6 +98,8 @@ def tick(second: int) -> None:
     schedule.every().minute.at(":00").do(tick, second=0)
     schedule.every().minute.at(":20").do(tick, second=20)
     schedule.every().minute.at(":40").do(tick, second=40)
+    schedule.every().minute.at(":45").do(get_configs)
```
lauriemerrell, line 101: This syntax makes it look like it would happen every minute (as opposed to every 5 minutes) is that wrong?

atvaccaro: Yeah I just made it run every minute since it's not going to affect delay; let me fix the PR name

lauriemerrell (approved):

PR #2880, opened 2023-08-09
Fix dbt ci compile warnings

lauriemerrell:
# Description

This PR addresses a few issues that were identified by warnings in the dbt CI GitHub action:

* Deletes Payments tests that reference v1 payments tables -- @charlie-costanzo can you confirm that the two payments rides tests are ok to delete? The rest were actively warning that they reference nonexistent resources. 
* The YAML for `fct_transit_data_quality_issues` had not been updated with the table's new name, so tests on that table were throwing warnings about a nonexistent model. Once that was fixed, those tests started firing because the model wasn't actually limiting to just the latest rows, it was appending every day so we had a daily copy of mostly-identical data. Resolved by:
   * Adding the latest only macro to get only latest data for that model.
   * Also cleaned up the SQL in that model to make it less verbose.
   * Also cleaned up the types model; it was unnesting inappropriately so that its grain was issue/type combinations instead of being a model of types only.  

Resolves warnings thrown in for example this action run: https://github.com/cal-itp/data-infra/actions/runs/5812794501/job/15758944572?pr=2878

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Verified warnings no longer show on `dbt compile`. Verified that tests now pass on `fct_transit_data_quality_issues` and its upstream models have a more appropriate number of rows. 

## Post-merge follow-ups

- [x] No action required


atvaccaro:
I can look into the failure on `payments_tests_monthly_date_spine` while we're at it

charlie-costanzo:
> can you confirm that the two payments rides tests are ok to delete?

yep! they are ok to delete @lauriemerrell

PR #2893, opened 2023-08-16
Support GTFS validator v4.1.0 in validation tasks

SorenSpicknall:
# Description

Adds a new version reference in our validator wrapper, v4.1.0, beginning 9/1. Introduces a corresponding new environment variable labeling scheme for minor versions, `V[major]_[minor]_VALIDATOR_JAR`, which mirrors actual version labeling except for the use of underscores rather than periods.

Also updates the README to briefly describe JAR version selection in the current validation process.

## Type of change

- [x] New feature
- [x] Documentation

## How has this been tested?

Used new validator version on recent test data, and verified working state on older test data.

## Post-merge follow-ups

- [ ] No action required

Monitor ongoing validations for any unexpected issues from the minor version change.


jobs/gtfs-schedule-validator/gtfs_schedule_validator.py
lauriemerrell, line 1: Can this file be deprecated? Did I miss it in my Airflow updates? cc @atvaccaro

lauriemerrell (commented):
Just one question, not sure we need to be maintaining both files here

jobs/gtfs-schedule-validator/gtfs_schedule_validator.py
atvaccaro: yep good call

SorenSpicknall: Resolved in [a6dc091](https://github.com/cal-itp/data-infra/pull/2893/commits/a6dc091a6f616a63febdc0911e986510dadb6802)

jobs/gtfs-schedule-validator/README.md
```diff
@@ -2,10 +2,24 @@
...
 of the underlying data. The RT code was written prior to substantial changes to
 `calitp-py` and therefore bundles together shared functionality that will eventually
 live in `calitp-py`.
+
+See [this PR](https://github.com/cal-itp/data-infra/pull/2893) for an example of how
+to integrate new versions of the underlying MD-stewarded validator into our validation
+process - specifically, changes to [`gtfs_schedule_validator_hourly.py`](https://github.com/cal-itp/data-infra/pull/2893/files#diff-f970d61c9ce49b3899e7b309610f43fca51d29eaae4d5db192208540bc4da702) and
+[the `Dockerfile`](https://github.com/cal-itp/data-infra/pull/2893/files#diff-de54f054016ca4e44544d4081d2798a21aa43790c3faea3967f6f5b255529f1a), and the
+addition of [a new JAR file](https://github.com/cal-itp/data-infra/pull/2893/files#diff-628dcc44a01286366e42ab8fd440d213d764d8a76f41f268e378a14628e8447c)
+corresponding to the new validator version being referenced (taken from the Assets
```
atvaccaro, line 18: nit: I don't know if the commit-level links are that useful

atvaccaro (approved):
lgtm ty! can you post an announcement saying we'll start validating with the new version starting Sep 1?

jobs/gtfs-schedule-validator/README.md
SorenSpicknall: I understand where you're coming from, but I think I'm gonna leave these as-is. The reason I'm including these file-changes-on-PR-level links is because we wanted to use this PR as an example of a validator upgrade, but the PR also includes some things not directly related to the upgrade (like getting rid of the non-hourly validation script). So to avoid confusion for a reader down the road, I decided to point out these specific changes as the ones relevant to the upgrade.

I also considered just using an older upgrade PR as an example, but the underlying code has changed so much since then (the previous upgrade included the entire implementation of the time-based JAR switch along with many unrelated changes bundled into a big PR) that this seems like the best path to me. Not ideal, but I think it's more demonstrative of the work involved in an upgrade than just gesturing at the whole PR.

PR #2901, opened 2023-08-18
bug fix: missing `mart_gtfs.fct_daily_scheduled_stops`

### LINKED ISSUE #2900: Bug: fewer feeds present in `mart_gtfs.fct_daily_scheduled_stops`
**Describe the bug**
While creating Aug's hqta dataset, I noticed some ferry stops were missing. Upon investigation, I had only a fraction of unique feeds in my stops table. At some point between 7/12/23 has 100+, nearly 200 `feed_keys` and 8/1/23, feeds start dropping in `fct_daily_scheduled_stops`. 

Those feeds are active in `fct_scheduled_trips` though. For 8/16/23, there were 178 `feed_keys`.

**To Reproduce**

Checked for a count of unique feeds present for 8/16/23, 8/1/23, and 7/12/23 (last month's update for hqta).
```
SELECT 
  COUNT(DISTINCT feed_key) 
FROM `cal-itp-data-infra.mart_gtfs.fct_daily_scheduled_stops` 
WHERE `service_date` = "2023-08-16"
```
one
two
three

**Expected behavior**
Over 150+ distinct feed keys for `fct_daily_scheduled_stops`

**Additional context**

tiffanychu90:
# Description

Since we want `fct_daily_scheduled_stops` to show stops for all days available, move `fct_scheduled_trips` to use as base table on which `dim_stop_times` is joined.

Resolves #2900 

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

`poetry run dbt run --select fct_daily_scheduled_stops`
`poetry run dbt test --select fct_daily_scheduled_stops`

Also, checked Big Query to make sure there are more feeds available in Aug. 
bq

## Post-merge follow-ups

- [ ] No action required
full refresh for `mart_gtfs.fct_daily_scheduled_stops`

### DIFF
--- a/warehouse/models/mart/gtfs/fct_daily_scheduled_stops.sql
+++ b/warehouse/models/mart/gtfs/fct_daily_scheduled_stops.sql
@@ -19,10 +19,6 @@ WITH fct_scheduled_trips AS (
 dim_stop_times AS (
     SELECT *
     FROM {{ ref('dim_stop_times') }}
-    WHERE {{ incremental_where(default_start_var='GTFS_SCHEDULE_START',
-                               this_dt_column='_feed_valid_from',
-                               filter_dt_column='_feed_valid_from',
-                               dev_lookback_days = None) }}
 ),
 
 dim_stops AS (
--- a/warehouse/models/mart/gtfs/fct_daily_scheduled_stops.sql
+++ b/warehouse/models/mart/gtfs/fct_daily_scheduled_stops.sql
@@ -65,14 +65,14 @@ stops_by_day_by_route AS (
             trips.contains_warning_missing_foreign_key_stop_id
         ) AS contains_warning_missing_foreign_key_stop_id
 
-    FROM dim_stop_times AS stop_times
+    FROM fct_scheduled_trips AS trips
+    LEFT JOIN dim_stop_times AS stop_times
+        ON trips.feed_key = stop_times.feed_key
+            AND trips.trip_id = stop_times.trip_id
     LEFT JOIN int_gtfs_schedule__frequencies_stop_times freq
         ON stop_times.feed_key = freq.feed_key
         AND stop_times.trip_id = freq.trip_id
         AND stop_times.stop_id = freq.stop_id
-    LEFT JOIN fct_scheduled_trips AS trips
-        ON trips.feed_key = stop_times.feed_key
-            AND trips.trip_id = stop_times.trip_id
     GROUP BY 1, 2, 3, 4, 5, 6
 
 ),
--- a/warehouse/models/mart/gtfs/fct_daily_scheduled_stops.sql
+++ b/warehouse/models/mart/gtfs/fct_daily_scheduled_stops.sql
@@ -10,12 +10,19 @@ WITH fct_scheduled_trips AS (
 
     SELECT *
     FROM {{ ref('fct_scheduled_trips') }}
+    WHERE {{ incremental_where(default_start_var='GTFS_SCHEDULE_START',
+                               this_dt_column='service_date',
+                               filter_dt_column='service_date',
+                               dev_lookback_days = None) }}
 ),
 
 dim_stop_times AS (
     SELECT *
     FROM {{ ref('dim_stop_times') }}
-    WHERE {{ incremental_where(default_start_var='GTFS_SCHEDULE_START', this_dt_column='_feed_valid_from', filter_dt_column='_feed_valid_from', dev_lookback_days = None) }}
+    WHERE {{ incremental_where(default_start_var='GTFS_SCHEDULE_START',
+                               this_dt_column='_feed_valid_from',
+                               filter_dt_column='_feed_valid_from',
+                               dev_lookback_days = None) }}
 ),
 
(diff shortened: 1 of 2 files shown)


tiffanychu90:
@lauriemerrell: where should i put the test of stop's feeds = trip's feeds?

lauriemerrell (changes requested):
Ope sorry -- the other key change is to move incremental where statement from the dim stop times CTE to the fct scheduled trips CTE at the very top (which will also require changing with date columns are referenced), otherwise we will still only join in stop information for new feeds downloaded today. 

Re: where the test would be located: This would probably actually be a new custom test! Here's an example of something fairly similar: https://github.com/cal-itp/data-infra/blob/main/warehouse/tests/validate_guidelines_checks_rows_per_check_matches_index.sql. I think what we'd want to check is that the count of feed keys by date is the same between the two models. There is more about defining custom tests and how to make them generic on this page: https://docs.getdbt.com/docs/build/tests#generic-tests. I think we can just start with a singular test for this and then if you all want to reuse it (to check other daily models the same way), you could consider making it generic later.

tiffanychu90:
@lauriemerrell 
> the other key change is to move incremental where statement from the dim stop times CTE to the fct scheduled trips CTE at the very top (which will also require changing with date columns are referenced), otherwise we will still only join in stop information for new feeds downloaded today.

Might need a call to figure this out. I think I'm getting further with these 2 CTEs to filter `fct_scheduled_trips` down to only the latest batch of dates: 
```
WITH dim_schedule_feeds AS (
    SELECT *
    FROM {{ ref('dim_schedule_feeds') }}
    WHERE {{ incremental_where(default_start_var='GTFS_SCHEDULE_START', 
                               this_dt_column='_valid_from', 
                               filter_dt_column='_valid_from', 
                               dev_lookback_days = None) }}
),

fct_scheduled_trips AS (

    SELECT *
    FROM {{ ref('fct_scheduled_trips') }}
    WHERE fct_scheduled_trips.service_date >= dim_schedule_feeds._valid_from
),
```

lauriemerrell:
@tiffanychu90 happy to have a call if needed! I think the solution would be:

```sql
WITH fct_scheduled_trips AS (
    SELECT *
    FROM {{ ref('fct_scheduled_trips') }}
    WHERE {{ incremental_where(default_start_var='GTFS_SCHEDULE_START', 
                               this_dt_column='service_date', 
                               filter_dt_column='service_date', 
                               dev_lookback_days = None) }}
),

```

Basically, the operation that we want to do every day is load the new daily service from the trips table. The filtering on feeds and stop times will happen based on which feeds had **service** on the new day, rather than which feeds were **downloaded** on the new day. But yeah happy to chat if needed!

tiffanychu90:
@lauriemerrell 

Ahh, I see, I didn't catch that distinction between downloaded feed batches from `dim_scheduled_feeds` vs what's in service is in `fct_scheduled_trips`. 

Added a test, but it's failing when I run `poetry run dbt test --select validate_fct_daily_tables`.  Running that test in BQ to see yields differences that seem to be reasonable though.  Some feeds dropping out across the 4 common schedule tables has been something that's been around. Is there a way to make this less strict of a test and capture when something is glaringly off?
```
WITH trip_check_cts AS (
    SELECT service_date, COUNT(DISTINCT feed_key) AS trip_n_feeds
    FROM `cal-itp-data-infra.mart_gtfs.fct_scheduled_trips`
    GROUP BY `service_date`
),

stop_check_cts AS (
    SELECT service_date, COUNT(DISTINCT feed_key) AS stop_n_feeds
    FROM `cal-itp-data-infra-staging.tiffany_mart_gtfs.fct_daily_scheduled_stops`
    GROUP BY `service_date`
)

SELECT
    trip_check_cts.service_date,
    trip_n_feeds,
    stop_n_feeds
FROM trip_check_cts
LEFT JOIN stop_check_cts 
    ON trip_check_cts.service_date = stop_check_cts.service_date
WHERE trip_n_feeds > stop_n_feeds
ORDER BY service_date DESC
```
bq

lauriemerrell:
In general a threshold can make sense and I think you've implemented it correctly, but I didn't know that feeds were dropping out between these two.... That doesn't necessarily make sense from an overall data perspective, if a feed has trips scheduled it should also have stops on those trips 👀

tiffanychu90:
Perhaps this PR can just address this one bug. 

I suppose digging into why number of feeds doesn't match across stops could happen in the future. But, the tests failed with 284 rows, definitely not isolated. I have noticed that the number of feeds differ in `trips`, `shapes` (expected, bc it's optional), `stops` and `stop_times`, but the difference is in the range of 10-20.

lauriemerrell:
I looked into it and the mismatches are Flex feeds, which does make sense! 

Re:
> do I want to be the one to kick off a refresh for this as a very high stakes practice?

I'm not sure if that's a question for you all on the CT side to decide who would do this in future? It might be a good idea for someone not on Jarvus to do this, perhaps I can join a call to help guide but it would be good for someone else to know how to do it!

lauriemerrell (approved):

PR #2903, opened 2023-08-21
include `incomplete_variable_fare` rides in `fct_payments_rides_v2` table

charlie-costanzo:
# Description
As discovered by Ali Attari, there are rides being excluded in `fct_payments-rides_v2` where a rider didn't tap on but did tap off, and we still want to include these taps in our calculations as they charge a penalty fare. After some investigation, it appears due to a `INNER JOIN` being used where a `LEFT JOIN` should be used instead.

## How has this been tested?
in the GCS console

## Post-merge follow-ups
full refresh of `fct_payments_rides_v2`


lauriemerrell (approved):
Thanks Charlie! Also wonder about the other inner join on line 328, assume we only want micropayments where customer is populated? But assume that's unrelated to this change

PR #2924, opened 2023-08-25
revise payments date spine to exclude null values

charlie-costanzo:
# Description

Since merging the PR which includes payments rides that have `off` taps but don't have `on` taps, we introduced rides with no transaction date which is interfering with the date spine we created. This PR excludes null values for transaction dates in the date spine.

## Type of change

## How has this been tested?

locally with dbt


atvaccaro (approved):

lauriemerrell:
Can we merge this today? It's still causing the dbt jobs to log failures to Airflow, and tests don't run

PR #2928, opened 2023-08-29
Add Anaheim Transportation Network to LittlePay pipeline

SorenSpicknall:
# Description

Adds code to bring Anaheim Transportation Network LittlePay data (already available in a vendor-controlled S3 bucket) into the warehouse. The agency is referred to here as `atn` for consistency with the name of the vendor data at the source (the bucket is `littlepay-prod-atn-datafeed`, rather than using an expanded form of the agency's name). That can be changed if we prefer to use the full name - I see there's a mixture of names and acronyms among the current LittlePay syncs.

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

Tested on local Airflow from raw data ingestion to external table creation

## Post-merge follow-ups

- [x] No action required

### DIFF
--- a/warehouse/models/mart/payments/fct_payments_rides_v2.sql
+++ b/warehouse/models/mart/payments/fct_payments_rides_v2.sql
@@ -46,6 +46,11 @@
     principals = ['serviceAccount:redwood-coast-transit@cal-itp-data-infra.iam.gserviceaccount.com']
 ) }}",
 " {{ create_row_access_policy(
+    filter_column = 'participant_id',
+    filter_value = 'atn',
+    principals = ['serviceAccount:atn@cal-itp-data-infra.iam.gserviceaccount.com']
+) }}",
+" {{ create_row_access_policy(
     principals = ['serviceAccount:metabase@cal-itp-data-infra.iam.gserviceaccount.com',
                   'serviceAccount:bq-transform-svcacct@cal-itp-data-infra.iam.gserviceaccount.com',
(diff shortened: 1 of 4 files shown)


warehouse/models/mart/payments/fct_payments_rides_v2.sql
```diff
@@ -45,6 +45,11 @@
     filter_value = 'redwood-coast-transit',
     principals = ['serviceAccount:redwood-coast-transit@cal-itp-data-infra.iam.gserviceaccount.com']
 ) }}",
+" {{ create_row_access_policy(
+    filter_column = 'participant_id',
+    filter_value = 'atn',
+    principals = ['serviceAccount:atn@cal-itp-data-infra.iam.gserviceaccount.com']
```
atvaccaro, line 51: I don't see this service account, can you create it before merge? might error if it doesn't exist, I don't remember exactly

atvaccaro (approved):

warehouse/models/mart/payments/fct_payments_rides_v2.sql
atvaccaro: Also I just realized we don't have the suffix `-payments-user` on all of these agency service accounts but it'd be useful especially on these abbreviated instance names

SorenSpicknall: Just created it, which also made me realize that the three-character name was too short for Google Cloud requirements, so it got a small rename in [9050782](https://github.com/cal-itp/data-infra/pull/2928/commits/90507826f8a0c8956316a3a1114f381a21bb293b)

PR #2934, opened 2023-08-31
change fct_payments_rides_v2 join to include missing transactions

charlie-costanzo:
# Description

After comparing HTA's July littlepay statement to the Metabase dashboard we realized that transactions were missing. After investigating, we realized that it was a result of an inner join on the `customer_id`field in `fct_payments_rides_v2`. This Pr changes that inner join to a left join, which causes the transaction values to match much more closely.

Resolves #\[issue\]

## Type of change

## How has this been tested?
locally in Bigquery and dbt


lauriemerrell (approved):
I think this is what I asked about last week 😅 https://github.com/cal-itp/data-infra/pull/2903#pullrequestreview-1594436211

charlie-costanzo:
> I think this is what I asked about last week 😅 https://github.com/cal-itp/data-infra/pull/2903#pullrequestreview-1594436211

lol it definitely is, I should have just done it then

PR #2953, opened 2023-09-28
handle empty strings as nulls in littlepay staging tables

### LINKED ISSUE #2943: Payments: Handle empty strings as nulls in Payments data
In the investigation of refunds data, I discovered that several payments staging tables include cases where a column contains an empty string (""). These columns are very problematic because they are effectively null but they will not be returned by queries on `column IS NULL`. We [have a dbt macro trim_make_empty_string_null](https://github.com/cal-itp/data-infra/blob/main/warehouse/macros/trim_make_empty_string_null.sql) designed to convert such cases (and also cases where a string is only whitespace) to nulls. We should apply this macro on all string columns in the Payments data:

- [x] LP authorisations
- [x] LP customer funding sources
- [x] LP device transaction purchases
- [x] LP device transactions
- [x] LP micropayment adjustments
- [x] LP micropayment device transactions
- [x] LP micropayments
- [x] LP product data
- [x] LP refunds
- [x] LP settlements
- [ ] Elavon transactions

charlie-costanzo:
# Description

As described in #2943, we need to add logic to handle empty strings in the payments pipeline as NULL values. This PR implements the `trim_make_empty_string_null` macro for littlepay staging tables.

Resolves #2943 

## Type of change

- [x] New feature

## How has this been tested?
locally with dbt

## Post-merge follow-ups

- [x] No action required


lauriemerrell (changes requested):
Looks good but I think we should be doing this also in cases where we cast, for example: `{{ safe_cast(trim_make_empty_string_null('transaction_amount'), type_numeric()) }}` -- the point being that if the thing is empty it should become null also (basically, looks good in the cases where macro applied, I just think we need to apply it more broadly too)

lauriemerrell:
Maybe `safe_cast` isn't the best example because that may handle nulls, but definitely should use for cases where we cast to other formats I think (like lat/long).... Just to be safe? 🤔 In GTFS we do this even inside of `SAFE_CAST`s

lauriemerrell (approved):
Chatted offline, decided that this is ok as-is. Tested a bit and confirmed that we'll get cast errors if something is null

PR #2954, opened 2023-09-28
Initial authorisations modeling

lauriemerrell:
# Description

This PR:
* Creates macros to use at the column level for extracting date & datetime from a Littlepay file name
* Creates macros to use in two-stage deduping of first entire files and then full-duplicate rows 
* Applies the macros to the authorisations staging table

TODO:
* ~Confirm CCJPA date handling is correct for the invalid January filenames~

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

New tests pass, have manually inspected tables throughout development to ensure deduplication working as expected

## Post-merge follow-ups

- [x] No action required


lauriemerrell:
Reverted to draft because I am trying to figure out how to handle the keying. My intended natural key construction is `aggregation_id + authorisation_date_time_utc`. There are 8 pairs of duplicates if you use that, which are rows with the same ID and same auth time but different RRN values.

For 6 of the pairs, only one of the rows has a RRN + status (i.e., one of the RRNs is null); it seems clear that the null RRN/status ones are just duplicates that basically doesn't get used. For the remaining 2 pairs, one is a debt reauth where there's no settlement. The final pair is two duplicate authorisations that both got authorized but only one actually got settled.

A few possible courses of action (not mutually exclusive):
1. Ask vendor about these to validate if we can safely drop the duplicates and just keep one.
2. Add RRN to the key construction; this is what @SorenSpicknall originally proposed. I am conflicted on this because these rows seem problematic anyway.
3. Just drop the duplicates; this could come in a few flavors:
   1. Where to do it:
        1. Staging table OR
        2. Intermediate table
   2. How aggressively to do it:
      1. Could drop only the unproblematic cases where one is missing RRN and the other has it.
      2. Could make a heuristic to handle all 8; this might involve joining in settlements data.


My instinct is to drop the unproblematic cases in the staging table, give the uniqueness test in staging a little wiggle room, and then drop the harder cases in the intermediate table to avoid a join in the staging table. But open to alternative ideas.

SorenSpicknall (approved):
This looks good to me!

PR #2957, opened 2023-09-29
Payments: Add string handling, refactor modeling of Elavon data

### LINKED ISSUE #2956: Payments: refactor elavon transactions and add handling for empty strings
Continuing from #2943, this PR refactors `stg_elavon__transactions` by breaking it out into staging and intermediate tables to then trim strings and handle empty strings as nulls in the staging table.

charlie-costanzo:
# Description

This PR resolves #2956  and is a continuation of the string handling outlined in #2943, but for Elavon data. This also includes refactoring of the Elavon tables.

* moved cleaning, typecasting, and deduplication from mart table to `stg_elavon__transactions`
* makes macro `parse_elavon_date` and implements in `stg_elavon__transactions`
* broke staging table out into intermediate tables to focus on billing and deposits exclusively (new tables: `int_elavon__billing_transactions`, `int_elavon__deposit_transactions`)
* removed non-relevant columns (always null) from `int_elavon__billing_transactions`
* refactored `fct_elavon__transactions` to focus on union of billing and deposits
* fix references in downstream tables
* dbt docs updates

Resolves #2956 

## Type of change

- [x] New feature

## How has this been tested?

tested locally with dbt

## Post-merge follow-ups

- [x] No action required


warehouse/models/staging/payments/elavon/stg_elavon__transactions.sql
```diff
@@ -1,108 +1,96 @@
...
-      transactions.ent_num,
-      transactions.dt,
-      transactions.execution_ts
-    FROM {{ source('elavon_external_tables', 'transactions') }}
-    WHERE batch_type = 'B'
-    AND execution_ts = (SELECT MAX(execution_ts) FROM {{ source('elavon_external_tables', 'transactions') }})
+
+        * EXCEPT (payment_date, transaction_date, settlement_date),
+
+        CASE WHEN
```
lauriemerrell, line 68: When you have three identical blocks like this, I would actually just make a macro and use that to reduce risk of someone editing and missing one. So like I'd make a macro called `parse_elavon_date` that has this case logic and takes column name as a variable and call that on these.

warehouse/models/staging/payments/elavon/int_elavon__deposit_transactions_deduped.sql
```diff
@@ -0,0 +1,85 @@
...
+),
+
+deposit_transactions AS (
+
+  SELECT * FROM stg_elavon__transactions
+  WHERE batch_type = 'D'
+
+),
+
+dedup_deposit_transactions AS (
```
lauriemerrell, line 18: I am kind of confused by this CTE, can we comment what it's doing? It kind of seems like it might be equivalent to 

```sql
SELECT * EXCEPT(execution_ts),
execution_ts AS max_ts
FROM deposit_transactions
QUALIFY ROW_NUMBER OVER (PARTITION BY trn_ref_num ORDER BY execution_ts DESC) = 1
```
Which seems a *lot* clearer to me?

warehouse/models/staging/payments/elavon/int_elavon__billing_transactions_deduped.sql
```diff
@@ -0,0 +1,79 @@
...
+  SELECT * FROM stg_elavon__transactions
+  WHERE batch_type = 'B'
+
+),
+
+dedup_billing_transactions AS (
+
+  SELECT *
+  FROM billing_transactions
+  WHERE execution_ts = (SELECT MAX(execution_ts) FROM billing_transactions)
```
lauriemerrell, line 22: I am a bit confused about why this one is different from the deposits one. For billing transactions do we only want things from the latest execution, but for deposits we want latest by `trn_ref_num`? And same as the other one it would be clearer to me as a `QUALIFY` statement, we have generally not really used subqueries elsewhere in our codebase but I can just deal if they're preferred in payments for whatever reason.

warehouse/models/staging/payments/elavon/int_elavon__billing_transactions_deduped.sql
```diff
@@ -0,0 +1,79 @@
+{{ config(materialized='table') }}
+
+WITH
+
+stg_elavon__transactions AS (
```
lauriemerrell, line 5: For conciseness, could just combine this with the following CTE (`SELECT * WHERE batch type`) and same with deposits

warehouse/models/mart/payments/fct_elavon__deposit_billing_transactions.sql
```diff
@@ -0,0 +1,78 @@
...
+        term_id,
+        ent_num,
+        dt,
+        execution_ts
+
+    FROM union_deposits_and_billing
+
+)
+
+SELECT * FROM fct_elavon__deposit_billing_transactions
```
lauriemerrell, line 78: I think that *in general* it is preferable that mart tables maybe don't have a specific vendor name, and instead just reflect the concept. I think it's probably fine in this case, but could be something to consider in future (like, should we have `fare_processor_transactions` and `fare_calculator_transactions`? Something more generic and vendor-agnostic.) Maybe just longer-term food for thought

lauriemerrell (commented):
Just a few ideas!

warehouse/models/staging/payments/elavon/int_elavon__billing_transactions_deduped.sql
charlie-costanzo: So they're different because billing transactions don't have a `trn_ref_num` associated with them, so the only thing we can dedupe on is the latest execution. this is the heart of the issue with reconciling Elavon and Littlepay, there's no `trn_ref_num` to join on when there should be

charlie-costanzo: I agree with the `QUALIFY` statement though, I can implement that

lauriemerrell: That makes sense, though I thought that we were using `purch_id` to map to Littlepay rather than `trn_ref_num`. Thank you for explaining!

charlie-costanzo: sorry yes you're right, purch_id is used for the mapping and isn't present in billing transactions, along with trn_ref_num and a slew of other fields

warehouse/models/staging/payments/elavon/int_elavon__deposit_transactions_deduped.sql
lauriemerrell: Updated per convo 10/5: we should just do `QUALIFY DENSE_RANK() OVER (ORDER BY execution_ts DESC)` and do it once for *all* the rows and handle duplicate `trn_ref_num` values downstream if necessary

charlie-costanzo:
hey @lauriemerrell just tagged you in for re-review. I can handle dbt docs once it's settled. I can also add a uniqueness test for `trn_ref_num`? if we want to capture that now that we've removed it in the deduplication

warehouse/models/staging/payments/elavon/stg_elavon__transactions.sql
```diff
@@ -1,108 +1,82 @@
-WITH primary_batch_transactions AS (
+{{ config(materialized='table') }}
+
+WITH
+
+source AS (
+    SELECT * FROM {{ source('elavon_external_tables', 'transactions') }}
```
lauriemerrell, line 6: I would put the `QUALIFY DENSE_RANK() OVER (ORDER BY execution_ts DESC) = 1` here (instead of in the intermediate tables) and add a comment that we pull the whole world every day so we get only the latest extract. 

We may then also want to add a custom test to confirm that all the dates with data in the external table are present in the staging table to make sure we aren't missing anything with this filter

lauriemerrell: I guess that second part may not be possible since I don't see a filename attribute here 🤔 Ideally we would have a filename attribute for these like for LP

warehouse/models/staging/payments/elavon/int_elavon__deposit_transactions_deduped.sql
lauriemerrell: Started a new thread for this

warehouse/models/staging/payments/elavon/int_elavon__billing_transactions_deduped.sql
```diff
@@ -0,0 +1,73 @@
...
+      amount,
+      surchg_amount,
+      convnce_amt,
+      card_type,
+      charge_type,
+      charge_type_description,
+      card_plan,
+      card_no,
+      chk_num,
+      transaction_date,
```
lauriemerrell, line 47: For billing transactions, I would not include the columns that are always null-- Things like `purch_id`, `trn_ref_num`, etc. (the transaction-level data that is not present for billing rows)

warehouse/models/staging/payments/elavon/int_elavon__billing_transactions_deduped.sql
lauriemerrell, line 1: I would rename this and the other one to just `int_elavon__<type>_transactions` and drop the `deduped`.

warehouse/models/mart/payments/fct_elavon__deposit_billing_transactions.sql
lauriemerrell, line 1: I actually think that for the short term maybe it's not worth renaming this model at this point (so you don't have to update the Metabase dashboards yet) and we do a bit more refactoring once the LP models are ready around transaction level stuff. Because I actually kind of think that we will want separate mart tables `fct_elavon_billing_batches` and `fct_elavon_deposits` eventually

lauriemerrell (changes requested):

lauriemerrell (commented):
Code changes LGTM! Only requests would be:

* Document the new intermediate models in the YAML
* Update PR description to match the current state (I think it's now a bit out of date)

lauriemerrell (approved):

Issue #2945, opened 2023-09-18
Payments: Dedupe rows on all incoming tables


lauriemerrell:
We currently [dedupe Payments micropayments data](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/staging/payments/littlepay/stg_littlepay__micropayments.sql#L25) to account for the fact that the same data can be uploaded multiple times (and because we sometimes get multiple full duplicate rows). We should add the same handling on all Payments tables. ( ⚠️ Primary key to use for deduplication will vary; we should aim to dedupe down to the entry/file level in staging tables, I think.)

- [ ] LP authorisations
- [ ] LP customer funding sources
- [ ] LP device transaction purchases
- [ ] LP device transactions
- [ ] LP micropayment adjustments
- [ ] LP micropayment device transactions
- [ ] LP product data
- [ ] LP refunds
- [ ] LP settlements
- [ ] Elavon transactions (double check whether applicable to Elavon data)

lauriemerrell:
Transferring from offline conversation: We have discovered that the current micropayments `QUALIFY` probably deserves a bit more scrutiny. We are finding a fairly large number of cases where it combines rows with distinct `transaction_time` values, whose interpretation is unclear to us.

lauriemerrell:
This is closed in #2993 and #2994

PR #2993, opened 2023-10-06
Littlepay Table Deduplication Adjustments

SorenSpicknall:
# Description

This represents a partial solution to #2945, encompassing deduplication and testing changes to the customer_funding_source, device_transaction_purchases, device_transactions, micropayment_adjustments, micropayment_device_transactions, and product_data tables from Littlepay. Laurie is taking care of the three remaining Littlepay pipeline tables.

Our intent was to rid ourselves of true duplicates and obvious stale data that is superseded by later, updated data. Every row in these tables has a unique combination of `_key` and `_payments_key`. A couple of these tables contain QUALIFY statements that utilize `littlepay_export_ts` in addition to the new standard deduplication logic, because we've detected that these tables occasionally receive updated versions of previous rows.

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

All tables created in a staging environment and dbt tests run for each. Downstream models also run where relevant.

## Post-merge follow-ups

- [ ] No action required

Triage remaining payments test failures that exist after merging these uniqueness changes and Laurie's changes to the micropayments, refunds, and settlements tables


warehouse/models/staging/payments/littlepay/_littlepay.yml
```diff
@@ -243,10 +235,20 @@ models:
         description: The number of records on which the `customer_id` shows up in the table
       - name: calitp_customer_id_rank
         description: A ranking of the records using the `customer_id`, ordered newest to oldest
+      - *lp_export_date
+      - *lp_export_ts
+      - *lp_line_number
+      - *payments_input_row_key
+      - name: _payments_key
```
lauriemerrell, line 242: For this and any others with `unique_proportion`, can we use YAML anchors and just override minimally to change the unique proportion? Seems preferable to copying identical description many times, in case we want to modify. See for example: https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/gtfs/_mart_gtfs_fcts.yml#L366-L370

warehouse/models/staging/payments/littlepay/_littlepay.yml
```diff
@@ -404,10 +427,19 @@ models:
           The ID of the product with the incentive that was applied to micropayments.
 
           Same ID as `product_id`.
+      - *lp_export_date
+      - *lp_export_ts
+      - *lp_line_number
+      - *payments_input_row_key
+      - name: _payments_key
```
lauriemerrell, line 434: In my PR I have a fully unique anchor for this column that can be used here

warehouse/models/staging/payments/littlepay/stg_littlepay__customer_funding_source.sql
```diff
@@ -1,8 +1,8 @@
 WITH source AS (
-    SELECT * FROM {{ littlepay_source('external_littlepay', 'customer_funding_source') }}
+    SELECT * FROM {{ source('external_littlepay', 'customer_funding_source') }}
 ),
 
-stg_littlepay__customer_funding_source AS (
+clean_columns_and_dedupe_files AS (
```
lauriemerrell, line 5: In my PR #2994 I changed the corresponding CTE names to just `clean_columns` because we are no longer deduping files

warehouse/models/staging/payments/littlepay/stg_littlepay__device_transactions.sql
```diff
@@ -32,12 +32,69 @@ stg_littlepay__device_transactions AS (
...
+            littlepay_export_ts,
+            littlepay_export_date,
+            ts,
+            _key,
+            _payments_key,
+            _content_hash,
+    FROM add_keys_drop_full_dupes
+    -- Some transactions have placeholder information for routes, stops, and directions in their first export,
+    -- then a later export contains the canonical version of the transaction with that information corrected
     QUALIFY ROW_NUMBER() OVER (PARTITION BY littlepay_transaction_id ORDER BY littlepay_export_ts DESC, transaction_date_time_utc DESC) = 1
```
lauriemerrell, line 98: In PR description it says that this is one of the tables that still has natural key duplicates, I am particularly interested in that given this extra qualify.... Can you tell why they are happening?

lauriemerrell (commented):
Thanks for tackling, just a few things:

> In about half the tables, the assumed natural key of the table is fully unique. Some retain duplicates on the assumed natural key, which we intend to deal with in downstream models (or in staging models) if we learn more about the conditions causing the duplicates.

I was approaching this as if this round of updates **was** the investigation where we would try to learn more about the conditions causing the duplicates. Like for micropayments, where we have a few remaining duplicates, I identified exactly what is causing them and identified what would be needed to address and why it's not appropriate to do in staging (specifically, would require a join with settlements); I am realizing now that I should write it into a ticket to capture outside of Slack conversations. 

Can we try to at least write tickets identifying the causes for these tables as well? Like, what columns are differing when the natural key is duplicated? Can we tell if they seem like true duplicates, or whether we should consider adding a new column to the key (even if that goes against LP documentation)? Sorry if I missed this documentation being added elsewhere. 

> A couple of these tables retain QUALIFY statements from their previous form, in addition to the new deduplication logic. It is likely valuable to put a careful set of extra eyes on those statements to ensure we're not introducing any unintended behavior or doing anything redundant.

Did you check whether these statements are still needed with the full-row duplicate dropping? (Does the row count change?) It looks like `device_transactions` at least has a commented explanation, but I am a little interested to see that that qualify is using `littlepay_transaction_id` and the primary key seems to be `device_transaction_id`, did we confirm why that's the case?

warehouse/models/staging/payments/littlepay/stg_littlepay__device_transactions.sql
SorenSpicknall: This has been superseded by further investigation - no more partial dupes.

warehouse/models/staging/payments/littlepay/stg_littlepay__customer_funding_source.sql
```diff
@@ -29,7 +41,47 @@ stg_littlepay__customer_funding_source AS (
...
+        littlepay_export_ts,
+        littlepay_export_date,
+        calitp_funding_source_id_rank,
+        calitp_funding_source_vault_id_rank,
+        calitp_customer_id_rank,
+        _key,
+        _payments_key,
+        _content_hash,
+    FROM add_keys_drop_full_dupes
+    -- Some funding sources have incomplete information when first present in data, like missing
```
lauriemerrell, line 76: These comments make sense and I *think* this handling seems appropriate but for this data in particular I am having a 💡 moment looking at this -- my only question is whether there's any chance this needs to be handled more like type 2 slowly-changing data. And same question for `product_data`. 

Looking at the data dictionary I see that things like the card's associated zip code and expiration date don't seem to be included so maybe not, but can you expand (probably offline) on which filled-in fields do change? 

Just realized that this table seems a bit dissimilar from the others in that it's more of a dimension than a fact, and I'm not sure if/how we need to handle accordingly.

warehouse/models/staging/payments/littlepay/stg_littlepay__device_transactions.sql
```diff
@@ -32,13 +32,73 @@ stg_littlepay__device_transactions AS (
...
+            littlepay_export_ts,
+            littlepay_export_date,
+            ts,
+            _key,
+            _payments_key,
+            _content_hash,
+    FROM add_keys_drop_full_dupes
+    -- Some transactions have placeholder information for routes, stops, and directions in their first export,
+    -- then a later export contains the canonical version of the transaction with that information corrected
+    QUALIFY ROW_NUMBER() OVER (
```
lauriemerrell, line 98: Does this persist the 4 duplicate rows we talked about where there are two transaction IDs for otherwise identical info? Should those be dropped?

warehouse/models/staging/payments/littlepay/stg_littlepay__product_data.sql
```diff
@@ -29,12 +29,79 @@ stg_littlepay__product_data AS (
...
+        `instance`,
+        extract_filename,
+        littlepay_export_ts,
+        littlepay_export_date,
+        ts,
+        _key,
+        _payments_key,
+        _content_hash,
+    FROM add_keys_drop_full_dupes
+    -- Some products change in form over time, e.g. getting different 'capping_type' values or
```
lauriemerrell, line 98: Similar question to above.... I am wondering whether these need to be handled as type 2 or if we should treat these as corrections. 🤔 Because the question would be if we're looking at older data, should we access the capping conditions that were in effect at the time? Or the capping conditions as they are now?

lauriemerrell (commented):
Thanks for addressing the first round, I think everything mostly looks good but re-reviewing I kinda realized that funding sources and product data seem a bit different in **kind** from the other data and I am wondering if we should think about treating those as type 2 dimensions rather than just getting the latest version of the record? If so, perhaps they can/should be spun out separately.

warehouse/models/staging/payments/littlepay/stg_littlepay__device_transactions.sql
SorenSpicknall: I didn't want to add logic that would drop those until we knew more about what we should be seeing. Especially in light of your notes from the recent meeting with Littlepay, it seems like it might be premature to write logic that collapse on shared values in other fields when our current docs tell us to treat `littlepay_transaction_id` as unique.

lauriemerrell: Yeah I can see that, I do wonder if ultimately some handling akin to authorisations where rows are dropped later if they don't map to something could be appropriate? Should we at least write an issue to document that this is being left in place?

lauriemerrell: Confirmed offline that @SorenSpicknall will write an issue after PR merges.

lauriemerrell (approved):

Issue #2958, opened 2023-10-02
Payments: Develop `fct_aggregations` model


lauriemerrell:
As we focus more on the outcomes of (attempted) fare payment transactions, it seems appropriate to develop a model of aggregations, since those are the entity at which the dollar amount reconciliation occurs. 

Proposed schema:

- `aggregation_id`
- `distinct_auth_actions` = count of distinct authorisation rows for this aggregation
- `auth_type` = most recent auth request_type
- `auth_status` = most recent status 
- `auth_datetime` = most recent authorisation_date_time_utc value
- `card_check_status` = most recent card check outcome for this aggregation
- `card_check_datetime` = authorisation_date_time_utc value for most recent card check
- `authorisation_status` = most recent authorisation outcome 
- `authorisation_datetime` = authorisation_date_time_utc value for most recent authorisation
- `debt_recovery_status` = most recent debt recovery authorisation outcome
- `debt_recovery_datetime` = authorisation_date_time_utc value for most recent debt recovery attempt 
- `settlement_requested_datetime` = settlement_requested_date_time_utc
- A few columns about refunds? Maybe a boolean for `has_refund` and then some details about the refund if one occurred?

This must take account of cases where micropayment transaction_amount is 0 so no further authorisation or settlement is required. 

To create such a table, it is likely that we will need to create some parent models, things like:
- Pivoted / aggregated authorisations 
- A unified refunds model that combines the two types

PR #2994, opened 2023-10-06
Payments staging models refactor: settlements, micropayments, refunds

lauriemerrell:
# Description

This PR continues working on deduping tables that will contribute to a final `fct_aggregations` table.

* Fixes a filtering issue on the authorisations table that was incorrectly dropping both copies of duplicated rows instead of just one copy
* Brings micropayments deduping more in line with other tables and adds contextual comments
* Dedupes settlements
   * There is one duplicate on settlement ID with two rows at different timestamps -- it doesn't seem to be associated with a refund. I just dropped this one row manually because it seems like we'd want to know if it happens again (since otherwise there is full 100% uniqueness.) 
* Adds some extra casting and keying for refunds; there are no full-row duplicates for refunds anyway, so no deduping is actually needed

Progress towards #2958 and #2945

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

Ensured that dbt tests pass on all affected tables. For micropayments, confirmed no new test failures on downstream tables either. 

## Post-merge follow-ups

- [x] No action required


charlie-costanzo (approved):
Amazing - LGTM! Thank you for taking care of this @lauriemerrell.

PR #2996, opened 2023-10-06
Add new Littlepay columns + correct mistyped columns

### LINKED ISSUE #2995: Bug: Incorrectly named Littlepay columns prevent data movement
**Describe the bug**
Due to mislabeled column names in our external table creation process (stemming from copy/pastes from an old Littlepay-provided data dictionary that itself had some errors), a couple of our Littlepay tables include columns that are erroneously empty. The ones I've identified are as follows:

micropayment_adjustments - zone_ids_us should be zone_ids_used
device_transactions - transction_deny_reason should be transaction_deny_reason
product_data - capping_overlap should be capping_overlap_time

**To Reproduce**
Check these column names in the external table creation YAML files and downstream models against the titles of those columns in raw copied source files in GCS.

**Expected behavior**
Once these column names are corrected, source data should flow in correctly for these columns.

**Additional context**
A full fix will require us to reprocess old Littlepay source files into their `parsed` form

SorenSpicknall:
# Description

Reconciles Littlepay external tables with the column lists provided to us in the most recent version of the Littlepay data dictionary, and corrects a few column names in external tables and associated models in order to resolve #2995.

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

Tested in local Airflow, staging GCS, and local dbt.

## Post-merge follow-ups

- [ ] No action required

Since this changes a column name in `payments_rides_v2`, dashboards downstream of that model may need a small update. Additionally, on merge, external tables should be run right away in order to bring in new columns that are referenced by models (to avoid erroring on the next dbt run).


warehouse/models/staging/payments/littlepay/stg_littlepay__micropayment_adjustments.sql
```diff
@@ -14,7 +14,8 @@ clean_columns AS (
         {{ trim_make_empty_string_null('amount') }} AS amount,
         {{ trim_make_empty_string_null('time_period_type') }} AS time_period_type,
         {{ safe_cast('applied', type_boolean()) }} AS applied,
-        {{ trim_make_empty_string_null('zone_ids_us') }} AS zone_ids_us,
+        {{ trim_make_empty_string_null('zone_ids_used') }} AS zone_ids_used,
+        {{ trim_make_empty_string_null('incentive_product_id') }} AS incentive_product_id,
```
lauriemerrell, line 18: `incentive_product_id` seems to be null in 100% of our data. Do we actually get data with this column? (i.e., is it actually present in any files we're currently getting?)
image

warehouse/models/staging/payments/littlepay/stg_littlepay__product_data.sql
```diff
@@ -79,12 +99,24 @@ stg_littlepay__product_data AS (
         scheduled_end_date_time,
         all_day,
         weekly_cap_start_day,
+        weekly_cap_end_day,
         number_of_days_in_cap_window,
         capping_duration,
         number_of_transfer,
         capping_time_zone,
-        capping_overlap,
+        capping_overlap_time,
```
lauriemerrell, line 107: This seems to always be null, is this field present in the actual files we're getting?

warehouse/models/staging/payments/littlepay/stg_littlepay__product_data.sql
```diff
@@ -79,12 +99,24 @@ stg_littlepay__product_data AS (
...
         capping_application_level,
+        route_capping_enabled,
+        routes,
+        zoned_capping_enabled,
+        zoned_capping_mode,
+        zoned_capping_price_type,
```
lauriemerrell, line 113: Always null -- same question, is this present in the files?

warehouse/models/staging/payments/littlepay/stg_littlepay__product_data.sql
```diff
@@ -79,12 +99,24 @@ stg_littlepay__product_data AS (
...
+        on_peak_zones,
+        off_peak_zones,
+        incentive_enabled,
+        incentive_type,
```
lauriemerrell, line 117: Always null, same question

warehouse/models/staging/payments/littlepay/stg_littlepay__product_data.sql
```diff
@@ -79,12 +99,24 @@ stg_littlepay__product_data AS (
...
+        discount_qualifier,
```
lauriemerrell, line 118: Always null, same question

warehouse/models/staging/payments/littlepay/stg_littlepay__product_data.sql
```diff
@@ -79,12 +99,24 @@ stg_littlepay__product_data AS (
...
+        configuration,
```
lauriemerrell, line 119: Always null, same question

warehouse/models/staging/payments/littlepay/_littlepay.yml
lauriemerrell, line 1: Can the other added columns be documented? (ex. `previous_version_id` etc.)

lauriemerrell (commented):
It seems like some of the added columns are always null. Can you confirm whether these columns are actually appearing in the files that we're getting (like, are these fields present but always null vs. not listed in the files at all)? If these fields are not present in the files that we're actually getting, but are just listed in the LP documentation, I think it would be useful to add comments and YAML to that effect, just to distinguish... We could even consider not listing them yet and/or commenting them out until we get files that actually contain these fields.

warehouse/models/staging/payments/littlepay/_littlepay.yml
SorenSpicknall: Almost all of the new columns were previously documented in YAML, but weren't present in the models. [f9f3b2c](https://github.com/cal-itp/data-infra/pull/2996/commits/f9f3b2c5b0e09ecaf5410641dd4e43fd0f547f15) took care of the remainders.

SorenSpicknall:
> It seems like some of the added columns are always null. Can you confirm whether these columns are actually appearing in the files that we're getting (like, are these fields present but always null vs. not listed in the files at all)? If these fields are not present in the files that we're actually getting, but are just listed in the LP documentation, I think it would be useful to add comments and YAML to that effect, just to distinguish... We could even consider not listing them yet and/or commenting them out until we get files that actually contain these fields.

Each of these is legitimately present in our source data and all-null to date (though it appears that some ATN files are coming in with header rows only and no data within - I'll open a separate issue to investigate that, since it seems like it might be a problem at the source [or ATN just doesn't have any data yet for the relevant resources]). This question did prompt me to recheck all the column names and issue one correction in [f9f3b2c](https://github.com/cal-itp/data-infra/pull/2996/commits/f9f3b2c5b0e09ecaf5410641dd4e43fd0f547f15), though!

lauriemerrell (approved):
> Almost all of the new columns were previously documented in YAML, but weren't present in the models.

I am very curious how that came to pass 😛 But cool, sorry I missed it.

PR #3024, opened 2023-10-17
update payments_gtfs_datasets seed to include elavon customer_name

charlie-costanzo:
# Description

This PR creates a canonical reference between Littlepay's `participant_id` and Elavon's `customer_name` for reconciliation between the two data sources. It also,

* renames `participant_id` to `littlepay_participant_id` for added description 
* `customer_name` also exists as `elavon_customer_name` in the seed for added description
* This PR also updates downstream models to use the new column name `littlepay_participant_id`

Resolves #\[issue\]

## Type of change

- [x] New feature

## How has this been tested?

`dbt_run` and `dbt_test`

## Post-merge follow-ups

- [ ] No action required


lauriemerrell (commented):
Just one question: For Elavon, it seemed like there were multiple customer names associated with a single agency. Are the ones not listed in the seed definitely not needed? I wonder if we'll want to add them too 🤔 

See the below for what I mean -- the query returns 15 agencies, not just 10.
```sql
SELECT DISTINCT customer_name FROM `cal-itp-data-infra.mart_payments.fct_elavon__transactions`
```
And even looking just at deposits it seems like there are 14. 
```sql
SELECT DISTINCT customer_name FROM `cal-itp-data-infra.mart_payments.fct_elavon__transactions`
WHERE batch_type = "D"
```

charlie-costanzo:
> Just one question: For Elavon, it seemed like there were multiple customer names associated with a single agency. Are the ones not listed in the seed definitely not needed? I wonder if we'll want to add them too 🤔
> 
> See the below for what I mean -- the query returns 15 agencies, not just 10.
> 
> ```sql
> SELECT DISTINCT customer_name FROM `cal-itp-data-infra.mart_payments.fct_elavon__transactions`
> ```
> 
> And even looking just at deposits it seems like there are 14.
> 
> ```sql
> SELECT DISTINCT customer_name FROM `cal-itp-data-infra.mart_payments.fct_elavon__transactions`
> WHERE batch_type = "D"
> ```

Yeah @lauriemerrell so I think those are the sales that are appearing in Elavon data but not littlepay data like we talked about during that call with Rebel... bulk purchases and office sales and things like that. If we look at only customers appearing in both vendors it's those included in the seed currently. But I suppose we should account for everything in the Elavon data?

```
SELECT
  a.participant_id,
  b.customer_name
FROM `cal-itp-data-infra.staging.stg_littlepay__settlements` AS a
INNER JOIN `cal-itp-data-infra.mart_payments.fct_elavon__transactions` as b
ON a.retrieval_reference_number = b.purch_id
GROUP BY 1,2 
ORDER BY 1
```

lauriemerrell:
> Yeah @lauriemerrell so I think those are the sales that are appearing in Elavon data but not littlepay data like we talked about during that call with Rebel... bulk purchases and office sales and things like that. If we look at only customers appearing in both vendors it's those included in the seed currently. But I suppose we should account for everything in the Elavon data?

Yeah I figured it was related but yes I think we need to account for everything in Elavon. What I think we should do in these cases then is **map the Elavon customer name to a GTFS dataset record but not to Littlepay**, since it sounds like that best captures the situation. And then ideally mart tables / dashboards would use the GTFS dataset record for labeling so that there is consistency. That way we could `group by gtfs_dataset_source_record_id` and get overall summaries on both the LP and Elavon data. Does that make sense?

charlie-costanzo:
> > Yeah @lauriemerrell so I think those are the sales that are appearing in Elavon data but not littlepay data like we talked about during that call with Rebel... bulk purchases and office sales and things like that. If we look at only customers appearing in both vendors it's those included in the seed currently. But I suppose we should account for everything in the Elavon data?
> 
> Yeah I figured it was related but yes I think we need to account for everything in Elavon. What I think we should do in these cases then is **map the Elavon customer name to a GTFS dataset record but not to Littlepay**, since it sounds like that best captures the situation. And then ideally mart tables / dashboards would use the GTFS dataset record for labeling so that there is consistency. That way we could `group by gtfs_dataset_source_record_id` and get overall summaries on both the LP and Elavon data. Does that make sense?

@lauriemerrell yeah that makes a lot of sense, I just made those changes

lauriemerrell (approved):
Seed file updates look good to me! But I'm wondering if something funky happened in the git history because the diff is showing some of Soren's changes from other PRs as well 👀 May need to rebase against main or something?

charlie-costanzo:
> Seed file updates look good to me! But I'm wondering if something funky happened in the git history because the diff is showing some of Soren's changes from other PRs as well 👀 May need to rebase against main or something?

yeah weird, rebasing now and merging

PR #3036, opened 2023-10-19
Payments: intermediate refunds tables

### LINKED ISSUE #2978: Payments: Ensure refunds are being captured correctly Littlepay/Elavon data
And update the corresponding questions in Metabase

charlie-costanzo:
# Description
As a result of the [payments refunds investigation ](https://docs.google.com/document/d/1ijgAcY2Wc07kIaW16yos6bYIv_eEFji_e1ZTtgWU1R8/edit#heading=h.9aqwgu2w1vfi) this PR seeks to create two intermediate refunds tables and lightly modify the staging table:

* `int_payments__refunds_deduped`
  * table which unions the refunds observed as `type = 'CREDIT'` in the `micropayments` table and the refunds from the `refunds` table.
  * Reconciles a common schema between the two
* `int_payments__refunds_to_aggregations`
  * table which summarizes refund amount activity on `aggregation_id` and `retrieval_reference_number` at a top level as well as by `approval_status`
* `stg_littlepay__refunds`
  * Adds some casting and trimming as necessary to facilitate downstream models
  * Adds relationship tests with `int_payments__refunds_deduped` on `aggregation_id` and `retrieval_reference_number` 

* dbt documentation

Resolves #2978

## Type of change

- [x] New feature
- [x] Documentation

## How has this been tested?

`dbt run` locally and `dbt test` as necessary

## Post-merge follow-ups

- [x] No action required


warehouse/models/intermediate/payments/int_payments__refunds.sql
```diff
@@ -0,0 +1,88 @@
...
+    SELECT
+
+        aggregation_id,
+        micropayment_id,
+        participant_id,
+        customer_id,
+        ABS(charge_amount) AS refund_amount,
+        EXTRACT(DATE FROM transaction_time AT TIME ZONE "America/Los_Angeles") AS transaction_date,
+        _line_number,
+        SAFE_CAST(currency_code AS NUMERIC) AS currency_code,
```
lauriemerrell, line 15: I actually wonder if this `SAFE_CAST` should happen in the staging table? Normally casting should happen in staging

lauriemerrell (commented):
Pre-reviewing before an offline check-in about this tomorrow -- I think this generally looks good! I actually think that maybe we do want the refund amount to be negative? In https://github.com/cal-itp/data-infra/blob/main/warehouse/models/intermediate/payments/int_payments__settlements_summarized.sql I made credited (refunded) dollar amounts negative

warehouse/models/intermediate/payments/int_payments__refunds_deduped.sql
```diff
@@ -0,0 +1,156 @@
...
+        littlepay_export_date,
+        _content_hash,
+        _key,
+        _payments_key,
+        source_table
+
+    FROM refunds_union
+
+    -- for refunds that appear in both micropayments and refunds source tables,
+    -- is this a functional way to only keep duplicates that are in refunds source table?
```
lauriemerrell, line 150: I think that the number of refunds in this situation is like 3, so I would actually just manually drop them (`where _key NOT IN ([list their keys])`) from the micropayment refunds CTE above rather than doing this here

warehouse/models/intermediate/payments/int_payments__refunds_deduped.sql
```diff
@@ -0,0 +1,156 @@
...
+
+        aggregation_id,
+        micropayment_id,
+        participant_id,
+        customer_id,
+        ABS(charge_amount) AS refund_amount,
+        EXTRACT(DATE FROM transaction_time AT TIME ZONE "America/Los_Angeles") AS transaction_date,
+
+        -- Add columns that we want to preserve from refunds table after union as null strings
+        -- Is this an appropriate way to add new columns that are always NULL as the
```
lauriemerrell, line 16: LGTM!

lauriemerrell (commented):
Overall looks good, just a few things, also I think you may need to rebase again -- diff is lookin' funky

warehouse/models/intermediate/payments/int_payments__refunds_deduped.sql
```diff
@@ -0,0 +1,156 @@
...
+        _content_hash,
+        _key,
+        _payments_key,
+        source_table
+
+    FROM refunds_union
+
+    -- for refunds that appear in both micropayments and refunds source tables,
+    -- is this a functional way to only keep duplicates that are in refunds source table?
+    -- This is currently filtering out refunds table entries that have the same aggregation_id/refund_amount but different refund_ids
```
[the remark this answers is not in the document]
lauriemerrell: Ok, after re-checking the example you provided offline I am less sure about this, it seems like some of these are legit just duplicates.

warehouse/models/intermediate/payments/int_payments__refunds_deduped.sql
```diff
@@ -0,0 +1,150 @@
...
+        littlepay_export_ts,
+        littlepay_export_date,
+        _content_hash,
+        _key,
+        _payments_key,
+        'micropayments' AS source_table
+
+    FROM {{ ref('stg_littlepay__micropayments') }}
+    WHERE type = 'CREDIT'
+        AND _key NOT IN ('043ecc000223a299ce17f6a342b1d240', '3536fb2035bbcf4dcb1f3abf001b5185')
```
lauriemerrell, line 49: I would add a comment explaining that these are two that appear in both tables and it was easier to drop them manually than trying to make an overfit filter

warehouse/models/intermediate/payments/int_payments__refunds_deduped.sql
```diff
@@ -0,0 +1,150 @@
...
+        ts,
+        littlepay_export_ts,
+        littlepay_export_date,
+        _content_hash,
+        _key,
+        _payments_key,
+        'refunds' AS source_table
+
+    FROM {{ ref('stg_littlepay__refunds') }}
+    QUALIFY DENSE_RANK() OVER (PARTITION BY refund_id ORDER BY littlepay_export_ts DESC) = 1
```
lauriemerrell, line 92: Would add a comment here noting that this dedupes on refund ID because individual refunds sometimes appear multiple times with multiple statuses and the goal here is to get the latest update per refund

warehouse/models/intermediate/payments/int_payments__refunds_deduped.sql
```diff
@@ -0,0 +1,150 @@
...
+        participant_id,
+        customer_id,
+        ABS(charge_amount) AS refund_amount,
+        EXTRACT(DATE FROM transaction_time AT TIME ZONE "America/Los_Angeles") AS transaction_date,
+
+        -- add columns that we want to preserve from refunds table after union as null strings
+        SAFE_CAST(NULL AS STRING) AS refund_id,
+        SAFE_CAST(NULL AS STRING) AS settlement_id,
+        SAFE_CAST(NULL AS STRING) AS retrieval_reference_number,
+        SAFE_CAST(NULL AS STRING) AS coalesced_id,
```
lauriemerrell, line 19: For these refunds the aggregation ID should be used as the coalesced ID since they won't have an RRN

warehouse/models/intermediate/payments/int_payments__refunds_deduped.sql
```diff
@@ -0,0 +1,150 @@
...
+        ts,
+        littlepay_export_ts,
+        littlepay_export_date,
+        _content_hash,
+        _key,
+        _payments_key,
+        source_table
+
+    FROM refunds_union
+    QUALIFY ROW_NUMBER() OVER (PARTITION BY retrieval_reference_number, refund_amount ORDER BY littlepay_export_ts DESC) = 1
```
lauriemerrell, line 146: I think this should be partitioned by `coalesced_id` because otherwise it will be dropping most copies of the rows missing RRN

lauriemerrell (commented):
Ok I think the SQL broadly looks good, biggest things are just to comment explanations of the various deduping steps and I think we need coalesced ID in the final qualify.

After that I think it's just documentation & testing.

For testing, I think that we want a test on `stg_littlepay__refunds` to test that there are no RRNs present in that table that aren't present here, and same for aggregation IDs. We want to make sure this table is capturing all the refunds.

warehouse/models/intermediate/payments/int_payments__refunds_deduped.sql
```diff
@@ -0,0 +1,167 @@
...
+        ts,
+        littlepay_export_ts,
+        littlepay_export_date,
+        _content_hash,
+        _key,
+        _payments_key,
+        source_table
+
+    FROM refunds_union
+    QUALIFY ROW_NUMBER() OVER (PARTITION BY coalesced_id, refund_amount ORDER BY littlepay_export_ts DESC) = 1
```
lauriemerrell, line 163: Would add a comment here about the purpose of this final qualify

warehouse/models/intermediate/payments/int_payments__refunds_to_aggregations.sql
```diff
@@ -0,0 +1,58 @@
...
+    SELECT *
+    FROM {{ ref('int_payments__refunds_deduped') }}
+),
+
+summarize_by_type AS (
+    SELECT
+        aggregation_id,
+        retrieval_reference_number,
+        approval_status,
+        SUM(refund_amount) AS refund_amount,
```
lauriemerrell, line 13: I think this needs to be `proposed_amount` (this is what I've been complaining about, their column naming is incredibly counterintuitive -- `proposed_amount` is `The actual amount refunded to the customer.` while `refund_amount` is `The amount equalling transaction_amount - proposed_amount.`)

lauriemerrell (commented):
Two code comments and then for the PR description I think that you list `int_payments__refunds_deduped` twice, second one should be `int_payments__refunds_to_aggregations`, and at the bottom you reference `stg_littlepay__micropayments` but I think that should be `stg_littlepay__refunds`

warehouse/models/intermediate/payments/int_payments__refunds_to_aggregations.sql
charlie-costanzo: ah jeeze okay then I need to change how I prepare the micropayments table for the union in `int_payments__refunds_rdeduped`, doing that now

lauriemerrell: Ope wow yeah sorry I missed it above!

warehouse/models/intermediate/payments/int_payments__refunds_deduped.sql
```diff
@@ -0,0 +1,169 @@
...
+micropayments_table_refunds AS (
+    SELECT
+
+        aggregation_id,
+        micropayment_id,
+        participant_id,
+        customer_id,
+        ABS(charge_amount) AS proposed_amount,
+        EXTRACT(DATE FROM transaction_time) AS transaction_date,
+        aggregation_id AS coalesced_id,
```
lauriemerrell, line 14: I think we also need to preserve `aggregation_id` as itself for joins later (in the `_to_aggregations` model, we group by `aggregation_id` so we need that column populated for these rows)

lauriemerrell (commented):
Just one final thing I think!

lauriemerrell (approved):
LGTM! 🎉

warehouse/models/intermediate/payments/int_payments__refunds_deduped.sql
lauriemerrell: My bad it's a few rows up

PR #3041, opened 2023-10-20
Drop a few dupe micropayments

SorenSpicknall:
# Description

Responsive to Sentry issue [#69798](https://sentry.calitp.org/organizations/sentry/issues/69798/), and in keeping with recent changes in which we've identified and dropped rows that clearly represent duplicate information sending or corrupt/testing data from Littlepay, this PR proposes to rid us of two micropayments that are identical to other micropayments but for their `micropayment_id`. Each rolls up into a `transaction_id` that maps to another micropayment with exactly the same metadata and activity. They seem to be the result of the system erroneously creating two identical micropayment entries. If we kept them, we'd likely be double-counting the associated payment activities.

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

All dbt models run in staging, and table row counts and shapes compared to prod.

## Post-merge follow-ups

- [ ] No action required

Mark the corresponding Sentry issue as resolved after ensuring that it doesn't repeat.

### DIFF
--- a/warehouse/models/staging/payments/littlepay/stg_littlepay__micropayment_adjustments.sql
+++ b/warehouse/models/staging/payments/littlepay/stg_littlepay__micropayment_adjustments.sql
@@ -40,6 +40,15 @@ add_keys_drop_full_dupes AS (
     {{ qualify_dedupe_full_duplicate_lp_rows() }}
 ),
 
+drop_additional_dupes AS (
+    SELECT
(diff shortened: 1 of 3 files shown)


charlie-costanzo (approved):
looks good to me, runs locally and tests pass - thanks for taking care of this @SorenSpicknall!

PR #3052, opened 2023-10-26
Take most recent authorisation with non-null status for aggregation

SorenSpicknall:
# Description

This changes the `int_payments__latest_authorisations_by_aggregation` to disregard authorisation rows that lack a status, allowing us to move down the line to other models like `fct_payments_aggregations` with a "latest" authorisation that reflects useful information.

For instance, some aggregations have a series of authorisation denials followed by a null-status authorisation record. We want to be able to interpret these rows as non-authorised in downstream models, but can't interpret from the null status of the latest record alone.

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Tested via local dbt runs and tests

## Post-merge follow-ups

- [ ] No action required

The changed model will need a rerun.


lauriemerrell (commented):
Did you still want to add a flag that there *was* a row with null status later and maybe get its datetime or something?

SorenSpicknall:
> Did you still want to add a flag that there _was_ a row with null status later and maybe get its datetime or something?

I added a flag - I didn't want to overburden the `fct_payments_aggregation` model with columns that wouldn't be used directly in dashboards, so didn't add the subsequent datetime directly. Warehouse users can pull it out of the intermediate authorisations model via aggregation_id. Is that alright?

lauriemerrell (changes requested):
Approach over all looks good & I'm fine with just the boolean but can you add docs in the YAML for `fct_payments_aggregations` for the new column?

lauriemerrell (approved):

PR #3066, opened 2023-11-02
Create new table to be used for Mobility Marketplace's new provider map

allejo:
# Description

This PR is one of many steps towards the Mobility Marketplace (Mobi Mart) [Provider Map Refresh](https://github.com/cal-itp/mobility-marketplace/milestone/23). Currently, the Provider Map/data on Mobi Mart is dated and was created manually years ago. One of the goals of this refresh is to automate the aggregation of this data using the warehouse.

This is my first PR to the warehouse infrastructure, and I would appreciate all feedback on moving forward with this PR. I have a few questions/concerns/thoughts listed below:

1. Should this query exist as a dimension or fact in the warehouse?
2. Should this be a table or a view?
   I am thinking of this being a view since all of the data that is presented via this query is pulled from other tables, so no data is living here. Also, this data should ideally be available in Metabase, so does that impact the choice between being a view or a table, or are there any special considerations needed?
3. Is putting this view under the `mart/transit_database` collection (idk what this namespacing is called) the correct spot? Or should it be moved elsewhere?
4. I'm biased in the YAML documentation for this new table. Are there any columns I should document explicitly, or are the column names self-explanatory?

/cc @thekaveman 

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

image

The agency list provided by @mrose914 has 238 entries and this query results in the same number of rows with the same agencies.

## Post-merge follow-ups

- [x] No action required

### DIFF
--- a/warehouse/models/mart/transit_database/_mart_transit_database.yml
+++ b/warehouse/models/mart/transit_database/_mart_transit_database.yml
@@ -539,10 +539,16 @@ models:
       - name: _valid_to
   - name: dim_mobility_mart_providers
     description: |-
-      Listing of all Transit Agencies (TAs) in California with respective data fields needed for the Mobility Marketplace provider map.
+      Listing of all Transit Agencies (TAs) in California with respective data fields needed for the Mobility Marketplace provider map. The number of TAs in this table should match up with the number of TAs that service California as per the National Transit Database (NTD).
+
+      Some information in this table may not be directly related to the TA itself, but rather the parent organization that the TA belongs to. Columns that are affected by this logic are:
+
+      - counties_served
+      - funding_sources
+      - gtfs_schedule_uris
     columns:
       - name: agency_name
-        description: The name of the TA as specified in the National Transit Database (NTD).
+        description: The name of the TA as specified in NTD.
       - <<: *org_ntd_id
         tests:
           - not_null
@@ -552,7 +558,7 @@ models:
       - name: hq_county
         description: The county where this TA's headquarters is located.
       - name: counties_served
-        description: An array of counties that are serviced by the parent organization this TA belongs to.
+        description: A semi-colon delimited list of counties that are serviced by the parent organization this TA belongs to; this TA may not service all counties listed.
       - name: agency_website
         description: The website for this TA.
       - name: caltrans_district_id
@@ -561,14 +567,14 @@ models:
       - <<: *public_currently_operating
         name: is_publicly_operating
       - name: funding_sources
-        description: An array of funding program names that this TA's parent organization receives.
+        description: A semi-colon delimted list of numerical funding program names that this TA's parent organization receives; all non-numerical (e.g. `Public`, `Caltrans`) funding programs are excluded in this column.
       - name: vehicles_at_max_service
         description: (abbreviated as "voms") The number of revenue vehicles operated to meet the annual maximum service requirement. This is the revenue vehicle count during the peak season of the year on the week and day that maximum service is provided.
       - name: on_demand_vehicles_at_max_service
         description: |-
           (abbreviated as "voms_dr") The number of revenue vehicles operated to meet the annual maximum service requirement for NTD categories: Demand Response (DR) and Demand Response Taxi (DT).
       - name: gtfs_schedule_uris
-        description: An array of URLs for GTFS schedules this TA's parent organization has under its umbrella of services.
+        description: A semi-colon delimited list of URLs for GTFS schedules this TA's parent organization has under its umbrella of services; there may be feeds that do not belong to this TA but instead from a sibling service provider within the same parent organization.
   - name: bridge_components_x_properties_and_features
     description: |
       Mapping table between components and properties_and_features.
--- a/warehouse/models/mart/transit_database/_mart_transit_database.yml
+++ b/warehouse/models/mart/transit_database/_mart_transit_database.yml
@@ -543,7 +543,10 @@ models:
     columns:
       - name: agency_name
         description: The name of the TA as specified in the National Transit Database (NTD).
-      - *org_ntd_id
+      - <<: *org_ntd_id
+        tests:
+          - not_null
+          - unique
       - name: hq_city
         description: The city where this TA's headquarters is located.
       - name: hq_county
--- a/warehouse/models/mart/transit_database/_mart_transit_database.yml
+++ b/warehouse/models/mart/transit_database/_mart_transit_database.yml
@@ -539,14 +539,13 @@ models:
       - name: _valid_to
   - name: dim_mobility_mart_providers
     description: |-
-      Hello world
+      Listing of all Transit Agencies (TAs) in California with respective data fields need for the Mobility Marketplace provider map.
     columns:
       - name: agency_name
       - *org_ntd_id
       - name: hq_city
       - name: hq_county
       - name: counties_served
-        description: A semi-colon delimited list of counties this agency serves
       - name: agency_website
       - name: caltrans_district_id
       - name: caltrans_district_name
@@ -554,11 +553,9 @@ models:
       - <<: *public_currently_operating
         name: is_publicly_operating
       - name: funding_sources
-        description: A semi-colon delimited list of funding sources
       - name: on_demand_vehicles_at_max_service
       - name: vehicles_at_max_service
       - name: gtfs_schedule_uris
-        description: A semi-colono delimited list of GTFS schedule URIs
   - name: bridge_components_x_properties_and_features
     description: |
       Mapping table between components and properties_and_features.
--- a/warehouse/models/mart/transit_database/_mart_transit_database.yml
+++ b/warehouse/models/mart/transit_database/_mart_transit_database.yml
@@ -74,7 +74,8 @@ models:
     tests: *mutually_exclusive_ranges
     columns:
       - *key
-      - name: ntd_id
+      - &org_ntd_id
+        name: ntd_id
         description: |
           The organization's ID in the National Transit Database / FTA agency ID number.
           Data entry processes / sources of truth for this field have changed over time.
@@ -155,7 +156,8 @@ models:
         description: |
           Formula in source system returns true value under same conditions as `public_currently_operating`
           plus additional requirement that service has `fixed_route = true` (for organizations, only at least one managed service must meet this criterion).
-      - name: is_public_entity
+      - &org_is_public_entity
+        name: is_public_entity
         description: |
           Formula in source system returns true if the `organization_type` column has been set to one of the following:
             City/Town,
@@ -535,6 +537,28 @@ models:
       - name: _is_current
       - name: _valid_from
       - name: _valid_to
+  - name: dim_mobility_mart_providers
+    description: |-
+      Hello world
+    columns:
+      - name: agency_name
+      - *org_ntd_id
+      - name: hq_city
+      - name: hq_county
+      - name: counties_served
+        description: A semi-colon delimited list of counties this agency serves
+      - name: agency_website
+      - name: caltrans_district_id
+      - name: caltrans_district_name
+      - *org_is_public_entity
+      - <<: *public_currently_operating
+        name: is_publicly_operating
+      - name: funding_sources
+        description: A semi-colon delimited list of funding sources
+      - name: on_demand_vehicles_at_max_service
+      - name: vehicles_at_max_service
+      - name: gtfs_schedule_uris
+        description: A semi-colono delimited list of GTFS schedule URIs
   - name: bridge_components_x_properties_and_features
     description: |
       Mapping table between components and properties_and_features.
--- a/warehouse/models/mart/transit_database/_mart_transit_database.yml
+++ b/warehouse/models/mart/transit_database/_mart_transit_database.yml
@@ -539,23 +539,33 @@ models:
       - name: _valid_to
   - name: dim_mobility_mart_providers
     description: |-
-      Listing of all Transit Agencies (TAs) in California with respective data fields need for the Mobility Marketplace provider map.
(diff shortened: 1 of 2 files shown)


lauriemerrell (commented):
Welcome to warehouse world @allejo! I'll take a stab at your questions and just a few general inline comments about dbtonic SQL. If it's ok I'd appreciate a second round review on the actual SQL logic in case any of these comments change your approach. 

> Should this query exist as a dimension or fact in the warehouse?

It seems to me that a dimension is appropriate; per [our dbt docs](https://dbt-docs.calitp.org/#!/overview) "naming and types of models" section: 
   > These tables are prefixed with dim_ and are ["typically likened to nouns"](https://docs.getdbt.com/terms/dimensional-modeling#dimensions), i.e., the entity associated with an event. Many of our dimension tables are [type-2 slowly changing dimensions](https://en.wikipedia.org/wiki/Slowly_changing_dimension#Type_2:_add_new_row) and have _valid_from and _valid_to columns that indicate the period of time that that version of a record was in effect.

> Should this be a table or a view? I am thinking of this being a view since all of the data that is presented via this query is pulled from other tables, so no data is living here. Also, this data should ideally be available in Metabase, so does that impact the choice between being a view or a table, or are there any special considerations needed?

All models added to existing `mart_` schemas will be available in Metabase regardless of materialization type, but we usually try to have models in `mart` data schemas be materialized as tables precisely because they are accessible in Metabase (see dbt's [explanation of the reasoning here](https://docs.getdbt.com/docs/build/materializations#table)). In this case you're right that it probably does not make much of a material difference but for convention/consistency I'd probably just make it a table. 

> Is putting this view under the mart/transit_database collection (idk what this namespacing is called) the correct spot? Or should it be moved elsewhere?

Yes this is appropriate since it looks like it's basically all transformations on other data in this schema. (Folders roughly match to [schemas](https://docs.getdbt.com/reference/resource-configs/schema#definition) i.e. BigQuery datasets.)

> I'm biased in the YAML documentation for this new table. Are there any columns I should document explicitly, or are the column names self-explanatory?

Ideally all columns would be documented explicitly simply because the warehouse has all sorts of users who may stumble across something and it's preferable to have things written explicitly. If a column is inherited from something upstream you can use YAML anchors as you did already, but if you are calculating things or creating new columns it's definitely ideal to have their logic spelled out. Can be as simple as something like `column_1 from dim_organizations where populated with fallback to column_1_but_different from dim_services`.

warehouse/models/mart/transit_database/dim_mobility_mart_providers.sql
```diff
@@ -0,0 +1,125 @@
...
+-- 2018), so let's narrow down that table to *just* the most recent year so
+-- we don't get duplicate rows for each year.
+--
+-- We cannot use `_is_current` here because every year is marked as "current"
+-- since it's the "current" record for the respective year.
+annual_ntd AS (
+    SELECT *
+    FROM mart_ntd.dim_annual_ntd_agency_information
+    WHERE
+        year = (SELECT MAX(year) FROM mart_ntd.dim_annual_ntd_agency_information)
```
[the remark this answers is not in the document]
allejo: Thank you for this!! I had found the `QUALIFY` statement, but I couldn't figure out how to use it here. For this query, I want to match with *just* NTD reporters from the current year, so I used what you suggested,

```sql
QUALIFY DENSE_RANK() OVER(ORDER BY year DESC) = 1
```

warehouse/models/mart/transit_database/dim_mobility_mart_providers.sql
```diff
@@ -0,0 +1,125 @@
+-- An Organization (i.e. dim_organizations) can provide multiple different
+-- Services (i.e. dim_services). Each Service will have its own funding sources,
+-- GTFS feeds, and target counties.
+
+-- This query will collect every funding source an Organization has via the
+-- services it provides.
```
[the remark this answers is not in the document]
allejo: Thanks for pointing me in the direction of `ref()`, that helped a lot, and noticed it being used everywhere else. I did run into an issue with `dim_organizations` not being found though.

```
Completed with 1 error and 0 warnings:

Runtime Error in model dim_organizations (models/mart/transit_database/dim_organizations.sql)
  404 Not found: Dataset cal-itp-data-infra-staging:vlad was not found in location us-west2
```

warehouse/models/mart/transit_database/dim_mobility_mart_providers.sql
```diff
@@ -0,0 +1,160 @@
...
+    WHERE state = "CA"
+
+    -- We only want data from the latest data from NTD. In the rare edge case
+    -- that an NTD ID is removed and no longer exists in the current year's
+    -- dataset, then replace `DENSE_RANK()` with `ROW_NUMBER()` and add a
+    -- `PARTITION BY ntd_id` inside of the `OVER()`.
+    QUALIFY DENSE_RANK() OVER(ORDER BY year DESC) = 1
+),
+
+gtfs_data_by_organization AS (
```
lauriemerrell, line 78: I think you might be able to replace this (and maybe the orgs to services join above as well?) with a reference to [`dim_provider_gtfs_data`](https://dbt-docs.calitp.org/#!/model/model.calitp_warehouse.dim_provider_gtfs_data) which might just simplify things a little (.... sorry for not suggesting this earlier 😅)

warehouse/models/mart/transit_database/dim_mobility_mart_providers.sql
```diff
@@ -0,0 +1,160 @@
...
+        orgs.key
+),
+
+mobility_market_providers AS (
+    SELECT
+        annual_ntd.agency_name AS agency_name,
+        annual_ntd.ntd_id AS ntd_id,
+        annual_ntd.city AS hq_city,
+        orgs_hq_bridge.county_geography_name AS hq_county,
+        ARRAY(
```
lauriemerrell, line 117: Just confirming -- the downstream use case for this is programmatic ingestion of some kind (to make the map)? Asking because we sometimes try to avoid array-type columns in mart tables because Metabase doesn't handle arrays well -- if there's a goal for this table to be useful in Metabase, it may be easier to make a bridge table for the three array columns (counties, funding sources, and schedule URIs) and leave the orgs/NTD mapping in the `dim`. But if there's basically one defined non-Metabase use case it might be ok to leave the arrays.

warehouse/models/mart/transit_database/dim_mobility_mart_providers.sql
```diff
@@ -0,0 +1,160 @@
...
+            ORDER BY source
+        ) AS funding_sources,
+        annual_ntd.voms_do AS on_demand_vehicles_at_max_service,
+        annual_ntd.total_voms AS vehicles_at_max_service,
+        ARRAY(
+            SELECT DISTINCT
+            uris
+            FROM UNNEST(gtfs_data_by_organization.gtfs_uris) uris
+            ORDER BY uris
+        ) AS gtfs_schedule_uris
```
lauriemerrell, line 141: Ok this is a very weird quirk of our data model, but I think that `uri` may not be the best field here, depending on what you're aiming for; the field that Cal-ITP maintains for data scraping is `pipeline_url` but unfortunately that's not available in the mart -- you can recover it by decoding `base64_url` using our macro: `{{ from_url_safe_base64(base64_url) }}`. But I am not sure if the `uri` field is actually maintained at this point... I will write a ticket to consider removing it from the model because it is deceptive.

warehouse/models/mart/transit_database/dim_mobility_mart_providers.sql
```diff
@@ -0,0 +1,160 @@
...
+   WHERE _is_current
+),
+
+services AS (
+    SELECT * FROM {{ ref('dim_services') }}
+    WHERE _is_current
+),
+
+county_geogs AS (
+    SELECT * FROM {{ ref('dim_county_geography') }}
```
lauriemerrell, line 18: It may be worth adding the `_is_current` filter on all these just to future proof

lauriemerrell (changes requested):
Changes look great, thank you! Just a few comments at this point, I think the `uri` one is potentially blocking, the `array` one is mostly a question about your use case, and then the rest are stylistic/simplification proposals

warehouse/models/mart/transit_database/dim_mobility_mart_providers.sql
allejo: In this case, I used an array because I thought Metabase could handle them. The use case for this data is on the Mobility Marketplace website, which will make this data available as a CSV file and the array values will be flattened + delimited by semi-colons. So I can flatten this data in this table if Metabase isn't going to handle it well, is that how I should proceed?

lauriemerrell: Yeah, that (flattening) might be preferable -- for strings in Metabase you can do a `contains` filter but Metabase can't filter on arrays at all. Here's an [example table in Metabase](https://dashboards.calitp.org/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7ImRhdGFiYXNlIjoyLCJxdWVyeSI6eyJzb3VyY2UtdGFibGUiOjIzODAsImZpZWxkcyI6W1siZmllbGQiLDE4Nzg4LG51bGxdLFsiZmllbGQiLDE4NzkyLG51bGxdLFsiZmllbGQiLDE4Nzg2LG51bGxdXX0sInR5cGUiOiJxdWVyeSJ9LCJkaXNwbGF5IjoidGFibGUiLCJ2aXN1YWxpemF0aW9uX3NldHRpbmdzIjp7fX0=) with a few array columns if you want to test what kinds of functionality you'd want.

lauriemerrell: [Here's a few examples](https://github.com/search?q=repo%3Acal-itp%2Fdata-infra%20array_to_string&type=code) of using the `ARRAY_TO_STRING` function that I think you'd want to use

warehouse/models/mart/transit_database/dim_mobility_mart_providers.sql
```diff
@@ -0,0 +1,163 @@
...
+    WHERE
+        gtfs_schedules.base64_url IS NOT NULL
+    GROUP BY
+        orgs_x_services_x_gtfs.organization_key
+),
+
+serviced_counties_by_org AS (
+    SELECT
+        orgs.organization_key AS org_key,
+        ARRAY_CONCAT_AGG(services.operating_counties) AS operating_counties
```
lauriemerrell, line 97: This one is the most ambiguous re: whether it's worth doing the `distinct` here or later

warehouse/models/mart/transit_database/_mart_transit_database.yml
```diff
@@ -535,6 +537,35 @@ models:
       - name: _is_current
       - name: _valid_from
       - name: _valid_to
+  - name: dim_mobility_mart_providers
+    description: |-
+      Listing of all Transit Agencies (TAs) in California with respective data fields needed for the Mobility Marketplace provider map.
+    columns:
+      - name: agency_name
+        description: The name of the TA as specified in the National Transit Database (NTD).
+      - *org_ntd_id
```
lauriemerrell, line 546: I'm wondering if you might want unique and not null tests on this column, since my understanding is that you want one row per NTD ID

lauriemerrell (approved):
All remaining comments are optional so defer to you on whether you want to do them (and I'm happy to review again if you make changes and want a final pair of eyes). Most important would probably be adding uniqueness tests somewhere to monitor that the table is reflecting the granularity that you want.

warehouse/models/mart/transit_database/dim_mobility_mart_providers.sql
```diff
@@ -0,0 +1,163 @@
...
+
+services_x_funding AS (
+    SELECT * FROM {{ ref('bridge_services_x_funding_programs') }}
+    WHERE _is_current
+),
+
+gtfs_schedules AS (
+    SELECT
+        key,
+        {{ from_url_safe_base64('base64_url') }} AS base64_url
```
lauriemerrell, line 37: [belated] Just noting I found this a bit confusing since at this point this is no longer base64 encoded, might just call it `schedule_url` or similar?

warehouse/models/mart/transit_database/dim_mobility_mart_providers.sql
```diff
@@ -0,0 +1,163 @@
...
+    SELECT * FROM {{ ref('dim_provider_gtfs_data') }}
+    WHERE _is_current
+),
+
+-- This query will collect every funding source an Organization has via the
+-- services it provides.
+funding_by_org AS (
+    SELECT
+        orgs.organization_key AS org_key,
+        ARRAY_AGG(services_x_funding.funding_program_name) AS funding_sources
```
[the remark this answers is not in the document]
allejo: How did I not realize that `ARRAY_AGG` had a `DISTINCT` option??? This was exactly what I was looking for but had to settle with the current approach I've got. Definitely going to change this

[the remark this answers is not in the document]
lauriemerrell: The way the BigQuery documentation displays the options is not intuitive to me

lauriemerrell (approved):
LGTM! Thank you for iterating through this! 

One possible optional nitpick would be if you want to perhaps a bit more context to the documentation, specifically on filtering logic (so things like: will only list numeric funding programs, will only list organizations with services that have some funding information). Basically I like to try to document any `where` clause or `inner` join so people understand what subset of the upstream data they should expect to see. But it's not a huge deal, especially since you do have good inline comments in the SQL.

allejo:
I added a few additional notes in 4de938ac15ff18a404a210e9c1e1c66bb1804dbb that should explain some of the logic and design of the data 😄 Thank you for all the feedback and review, I learned a lot from this PR

PR #3070, opened 2023-11-02
Remove Principal Customer ID Tests W/Poor Assumptions

### LINKED ISSUE #3039: Payments: Some Littlepay Customers Have Improper Principal Links
**Describe the bug**

We bring a `principal_customer_id` through from `stg_littlepay__customer_funding_source` to other models. In doing so, we assume that a principal customer ID can only exist one level above a given customer ID - if a customer ID is a principal to another customer ID, it _must_ have itself as its principal customer ID. This doesn't seem to hold in some of our data, as evidenced by Sentry issues [#71485](https://sentry.calitp.org/organizations/sentry/issues/71485/) and [#71486](https://sentry.calitp.org/organizations/sentry/issues/71485/).

**To Reproduce**

The linked Sentry issues provide some test queries that can be run to surface the unexpected relationships.

**Expected behavior**

We initially interpreted these relationships through guesswork, so it's possible that multiple levels of principal customer ID relationships _could_ exist. However, we don't believe that should be the case. Ideally, we should be able to identify customer ID relationships to correct or to suppress (if we determine that we have flawed source data, malformed or used for testing purposes), as we did in #3021.

SorenSpicknall:
# Description

As described in #3039, we currently experience two daily Sentry alerts corresponding to four Littlepay customer records that are apparently the principal customer for a different record _and_ subject to a different principal customer relationship, which we previously believed shouldn't happen.

Investigating based on an updated Littlepay dictionary, it appears that this relationship usually stems from a transfer of the principal relationship when a new card (represented by `funding_source_vault_id`) is associated with a preexisting registered customer, and then transferred to a different registered customer down the line:

1. Customer 1 is created and configured with Card A, which has customer_id `a` and principal_customer_id `a`.
2. Customer 1 registers a second card, Card B, which automatically generates two rows in Littlepay data: one with customer_id `a` and principal_customer_id `a`, and another with customer_id `b` and principal_customer_id `a`. customer_id `b` is never used on its own anywhere else. This seems to happen because Card B was not previously registered or used anywhere else in Littlepay's systems - it gets assigned a `customer_id` even _as_ it's being actively assigned to a different principal customer.
3. A new customer, Customer 2, is created separately and eventually takes over registration of both cards. This instantaneously generates four new `customer_funding_source` rows - Card 1 Customer `c` Principal `c`, Card 2 Customer `c` Principal `c`, Card 1 Customer `a` Principal `c`, and Card 2 Customer `a` Principal `c`. Because all these cards were preexisting in Littlepay systems, no new `customer_id` is generated for them.
4. Our downstream logic in `int_littlepay__customers` takes the latest exported rows for each combination of `funding_source_vault_id` (the card) and `customer_id` (the associated customer), and finds distinct combinations of `customer_id` and `principal_customer_id`, giving us these rows - Customer `b` Principal `a`, Customer `a` Principal `c`, and Customer `c` Principal `c`.

The test fails because customer_id `b` doesn't follow the registration to the new Customer 2, the way that customer_id `a` does. In the example above, this appears to be because of an unused customer record implicitly created by Littlepay systems but not actively used by anything.

However, that circumstance is not exhaustive: there are a handful of transactions associated with cards that don't seem to show up in Littlepay data for the first time until after their first use - they have micropayments associated with their initial, implicit customer ID, which becomes associated with a different customer later (the first time the customer ID appears in customer funding vaults data we have access to). An initial search seems to indicate that this isn't due to missing exports in our pipeline, so we may need to approach the vendor about it. See:

```
select * from cal-itp-data-infra.staging.stg_littlepay__micropayments
where _key in ('4d0d1fb75353bfd0b9e37782eda84081', '112741fda9f178d780e2bc72609c0c13')
```

Because of uncertainty around cause for these two rows, and a general possible utility to tracking the full lineage of a given customer record, it seems most desirable to discard these two tests and permit the two-layer relationships that the test attempted to flag (alternatively, we could turn these into proportional threshold tests if desired). We could add logic to discard or remap the "dangling" rows for the two circumstances described above, but that would lose fidelity of information about how a customer record has traversed through different principal relationships over time, so I'd recommend against it.

Resolves #3039

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Relevant dbt tests run locally to make sure removing these tests hasn't done anything unexpected.

## Post-merge follow-ups

- [ ] No action required

Resolve associated Sentry tests, and create a new Issue to more thoroughly investigate the two customer records that we seem to be missing some initial funding source creation records for.


lauriemerrell (commented):
I'm kind of conflicted about this. I would like us to consider reaching out to Littlepay about this. Because looking at the data, I am wondering: Should we just manually map the problematic customer to its principal's principal in `int_customers`, rather than adjusting the logic in the test? Because if the purpose of the test is more about ensuring integrity in `fct_payments_rides`, we want to actually manage the relationship rather than just suppress the test. Because I'm not really sure that conceptually this situation is that different than customers who sometimes have themselves as principals and other times have a different principal, and right now we just take the latest such relation in `int_customers`.

SorenSpicknall:
@lauriemerrell, after our discussion earlier today, I separated out the failing rows into two different circumstances. I currently recommend getting rid of the tests entirely based on what I found, but we could also draw a line from one principal to another as an alternative. I've outlined the tradeoffs in an updated PR description above.

lauriemerrell (approved):
I am good to remove these tests. Only question is whether it's worth updating the documentation for this field to perhaps link to this investigation or outline any of the considerations here. (I am thinking about future discoverability.)

PR #3110, opened 2023-11-14
DAG: copy_prod_archiver_configs_to_test

### LINKED ISSUE #3057: RT Archiver: Copy prod configs to test occasionally
As a Cal-ITP GTFS data user, I want the test RT archiver not to report false-alarm Sentry issues because of outdated configurations so that I don't waste time troubleshooting something that isn't a real issue. 

Today we ran into an issue where the RT archiver test instance was using a really old config file that contained deprecated URLs. We don't want to run a service that's hitting unused URLs. The challenge is that the test archiver [looks](https://github.com/cal-itp/data-infra/blob/main/services/gtfs-rt-archiver-v3/docker-compose.yml#L8) in the `test-calitp-gtfs-download-config` bucket for its configs, and we don't regularly update the data in there.

We could consider setting up a ~weekly job to copy the latest prod config into the test bucket so the test archiver doesn't get *too* out of sync.

charlie-costanzo:
# Description

To avoid false-alarm Sentry issues because of outdated configurations, this PR sets up a weekly job to copy the latest prod archiver config into the test bucket so the test archiver doesn't get too out of sync.

Resolves #3057 

## Type of change

- [x] New feature

## How has this been tested?
in local airflow
Screenshot 2023-11-14 at 5 37 26 PM

## Post-merge follow-ups

new composer environment variables need to be added by a user with appropriate permissions, I (Charlie) do not have them


airflow/dags/copy_prod_archiver_configs_to_test/METADATA.yml
```diff
@@ -0,0 +1,20 @@
...
+tags:
+  - all_gusty_features
+default_args:
+    owner: airflow
+    depends_on_past: False
+    start_date: !days_ago 1
+    email:
+      - "evan.siroky@dot.ca.gov"
+      - "hunter.owens@dot.ca.gov"
+      - "jameelah.y@jarv.us"
```
lauriemerrell, line 12: Would remove Jameelah & Andrew

airflow/dags/copy_prod_archiver_configs_to_test/copy_prod_archiver_configs_to_test.py
```diff
@@ -0,0 +1,55 @@
+from calitp_data_infra.storage import GTFSDownloadConfigExtract, get_latest
+from google.cloud import storage
+
+prod_bucket = "calitp-gtfs-download-config"
```
lauriemerrell, line 4: I would consider defining these as environment variables in the docker-compose YAML so that they're configurable

airflow/dags/copy_prod_archiver_configs_to_test/copy_prod_archiver_configs_to_test.py
```diff
@@ -0,0 +1,55 @@
+from calitp_data_infra.storage import GTFSDownloadConfigExtract, get_latest
+from google.cloud import storage
```
lauriemerrell, line 2: Other places we use [get_fs](https://github.com/cal-itp/data-infra/blob/main/packages/calitp-data-infra/calitp_data_infra/storage.py#L49-L55) which I think handles a bit of the local vs. cloud auth logic, and it uses [`gcsfs`](https://gcsfs.readthedocs.io/en/latest/api.html) which has a `copy` function that I think we can just use

airflow/dags/copy_prod_archiver_configs_to_test/README.md
```diff
@@ -0,0 +1,9 @@
+# `copy_prod_archiver_configs_to_test`
+
+Type: [Now / Scheduled](https://docs.calitp.org/data-infra/airflow/dags-maintenance.html)
+
+This DAG performs a weekly job to copy the latest prod config into the test bucket so the test archiver doesn't get too out of sync.
+
+## Secrets
```
lauriemerrell, line 7: Curious about this - I thought that the test archiver uses prod secrets :eyes:

lauriemerrell (commented):
Skeleton looks good, a few inline comments -- happy to chat offline!

lauriemerrell (approved):
LGTM! 🎉

PR #3129, opened 2023-11-18
New pipeline to load BlackCat API data

kengie:
# Description

This code creates a new data pipeline with 2 dags to ingest NTD submitted forms from BlackCat, a third-party vendor, into the Cal-ITP data warehouse. The first dag hits the API and saves all of its contents into one json zipped file in Google Cloud Storage, using the new operator `blackcat_to_gcs.py`. This code follows the same structure as the [`airtable_to_gcs.py`](https://github.com/cal-itp/data-infra/blob/main/airflow/plugins/operators/airtable_to_gcs.py) operator. Hive partitions are created by date and timestamp, and date and timestamp also make up part of the file path to each report's jsonl zipped file. Even though it is a small amount of data and the pipeline will usually not run more than once per day, we included hive-partitioning by timestamp in the event that it is run > 1x/day so that filepaths will differ for each run. Otherwise, the Airflow job will fail since it cannot overwrite a JSONL file of the same name.

Secondly, we create an external table so it can be viewed in BigQuery. This is done just by adding a new folder in the `create_external_tables` dag with YAMLs for each table. They call upon the existing [`external_table.py` ](https://github.com/cal-itp/data-infra/blob/main/airflow/plugins/operators/external_table.py)operator.

Thirdly, we have 3 levels of dbt models to transform this data and conduct some validation checks on the submitted answers. 
- The `staging/ntd_validation` folder has SQL statements that take the API data from the external table and unnest the columns into individual tables (In the external table, each NTD table from the API is a nested column). 
- The `intermediate/ntd_validation` folder has dbt models (either SQL or python) that further massage the data into the form needed for the final validation checks.
- the `mart/mtd_validation` folder has validation checks for RR-20. They are all of those that are in the [`rr20_service_check.py`](https://github.com/cal-itp/ntd-modernization/blob/main/validation_tool/rr20_service_check.py) script and the ['rr20_financials_check.py`](https://github.com/cal-itp/ntd-modernization/blob/main/validation_tool/rr20_financials_check.py) scripts, except for the one that checks against the vehicle inventory since the API doesn't have that info yet. 

I also added columns for the date checked and added the NTD error ID into the validation tables.

image

Resolves #\[issue\]

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

The new `blackcat_to_gcs.py` operator have been tested using local Airflow, successfully writing JSONL files to GCS. One can view the API data saved in test runs, in the test GCS folder  `test-calitp-ntd-report-validation` The dbt models have been tested using local dbt, and exist in BigQuery on the `cal-itp-data-infra-staging` project in the `staging_staging` database.

## Post-merge follow-ups

- [ ] No action required

This Airflow DAG will remain turned off. The NTD reporting season officially closed on October 31, so there is no purpose in keeping it running since it only pulls NTD-related data from BlackCat. @csuyat-dot on Cal-ITP team will own this pipeline in 2024, and turn it on when sub-recipients start submitting 2024 data, around September 2024.


airflow/plugins/operators/blackcat_to_gcs.py
```diff
@@ -0,0 +1,193 @@
...
+        self.logger.info(
+            f"Downloading BlackCat data for {self.extract_time.format('YYYY')}_{self.bq_table_name}."
+        )
+        response = requests.get(self.api_url, verify=False)
+        blob = response.json()
+
+        org_data = []
+
+        # Cyling through and processing each org's data
+        for x in blob:
```
evansiroky, line 86: This seems like it deviates from some other patterns we have when downloading raw data. In some other DAGs we save the raw JSON or GTFS files to cloud storage before extracting anything from them. After saving the raw contents, there can be further operations that happen on top of that. If only a part of the JSON is saved after getting a response from the server, any other elements in the JSON file are lost forever. 

In a much earlier version of the pipeline we downloaded GTFS data and extracted the files instead of saving the entire zip file. It came back to bite us when we wanted to examine the entire zip file.

airflow/dags/ntd_report_validation/a10_submitted_for_ntd.yml
```diff
@@ -0,0 +1,7 @@
+operator: operators.BlackCatApiToGCSOperator
+
+bucket: "gs://calitp-ntd-report-validation"
+api_url: "https://services.blackcattransit.com/api/APIModules/GetNTDReportsByYear/BCG_CA/2023"
```
evansiroky, line 4: This appears to be the same URL for each DAG task which means the same file is downloaded multiple times. To build on top of [my comment about the downloading and extracting of the data](https://github.com/cal-itp/data-infra/pull/3129/files#r1398015231), it seems like there should be a DAG task to download this file once. Also, a single external table could be used as source data for dbt modeling.

warehouse/models/intermediate/ntd_validation/int_ntd_rr20_service_ratios.py
```diff
@@ -0,0 +1,71 @@
...
+    stream_handler = logging.StreamHandler()
+    stream_handler.setFormatter(formatter)
+
+    if not logger.hasHandlers():
+        logger.addHandler(file_handler)
+        logger.addHandler(stream_handler)
+
+    return logger
+
+def make_ratio_cols(df, numerator, denominator, col_name, logger, operation="sum"):
```
lauriemerrell, line 28: I am a little confused why this is a Python model; it seems that these ratio operations can be performed relatively straightforwardly in SQL (generally a Python model is only used if the relevant operations cannot be performed in SQL or would be very verbose)

warehouse/models/mart/ntd_validation/fct_ntd_rr20_service_checks.py
```diff
@@ -0,0 +1,223 @@
...
+    voms0_check = check_single_number(allyears, 'VOMX', this_year, last_year, logger)
+
+    # Combine checks into one table
+    rr20_checks = pd.concat([cph_checks, mpv_checks, vrm_checks, 
+                             frpt_checks, rev_speed_checks, 
+                             tph_checks, voms0_check], 
+                             ignore_index=True).sort_values(by="Organization")
+    
+    ## Part 1: save Excel file to GCS
+    GCS_FILE_PATH_VALIDATED = f"gs://calitp-ntd-report-validation/validation_reports_{this_year}"
```
lauriemerrell, line 181: As noted in Slack, the intention for a Python model in dbt is *not* to output files to other locations but is [simply to perform operations on data models](https://docs.getdbt.com/docs/build/python-models#defining-a-python-model) (totally akin to SQL models). Exports should be orchestrated using Airflow

warehouse/models/mart/ntd_validation/fct_ntd_rr20_service_checks.py
```diff
@@ -0,0 +1,223 @@
...
+
+    if not logger.hasHandlers():
+        logger.addHandler(file_handler)
+        logger.addHandler(stream_handler)
+
+    return logger
+
+
+def check_rr20_ratios(df, variable, threshold, this_year, last_year, logger):
+    '''Validation checks where a ratio must be within a certain threshold limit
```
lauriemerrell, line 29: Is the goal to create columns with these ratio checks, or to treat this as a validation that the rows are valid? (I.e., do you want to have a column with a boolean value for pass/fail, or do you want to just alert if some fail? If the latter, could [leverage dbt tests](https://github.com/dbt-labs/dbt-utils#accepted_range-source).)

lauriemerrell (commented):
Thanks Kim! Sorry for delayed review since I was out of office. I agree with Evan's comments above and have a few more -- I mostly just want to clarify the intention for using Python models. While we do technically support them, we don't actually use them anywhere in the project yet and it seems like most of the operations here can be performed in SQL in ways that might be easier to maintain (since it would be more in keeping with the rest of the warehouse.)

lauriemerrell:
One stray comment -- I see that the `lint` GitHub Action is failing, you may want to [use pre-commit locally](https://github.com/cal-itp/data-infra#linting-and-type-checking) to resolve any linting issues

kengie:
Thanks @lauriemerrell , @evansiroky , I updated the code to pull everything from the API (which is indeed just one URL, and then save it as one external table. In this table, each NTD report is one a nested column in BigQuery. Then, dbt models will parse out the reports into separate tables. 

In terms of python vs SQL, at this point its a partially a pragmatic action to start out with, since the python is already written and understood by @csuyat-dot who will be taking over, and I have about one week of dev time left on this project. My intention is to transfer all of the checks into dbt first, then as time allows convert as much python as possible into SQL. Imperfect but ensures coverage.

warehouse/models/mart/ntd_validation/fct_ntd_rr20_service_checks.py
kengie: moving over to a different script, will try to deploy with Airflow as suggested.

warehouse/models/intermediate/ntd_validation/int_ntd_rr20_service_ratios.py
KatrinaMKaiser: Confirming that @csuyat-dot and I talked to Kim about translating the ratio calculation step to SQL next year, Kim will prioritize automating the other checks.

airflow/plugins/operators/blackcat_to_gcs.py
kengie: changed as suggested!

airflow/dags/ntd_report_validation/a10_submitted_for_ntd.yml
kengie: changed as suggested

evansiroky:
It looks like there are other eyes on this. I like the refactor of downloading the raw data. Also, I generally agree with @lauriemerrell that dbt should be leveraged as much as possible to keep the coding style similar to the rest of the warehouse scripts.

KatrinaMKaiser (approved):

lauriemerrell (approved):

Issue #3111, opened 2023-11-14
Investigate differences between on/off taps and ask Littlepay about them 


lauriemerrell:
As an analytics engineer, I want to understand cases where key attributes differ between on and off taps in a variable fare payments transaction so that I can handle the differences appropriately in modeling. 

This is a follow up to #2663. 

AC:
- [ ] Investigate specifically which fields currently differ between tap on/tap off transactions
- [ ] Ask Littlepay about the interpretation of these differences 
- [ ] Write follow up tickets as needed based on responses

PR #3159, opened 2023-11-30
Refactor payments rides for code readability and modularity: device transactions portion

### LINKED ISSUE #2959: Break up payments rides
`fct_payments_rides_v2` and its immediate parents could use some refactoring for clarity, testability, and reusability.

* All inner joins in payments rides should be reconsidered or at least commented to explain why they are appropriate. (Are upstream not null tests on the relevant key(s) always passing? Risk of inner joins is dropping rows where a key is null on one side; are we sure that this is not happening, or that if it did happen it would be appropriate?) 

* CTEs in payments rides should be evaluated for whether they can be migrated to standalone intermediate models that can be tested separately (for example, with row equality tests or foreign key tests or other validity checks). For example, the first/second transactions logic should probably become a separate intermediate model because it has some pretty complicated transformations that should be tested for validity.

### LINKED ISSUE #2960: Make a standalone `int_payments__micropayments` and/or `int_payments__micropayment_device_transactions` model
Related to #2959, we have a few operations that are spread out across a few models that probably should be brought together into one unified micropayments or micropayments device transactions model. 

* `int_littlepay__cleaned_micropayment_device_transactions` -- the goal here seems to be to dedupe micropayments that are associated with the same LP transaction
* The above seems pretty related to the suggestion in #2959 to match on/off micropayments for the same ride somewhere other than a CTE inside payments rides
* And both are related to `int_littlepay__device_transaction_types`, which seems like it could pretty easily be refactored to use `CASE` statements on the relevant columns instead of doing a union... Case statements seem much clearer

It seems like all of these modifications could happen in the context of a unified intermediate table that basically cleans up micropayments and/or micropayment device transactions into a useable form for downstream use.

lauriemerrell:
# Description

Continue work started in #3123 to refactor the payments rides model for readability, testability, and modularity. 

Specific changes:
* Refactors `int_littlepay__device_transaction_types` to instead just pair up on/off transactions in an intermediate table instead of assigning types and unioning those back together and then doing the pairing in `fct_payments_rides_v2`.
* Moves all intermediate models into the intermediate folder in dbt and renames more generically for consistency. 
* Makes one step towards #3111 by suppressing mismatched device ID from the criteria for the common fields test model -- AFAIK there is not really a reason to expect that someone would always tap out on the save device they tapped in on 

Questions (probably for @charlie-costanzo):
* Are `transaction_type` and `off_transaction_type` used anywhere in Metabase? As written this PR changes the handling of those fields: instead of assigning them based on the associated micropayment type, which was what was done before, this just passes through the original transaction types from the transactions themselves. Can compare the outcomes here using these two queries:
```sql
SELECT 
  charge_type,
  transaction_type,
  off_transaction_type,
  COUNT(*)
FROM `cal-itp-data-infra.mart_payments.fct_payments_rides_v2`
GROUP BY 1,2,3
ORDER BY 1,2,3

SELECT 
  charge_type,
  transaction_type,
  off_transaction_type,
  COUNT(*)
FROM `cal-itp-data-infra-staging.laurie_mart_payments.fct_payments_rides_v2`
GROUP BY 1,2,3
ORDER BY 1,2,3
```
* Do we need the `pending` flag? It's added in the old `_types` model but it doesn't seem to be used anywhere and it's not surfaced in `payments rides`

Resolves #2959 
Resolves #2960 

## Type of change

This is kind of N/A, this is a refactor for technical clarity, very limited functional changes

- [ ] New feature
- [ ] Documentation

## How has this been tested?

`poetry run dbt build -s source:external_littlepay+` passes

Confirmed row count of `fct_payments_rides_v2` is identical between prod & my staging environment

## Post-merge follow-ups

- [ ] No action required

Delete deprecated models (`int_littlepay__*`).


charlie-costanzo (approved):
Sorry for the delay here, I was just spending extra time going through it for my own edification. But this looks great, thank you for taking care of this! As far as your questions:

> Are transaction_type and off_transaction_type used anywhere in Metabase?

No they are not

> Do we need the pending flag? It's added in the old _types model but it doesn't seem to be used anywhere and it's not surfaced in payments rides

We do not

PR #3170, opened 2023-12-07
fix(k8s): increase Sentry Clickhouse storage to 50Gi

themightychris:
# Description

A Grafana alert has been consistently firing as of late about one of Sentry's storage volumes hovering around 80% utilization. Investigating this I found that this was _not_ to do with Zookeeper which [routinely needs old log/snapshot data purged](https://github.com/cal-itp/data-infra/blob/main/runbooks/workflow/disk-space.md), but rather was an issue with Clickhouse's data volume which stores Sentry's events legitimately growing over time. Reducing disk usage here would require actually deleting some of Sentry's event history. Given the long time period covered by the initially-provisioned 30Gi of data storage (the [upstream default](https://github.com/sentry-kubernetes/charts/blob/20.0.0/sentry/values.yaml#L990) for the Helm chart we use to deploy Sentry) and relative cost of storage, I suggest it would be best to just expand the storage volume to 50Gi

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

I executed the following command within the `kubernetes/apps/charts/sentry` directory:

```bash
helm template sentry . -n sentry -f ../../values/sentry_sensitive.yaml -f ./values.yaml --debug
```

This renders the complete manifests offline for our Sentry deployment, and by diffing the before and after output I verified that only the storage request for the `sentry-clickhouse` StatefulSet was changed:

```diff
<           storage: "30Gi"
---
>           storage: "50Gi"
```

## Post-merge follow-ups

Ensure that the following 3 PersistentVolumeClaims change from 30Gi to 50Gi:

- `sentry-clickhouse-data-sentry-clickhouse-0`
- `sentry-clickhouse-data-sentry-clickhouse-1`
- `sentry-clickhouse-data-sentry-clickhouse-2`

It may additionally be necessary to log into the associated pods or nodes and use `resize2fs` to expand the filesystem to make use of the newly available physical disk space. Use `df -h` within each of the three pods to then confirm that the filesystems have 50Gi available and are now well below 80% utilization.

### DIFF
--- a/runbooks/workflow/disk-space.md
+++ b/runbooks/workflow/disk-space.md
@@ -12,7 +12,7 @@ In the expanded details for the alert message posted to Slack, look for a line i
 persistentvolumeclaim=sentry-clickhouse-data-sentry-clickhouse-1,
 ```
 
-Using Lens, you can easily track down what service this PVC is associated with by first opening the **Storage** > **Persistent Volume Claims** section in Lens, then finding or searching for the named PVC. Once found, you can click the associated **Pod** and the **Controlled By** field will tell you what higher-level `Deployment` or `StatefulSet` that pod was created by—telling you which service is affected. The PVC shown above, for example, leads to the `StatefulSet` named `sentry-clickhouse`.
+If you're using [Lens](https://k8slens.dev/), you can easily track down what service this PVC is associated with by first opening the **Storage** > **Persistent Volume Claims** section in Lens, then finding or searching for the named PVC. Once found, you can click the associated **Pod** and the **Controlled By** field will tell you what higher-level `Deployment` or `StatefulSet` that pod was created by—telling you which service is affected. The PVC shown above, for example, leads to the `StatefulSet` named `sentry-clickhouse`.
 
 ## Possible actions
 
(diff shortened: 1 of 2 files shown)


evansiroky (approved):
The logic makes sense, but I don't know enough about kubernetes config to know if the proposed change will work, so I'll trust the process.

runbooks/workflow/disk-space.md
```diff
@@ -2,25 +2,54 @@
...
-## ZooKeeper
+## Identifying affected service
+
+In the expanded details for the alert message posted to Slack, look for a line indicating the `persistentvolumeclaim` (PVC), e.g.
+
+```console
+persistentvolumeclaim=sentry-clickhouse-data-sentry-clickhouse-1,
+```
+
+Using Lens, you can easily track down what service this PVC is associated with by first opening the **Storage** > **Persistent Volume Claims** section in Lens, then finding or searching for the named PVC. Once found, you can click the associated **Pod** and the **Controlled By** field will tell you what higher-level `Deployment` or `StatefulSet` that pod was created by—telling you which service is affected. The PVC shown above, for example, leads to the `StatefulSet` named `sentry-clickhouse`.
```
SorenSpicknall, line 15: Since this is the first mention of Lens in this runbook, I'd modify the initial text to something like "If you're using Lens", and make the word "Lens" into a link to the website for the IDE.

runbooks/workflow/disk-space.md
```diff
@@ -2,25 +2,54 @@
...
+- Reduce disk usage
+- Increase disk capacity
+
+Some services are designed to be able to operate within finite storage and may need intervention to help them do that, see the [Service-specific additional steps](#service-specific-additional-steps) section below for guidance on some such services.
+
+If however you determine that the growth in disk usage is likely legitimate and consisting of data we generally want to retain, it may just be time to increase the requested disk size for the affected volume.
+
+## Increasing disk capacity
+
+In a pinch, you can edit a PVC directly in Kubernetes to see if that resolves an issue. After a PVC's request size is increased, the pod it is bound too will need to be manually destroyed and then the new requested size will be applied by the cloud provider's storage controlled before the volume gets re-bound to the next instance of the pod. It may be additionally necessary to shell into the pod or node it's hosted on and execute `resize2fs` on the mounted volume to resize the filesystem within the volume to fill the newly available space.
```
SorenSpicknall, line 30: "the pod it is bound too" -> "the pod it is bound to"

runbooks/workflow/disk-space.md
```diff
@@ -2,25 +2,54 @@
...
+
+However it is important to apply the new requested disk capacity value upstream in source control so that future syncs from GitHub to Kubernetes will not reset it. This typically involved identifying the relevant Helm chart and configuration value to change within one of this repositories Helm values YAML files. [PR #3170](https://github.com/cal-itp/data-infra/pull/3170) provides an example of this being done for the Clickhouse data volume.
```
SorenSpicknall, line 32: Nit: a comma after "However" would help

SorenSpicknall (approved):
Tentatively approving because my only requested changes are copy changes, and I anticipate you'll be able to make them without requiring a re-review to merge.

PR #3175, opened 2023-12-08
Bugfix/kim ntdvalidation

kengie:
# Description

Firstly sorry about the commits showing in this PR, I tried to rename my old branch locally because I already had commits on it to push, but it ended up showing the whole history here which is confusing. Just ignore all the commits, its the old ones but the real new ones are only since Dec 8.

Resolves #\[issue\]

* fixes intermediate model `int_ntd_rr20_service_ratios` so that the table reflects all the columns that should be calculated. Replace a python file with a simpler SQL one here
* Rearranges some columns in the fct models to how they should be ordered
* add new `int` and `fct` model for the a10 to produce 2 more validation checks that we can do without data from the "prior year", which we don't have 
* delete the half-done dag that was for pushing the BigQuery final fct model tables to Excel. This could be developed in the future though.

- [x] New feature
- [ ] Documentation

## How has this been tested?

Tested locally successfully. 
image

image
 

## Post-merge follow-ups

- [ ] No action required


warehouse/models/mart/ntd_validation/fct_ntd_a10_facilitiescheck.sql
lauriemerrell, line 1: Would be good to have YAML documentation for this new model

lauriemerrell (approved):
I mostly skimmed for general dbt validity -- looks ok at a high level, would prefer to have documentation for the new fct model but wouldn't block merging on that given time crunch.

warehouse/models/mart/ntd_validation/fct_ntd_a10_facilitiescheck.sql
kengie: Good call, I added it to the YAML

PR #3190, opened 2023-12-15
Implement Airtable column renaming within transit data quality issues table

evansiroky:
# Description

Fixes a dbt problem that is occurring due to a column that was deleted and then recreated under a different name.

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

dbt ran locally via the 2 commands:
- `poetry run dbt run -s +stg_transit_database__transit_data_quality_issues+`
- `poetry run dbt docs generate`

## Post-merge follow-ups

- [x] No action required

If the transit data quality issues base is refactored to have synthetic `valid_from`, `valid_to` and `is_current`, then there may need to be some extra work in order to coalesce the two column names in to one.


lauriemerrell (approved):
If this worked in testing then LGTM! Just noting that I think that part of the cause here is that the [associated external table does not specify a schema](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/create_external_tables/airtable/external_airtable_transit_data_quality_issues_transit_data_quality_issues.yml) so BigQuery is [imputing the schema](https://cloud.google.com/bigquery/docs/schema-detect#schema_auto-detection_for_external_data_sources) of the files in GCS and I think that it's technically possible that that auto-detect is nondeterministic (i.e., I think it could theoretically occur that BQ uses an older file to impute the schema) and that this issue could reappear and the fix would be to explicitly define the schema in the external table. But we can cross that bridge when we come to it and the changes in this PR would be needed regardless.

PR #3197, opened 2023-12-20
Align schemas between old and new cloudaudit data access logs

### LINKED ISSUE #3179: DbtModelError in `stg_audit__cloudaudit_googleapis_com_data_access`
It seems that [the union in stg_audit__cloudaudit_googleapis_com_data_access](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/staging/audit/stg_audit__cloudaudit_googleapis_com_data_access.sql#L31-L37) has broken, presumably because of an upstream schema change in the raw data. Data will stop updating in the BQ usage tables until this is fixed.

Sentry Issue: [CAL-ITP-DATA-INFRA-26WS](https://sentry.calitp.org/organizations/sentry/issues/73177/?referrer=github_integration)

```
DbtModelError: model.calitp_warehouse.stg_audit__cloudaudit_googleapis_com_data_access - Database Error in model stg_audit__cloudaudit_googleapis_com_data_access (models/staging/audit/stg_audit__cloudaudit_googleapis_com_data_access.sql)
  Query error: Queries in UNION ALL have mismatched column count; query 1 has 16 columns, query 5 has 15 columns at [87:5]
  compiled Code at target/run/calitp_warehouse/models/staging/audit/stg_audit__cloudaudit_googleapis_com_data_access.sql
```

SorenSpicknall:
# Description

Google recently rolled out several new top-level columns and nested fields in the data access audit log that drives our `stg_audit__cloudaudit_googleapis_com_data_access` model. Because we use a `UNION ALL` to process older data up to a 90-day threshold, this caused failures stacking query results on top of each other.

This PR resolves the issue by specifying a set of columns and fields to bring in that satisfy the desired outcomes for the resulting fact table. It will also be able to be fully undone 90 days after the audit log changes were made by Google (~February 2024), allowing us to return to a simpler `SELECT *` approach because the schemas of older logs under consideration by the union will be consistent with newer logs at that point.

Resolves #3179 

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Tested with local dbt runs on the staging project, through to the downstream fact table. Schemas of resulting tables inspected by hand and compared to the existing tables.

## Post-merge follow-ups

- [x] No action required

### DIFF
--- a/warehouse/models/staging/audit/stg_audit__cloudaudit_googleapis_com_data_access.sql
+++ b/warehouse/models/staging/audit/stg_audit__cloudaudit_googleapis_com_data_access.sql
@@ -16,7 +16,61 @@
 WITH latest AS (
     {% set yesterday = modules.datetime.date.today() - modules.datetime.timedelta(days=1) %}
 
-    SELECT *
+    SELECT
+        -- This was previously a SELECT *, but changes to the audit logs required us to align old
+        -- columns (including nested fields) with new ones in late 2023 for the 90-day lookback to
+        -- work as expected. A SELECT * can likely be restored by March of 2024.
+        logname,
+        STRUCT(
+            resource.type,
+            STRUCT(
+                resource.labels.dataset_id,
+                resource.labels.project_id,
+                resource.labels.location
+            )
+            AS labels
+        )
+        AS resource,
+        STRUCT(
+            protopayload_auditlog.servicename,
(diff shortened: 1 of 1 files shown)


charlie-costanzo (approved):
Looks good to me, runs locally, thanks for taking care of this!!

PR #3201, opened 2023-12-21
make dbt models select only latest uploaded

kengie:
# Description

1. Realizing that the external table containing the NTD Report data can potentially have duplicates, we change the SQL statements in the dbt intermediate models slightly to select the data with the latest upload date.
  a. NOTE the linter will not allow me to select the `Max(api_report_last_modified_date)` AND to `SELECT DISTINCT` so I could not. In theory yes if you select the max date uploaded, as long as its not the same down to the tenth of a millisecond, you will not get duplicates.

3. Remove `2023` from the names of any models since they will work for all years.
 

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Tested all dbt models locally, they all passed.
image

## Post-merge follow-ups

- [x] No action required


csuyat-dot (approved):
discussed in person. looks good.

PR #3204, opened 2024-01-02
Move setup-gcloud to floating V2 release

SorenSpicknall:
# Description

Resolves #3203 by upgrading the `google-github-actions/setup-gcloud` Action to use a supported version, which also required adding the `google-github-actions/auth` action since the current version of `setup-gcloud` cleaves away authentication.

Additionally, moves our linting step back to use Python 3.11.7 instead of 3.12 due to [PEP 632](https://peps.python.org/pep-0632/) impacting an underlying tool that we don't have direct control over.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

## How has this been tested?

Many of the relevant workflows were tested automatically within this PR. Changes made in the others are identical, and those workflows have access to the same creds, so no differing behavior is expected.

## Post-merge follow-ups

- [x] No action required
- [ ] Actions required (specified below)

PR #3208, opened 2024-01-03
Fix Up PyPI Upload Process

SorenSpicknall:
# Description

Upon merge of #3204, publication to PyPI was unsuccessful for the `calitp-data-infra` package (it was skipped for `calitp-data-analysis` because the current version number matched the published version number). This PR makes a change to a mistyped environment variable assignment in the publication YML for `calitp-data-analysis` which _should_ resolve the issue based on my understanding of [Poetry's environment variable construction norms](https://python-poetry.org/docs/configuration#pypi-tokenname). If not, there is an additional step we can take next.

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Uploads made to TestPyPI based on similar environment variable structure - since this workflow only fires to prod upon merge, if this set of changes doesn't work an additional PR will be created for the next set of changes.

## Post-merge follow-ups

- [x] No action required

### DIFF
--- a/.github/workflows/build-calitp-data-infra.yml
+++ b/.github/workflows/build-calitp-data-infra.yml
@@ -47,7 +47,7 @@ jobs:
         run: |
           poetry config repositories.test-pypi https://test.pypi.org/legacy/
           poetry config pypi-token.test-pypi $POETRY_TEST_PYPI_TOKEN_PYPI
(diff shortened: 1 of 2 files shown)


charlie-costanzo (approved):
LGTM!

PR #3210, opened 2024-01-04
convert `end_of_month_date` calculations in payments tables to Pacific Time

### LINKED ISSUE #3209: Align `end_of_month_date` calculation with Metabase localization
As described in [this Slack thread](https://cal-itp.slack.com/archives/C02CYPQKHH8/p1702895077295269?thread_ts=1701114078.642689&cid=C02CYPQKHH8), the datetime values for `fct_payments_settlements` and `fct_payments_aggregations` are in `UTC`  in the warehouse but Metabase is performing automatic localization to convert dates to `Pacific Time`. This can be helpful for analysis, but is proving the calculation for `end_of_month_date` to be incorrect. 

We need to correct the logic used in `end_of_month_date` to calculate based on Pacific Time, and rename `end_of_month_date_pacific`. We can also preserve the original calculation and rename as `end_of_month_date_utc`.

After, we will need to make sure to fix any metabase questions using this field.

Alternative option considered: remove Metabase automatic localization

charlie-costanzo:
# Description

As described in [this Slack thread](https://cal-itp.slack.com/archives/C02CYPQKHH8/p1702895077295269?thread_ts=1701114078.642689&cid=C02CYPQKHH8), the datetime values for `fct_payments_settlements` and `fct_payments_aggregations` are in `UTC`  in the warehouse but Metabase is performing automatic localization to convert dates to Pacific Time. This can be helpful for analysis, but is proving the calculation for `end_of_month_date` to be incorrect. 

This PR changes the timezone used for the `end_of_month_date` calculation to use Pacific Time and renames `end_of_month_date` to `end_of_month_date_pacific`. It also preserves the original end of month calculation as a new field, `end_of_month_date_utc`

Resolves #3209

Alternative option: remove Metabase automatic localization

## Type of change

- [x] New feature

## How has this been tested?
locally using `dbt run` and `dbt test`

## Post-merge follow-ups
Identify Metabase questions that use `end_of_month_date` and substitute with new fields


warehouse/models/mart/payments/_payments.yml
```diff
@@ -269,7 +276,11 @@ models:
...
       - *organization_source_record_id
-      - name: end_of_month_date
+      - name: end_of_month_date_pacific
+        description: |
+          The last day of the month in Pacific Time (involves conversion of record_updated_timestamp_utc to Pacific Time).
+          This column is primarily for use in BI tooling to support consistent dates across activity types.
+      - name: end_of_month_date_utc
         description: |
           The last day of the month of the record_updated_timestamp_utc.
           This column is primarily for use in BI tooling to support consistent dates across activity types.
```
SorenSpicknall, line 286: Should this description line for `end_of_month_date_utc` be modified to reflect updated patterns of use?

warehouse/models/mart/payments/_payments.yml
```diff
@@ -167,9 +167,16 @@ models:
...
-          The last day of the month of the aggregation_datetime.
+          The last day of the month of the aggregation_datetime in Pacific Time (involves conversion of aggregation_datetime to Pacific Time).
+          This column is primarily for use in BI tooling to support consistent dates across activity types.
+          Aggregation activity happens throughout the month.
+        tests:
+          - not_null
+      - name: end_of_month_date_utc
+        description: |
+          The last day of the month of the aggregation_datetime in UTC.
           This column is primarily for use in BI tooling to support consistent dates across activity types.
```
SorenSpicknall, line 180: Should this description line for `end_of_month_date_utc` be modified to reflect updated patterns of use?

SorenSpicknall (approved):
Generally, this seems like a valid approach. Not knowing much about Metabase behavior, though, some extra questions to make sure we understand before merging, since we've generally tried to keep timestamps with time zones out of the warehouse in the past:

- Do we know that we're always going to get this column as PT? Can localization change based on factors like where a Metabase user is located, where a row comes from, etc.?
- Have we determined from example data that the new calculation is accurate?

charlie-costanzo:
Hey thanks for the review, good questions. Responding in this comment:

> Do we know that we're always going to get this column as PT? Can localization change based on factors like where a Metabase user is located, where a row comes from, etc.?

It looks like the display timezone will come from the `Localization -> Report Timezone` Admin setting in Metabase, which is currently set to `America/Los_Angeles`. A conversion to PT will happen if the warehouse datetime value includes a timezone, in this case `response_created_timestamp_utc` includes UTC in the values (ex `2024-01-01 02:34:04 UTC`). More info from Metabase [here](https://www.metabase.com/docs/latest/configuring-metabase/localization#report-timezone) and [here](https://www.metabase.com/docs/latest/troubleshooting-guide/timezones#is-the-report-time-zone-set-incorrectly)

Screenshot 2024-01-04 at 4 04 10 PM

> Have we determined from example data that the new calculation is accurate?

I found a settlement that could be recorded as two different days based on time zone (`_key = '7dc29b619bc63f765e30f72106c441b8'`) and the calculation appears to be working

Screenshot 2024-01-04 at 4 19 44 PM

Observable here:

```
SELECT
  * 
FROM
`cal-itp-data-infra-staging.charlie_mart_payments.fct_payments_settlements`
WHERE _key = '7dc29b619bc63f765e30f72106c441b8'
```

Working on the docs updates now

warehouse/models/mart/payments/_payments.yml
charlie-costanzo: pushed update:

> This column preserves the end of month date value for the original UTC date as received by Littlepay, however, `end_of_month_date_pacific` is primarily used in BI tooling instead to support consistent dates across activity types.

warehouse/models/mart/payments/_payments.yml
charlie-costanzo: pushed update:

> This column preserves the end of month date value for the original UTC date as received by Littlepay, however, `end_of_month_date_pacific` is primarily used in BI tooling instead to support consistent dates across activity types.

Issue #3215, opened 2024-01-05
Bug: deploy_airflow GitHub Workflow Fails to Update Composer PyPI Packages List


SorenSpicknall:
**Describe the bug**
After merging #3207, we noticed that the requirements deployment step was skipped in CI despite there being qualifying changes to the `requirements.txt` Composer PyPI packages list. This should not happen.

**To Reproduce**
Merge any change that modifies the package versions referenced in `requirements.txt`

**Expected behavior**
The deployment step should write the updated requirements list to Composer whenever there's a change to the package list, whether that's addition/subtraction of a package or an update to a package version.

**Additional context**
In the meantime, anybody with proper permissions can manage the package version numbers by hand [here](https://console.cloud.google.com/composer/environments/detail/us-west2/calitp-airflow2-prod/packages?authuser=1&project=cal-itp-data-infra) if they need to merge changes to the PyPI packages list for Composer.

PR #3216, opened 2024-01-05
Add pydantic 1.9 to Composer PyPI Packages

SorenSpicknall:
# Description

When we upgraded to use pyairtable V2 in our Composer PyPI package list for auth reasons, it introduced a dependency conflict: pyairtable has a version-unbounded pydantic requirement that in turn brings in typing-extensions. Because the pydantic version being fetched relied on a newer version of typing-extensions than the one that comes with our Composer-Airflow instance (which we can't modify), we need to pin the pydantic version imported in Composer to avoid the conflict. I used 1.9, which is used elsewhere in `data-infra`.

Merging this PR will also conveniently allow us to test some additional current behavior related to #3215 - does the Action fail to run when a new _package_ is added to the requirements list, or just when the package version numbers are changed?

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Tested via rollout on production Composer - a safe move because Composer automatically rejects changes that produce dependency conflicts, rather than getting into a bad state.

## Post-merge follow-ups

- [ ] No action required

Monitor merge CI behavior to continue investigating #3215.

### DIFF
--- a/airflow/requirements.txt
+++ b/airflow/requirements.txt
@@ -1,6 +1,7 @@
 calitp-data-infra==2024.1.3
 gusty==0.6.0
 pyairtable==2.2.1
+pydantic==1.9
 typer==0.4.1
 sentry-sdk==1.17.0
 platformdirs<3,>=2.5


charlie-costanzo (approved):
LGTM, thank you for taking care of this!

PR #3220, opened 2024-01-08
Update changed-files Action dependency

SorenSpicknall:
# Description

Mirrors changes in #3202, but on a branch rather than a fork in order to run all relevant Actions successfully (forks are not permitted to run `google-github-actions/auth`

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

Workflow runs were completed in CI as a result of changed introduced temporarily in https://github.com/cal-itp/data-infra/pull/3220/commits/e8413ef847d145d3fa3f0a3ca6a43f559612a992. The file change detection automation fired correctly where expected, and correctly _didn't_ fire for the changed `/airflow/requirements.txt` file (which is only supposed to detect changes upon merge to main).

## Post-merge follow-up

- [ ] No action required

Monitor workflow runs in prod in the next couple weeks to ensure proper function after this version change


charlie-costanzo (approved):
LGTM!

PR #3227, opened 2024-01-10
add exception handling to ckan upload script

### LINKED ISSUE #3182: Open data publishing (GTFS): Handle file-level errors so subsequent files can be attempted
As a Cal-ITP data maintainer, I want the GTFS schedule open dataset to get updated to the greatest extent possible on each publish run so that our open data is up to date. 

Currently, if one file fails (specifically, stop times tends to fail because of its size), then the subsequent files (in alphabetical order) are not even attempted. At time of writing, stop times has been failing since November 27, which means that stops, transfers, translations, and trips have not even been attempted since November 27 even though likely would have succeeded (since the issue seems to be specifically with the multi part upload for stop times). 

Basically I think we should add some exception handling ~here: https://github.com/cal-itp/data-infra/blob/main/warehouse/scripts/publish.py#L478-L488 so that if one file fails we do continue trying to update the subsequent files.

charlie-costanzo:
# Description

As described in #3182, when trying to upload data to the open data portal (CKAN), if one file fails (specifically, stop times tends to fail because of its size), then the subsequent files (in alphabetical order) are not even attempted.

This PR adds exception handling to `warehouse/scripts/publish.py` so that if one file upload fails the others are attemped.

Resolves #3182 

## Type of change

- [x] New feature

## How has this been tested?

Will have to test in prod as this is an error generated by Caltrans servers

## Post-merge follow-ups
- [x] No action required

### DIFF
--- a/warehouse/scripts/publish.py
+++ b/warehouse/scripts/publish.py
@@ -488,7 +488,7 @@ def _publish_exposure(
                                 )
                         except requests.exceptions.RequestException as e:
                             print(e)
-                            pass
+                            continue
 
                     else:
                         typer.secho(
--- a/warehouse/scripts/publish.py
+++ b/warehouse/scripts/publish.py
@@ -486,9 +486,11 @@ def _publish_exposure(
                                     file=fp,
                                     resource_id=resource.id,
                                 )
-                        except requests.exceptions.RequestException as e:
-                            print(e)
-                            continue
+                        except requests.exceptions.HTTPError as e:
+                            typer.secho(
+                                f"Failed to upload to {fpath} due to error: {e}",
+                                fg=typer.colors.RED,
+                            )
 
                     else:
                         typer.secho(
--- a/warehouse/scripts/publish.py
+++ b/warehouse/scripts/publish.py
@@ -477,14 +477,19 @@ def _publish_exposure(
 
                     if publish:
                         typer.secho(publish_msg, fg=typer.colors.GREEN)
-                        with open(fpath, "rb") as fp:
-                            upload_to_ckan(
-                                url=destination.url,
-                                fname=fname,
-                                fsize=fsize,
-                                file=fp,
-                                resource_id=resource.id,
-                            )
+                        try:
+                            with open(fpath, "rb") as fp:
+                                upload_to_ckan(
+                                    url=destination.url,
+                                    fname=fname,
+                                    fsize=fsize,
+                                    file=fp,
+                                    resource_id=resource.id,
+                                )
+                        except requests.exceptions.RequestException as e:
+                            print(e)
+                            pass
+
                     else:
                         typer.secho(
                             f"would be {publish_msg} if --publish",


warehouse/scripts/publish.py
```diff
@@ -477,14 +477,19 @@ def _publish_exposure(
...
+                        try:
+                            with open(fpath, "rb") as fp:
+                                upload_to_ckan(
+                                    url=destination.url,
+                                    fname=fname,
+                                    fsize=fsize,
+                                    file=fp,
+                                    resource_id=resource.id,
+                                )
+                        except requests.exceptions.RequestException as e:
```
SorenSpicknall, line 489: I believe that this should be `requests.exceptions.HTTPError`, which is the specific class of error produced [when an upload times out](https://o1d2fa0877cf3fb10p-tp.appspot.com/log?dag_id=publish_open_data&task_id=publish_california_open_data&execution_date=2024-01-01T00%3A00%3A00%2B00%3A00&map_index=-1).

warehouse/scripts/publish.py
```diff
@@ -477,14 +477,19 @@ def _publish_exposure(
...
+                            print(e)
```
SorenSpicknall, line 490: We're already using `typer.echo` statements instead of the built-in `print` tool elsewhere in this script, so I'd make this into something like:

`typer.echo(f"Failed to upload to {fpath} due to error: {e}")`

We use styled logs in this script (which don't actually display with styling in Airflow, so jury's out on whether it's worth continuing to do so). But if you want to align this with most of the other logging in the same script, you'd use the slightly more complex `typer.secho` command like this one:

`typer.echo(f"Failed to upload to {fpath} due to error: {e}", fg=typer.colors.RED)`

warehouse/scripts/publish.py
```diff
@@ -477,14 +477,19 @@ def _publish_exposure(
...
+                            continue
```
SorenSpicknall, line 491: I don't think you need this statement here. The end of the loop iteration, a few lines further down, manually deletes the `df` dataframe after the if-else logic but before the loop iteration completes, which was probably done for resource constraint reasons. We likely want to still be able to reach that line after an error is logged, after which point the loop will still move on to its next iteration as normal.

SorenSpicknall (changes requested):
A few changes to request. Let me know if you want more context for any of this, or if there's anything I'm missing that informs the choices you made in the first draft!

warehouse/scripts/publish.py
charlie-costanzo: yes this makes sense! I almost did that but defaulted to a more general exception for some unknown reason

warehouse/scripts/publish.py
charlie-costanzo: makes a lot of sense, I went back on forth on this for a long time too hah

warehouse/scripts/publish.py
charlie-costanzo: ah yes good point!

charlie-costanzo:
> A few changes to request. Let me know if you want more context for any of this, or if there's anything I'm missing that informs the choices you made in the first draft!

This review was very helpful Soren, thank you! I just pushed the changes now. Let me know if anything else looks off

SorenSpicknall (approved):

PR #3261, opened 2024-01-31
Calculate distance traveled for CCJPA rides

charlie-costanzo:
# Description
This PR facilitates the calculation of distance traveled (in miles) between `location_name` and `off_location_name` for CCJPA rides in the `fct_payments_rides_v2` table. It does this by:

* Creating an O/D pair matrix for CCJPA stations with the distance between each pair calculated in miles
* Uploads this as a dbt seed
* Joins this column to the `fct_payments_rides_v2` table
* Adds necessary documentation for the new seed, and new column in the mart table

## Type of change
- [x] New feature
- [x] Documentation

## How has this been tested?
Locally using dbt


SorenSpicknall:
I've of two minds about this one. My initial preference is to spin this new information out into another CCJPA-specific model, or even just an enhanced version of `fct_payments_rides_v2`, to avoid entrenching a precedent of sparse agency-specific columns in `fct_payments_rides_v2`. However, I can see an argument for inclusion in `fct_payments_rides_v2` from a standpoint of avoiding proliferation of agency-specific _models_ or agency-specific joins, and I don't have a ton of the prior context for how agency-specific information has been treated in payments models in the past.

I'll leave the final decision to you, Charlie, since you have a longer history working on payments models than I do, but from my first-pass standpoint I would prefer this column exist somewhere other than `fct_payments_rides_v2` - an intermediate model based on the seed, or an agency-specific model that takes only the CCJPA rows of `fct_payments_rides_v2` and adds the new column in that more limited context.

Let me know what you decide to do, and I'll put in a formal review after that decision point.

charlie-costanzo:
okay @SorenSpicknall I've spent some time thinking about this

I did move the joining of the seed to the intermediate table, `int_payments__matched_device_transactions`

However, I left the column on `fct_payments_rides_v2` and made the column name agency-agnostic (`distance_miles`), and my thinking for this is two-fold:

* We actually have precedent in doing this for the column `distance_meters` on the `fct_payments_rides_v2` table — it's essentially the same concept except in meters. It's also only relevant for a couple of agencies for the time being.
* It is the hope/goal that this is relevant for more agencies soon. This is also the preference of the stakeholders who created this ask.

Let me know if you don't agree with this and I'm happy to reconsider.

SorenSpicknall (approved):
I'm happy with this approach. Approved!

PR #3263, opened 2024-02-02
Ignore frequency rows with 0-value headways in frequencies_stop_times

SorenSpicknall:
# Description

In today's transform_warehouse DAG run, the dbt_run_and_upload_artifacts task tripped up on the `int_gtfs_schedule__frequencies_stop_times` model because two `headway_secs` values of 0 came in from an LAX FlyAway `frequencies.txt`, violating a GTFS standard rule that all such values need to be positive integers. This broke some array creation code.

This fix ignores `dim_frequencies` rows with improper 0-value headways when making the `frequencies_stop_times` model. No valid data will be ignored by the fix, since none of the impacted rows are compliant with the GTFS spec.

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Model built and tested in the staging warehouse environment.

## Post-merge follow-ups

- [ ] No action required

After merging, I'll clear and rerun today's transform_warehouse DAG run.


charlie-costanzo (approved):
LGTM!

PR #3270, opened 2024-02-08
Bump Metabase Version to v0.48.3

SorenSpicknall:
# Description

Resolves #3247

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

## How has this been tested?

On the test Metabase instance, I substituted the new version number for the existing one by hand. We currently don't deploy PR changes to the test instance because doing so on every commit to a PR would be costly and potentially harmful. After making the version change by hand, I monitored for application errors and confirmed that the postgres backup job is still functioning correctly for `metabase-test`. Everything seems to be in order.

## Post-merge follow-ups

- [ ] No action required
- [x] Actions required (specified below)

Monitor prod deployment for errors, including application errors and postgres backup issues.

PR #3279, opened 2024-02-15
Upgrade dbt-metabase Package in Warehouse to Version That Supports Current Metabase Version

SorenSpicknall:
# Description

We use the dbt-metabase Python package for our Metabase data syncs. The version we had in use came before [this fix](https://github.com/gouline/dbt-metabase/pull/199), which addresses a boolean value interpretation change in Metabase x.48.x. We must now adopt a newer version of dbt-metabase in order to make use of that fix, since we upgraded out Metabase application version in #3270. This change will resolve [this Sentry error](https://sentry.calitp.org/organizations/sentry/issues/73589/).

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

New image built locally, and will be built on CI as well. Changelogs on the dbt-metabase package inspected for impacts to our code calls - didn't find any. The upgrade should be backwards compatible with what we already have in place.

## Post-merge follow-ups

- [ ] No action required

Rerun the Metabase sync in today's run_and_upload Airflow task, and monitor for successful completion.


warehouse/pyproject.toml
```diff
@@ -21,13 +21,13 @@ python-slugify = "^6.1.2"
 sentry-sdk = "^1.9.8"
 backoff = "^2.2.1"
 gcsfs = "^2023.1.0"
-dbt-metabase = "^0.9.14"
+dbt-metabase = "0.10.3"
```
evansiroky, line 24: Are future versions incompatible which is why this is pinned?

SorenSpicknall: The `dbt-metabase` package recently moved to a 1.0, which I'm going to create a new issue to move to - it includes breaking changes for several arguments, and so it becomes a slightly larger project to switch it out.

I pinned to a specific version here as a sort of implicit flag that 0.10.3 appears to be the end of the line for pre-1.0 versions of the package, but that's better addressed via a note, so I'm putting in another commit shortly to float the minor version once again and document that we don't anticipate further pre-1.0 updates to the package.

evansiroky (approved):

PR #3284, opened 2024-02-23
Added a minimum schema for fixing #3283

### LINKED ISSUE #3283: Bug: Numeric Column Name Breaking Airtable Quality Pipeline
**Describe the bug**
As described [here in the Cal-ITP Slack](https://cal-itp.slack.com/archives/C0332307HKM/p1708442657824929), the external table creation process for the Airtable-sourced `transit_data_quality_issues__services` model has been erroring for a couple days, which is causing a couple additional downstream errors. This is because of an unhandled column in the source Airtable whose name begins with a number, unsupported by BQ column naming conventions.

**To Reproduce**
Any attempt to create an external table using a column whose name begins with a numeric character will fail in the same way that the `external_airtable_transit_data_quality_issues_services` task is currently failing in Airflow - example logs [here](https://b2062ffca77d44a28b4e05f8f5bf4996-dot-us-west2.composer.googleusercontent.com/log?dag_id=create_external_tables&task_id=external_airtable_transit_data_quality_issues_services&execution_date=2024-02-20T09%3A00%3A00%2B00%3A00&map_index=-1).

**Expected behavior**
Evan has indicated that the column in question need not be present on the BQ version of this table, so it can either be handled via a rename or ignored entirely. Once that's done, downstream modeling impacts on the `dbt_run_and_upload_artifacts` DAG and dbt tests should passively resolve.

**Additional context**
Airtable column names that begin with numeric characters are not new to this ecosystem, and we've handled them gracefully before. Refer to treatment of e.g. `external_airtable.california_transit__services` for how to navigate such column names in a fashion that's canonically aligned with the rest of our data infra.

This task will be completed in concert with Vivek.

vevetron:
# Description

Adding a minimal custom schema so a certain field, `2021_ntd_id__from_provider_`, doesn't get pulled from the source data in cloud storage.

Resolves #3283

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

This has not been tested.

## Post-merge follow-ups

Check to see if slack alerts continue after rebuild.

- [x] No action required

We can either wait for the next airflow run or somebody can trigger it manually.


SorenSpicknall (approved):
This should do it, though we'll monitor to be sure. Here's what comes next after merge:

1. We wait for this CI Action to run,
2. We retry the failed task in today's `create_external_tables` DAG run
3. We either retry the failed `dbt_run_and_upload_artifacts` task in today's `transform_warehouse` DAG run, or we rerun each impacted model in a more targeted fashion

For (3), a full retry would run a bunch of models we don't actually need to rerun and would take a long time. However, @vevetron, you may not be set up yet to run targeted models in prod without hitting permissions issues. One the external table creation task goes green, I can make those targeted runs and let you know what commands I used to do so, so that once you're all set up on your machine you'll be able to do similar.

PR #3290, opened 2024-02-27
transform_warehouse DAG to run after new Littlepay publishing time

charlie-costanzo:
# Description

Littlepay recently updated the cadence of their data publishing to better support our analytics needs, but in doing so threw the cadence of our payments ingest/parse/transform jobs out of sync. This PR pushes the running of the `transform_warehouse` DAG back 4 hours (14:00 UTC).

A follow-up ticket will be created to break apart our `transform_warehouse` DAG runs so that all other models are run earlier in the morning (as they were previously), and payments is run at 14:00 UTC.

## Type of change

- [x] New feature

## How has this been tested?
docker image builds locally

## Post-merge follow-ups

- [x] No action required

### DIFF
--- a/airflow/dags/transform_warehouse/METADATA.yml
+++ b/airflow/dags/transform_warehouse/METADATA.yml
@@ -1,5 +1,5 @@
 description: "Builds and tests the warehouse, and uploads artifacts"
-schedule_interval: "0 10 * * *"
+schedule_interval: "0 14 * * *"
 tags:
   - all_gusty_features
 default_args:


SorenSpicknall (approved):
Once this merges, we'll want to closely monitor data freshness over the next few days to make sure there isn't a downstream linkage (another DAG, etc.) that needs to also be adjusted because of the new timing. I didn't identify any upfront, but always good to have an eye on the whole DAG ecosystem when releasing adjustments like this one.

charlie-costanzo:
> Once this merges, we'll want to closely monitor data freshness over the next few days to make sure there isn't a downstream linkage (another DAG, etc.) that needs to also be adjusted because of the new timing. I didn't identify any upfront, but always good to have an eye on the whole DAG ecosystem when releasing adjustments like this one.

sweet — merging now, and setting a reminder to monitor this week

PR #3294, opened 2024-02-27
Move Elavon Secret Fetch Away from Top Level of Raw Sync DAG

### LINKED ISSUE #3289: Reconfigure Elavon Loader to Fetch Secrets at Execution

As a warehouse developer, I want to be able to utilize a local instance of Airflow for development even if I don't have access to all the creds used by each DAG. Because Airflow's DagBag import fails if even one incoming file experiences errors at import time, any issue that comes up during the DAG import process prevents interacting with all the other DAGs as normal.

One such common blocker is the sync_elavon DAG, which uses [elavon_to_gcs_raw.py](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/sync_elavon/elavon_to_gcs_raw.py) to complete its work. That Python file calls `get_secret_by_name()` at the top level when constructing constant, and the call fails if the user running the Airflow instance doesn't have access to the `CALITP__ELAVON_SFTP_PASSWORD` secret in Google Secret Manager.

### Acceptance Criteria

The `get_secret_by_name()` function call should be moved somewhere that doesn't fail the DagBag import process, likely bypassing construction of a constant altogether and feeding it into the `password` variable construction [here](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/sync_elavon/elavon_to_gcs_raw.py#L32).

SorenSpicknall:
# Description

This moves a function call to `get_secret_by_name()` out of the top level of the `sync_elavon` DAG's Python code, where it was getting run regularly as Airflow parsed each DAG. The previous behavior was fine in prod, but stymied local developers who lacked permissions on the Google Secret Manager secret being fetched by that function call, preventing the DAG from loading correctly. This solution has the added advantage of reducing the number of API calls to Google Secret Manager, since this function won't be run under the hood by Airflow anymore.

Resolves #3289 

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Tested function of the new code placement via local Airflow

## Post-merge follow-ups

- [x] No action required

### DIFF
--- a/airflow/dags/sync_elavon/elavon_to_gcs_raw.py
+++ b/airflow/dags/sync_elavon/elavon_to_gcs_raw.py
@@ -12,7 +12,6 @@ from calitp_data_infra.storage import get_fs
 CALITP__ELAVON_SFTP_HOSTNAME = os.environ["CALITP__ELAVON_SFTP_HOSTNAME"]
 CALITP__ELAVON_SFTP_PORT = os.environ["CALITP__ELAVON_SFTP_PORT"]
 CALITP__ELAVON_SFTP_USERNAME = os.environ["CALITP__ELAVON_SFTP_USERNAME"]
-CALITP__ELAVON_SFTP_PASSWORD = get_secret_by_name("CALITP__ELAVON_SFTP_PASSWORD")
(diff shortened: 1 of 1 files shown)


charlie-costanzo (approved):
LGTM! Good idea, thanks for taking care of this!

PR #3318, opened 2024-03-27
Address direct dependency upgrades

SorenSpicknall:
# Description

Updates dependencies from the current automatic `chore` PRs that are directly controlled via pyproject.toml files and requirements.txt files, ignoring those that are a layer deep inside automatically configured poetry.lock files. Also ignores the suggested upgrade to black 24.3.0 across the board, since there's a conflict with a black version required by another dependency in Poetry (to be investigated separately).

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

All changed Docker images and packages were built locally before merge

## Post-merge follow-ups

- [ ] No action required

Any refs to specific version numbers for these packages, services, and images will need to be updated. This cannot be done _in_ this PR because those updates rely on publication of the changes from this PR (e.g. new image version and new package versions). This round of changes will include updates to two Kubernetes kustomization files.


charlie-costanzo (approved):
LGTM! Thanks for taking care of the README updates as well

PR #3326, opened 2024-04-09
Extending scrape_ntd.py, and adding ridership to the mart

### LINKED ISSUE #3164: New Airflow DAG for NTD downloading

As an analyst,
I would love to be able to query NTD data
and be able to join the NTD data with other data from our warehouse,
So that I can do various analyses to classify transit agencies according to revenue, ridership, etc.

### Acceptance Criteria

1. There is a DAG that has a task that can be manually ran to download NTD data to Google Cloud
2. There is a DAG that has a task that runs monthly that downloads monthly ridership data
3. Maybe a separate ticket? There is documentation on how to do various dbt modeling to enable the exposure of NTD data in various mart tables that metabase can read.

### Notes

Laurie will be creating a document that discusses some of this.

vevetron:
# Description

Still a work in progress.  The main goal is to add NTD ridership to the MART.  I think I could use some feedback sooner rather than later.

- Updates to scrape_ntd.py to also handle the ridership excel notebook and load it to a bucket

- Initial queries to load ridership data to staging

- Code that pivots ridership dataset

I'm trying to make changes similar to [these changes](https://github.com/cal-itp/data-infra/commit/0f997fe95267e1e564fc33e452136f1c249fc810#diff-049da376c46f897ea20520e4e9fd03a778138d26ee9514bb8a27bf5278a12390).

Current problems:
- Pulls everything from the bucket, when we just want the latest file.  So there are 4 duplicates
- Some references are hard coded in `scrape_ntd.py`
- document columns
- need to add tests

Resolves #3164

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

You can query the final mart table at 
`SELECT * FROM `cal-itp-data-infra-staging.vb_mart_ntd.dim_monthly_ntd_ridership_with_adjustments_upt` LIMIT 1000`
Also it runs with dbt

## Post-merge follow-ups
Need to document and write tests before merging, then it should be good!

- [x ] No action required


charlie-costanzo (changes requested):
Hey Vivek – awesome, very glad to see you got this running. I was able to get it running locally as well. A few first thoughts,

* With dbt, we're going to want to remove the hard-coding of the tables in the SQL files as well, following the jinja `{{ ref('table_name') }}` convention. Some good examples [in this file here](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/gtfs/fct_daily_schedule_feeds.sql#L5).
* That file also is a good example of the SQL formatting best practices the team has used as far as spacing, indents, new lines, being explicit etc
* As far as pivoting in the staging table, my initial thought is that this is something that should be handled in the intermediate table
* Happy to think through the best way to pull the most recent files from the bucket in this case but feel like @SorenSpicknall will have some more helpful opinions...

SorenSpicknall:
>Pulls everything from the bucket, when we just want the latest file. So there are 4 duplicates

I want to understand more about the challenge you're referencing, and what behavior is desired here in the long run.

Generally speaking, we keep all versions of incoming data available via external tables (and usually in staging models, too) because it's important to be able to rebuild history when doing full refreshes, reporting on changes in data quality, etc. This infra was built so that all prior logic is traceable and repeatable from the ground up, so that instead of relying solely on downstream dashboards and reports to tell us what happened two years ago, developers can audit and rerun the logic that produced those downstream outputs starting at the raw data in GCS. This allows for flexibility to update logic on older data, make corrections, and produce new analyses from historic data holdings - it enables us to ask new questions about old data.

If the data incoming from the NTD monthly ridership products are fully cumulative (including data for past months within each export), you could do something like what we did with the Elavon transaction loader DAG - [filtering the external table by the outputs of the most recent DAG run](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/staging/payments/elavon/stg_elavon__transactions.sql#L11-L16) when constructing your staging model(s). That is **only** advisable if the source files are cumulative. Otherwise, we want to use dbt to construct data history with more care.

If the duplicates you're referring to are duplicates due to testing, present because you loaded the same month's data several times via the Python script during the source of development, you can just hand-delete the associated subfolders from testing and schedule your production DAG in such a way that it runs once a month and the Python script will only pick up the latest month's files at the source.

If it's anything else, you'll end up using a combination of Hive partitions and query filters to get the latest picture at any given time. All data will be available via external tables, but you'll interact with it primarily through queries that include a filter on one of the partitions. When we talked last week about the folder structure you're using in your bucket, my initial advice may have been a little misaligned with what I'd recommend now - but I want to know more about this data before making an updated recommendation.

Some follow-up questions that would help me give advice on how to revise this approach:

- How often does the schema change for these monthly files?
- Do those schema changes include columns you wish to use in modeling, or do you intend to produce a consistent, shared set of columns from every month's NTD ridership data?
- Is there any data overlap between monthly exports, or is each monthly source file discrete from the next?
- Am I correct in my assumption that you intend to run this ingest process monthly?
- Do _existing/previous_ monthly files ever get later-published updates, changing history as it may have previously been loaded?

charlie-costanzo:
also @SorenSpicknall not exactly sure why [this action](https://github.com/cal-itp/data-infra/actions/runs/8623683380/job/23637308140?pr=3326#step:14:72) is failing?

SorenSpicknall:
> also @SorenSpicknall not exactly sure why [this action](https://github.com/cal-itp/data-infra/actions/runs/8623683380/job/23637308140?pr=3326#step:14:72) is failing?

It looks like it's trying to select the whole project and failing to visualize because of the sheer number of nodes - let's see how it behaves after some changes are made in upcoming commits.

vevetron:
> If the data incoming from the NTD monthly ridership products are fully cumulative (including data for past months within each export), you could do something like what we did with the Elavon transaction loader DAG - [filtering the external table by the outputs of the most recent DAG run](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/staging/payments/elavon/stg_elavon__transactions.sql#L11-L16) when constructing your staging model(s). That is only advisable if the source files are cumulative. Otherwise, we want to use dbt to construct data history with more care.

i'll try this

> * How often does the schema change for these monthly files?

The schema of the original data changes every month. A new column is added for each new month.
> * Do those schema changes include columns you wish to use in modeling, or do you intend to produce a consistent, shared set of columns from every month's NTD ridership data?

In the source data, the first 10 columns are consistent which we want to keep.  And then there are 50+ year-month columns that keep growing each month, and we need to include all of them and then pivot them into a single column.
> * Is there any data overlap between monthly exports, or is each monthly source file discrete from the next?

All the data appears to be the same for each export except a new column gets added each month.  So all we really need is the latest file.
> * Am I correct in my assumption that you intend to run this ingest process monthly?

Yes.  Manually at first.
> * Do _existing/previous_ monthly files ever get later-published updates, changing history as it may have previously been loaded?

I don't know.  I assume not.  It seems like every 2 years there is a schema change. [source data](https://www.transit.dot.gov/ntd/data-product/monthly-module-adjusted-data-release)

vevetron:
@charlie-costanzo 

> * As far as pivoting in the staging table, my initial thought is that this is something that should be handled in the intermediate table

I tried that initially, but because every month the source would increase columns that we need to capture, I used a select * in staging.  But then the sqlfluff pre-commit kept killing my commit.  I'm not sure how to proceed.  I'm happy to override the pre-commit if that's okay.

SorenSpicknall:
RE: my questions, the cumulative nature of the file lends itself to using similar logic to our Elavon DAG to select the latest versions at any given time in the staging model.

For the sqlfluff issue, you could [add a noqa annotation](https://docs.sqlfluff.com/en/stable/configuration.html#ignoring-individual-lines) to the end of the line in question and perhaps silence the alert for now, but there's probably a more deterministic way of dealing with it.

SorenSpicknall:
Small change in 5cb84b9 - the "Version" tag on schema YML files refers to configuration structure versions, not to versioning _contents_ of the schema. See [this explainer](https://discourse.getdbt.com/t/why-does-every-schema-yml-file-need-version-2/2321) for more context.

SorenSpicknall:
I resolved the CI issue with [these changes in `/warehouse/scripts/visualize.py`](https://github.com/cal-itp/data-infra/pull/3326/files#diff-fb50f105402b017d8385d34564d06907974fc508961485cce3e403c74d655588). We had a latent bug in our CI visualization logic, impacting an edge case you tripped up for the first time since we last significantly changed visualization logic [about a year ago](https://github.com/cal-itp/data-infra/pull/2661). This explains the following log lines in failed runs of the visualization task, where the task gets an empty list of modified models and then tries to visualize the entire dbt project:

```
Visualizing the following models: []
Warning: graph has 1307 nodes...layout may take a long time.
```

If you look back through our merged PRs for [any which included new models](https://github.com/cal-itp/data-infra/pulls?page=1&q=is%3Apr+%F0%9F%8C%B1+is%3Aclosed+sort%3Aupdated-desc) (which get listed rather than visualized), you'll notice that it's virtually impossible to find one in which new models are added without referencing or changing an existing model, like you do in this PR. For this CI error to occur, the CI visualization task required new models to come in _without_ a modification or a reference to other models. Even when we've established brand new pipelines in the past, we've often made cross-references to other models in the project - mapping IDs back and forth, creating mart tables that combine with other datasets, etc. When those refs are made, they get visualized because [the "modified" state](https://docs.getdbt.com/reference/node-selection/methods#the-state-method) in dbt considers modifications to dependencies between models.

So in short, we haven't had a condition under which an empty list of models would get fed to our visualization logic in quite some time - maybe in the entire time since we started using this visualization tool. We can work around it easily by bypassing the visualization step when we see an empty list of modified nodes.

script/scrape_ntd.py
```diff
@@ -25,7 +27,9 @@
 )
 from pydantic import HttpUrl, parse_obj_as
 
-CALITP_BUCKET__NTD_DATA_PRODUCTS = os.environ["CALITP_BUCKET__NTD_DATA_PRODUCTS"]
+# CALITP_BUCKET__NTD_DATA_PRODUCTS = os.environ["CALITP_BUCKET__NTD_DATA_PRODUCTS"]
+# bucket: gs://calitp-ntd-data-products
+CALITP_BUCKET__NTD_DATA_PRODUCTS = "gs://test-calitp-ntd-data-products"
```
SorenSpicknall, line 32: Make sure to restore these environment variable references before merging, rather than using hardcoded values.

The actual hardcoded value can be defined in /airflow/docker-compose.yaml, where you'll find labeled environment variable definitions for many other variables used by scripts and jobs.

SorenSpicknall (changes requested):
One change requested, to get environment variable definitions where we expect them to be. Otherwise, this is coming along well and is probably close to merging!

script/scrape_ntd.py
vevetron: I doubt this job will get actually put into airflow since the source data and paths change somewhat regularly.

evansiroky:
I discussed this briefly on a call with @vevetron. Upon reviewing this PR, it does look like Andrew Vaccaro did create a scraper script that did not save the raw file and immediately extracted the contents to JSON. I do like the idea of following a certain philosophy to ensure consistency with other parts of our pipeline as @SorenSpicknall noted in https://github.com/cal-itp/data-infra/pull/3326#issuecomment-2047851497 about being able to rebuild history.

So unless there's a compelling reason to use the existing scraping script that bypasses the saving of the raw file, I think it would be great to follow a pattern we do in other processes where we have DAG tasks for:

1. Downloading and saving the raw excel files
2. Extracting the excel data into JSON
3. Defining the external tables to read the JSON

vevetron:
> So unless there's a compelling reason to use the existing scraping script that bypasses the saving of the raw file, I think it would be great to follow a pattern we do in other processes where we have DAG tasks for:
> 
> 1. Downloading and saving the raw excel files
> 2. Extracting the excel data into JSON
> 3. Defining the external tables to read the JSON

I don't think it's super necessary since i'm saving all the extracts from the excel unedited but I [added code](https://github.com/cal-itp/data-infra/pull/3326/commits/82a41a2d4f4fcb1df1d60c0b9a48595e34ca3529) to save the raw excel files regardless.

evansiroky:
> > So unless there's a compelling reason to use the existing scraping script that bypasses the saving of the raw file, I think it would be great to follow a pattern we do in other processes where we have DAG tasks for:
> > 
> > 1. Downloading and saving the raw excel files
> > 2. Extracting the excel data into JSON
> > 3. Defining the external tables to read the JSON
> 
> I don't think it's super necessary since i'm saving all the extracts from the excel unedited but I [added code](https://github.com/cal-itp/data-infra/pull/3326/commits/82a41a2d4f4fcb1df1d60c0b9a48595e34ca3529) to save the raw excel files regardless.

I will wait to get a second opinion from @SorenSpicknall, but I do think it would be worth separating the logic of downloading the file from extracting the data from the file into separate DAG tasks. As it is right now, the script still requires a manual run and it wouldn't be possible to recreate history from the raw file that was downloaded.

SorenSpicknall:
> > > So unless there's a compelling reason to use the existing scraping script that bypasses the saving of the raw file, I think it would be great to follow a pattern we do in other processes where we have DAG tasks for:
> > > 
> > > 1. Downloading and saving the raw excel files
> > > 2. Extracting the excel data into JSON
> > > 3. Defining the external tables to read the JSON
> > 
> > 
> > I don't think it's super necessary since i'm saving all the extracts from the excel unedited but I [added code](https://github.com/cal-itp/data-infra/pull/3326/commits/82a41a2d4f4fcb1df1d60c0b9a48595e34ca3529) to save the raw excel files regardless.
> 
> I will wait to get a second opinion from @SorenSpicknall, but I do think it would be worth separating the logic of downloading the file from extracting the data from the file into separate DAG tasks. As it is right now, the script still requires a manual run and it wouldn't be possible to recreate history from the raw file that was downloaded.

You're both right that the utility of saving the raw extracts is not _quite_ as present for an ad hoc, cumulative DAG like this one than it is for something that's scheduled and/or incremental. However, I still advocate for following the standard ingest pattern (including saving the raw artifacts) for a couple reasons:

1. It is desirable to guarantee consistent accessibility of our source data in its truest form. Publishers can remove or change files, and we want to be able to reconstruct our work from square one if desired.
2. Retaining the raw files allows us to audit our resulting JSONL for consistency with the source representation of data. Especially with sources like Excel and Airtable that are more complicated than basic text files (CSVs, etc.), it's easy for a tool like Pandas `read_excel` function to hit edge cases that result in unanticipated representations of data in our transformed files.
3. At the most human level, retaining the standard pattern simplifies onboarding and training at a later date. Where it's possible to do, we tend to prefer keeping each ingest as close to the [copy raw files] -> [turn raw files into JSONL] -> [load via external tables] -> [do data modeling and transformations] process as we can so that we don't proliferate exceptions and edge cases over time.

SorenSpicknall (approved):

charlie-costanzo (approved):

Issue #3352, opened 2024-05-23
Upgrade jupyterhub-singleuser custom image deps so it can build w/ Python 3.11


themightychris:
After upgrading JupyterHub, we need to upgrade our custom docker image to incorporate the latest jupyterhub library. It appears that this dependency comes is as part of the `jupyter/datascience-notebook` base image we use. In #3350 I upgraded it from `:python-3.9` to the latest `:python-3.11.6` since the `:python-3.9` tag hasn't been updated in over 2 years, but now the image fails to build:

```console
 => [15/23] COPY ./pyproject.toml /reqs/pyproject.toml                                                                                                                                                                                 0.2s
 => [16/23] COPY ./poetry.lock /reqs/poetry.lock                                                                                                                                                                                       0.0s
 => [17/23] RUN poetry config virtualenvs.create false                                                                                                                                                                                 0.7s
 => ERROR [18/23] RUN cd /reqs && poetry install --with=shared_utils --with=portfolio                                                                                                                                                  0.8s
------
 > [18/23] RUN cd /reqs && poetry install --with=shared_utils --with=portfolio:
0.600 The currently activated Python version 3.11.6 is not supported by the project (~3.9).
0.600 Trying to find and use a compatible version. 
0.691 
0.691 Poetry was unable to find a compatible version. If you have one, you can explicitly use it via the "env use" command.
------
Dockerfile:45
--------------------
  43 |     COPY ./poetry.lock /reqs/poetry.lock
  44 |     RUN poetry config virtualenvs.create false
  45 | >>> RUN cd /reqs && poetry install --with=shared_utils --with=portfolio
  46 |     RUN poetry config virtualenvs.create true
  47 |     
--------------------
ERROR: failed to solve: process "/bin/bash -o pipefail -c cd /reqs && poetry install --with=shared_utils --with=portfolio" did not complete successfully: exit code: 1
```

To reproduce locally, check out the latest `main` branch and:

```bash
cd images/jupyter-singleuser/Dockerfile
docker build .
```

It appears that some poetry work will need to be done to upgrade dependencies to be compatible with Python 3.11

mjumbewu:
[...]

PR #3362, opened 2024-05-28
Update dask versions in calitp-data-analysis

vevetron:
# Description

- This is part of the upgrade from python 3.9 to python 3.11.6
- This is so when building jupyter-single-user-image, the dask packages match and everything builds.  

continues work on #3352

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

poetry lock
poetry install

## Post-merge follow-ups

- [ ] No action required
When deployed we need to test things.

### DIFF
--- a/packages/calitp-data-analysis/poetry.lock
+++ b/packages/calitp-data-analysis/poetry.lock
@@ -488,30 +488,32 @@ tests = ["pytest", "pytest-cov", "pytest-xdist"]
 
 [[package]]
 name = "dask"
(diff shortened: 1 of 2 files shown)


tiffanychu90 (changes requested):
Super small comment - I think the convention is to use date for version, so instead of 2024-5-24b, it would just be 2024-5-28 for today?

tiffanychu90 (approved):
Thank you for upgrading!

PR #3364, opened 2024-05-31
Adds a column for the 2022 ntd ids

vevetron:
# Description
Trying to make it easier to get NTD ids.

Based off of: https://github.com/cal-itp/data-analyses/issues/1121, the only difference between our airtable ids and ntd ids was pulling out the 5 digits, so I did that algorithmically.

There are 27 orgs not on the list, deriving results for them regardless.

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

poetry run dbt run -s dim_organizations

SELECT *,
  IF(LENGTH(ntd_id) >= 10,
     SUBSTR(ntd_id, -5), 
     ntd_id) AS ntd_id_2022
FROM `cal-itp-data-infra.mart_transit_database.dim_organizations` 
where _is_current is true
and
ntd_id is not null

Both seemed fine.  There is one org where the ntd id is 9R02, basically CA DOT
## Post-merge follow-ups

- [x] No action required


evansiroky (approved):
Approving for now to get us going with the understanding that we may end up with a longer-term data modeling task to be able to join across multiple years of NTD data in some other kind of way.

tiffanychu90:
Follow-up notes and references:
* add seed table that's csv
* `poetry run dbt seed --select "ntd_id_to_source_record_id"`
* would adding a `state` and/or `uza` column in the table that could be useful for filtering
* https://docs.getdbt.com/docs/build/seeds

vevetron:
Basically this code was redacted since I felt it would be more useful to have the ntd ids in the source airtable, and more available throughout the pipelines and analysis tasks.

PR #3366, opened 2024-06-06
Add seed table for ntd_id (2022) and (organization) source_record_id

tiffanychu90:
# Description

Try another approach at getting `ntd_id_2022` added without truncating to 5 digits in every query. Add a seed table to help us bridge `ntd_id_2022` with `dim_organizations.source_record_id`.

Follow-up to #3364
Resolves #\[issue\]

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

## How has this been tested?

_Include commands/logs/screenshots as relevant._

## Post-merge follow-ups

_Document any actions that must be taken post-merge to deploy or otherwise implement the changes in this PR (for example, running a full refresh of some incremental model in dbt). If these actions will take more than a few hours after the merge or if they will be completed by someone other than the PR author, please create a dedicated follow-up issue and link it here to track resolution._

- [ ] No action required
- [ ] Actions required (specified below)

PR #3367, opened 2024-06-07
Updating ntd_id_2022 using airtable

vevetron:
# Description

In a previous https://github.com/cal-itp/data-infra/pull/3364 PR a formula was used to generate ntd_id_2022, but it's better to put it in airtable.  This way joins from airtable table have this id available easily especially when one might want to join to other more recent ntd data.

Separately using a csv join ntd_id_2022 was calculated and added to the orgs table.  This also removes that calculation.

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

local airflow

poetry run dbt run -s +dim_organizations

## Post-merge follow-ups

- [x] No action required

### DIFF
--- a/airflow/.gitignore
+++ b/airflow/.gitignore
@@ -1,1 +1,2 @@
 logs/*
+airflow.sh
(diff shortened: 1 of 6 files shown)


evansiroky (approved):
Linking to #3366 to capture relevant conversation that this is what the team has aligned on.

Issue #3401, opened 2024-07-23
EPIC: Ingest and Model Remaining NTD Tables


charlie-costanzo:

evansiroky:
I'm uploading a doc with various [NTD Questions](https://github.com/user-attachments/files/17805843/NTD.Questions.docx) that I'd like to be able to answer using Mart tables. These are organized by the file I think the data would be in and also the priority I think we should place on each item.

PR #3415, opened 2024-08-06
airflow: operator and dag/tasks to sync NTD data via DOT API and XLSX

### LINKED ISSUE #3402: NTD Scraping – Remaining Tables in Dataset

Part of #3401 

As a data engineer, I would like to facilitate the scraping of the remaining tables in the NTD dataset in similar fashion, extending the work found in `scrape_ntd.py` and `annual_database_service.yml` (found below).

Building upon the work completed in #3345

Existing NTD patterns:
- [script/scrape_ntd.py](https://github.com/cal-itp/data-infra/blob/main/script/scrape_ntd.py)
- [airflow/dags/create_external_tables/ntd_data_products/annual_database_service.yml](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/create_external_tables/ntd_data_products/annual_database_service.yml)

General Cal-ITP Pipeline Patterns
- [Data Pipelines Documentation](https://docs.calitp.org/data-infra/architecture/data.html#data-pipelines)
  - [Adding a New Data Source](https://docs.calitp.org/data-infra/architecture/data.html#adding-a-new-data-source)
  - [Previous PR Patterns](https://github.com/cal-itp/data-infra/pull/2376)

### Acceptance Criteria

I can successfully extract the data for the following tables from the https://www.transit.dot.gov/ntd/data-product website:
- Assets
- Expenses
- Fares/Funding
- Resources
- Security and Safety
- Service (previous years)

### Notes
---

charlie-costanzo:
# Description
This PR introduces new NTD data sources available through the federal Department of Transportation through their data API as well as XLSX file downloads.

Two Airflow operators were necessary for this work because although a large amount of NTD datasets are now available from the NTD API, there are still important datasets available only in XLSX format (monthly ridership, certain annual reports).

To accomplish this, two new Airflow operators (`scrape_ntd_api.py` and `scrape_ntd_xlsx.py`), two associated dags (`sync_ntd_data_api` and `sync_ntd_data_xlsx`), and a selection of NTD table endpoints as dag tasks were created.

Both operators utilize the `PartitionedGCSArtifact` class pattern used elsewhere in the pipeline.

NTD Data Sources scraped and stored in this PR include:
- 2022 Annual Reporting
- Monthly Ridership Data
- Safety, service, and security related data

We discovered that these tables are retroactively updated at a regular cadence, including annual reports for previous years, so a schedule has been configured to download from these endpoints on the first day of the month, every month.

Resolves #3402, part of Epic #3401

## Type of change

- [x] New feature

## How has this been tested?
Successful local Airflow runs, publishing to gcs buckets

## Post-merge follow-ups

- [x] Environment variables need to be added to composer
- [x] DAGs need to be manually triggered
- [ ] observe to verify expected behavior
- [ ] create exception handling follow-on ticket


airflow/plugins/operators/scrape_ntd_api.py
```diff
@@ -0,0 +1,148 @@
...
+            self.logger.info(
+                f"There is no data for {self.year} / {self.product}, not saving anything. Pipeline exiting."
+            )
+            pass
+        else:
+            fs.pipe(hive_path, gzip.compress(self.data))  # .encode()),
+        return hive_path
+
+
+class NtdDataProductAPIOperator(BaseOperator):
```
mjumbewu, line 112: Is there a reason we'd go the custom operator route here instead of, say, a `PythonOperator` pointing to a script? It may be a bias of mine, but I feel like custom operators are best left for operations that are going to be used in multiple places.

On that note, have we considered doing a separate task for each NTD API endpoint? E.g., have one `NtdDataProductEndpointOperator` or something, and use that once for each endpoint. I don't have context for how the NTD API might fail, and I could see it being frustrating to have the whole task fail because one endpoint failed.

airflow/plugins/operators/scrape_ntd_api.py
```diff
@@ -0,0 +1,148 @@
...
+        if response is None or len(response) == 0:
+            self.logger.info(
+                f"There is no data to download for {self.year} / {self.product}. Ending pipeline."
+            )
+            pass
+        else:
+            self.data = response
+            self.logger.info(
+                f"Downloaded {self.product} data for {self.year} with {len(self.data)} rows!"
+            )
```
mjumbewu, line 83: I don't know anything offhand about the failure modes of the NTD API, but it might be useful to check response status and log non-200 responses or something.

mjumbewu (commented):
@charlie-costanzo I took a first pass through this and left just a couple comments. I'd be down to set up a time to talk through it more.

airflow/data/ntd_endpoints.csv
```diff
@@ -0,0 +1,34 @@
...
+2022,Fuel and Energy,fuel_and_energy,https://data.transportation.gov/resource/,8ehq-7his,.json
+2022,Fuel and Energy (by Agency),fuel_and_energy_by_agency,https://data.transportation.gov/resource/,wwem-ata9,.json
+2022,Employees (by Mode and Employee Type),employees_by_mode_and_employee_type,https://data.transportation.gov/resource/,uyv8-9jek,.json
+2022,Employees (By Mode),employees_by_mode,https://data.transportation.gov/resource/,wsxw-2rpq,.json
+2022,Employees (By Agency),employees_by_agency,https://data.transportation.gov/resource/,brbd-9azc,.json
+2022,Capital Expenses (by Capital Use),capital_expenses_by_capital_use,https://data.transportation.gov/resource/,fphd-jyyj,.json
+2022,Capital Expenses (by Mode),capital_expenses_by_mode,https://data.transportation.gov/resource/,2667-vitc,.json
+2022,Capital Expenses (for Existing Service),capital_expenses_for_existing_service,https://data.transportation.gov/resource/,7kqv-yqbn,.json
+2022,Capital Expenses (for Expansion of Service),capital_expenses_for_expansion of_service,https://data.transportation.gov/resource/,nvgd-g6pj,.json
```
erikamov, line 32: Just noticed a typo here... there is a space in `capital_expenses_for_expansion of_service` instead of a _

vevetron:
It's probably okay, but I'm not entirely sure I understand why we are building operators and airflow dags for some of these data entities such as "2022_reporting/2022_capital_expenses_by_mode.yml" - is there an expectation the data will change?  Shouldn't it just be a one off data pull?

airflow/plugins/operators/scrape_ntd_xlsx.py
```diff
@@ -0,0 +1,174 @@
...
+
+            self.clean_excel_extract = NtdDataProductXLSXExtractClean(
+                year=self.year,
+                product=self.product + "/" + tab_name,
+                filename=f"{self.year}__{self.product}__{tab_name}.jsonl.gz",
+            )
+
+            self.clean_excel_extract.save_content(
+                fs=get_fs(), content=self.clean_gzipped_content
+            )
```
mjumbewu, line 174: **suggestion:** This feels like it should be broken out to a separate function (e.g. `self.upload_ntd_jsonl(df)` or something) as both this `else` clause and the `for` clause above contain the same code (unless I'm missing some small difference).

```suggestion
        if len(df_dict.keys()) > 1:
            for key, df in df_dict.items():
                self.upload_ntd_jsonl(df)
        else:
            self.upload_ntd_jsonl(df)
    
    def upload_ntd_jsonl(self, df):
        df = df.rename(make_name_bq_safe, axis="columns")
        typer.secho(
            f"read {df.shape[0]} rows and {df.shape[1]} columns",
            fg=typer.colors.MAGENTA,
        )

        self.clean_gzipped_content = gzip.compress(
            df.to_json(orient="records", lines=True).encode()
        )

        tab_name = ""

        tab_name = make_name_bq_safe(key)

        self.clean_excel_extract = NtdDataProductXLSXExtractClean(
            year=self.year,
            product=self.product + "/" + tab_name,
            filename=f"{self.year}__{self.product}__{tab_name}.jsonl.gz",
        )

        self.clean_excel_extract.save_content(
            fs=get_fs(), content=self.clean_gzipped_content
        )
```

airflow/plugins/operators/scrape_ntd_xlsx.py
```diff
@@ -0,0 +1,174 @@
...
+                    product=self.product + "/" + tab_name,
+                    filename=f"{self.year}__{self.product}__{tab_name}.jsonl.gz",
+                )
+
+                self.clean_excel_extract.save_content(
+                    fs=get_fs(), content=self.clean_gzipped_content
+                )
+
+        else:
+            df = df.rename(make_name_bq_safe, axis="columns")
```
mjumbewu, line 152: **question:** Where does `df` come from in this scope?

airflow/dags/sync_ntd_data_xlsx/ridership/historical_raw_monthly_ridership_no_adjustments_or_estimates.yml
```diff
@@ -0,0 +1,5 @@
+operator: operators.NtdDataProductXLSXOperator
+
+product: 'raw_monthly_ridership_no_adjustments_or_estimates'
+xlsx_file_url: 'https://www.transit.dot.gov/sites/fta.dot.gov/files/2024-09/S%26S%20Time%20Series-May%202024-Major%20Only_240903.xlsx'
```
evansiroky, line 4: I believe this file changes monthly. Are we planning on accounting for this and creating the needed automation for monthly downloads or is that something for a future effort?

charlie-costanzo: Hey Evan! Yep – we saw that all of these datasets are being updated regularly, even annual reports for previous years, so automation is built into this work. I just pushed changes with a placeholder value for the schedule - the first day of the month, every month. We can update that from here, but figured that was a reasonable suggestion. Let me know if you have any other thoughts.

airflow/plugins/operators/scrape_ntd_api.py
```diff
@@ -0,0 +1,125 @@
...
+            if response is None or len(response) == 0:
+                logging.info(
+                    f"There is no data to download for {self.year} / {self.product}. Ending pipeline."
+                )
+
+                pass
+            else:
+                logging.info(
+                    f"Downloaded {self.product} data for {self.year} with {len(response)} rows!"
+                )
```
mjumbewu, line 67: **suggestion (non-blocking):** We should probably raise an exception when there is a non-200 response from the API (e.g., a 404 response will still return some JSON, but not in the structure that's valid for the table).

This can be done in a follow-on issue.

mjumbewu (approved):
I'd say just make a follow-on issue related to the suggestion above, and then LGTM.

airflow/dags/sync_ntd_data_xlsx/ridership/historical_raw_monthly_ridership_no_adjustments_or_estimates.yml
charlie-costanzo: Hey @evansiroky – this PR is ready for merge, but my last remaining question relates to the scheduling of the DAG tasks in this PR. Based on the frequency that we see a lot of the endpoints updating (~monthly) I configured the DAG tasks to run once a month (first day of the month, every month, 3am PT), but open to suggestion here.

airflow/plugins/operators/scrape_ntd_api.py
charlie-costanzo: caputured in this ticket:

https://github.com/cal-itp/data-infra/issues/3474

PR #3418, opened 2024-08-12
update payments row access policy to include new contributors

charlie-costanzo:
# Description

Update payments row access polices in macro [create_row_access_policy](https://github.com/cal-itp/data-infra/blob/main/warehouse/macros/create_row_access_policy.sql) to include new GCP group: `mov-project-team@jarv.us` to facilitate querying and contributions to row access policy-protected tables (primarily [fct_payments__rides_v2](https://github.com/cal-itp/data-infra/blob/fd71594f3a59cc7fa44e2a128a45b034cfd090a7/warehouse/models/mart/payments/fct_payments_rides_v2.sql#L2) and [fct_elavon__transactions](https://github.com/cal-itp/data-infra/blob/fd71594f3a59cc7fa44e2a128a45b034cfd090a7/warehouse/models/mart/payments/fct_elavon__transactions.sql#L2))

## Type of change

- [x] New feature

### DIFF
--- a/warehouse/macros/create_row_access_policy.sql
+++ b/warehouse/macros/create_row_access_policy.sql
@@ -85,6 +85,7 @@ filter using (
                   'serviceAccount:bq-transform-svcacct@cal-itp-data-infra.iam.gserviceaccount.com',
                   'serviceAccount:github-actions-services-accoun@cal-itp-data-infra.iam.gserviceaccount.com',
                   'group:cal-itp@jarv.us',
+                  'group:mov-project-team@jarv.us',
                   'domain:calitp.org',
                  ]
 ) }};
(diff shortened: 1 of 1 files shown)


ohrite:
Hi @charlie-costanzo is it possible to add `domain:ministryofvelocity.com` or `group:consultants@ministryofvelocity.com` instead?

charlie-costanzo:
> Hi @charlie-costanzo is it possible to add `domain:ministryofvelocity.com` or `group:consultants@ministryofvelocity.com` instead?

Hey @ohrite  – when creating new groups to allow access to row access policies in GCP, I believe that the domain for a new group's email address is actually automatically assigned the domain of the 'owner' of the GCP organization, and I believe Jarvus is still technically the 'owner' of the Cal-ITP GCP instance. So I wasn't given an option to substitute domains, and I don't think it has much impact other than the settings for the group's access to the tables, but can look into this further if you'd like.

I could also potentially change to a domain-level access to the tables, such as `domain:ministryofvelocity.com`, if preferred, but that would be a less granular way to control who accesses these (more sensitive) payments tables. Currently the only domain that's given broad domain access is `calitp.org`, as the group that Jarvus users have been in names users specifically. Let me know if you'd like me to look further into this though, as well.

themightychris:
Compiler owns the existing GCP project, not Jarvus

erikamov (approved):
Since it will be reworked soon to be less specific on the email domain, we will keep the change.

PR #3419, opened 2024-08-16
payments: expose remaining elavon batch types and modify row access policy

### LINKED ISSUE #3411: Elavon Batch Data
From Rebel: Is it a feature or a bug that the "Fct Elavon Transactions" table only has batch type B and D transactions? I see that there are 4 other tables, splitting out the main Elavon table into A, B, C, D - screenshots below. If it's a bug, could you please modify the main table to have all (A, B, C, D) batch types? It would make things a bit easier for us (I'm not sure how we can re-add A and C to the main table to see the sum).

charlie-costanzo:
# Description

This PR creates new unique intermediate tables for the two remaining batch types of Elavon transactions found in `stg_elavon__transactions` that are not yet exposed in tables

New batch types and tables:
* `A` found in new table `int_elavon__adjustment_transactions`
* `C` found in new table `int_elavon__chargeback_transactions`

the YML file for these tables have also been updated in the dbt project

This PR also adds the service account for `metabase-payments-team` to the create_row_access_policy macro to more easily facilitate access to payments tables with row access policies by payments data team members

Resolves #3411 

As some background:
we used batch types `B` and `D` when creating the `fct_elavon__transactions` table. When constructing that, we tried to incorporate these other two batch types for reconciliation (the batch types above, `A` and `C` ) but they are aggregated at a different grain so we were unable to use them for those calculations.
However, as this batch data is valuable on it's own, once this PR is merged they will be available as their own tables in the warehouse and Metabase to be used for analysis.

## Type of change

- [x] New feature

## How has this been tested?

locally with dbt run and dbt test

## Post-merge follow-ups
- [ ] 
Metabase will need to be configured to expose these two new tables in the staging scheme to the `Payments Team Warehouse` in Metabase

We will also need to verify that the payments team service account is successfully accessing the payments tables protected by the row access policy

### DIFF
--- a/warehouse/macros/create_row_access_policy.sql
+++ b/warehouse/macros/create_row_access_policy.sql
@@ -82,6 +82,7 @@ filter using (
 
 {{ create_row_access_policy(
     principals = ['serviceAccount:metabase@cal-itp-data-infra.iam.gserviceaccount.com',
+                  'serviceAccount:metabase-payments-team@cal-itp-data-infra.iam.gserviceaccount.com',
                   'serviceAccount:bq-transform-svcacct@cal-itp-data-infra.iam.gserviceaccount.com',
                   'serviceAccount:github-actions-services-accoun@cal-itp-data-infra.iam.gserviceaccount.com',
                   'group:cal-itp@jarv.us',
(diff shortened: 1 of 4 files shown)


erikamov (approved):
Even though I don't know much yet about the Elavon data, it is easy to understand the change!
Great job!

PR #3422, opened 2024-08-19
fix(jupyterhub): updated admin and increases disk size

### LINKED ISSUE #3421: Bug: Disk space issues in jupyterhub images
**Describe the bug**
I'm hitting the space limit.

New users are limited to 10 gb.  Normal usage of data-analyses could be 6 gigs, and the cache can grow to 3-4 gigs.  I've hit the space limit a few times so far.

@edasmalchi was manually bumped to 100 gigs, since he uses 22 gigs right now.

**To Reproduce**
`du -sh /home/jovyan/.[!.]* /home/jovyan/* | sort -hr ` to see disk usage

**Expected behavior**
Having enough disk space to use our job!

**Additional context**
I think https://github.com/cal-itp/data-infra/pull/1651 was trying to fix it before but perhaps the wrong parameters were modified.

vevetron:
# Description

Updates admins and increases disk size for new jupyterhub images.

Resolves #3421

## Type of change

## How has this been tested?

YOLO (I really don't know how)

## Post-merge follow-ups

- [ ] No action required
Check if default disk sizes are updated with `df -h`
Currently it looks like:
```
jovyan@jupyter-vevetron:~/data-analyses$ df -h
Filesystem      Size  Used Avail Use% Mounted on
overlay          95G   17G   78G  18% /
tmpfs            64M     0   64M   0% /dev
tmpfs           7.9G     0  7.9G   0% /sys/fs/cgroup
/dev/sdc        9.8G  6.7G  3.2G  68% /home/jovyan
/dev/sda1        95G   17G   78G  18% /etc/hosts
shm              64M     0   64M   0% /dev/shm
tmpfs           7.9G     0  7.9G   0% /proc/acpi
tmpfs           7.9G     0  7.9G   0% /proc/scsi
tmpfs           7.9G     0  7.9G   0% /sys/firmware
```

### DIFF
--- a/kubernetes/apps/charts/jupyterhub/values.yaml
+++ b/kubernetes/apps/charts/jupyterhub/values.yaml
@@ -14,6 +14,8 @@ jupyterhub:
       limit: 10G
       # Very roughly I have seen most usage in the 2-3GB range
       guarantee: 3G
+    storage:
+      capacity: 32Gi
     cpu:
       # this is to put 2 pods per e2-highmem-2 which has 1.93 allocatable
       guarantee: 0.7
@@ -49,10 +51,11 @@ jupyterhub:
         authenticator_class: github
         Authenticator:
           admin_users:
-            - atvaccaro
-            - machow
+            - vevetron
+            - tiffanychu90
             - themightchris
-            - lottspot
+            - evansiroky
+            - charlie-costanzo
     # holds github OAuth creds
     existingSecret: jupyterhub-github-config
   ingress:


kubernetes/apps/charts/jupyterhub/values.yaml
```diff
@@ -14,6 +14,7 @@ jupyterhub:
       limit: 10G
       # Very roughly I have seen most usage in the 2-3GB range
       guarantee: 3G
+    capacity: 32Gi
```
themightychris, line 17: @vevetron it looks like this key should be under a `storage` key: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/HEAD/jupyterhub/values.yaml#L411

themightychris (changes requested):

kubernetes/apps/charts/jupyterhub/values.yaml
vevetron: Yea I messed up! Fixed now.

themightychris (approved):
Looks good to me. It appears that this section of config gets read at runtime by the `hub` service to use while dynamically spawning new user pods, rather than rendered via Helm into any underlying k8s manifests, so we won't see this config change directly reflected in the k8s diff and will have to deploy to verify.

We do see the k8s diff showing a length increase in the `values.yaml` it's dumping into a secret so we can be confident it's at least getting that far

vevetron:
So after merging in and running github actions, I logged out and stopped my server.  My admin rights changed.  But I don’t see my disk size changed.

 Went to Lens -> Cluster -> Storage -> PVC, 
 Updates Spec -> Storage to 100gb to match Edmasklies
 Now my server df -h matches with 100gb of storage

Side note:
Deleted PVC of walteryu, now i'm spinning up his server again.
His old volume disappeared, his new volume is 32gb, which matches our config.

I'm not sure if his files disappeared or he didn't have anything to being with.

I gave him some files and then shut down his server and then deleted his pvc, then had him start up again.
His files were gone, he got a new PV.  

Thus I think it makes the most sense just to manually increase the PV size in PVC of existing users (Andrew knew this).  Then any new user will have 32 gigs of storage.

Issue #3420, opened 2024-08-19
Bug: `transform_warehouse` DAG appears to not be running on Mondays


charlie-costanzo:
**Describe the bug**
Airflow/Composer `transform_warehouse` DAG appears to not be running on Mondays

**To Reproduce**
Airflow/Composer GUI schedule history shows `transform_warehouse` DAG not running on Mondays, although schedule interval is defined as `0 14 * * 1-6`

**Expected behavior**
DAG should run every day, Monday-Saturday

mjumbewu:
@charlie-costanzo **tl;dr: I think the easiest fix would be to set the [`start_date` for the DAG](https://github.com/cal-itp/data-infra/blob/796a3e178d6d214fa58bc2da67215d5ede577a49/airflow/dags/transform_warehouse/METADATA.yml#L8C5-L8C28) to a static value.**

----------

The Monday runs are getting executed -- it's the Saturday runs that are getting skipped. Because of how Airflow's scheduler works, [each run gets executed at the end of its interval](https://airflow.apache.org/docs/apache-airflow/2.0.1/scheduler.html#:~:text=Let%E2%80%99s%20Repeat%20That%2C%20the%20scheduler%20runs%20your%20job%20one%20schedule_interval%20AFTER%20the%20start%20date%2C%20at%20the%20END%20of%20the%20period.) (i.e., at the beginning of the next schedule interval).

It looks like the Saturday runs are getting _scheduled_ (as you can see for the Aug 03 run in the following screenshot) just like the Monday through Friday runs:
![image](https://github.com/user-attachments/assets/e8c139a2-ecf7-4447-95e3-3921c482c52c)

But they're never actually getting _executed_ (as you can see by the missing Aug 03 run in the following screenshot):
![image](https://github.com/user-attachments/assets/2dd127ca-907d-4694-ac2e-6706a1edb378)

My hypothesis is that this is because of the `start_date` definition in the DAG metadata:
https://github.com/cal-itp/data-infra/blob/796a3e178d6d214fa58bc2da67215d5ede577a49/airflow/dags/transform_warehouse/METADATA.yml#L1-L8

[...]

PR #3456, opened 2024-09-13
Fix transform_warehouse schedule to run on Mondays

erikamov:
# Description

The transform_warehouse DAG is not running on Mondays and this PR is intended to fix the issue #3420.

Based on Airflow concept, Mondays should be running the data interval Saturday-Sunday, but since the `start_date` is configured to only One Day Ago, it does not have anything to run on Mondays because Sundays are not part of the schedule.

As @mjumbewu mentioned on the Issue and going through the problem with him, we found out that setting up a real start date would fix the problem.

There is also a previous issue that was fixed the same way: https://github.com/cal-itp/data-infra/pull/3323

For more information about the problem see the screenshot bellow that shows (on the left side) a list of dates from Mondays to Fridays, but they are actually running from Tuesdays to Saturdays as the TIMESTAMP, Start date, and End date shows.
Screenshot 2024-09-10 at 11 17 00 AM

Airflow schedule works different than regular "jobs", to clarify here what the documentation says:

What does [execution_date](https://airflow.apache.org/docs/apache-airflow/stable/faq.html#what-does-execution-date-mean) mean?

Airflow was developed as a solution for ETL needs. In the ETL world, you typically summarize data. So, if you want to summarize data for 2016-02-19, you would do it at 2016-02-20 midnight UTC, which would be right after all data for 2016-02-19 becomes available. This interval between midnights of 2016-02-19 and 2016-02-20 is called the data interval, and since it represents data in the date of 2016-02-19, this date is also called the run’s logical date, or the date that this DAG run is executed for, thus execution date.

[Dates Concept from airflow](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dag-run.html#data-interval)
`All dates in Airflow are tied to the data interval concept in some way. The “logical date” (also called execution_date in Airflow versions prior to 2.2) of a DAG run, for example, denotes the start of the data interval, not when the DAG is actually executed.`

`Similarly, since the start_date argument for the DAG and its tasks points to the same logical date, it marks the start of the DAG’s first data interval, not when tasks in the DAG will start running. In other words, a DAG run will only be scheduled one interval after start_date.`

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Tested schedules on Airflow locally.

## Post-merge follow-ups

- [ ] No action required

Monitor production Airflow to ensure that the changes take effect as expected.

### DIFF
--- a/airflow/dags/transform_warehouse/METADATA.yml
+++ b/airflow/dags/transform_warehouse/METADATA.yml
@@ -5,7 +5,7 @@ tags:
 default_args:
     owner: airflow
     depends_on_past: False
-    start_date: !days_ago 1
+    start_date: "2024-09-17"
     email:
       - "hello@calitp.org"
     email_on_failure: False
--- a/airflow/dags/transform_warehouse/README.md
+++ b/airflow/dags/transform_warehouse/README.md
@@ -6,6 +6,10 @@ This DAG orchestrates the running of the Cal-ITP dbt project and deployment of a
 
 This DAG has some special considerations:
 
+- The schedule interval is 14:00 UTC Mondays to Saturdays. Mondays will be actually running the data interval for the previous Saturday-Sunday.
+
+- This DAG should not run on Sundays because of the other DAG `transform_warehouse_full_refresh_sunday`.
+
 - If a task fails, look carefully before assuming that clearing the task will help. If the failure was caused by a `DbtModelError`, there is an issue with the SQL or data in an individual model and clearing the task will not help until that issue is fixed.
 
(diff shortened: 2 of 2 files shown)


erikamov:
It was also pointed to change the start_date to a real date instead of a dynamic date, but I think we need to change it after this PR to see the results of each change.

charlie-costanzo (approved):
Thank you for all the time you spent looking into this Erika! I agree that this should be the resolution we're looking for, but we will monitor after merge and know soon enough.

PR #3457, opened 2024-09-13
fix: manually add cleverapi.bigbluebus.com intermediate certificate to image

themightychris:
# Description

Provides a short-term fix to https://github.com/cal-itp/data-infra/issues/3452.

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Running `curl -vvvv https://cleverapi.bigbluebus.com/` within the locally-built container image

## Post-merge follow-ups

- [ ] Deploy new image tag to cluster


evansiroky (commented):
It's likely that this won't be merged given that we generally don't want to have to add custom certificates to what should be generic HTTP requests. That being said, this proof-of-concept is a helpful demonstration.

Issue #3403, opened 2024-07-23
NTD Ingestion – Remaining Tables in Dataset


charlie-costanzo:
## User story / feature request

Part of #3401 

As a data engineer, I would like to ingest the data in the form of gcs blob storage and external tables for the remaining tables in the NTD dataset, extending the work found in `scrape_ntd.py` and `annual_database_service.yml` (found below).

Building upon the work completed in #3345

Existing NTD patterns:
- [script/scrape_ntd.py](https://github.com/cal-itp/data-infra/blob/main/script/scrape_ntd.py)
- [airflow/dags/create_external_tables/ntd_data_products/annual_database_service.yml](https://github.com/cal-itp/data-infra/blob/main/airflow/dags/create_external_tables/ntd_data_products/annual_database_service.yml)

General Cal-ITP Pipeline Patterns
- [Data Pipelines Documentation](https://docs.calitp.org/data-infra/architecture/data.html#data-pipelines)
  - [Adding a New Data Source](https://docs.calitp.org/data-infra/architecture/data.html#adding-a-new-data-source)
  - [Previous PR Patterns](https://github.com/cal-itp/data-infra/pull/2376)

### Acceptance Criteria

I can successfully access the data for the following tables from the https://www.transit.dot.gov/ntd/data-product website in the warehouse as external tables:
- Assets
- Expenses
- Fares/Funding
- Resources
- Security and Safety
- Service (previous years)

### Notes
---

erikamov:
External tables on branch [3403-ntd-api-external-tables](https://github.com/cal-itp/data-infra/compare/3403-ntd-api-external-tables)

erikamov:
* [External tables for 2022 API data](https://console.cloud.google.com/bigquery?ws=!1m4!1m3!3m2!1scal-itp-data-infra-staging!2sexternal_ntd__annual_reporting) were tested and working. I am still reviewing field types for each column, but could be deployed as is now (using default STRING) and as we work on modeling we can keep changing types as needed. 
 
* `historical__major_safety_events` and `historical__nonmajor_safety_and_security_events` external tables for Historical API data are failing because of the column names `:@computed_region_8fe2_rd7y` and `:@computed_region_m2nu_4dib`.
 @charlie-costanzo @mjumbewu  have you seen this problem before?

    The full error:
[...]

PR #3465, opened 2024-09-19
Create NTD external tables for 2022 API data

erikamov:
# Description

This PR creates [#3403] the NTD external tables for 2022 API data

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

## How has this been tested?

It was tested with local Airflow and the tables were created on cal-itp-data-infra-staging.external_ntd__annual_reporting

## Post-merge follow-ups

- [ ] No action required
- [x] Actions required (specified below)

Validate the creation of the tables on cal-itp-data-infra.external_ntd__annual_reporting

PR #3470, opened 2024-09-20
NTD external tables for Complete Monthly Ridership with Adjustments and Estimates Data

erikamov:
# Description

This PR creates the NTD external tables for XLSX Complete Monthly Ridership with Adjustments and Estimates Data
#3403

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

Tested with local Airflow and the tables were created on cal-itp-data-infra-staging.external_ntd__ridership

## Post-merge follow-ups

- [ ] No action required

Validate the creation of all tables on cal-itp-data-infra.external_ntd__ridership


vevetron:
Should we consider putting the data into calitp-ntd-data-products instead where the existing ntd data gets dropped?  And then we stop maintaining `monthly-ridership-with-adjustments`.

evansiroky:
Since there was also a new bucket created within #3465, I think we should probably start with a fresh bucket.

evansiroky (approved):

PR #3545, opened 2024-11-13
dag to scrape and save the current ridership URL from the NTD portal

charlie-costanzo:
# Description

When we created Airflow operator to ingest the NTD monthly ridership tables every month in XLSX format, we didn't realize that DOT changes the file name every time it publishes the data.

Instead of relying on someone to remember to manually run the data ingestion every month with the new link, this PR dynamically scrapes the webpage for the data download to retrieve the new link, save it in airflow as an environment variable, and allows the request library to utilize the updated link. To do this, it also makes the ingest airflow task dependent on the success of the URL scraper.

## Type of change

- [x] New feature

## How has this been tested?
local airflow
Screenshot 2024-11-18 at 6 14 13 PM

## Post-merge follow-ups

monitor for expected behavior


airflow/dags/sync_ntd_data_xlsx/scrape_ntd_ridership_xlsx_url.py
```diff
@@ -0,0 +1,41 @@
...
+    # print('https://www.transit.dot.gov' + file_link)
+
+    validated_url = parse_obj_as(HttpUrl, updated_url)
+
+    logging.info(f"Validated URL: {validated_url}.")
+
+    # Set or overwrite the variable
+    # Will this work?
+    # Variable.set("CURRENT_NTD_RIDERSHIP_URL", f"'{validated_url}'")
+    Variable.set("CURRENT_NTD_RIDERSHIP_URL", validated_url)
```
mjumbewu, line 41: @charlie-costanzo I don't actually know whether this will work (i.e. if the value of the variable will immediately and persistently be available to the rest of the tasks). I don't hate it, but for what it's worth, the [documented Airflow recommendation](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/variables.html#:~:text=Variables%20are%20global%2C%20and%20should%20only%20be%20used%20for%20overall%20configuration%20that%20covers%20the%20entire%20installation%3B%20to%20pass%20data%20from%20one%20Task/Operator%20to%20another%2C%20you%20should%20use%20XComs%20instead.) is to use XComs for this purpose. Have you had a chance to look into that?

charlie-costanzo: just switched to using xcoms if you're able to take a look

.pre-commit-config.yaml
```diff
@@ -15,9 +15,11 @@ repos:
     rev: 6.0.0
     hooks:
       - id: flake8
-        args: ["--ignore=E501,W503"] # line too long and line before binary operator (black is ok with these)
+        args: ["--ignore=E501,W503,E231"] # line too long and line before binary operator (black is ok with these) and explicitly ignore the whitespace after colon error
```
charlie-costanzo, line 18: Added `--ignore=E231` to flake8 args to ignore "missing whitespace after colon" errors in lambda functions, which were causing CI failures despite being properly formatted locally

.pre-commit-config.yaml
```diff
@@ -15,9 +15,11 @@ repos:
...
         types:
           - python
+        # Suppress SyntaxWarning about invalid escape sequence from calitp-data-infra dependency without modifying source
+        entry: env PYTHONWARNINGS="ignore::SyntaxWarning" flake8
```
charlie-costanzo, line 22: Added `entry: env PYTHONWARNINGS="ignore::SyntaxWarning" flake8` to suppress a SyntaxWarning about invalid escape sequence from the calitp-data-infra dependency without modifying its source code

evansiroky (approved):

PR #3564, opened 2024-11-26
revise misleading name of state highway network flag column on dim_stops_latest

charlie-costanzo:
# Description

Realized the column name for the state highway network boolean on the dim_stops_latest table was misleading, renamed to: `on_state_highway_network`

## Type of change
- [x] New feature

## How has this been tested?

locally with dbt

## Post-merge follow-ups
run table in production as requested


erikamov (approved):
Looks good.

PR #3663, opened 2025-01-23
fix: remaining ntd external table error

charlie-costanzo:
# Description

Unfortunately following #3658 there was a case of one more failure that needs to be rectified due to unpushed local changes, a two character fix

## Type of change

## How has this been tested?
Screenshot 2025-01-23 at 2 12 22 PM

## Post-merge follow-ups
clear task and make sure it's running appropriately


erikamov (approved):

vevetron:
magic!

PR #3655, opened 2025-01-21
Additional ntd time series external tables

charlie-costanzo:
# Description
This PR is to be merged once #3652 is merged

This PR creates the external tables for the NTD XLSX endpoints added to the pipeline in PR #3652. The external tables are 1:1 with the tabs available in the spreadsheets that are being ingested.

## Type of change
- [x] New feature

## How has this been tested?
Screenshot 2025-01-22 at 12 28 33 PM

## Post-merge follow-ups
- [x] Actions required (specified below)
ensure external tables are being created as expected

PR #3665, opened 2025-01-27
NTD time series data: create warehouse tables

charlie-costanzo:
# Description

Following up on #3655, this PR creates new warehouse tables and associated yml files for the new NTD time series endpoints that we are ingesting:
- staging tables with cleaning and casting
- intermediate tables that pivot on year
- mart tables that reflect the intermediate transformations

## Type of change

- [x] New feature

## How has this been tested?
Screenshot 2025-01-30 at 12 51 09 PM

## Post-merge follow-ups
- [x] No action required


warehouse/models/mart/ntd_funding_and_expenses/fct_capital_expenditures_time_series_facilities.sql
```diff
@@ -0,0 +1,62 @@
...
+    _2015,
+    _2016,
+    _2017,
+    _2018,
+    _2019,
+    _2020,
+    _2021,
+    _2022,
+    _2023,
+    _2023_mode_status,
```
evansiroky, line 44: When each year is a column in a database table it is really tedious to make a query that selects every one of these columns to make a metabase query. Is there a way to transform these tables so that we have a single column of `year` instead of each year being a column? This would create more rows and thus make it easier to make database queries.

charlie-costanzo: > When each year is a column in a database table it is really tedious to make a query that selects every one of these columns to make a metabase query. Is there a way to transform these tables so that we have a single column of `year` instead of each year being a column? This would create more rows and thus make it easier to make database queries.

Hey @evansiroky – without a doubt, I was definitely planning on adding that to the next round of work on these tables. I was just planning to get these into the warehouse quickly as MVPs in their current form and then do modeling iterations based on what's most pressing/useful immediately, but I can make these modeling changes before merging this PR if preferred

warehouse/models/mart/ntd_funding_and_expenses/_mart_ntd_funding_and_expenses.yml
```diff
@@ -0,0 +1,35 @@
+version: 2
+
+models:
+  - name: dim_capital_expenditures_time_series_read_me_data_dictionary
+  - name: dim_operating_and_capital_funding_time_series_read_me_data_dictionary
+  - name: dim_service_data_and_operating_expenses_time_series_by_mode_read_me_data_dictionary
```
erikamov, line 6: With cost reduction in my mind, I think we don't need these tables with "read_me" on the name, if you query the external table there are a lot of null columns. We would be spending time generating those tables, but they are not useful.

dim_capital_expenditures_time_series_read_me_data_dictionary
dim_operating_and_capital_funding_time_series_read_me_data_dictionary
dim_service_data_and_operating_expenses_time_series_by_mode_read_me_data_dictionary
stg_ntd__capital_expenditures_time_series__read_me_data_dictionary
stg_ntd__operating_and_capital_funding_time_series__read_me_data_dictionary
stg_ntd__service_data_and_operating_expenses_time_series_by_mode__read_me_data_dictionary

warehouse/models/mart/ntd_funding_and_expenses/fct_capital_expenditures_time_series_facilities.sql
erikamov: Charlie, there is an example you can follow to pivot the table and build the way Evan is asking: `int_ntd__monthly_ridership_with_adjustments_vrh.sql`

erikamov (changes requested):

erikamov (approved):
Looks great!

vevetron:
There is a resolved conversation - I'm okay with merging this in as is, but I don't like how we have all these columns listed out even at the cal-itp-data-infra.external_ntd__funding_and_expenses.historical__capital_expenditures_time_series__facilities stage - but realistically if we start using these tables more often, we can work on making it sustainable year after year at a later date.  For now we can hard code it in.

PR #3690, opened 2025-02-12
Switched out the default image for 2025.1.15, keeps older image

vevetron:
Switched out the default image for 2025.1.15, keeps older image

# Description

It appears the 2025.1.15 image which uses python 3.11 is well accepted by the analytics team, will make it to default.  We can keep the older image (it's only 2 GB).  Also the power user will also use the default image now.

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

Not tested.

## Post-merge follow-ups

- [ ] No action required
Just make sure jupyterhub is okay today.


tiffanychu90 (approved):
Thanks!

PR #3737, opened 2025-02-26
Update  Jupyter image so node version gets updated

vevetron:
# Description

The jupyter images are having serious issues with the node version in the current jupyter notebook.

This switched to a newer image on the same version of python. It also uses something like node ~22, while the old used node ~20.

Resolves #(https://github.com/cal-itp/data-analyses/issues/1388)

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Can't test locally, but the build.log looks good!

## Post-merge follow-ups

- [ ] No action required
Test in prod


vevetron:
Has too many weird reverts in the git log.

PR #3746, opened 2025-03-04
payments: create new operators and dags for littlepay feed v3 sync and parse

### LINKED ISSUE #3705: Payments: refactor sync and parse Littlepay operators for feed v3 data (new and existing providers)
Following the patterns of our previous agency Littlepay data ingestion, we need to ingest Littlepay payments data for Nevada County in order to provide dashboarding and other analysis

charlie-costanzo:
# Description
Littlepay has transitioned to their data feed v3, and data for our most recent open loop agency to come online (Nevada County Connects) is being provided in their v3 format.

This PR iterates on our Airflow sync and parse tasks for feed v1 to accommodate the changes found in feed v3: 
* deprecate dynamic creation of source bucket name due to changes in composition in v3
* stand up new raw and parsed buckets, secrets
* updates based on breaking changes in table names, add new tables

Resolves #3705 

## Type of change
- [x] New feature

## How has this been tested?
Screenshot 2025-03-17 at 4 01 30 PM
Screenshot 2025-03-17 at 4 13 15 PM

## Post-merge follow-ups
Turn on dags in GUI, observe for expected behavior

### DIFF
--- a/airflow/plugins/operators/littlepay_raw_sync_feed_v3.py
+++ b/airflow/plugins/operators/littlepay_raw_sync_feed_v3.py
@@ -48,10 +48,10 @@ class LittlepayFileKey(str):
     def entity(self) -> str:
         return self.split("/")[2]
 
-    # entity is now in location 3 (instead of 2)
+    # filename is now in location 3 (instead of 2)
     @property
     def filename(self) -> str:
-        return self.split("/")[2]
+        return self.split("/")[3]  # Return full filename with extension
 
 
 class LittlepayS3Object(BaseModel):
--- a/airflow/plugins/operators/littlepay_raw_sync_feed_v3.py
+++ b/airflow/plugins/operators/littlepay_raw_sync_feed_v3.py
@@ -156,14 +156,11 @@ class LittlepayRawSyncV3(BaseOperator):
         self,
         *args,
         instance: str,
-        # pull src_bucket from task yml
         src_bucket: str,
         access_key_secret_name: str,
         **kwargs,
     ):
         self.instance = instance
-        # deprecate because source bucket names are not predictable in v3
-        # self.src_bucket = f"littlepay-prod-{instance}-datafeed"
         self.src_bucket = src_bucket
         self.access_key_secret_name = access_key_secret_name
         super().__init__(**kwargs)
--- a/airflow/plugins/operators/littlepay_raw_sync_feed_v3.py
+++ b/airflow/plugins/operators/littlepay_raw_sync_feed_v3.py
@@ -1,0 +1,263 @@
+import concurrent
+import json
+import os
+import traceback
+from concurrent.futures import Future, ThreadPoolExecutor
+from datetime import datetime
+from typing import ClassVar, Dict, List, Optional
+
+import boto3
+import pendulum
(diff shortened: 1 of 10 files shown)


airflow/plugins/operators/littlepay_raw_sync_feed_v3.py
```diff
@@ -0,0 +1,260 @@
...
+from pydantic.class_validators import validator
+from pydantic.error_wrappers import ValidationError
+from pydantic.main import BaseModel
+from tqdm import tqdm
+from tqdm.contrib.logging import logging_redirect_tqdm
+
+from airflow.models import BaseOperator
+
+# new bucket because of schema changes, need to investigate backfill status with transitions to v3 or handling otherwise
+LITTLEPAY_RAW_BUCKET = os.getenv("CALITP_BUCKET__LITTLEPAY_RAW_V3")
```
mjumbewu, line 28: **suggestion:** Is this file the same as https://github.com/cal-itp/data-infra/blob/main/airflow/plugins/operators/littlepay_raw_sync.py except for the environment variable referenced here? If so, it seems like unnecessary duplication. It would be nice to pass the raw bucket value in to where it's needed instead of creating an entirely new module for it (or alternatively to derive the `RawLittlepayFileExtract` class and just override the `bucket` attribute).

airflow/plugins/operators/littlepay_to_jsonl_feed_v3.py
```diff
@@ -0,0 +1,132 @@
...
+            "customer-funding-sources",
+            "device-transaction-purchases",
+            "device-transactions",
+            "micropayment-adjustments",
+            "micropayment-device-transactions",
+            "micropayments",
+            "products",
+            "refunds",
+            "settlements",
+            "terminal-device-transactions",
```
mjumbewu, line 100: **suggestion:** Same note as [above](https://github.com/cal-itp/data-infra/pull/3746/files#r1999747397) here. Maybe we could make `entities` a class-level variable in [littlepay_to_jsonl.py](https://github.com/cal-itp/data-infra/blob/main/airflow/plugins/operators/littlepay_to_jsonl.py), and then just override the value here.

mjumbewu:
Left a couple notes above. If there's something I'm missing as a reason for the code duplication, please feel free to push back!

airflow/plugins/operators/littlepay_raw_sync_feed_v3.py
charlie-costanzo: @mjumbewu  so we're actually unable to dynamically generate the source bucket name anymore, like we were doing here, because they are now unpredictable. So I've made changes here as well to pull that information from the task, but can definitely explore handling in one operator if you think that's preferred.

charlie-costanzo: ah also worth mentioning @mjumbewu is that my plan was to have these up in parallel so that we could sync/parse our new open loop payments agency (Nevada County Connects, the first agency that is on the new feed v3) while we wait for Littlepay to turn on feed v3 for our current feed v1 agencies, and then turn feed v1 tasks off as soon as that's complete. but very open to reconsidering approach

PR #3750, opened 2025-03-05
Add documentation to NTD 2022 and 2023 Agency Information

erikamov:
# Description

This PR adds documentation to 2022 and 2023 NTD Agency Information tables and columns.

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

Tested running model locally and generating docs.

Screenshot 2025-03-05 at 4 35 52 PM

## Post-merge follow-ups

- [x] No action required

### DIFF
--- a/warehouse/models/mart/ntd_annual_reporting/_mart_ntd_annual_reporting.yml
+++ b/warehouse/models/mart/ntd_annual_reporting/_mart_ntd_annual_reporting.yml
@@ -1,6 +1,254 @@
 version: 2
 
+x-common-fields:
+  - &ntd_id
+    name: ntd_id
+    description: '{{ doc("ntd_id") }}'
+    tests:
+      - not_null
+  - &agency_name
+    name: agency_name
+    description: '{{ doc("ntd_agency_name") }}'
+  - &legacy_ntd_id
+    name: legacy_ntd_id
+    description: '{{ doc("ntd_legacy_id") }}'
+  - &state_parent_ntd_id
+    name: state_parent_ntd_id
+    description: '{{ doc("ntd_state_parent_ntd_id") }}'
+  - &reporter_type
+    name: reporter_type
+    description: '{{ doc("ntd_reporter_type") }}'
+  - &organization_type
+    name: organization_type
+    description: '{{ doc("ntd_organization_type") }}'
+  - &doing_business_as
+    name: doing_business_as
+    description: '{{ doc("ntd_doing_business_as") }}'
+  - &mailing_address_line_1
+    name: address_line_1
+    description: '{{ doc("ntd_mailing_address_line_1") }}'
+  - &mailing_address_line_2
+    name: address_line_2
+    description: '{{ doc("ntd_mailing_address_line_2") }}'
+  - &mailing_p_o__box
+    name: p_o__box
+    description: '{{ doc("ntd_mailing_p_o_box") }}'
+  - &mailing_city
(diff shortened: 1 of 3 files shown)


tiffanychu90 (approved):
Between 2022 and 2023 tables, do `ntd_id` or any other identifier columns change? 
For my understanding, is this just the best practice to separate out the years with  `[year]_[tablename]` (ex: `dim_2022/2023_agency_information`) for dimension tables, instead of a concatenated version where it's `dim_agency_information` with multiple years, and using that `year` or `dt` column to flag specific years? @erikamov

erikamov:
> Between 2022 and 2023 tables, do `ntd_id` or any other identifier columns change? For my understanding, is this just the best practice to separate out the years with `[year]_[tablename]` (ex: `dim_2022/2023_agency_information`) for dimension tables, instead of a concatenated version where it's `dim_agency_information` with multiple years, and using that `year` or `dt` column to flag specific years? @erikamov

@tiffanychu90  We have the specific year tables [dim_2022_agency_information](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/ntd_annual_reporting/dim_2022_agency_information.sql) and [dim_2023_agency_information](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/ntd_annual_reporting/dim_2023_agency_information.sql).

The `dim_agency_information` was created in 2023 as [dim_annual_database_agency_information](https://github.com/cal-itp/data-infra/commit/0f997fe95267e1e564fc33e452136f1c249fc810#diff-960d59eddf277c6d7312c8a8fcdbb4a24c5c36e44c4efab9eaa8b915fbee75e7) and renamed to `dim_annual_ntd_agency_information` and then to the current name. So I am just replacing the external table that generated the table, didn't change how it was built. If nobody uses this table, or should not use it anymore, we could delete it and users would use only the specific year tables.

tiffanychu90:
> > Between 2022 and 2023 tables, do `ntd_id` or any other identifier columns change? For my understanding, is this just the best practice to separate out the years with `[year]_[tablename]` (ex: `dim_2022/2023_agency_information`) for dimension tables, instead of a concatenated version where it's `dim_agency_information` with multiple years, and using that `year` or `dt` column to flag specific years? @erikamov
> 
> @tiffanychu90 We have the specific year tables [dim_2022_agency_information](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/ntd_annual_reporting/dim_2022_agency_information.sql) and [dim_2023_agency_information](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/ntd_annual_reporting/dim_2023_agency_information.sql).
> 
> The `dim_agency_information` was created in 2023 as [dim_annual_database_agency_information](https://github.com/cal-itp/data-infra/commit/0f997fe95267e1e564fc33e452136f1c249fc810#diff-960d59eddf277c6d7312c8a8fcdbb4a24c5c36e44c4efab9eaa8b915fbee75e7) and renamed to `dim_annual_ntd_agency_information` and then to the current name. So I am just replacing the external table that generated the table, didn't change how it was built. If nobody uses this table, or should not use it anymore, we could delete it and users would use only the specific year tables.

Ok! No problem, let's merge this, and if after using it awhile, it's confusing, we can always go and delete it. But I'm sure it'll be fine.

vevetron:
@erikamov should this be part of a sprint?

Issue #3751, opened 2025-03-06
Add new airtable data crosswalking agencies to MPO and RTPAs


evansiroky:
## User story / feature request

As a data analyst,
I would like to know which RTPA a transit agency is predominately associated with,
So that I can make various aggregate analyses at the RTPA level.

### Acceptance Criteria

New airtable data associating transit agencies with MPOs and RTPAs is added to warehouse.

### Notes

New columns have been added to the Calfiornia Transit Base in the organizations table. These are `MPO` and `RTPA`. These show which MPO and RTPA each transit agency is headquartered in. Since there are not MPOs covering the entire state, not all transit agencies are associated with an MPO, however all transit agencies will be associated with a single RTPA given that there are RTPAs covering the entire state of California.

evansiroky:
@3vivekb or @fsalemi could one of you please work on this?

fsalemi:
@evansiroky I will check this out.

evansiroky:
Re-opening after yesterday's discussion to try to add organization keys that can be used to reference RTPA and MPO records.

fsalemi:
This is completed.

PR #3759, opened 2025-03-10
Adding RTPA and MPO to external_airtable_california_transit_organizations.yml

fsalemi:
# Description

Adding RTPA and MPO to external_airtable_california_transit_organizations.yml in order to eventually add these fields to the dim_organizations in the GBC warehouse.

Resolves #\[[3751](https://github.com/cal-itp/data-infra/issues/3751)\]

## Type of change

- [X] New feature
- [ ] Documentation

## Post-merge follow-ups

- [ ] No action required

Once this step is completed, Staging, Intermediate and Marts models will be updated to include these fields in the dim_organizations.

### DIFF
--- a/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_organizations.yml
+++ b/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_organizations.yml
@@ -262,4 +262,4 @@ schema_fields:
     mode: REPEATED
   - name: mpo
     type: STRING
-    mode: NULLABLE
+    mode: REPEATED
--- a/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_organizations.yml
+++ b/airflow/dags/create_external_tables/airtable/external_airtable_california_transit_organizations.yml
@@ -57,9 +57,9 @@ schema_fields:
   - name: gtfs_schedule_status
     type: STRING
     mode: REPEATED
-  - name: mpo_rtpa
-    type: STRING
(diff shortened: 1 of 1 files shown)


fsalemi:
With this attempt:
1. Correctd the Mode for rtpa,
2. Deleted the mpo_rtpa,
3. verified that no historical information needed.

airflow/dags/create_external_tables/airtable/external_airtable_california_transit_organizations.yml
```diff
@@ -257,3 +257,9 @@ schema_fields:
   - name: ntd_id_2022
     type: STRING
     mode: NULLABLE
+  - name: rtpa
+    type: STRING
+    mode: REPEATED
+  - name: mpo
+    type: STRING
+    mode: NULLABLE
```
evansiroky, line 265: I think this should still be repeated since it's a linked record. I belive there will always be an array although, sometimes the array will be empty when there is no MPO set for the record and when there is an MPO set then there will be one value within the array.

evansiroky (approved):

airflow/dags/create_external_tables/airtable/external_airtable_california_transit_organizations.yml
fsalemi: @evansiroky, fixed.
I focused on the nullability of mpo as you have defined it in Airtable: not all transit agencies are associated with an MPO.

Issue #3765, opened 2025-03-11
Upgrade to composer-2.8.3-airflow-2.6.3 before March 19, 2025


mjumbewu:
## User story / feature request

We want to be on an extended support version of Cloud Composer. This is part of an incremental approach to getting onto the latest extended support version, `composer-2.10.2-airflow-2.10.2`

### Acceptance Criteria

- [ ] Cloud composer image is `composer-2.8.3-airflow-2.6.3`

mjumbewu:
There are a couple more dependencies to be updated...

mjumbewu:
I'm regrettably re-opening this issue, as the upgrade wasn't successful.

The new cluster's dependencies were consistent, but the workers were overloaded by an infinite recursion issue similar to https://github.com/apache/airflow/pull/35048. This issue happens when there is a logging error, but then one of the log handlers logs some information that causes another logging error, which creates the recursion.

An instance of that recursion starts [here](https://console.cloud.google.com/logs/query;cursorTimestamp=2025-04-02T14:04:40.417318488Z;endTime=2025-04-02T14:35:10.426Z;pinnedLogId=2025-04-02T14:04:40.426081504Z%2Fyvt92hxfjwrvuy8o;query=resource.type%3D%22k8s_container%22%0Aresource.labels.container_name%3D%22airflow-worker%22%0Aresource.labels.namespace_name%3D%22composer-2-8-3-airflow-2-6-3-f0ad6d77%22%0Aresource.labels.project_id%3D%22cal-itp-data-infra%22%0Aresource.labels.cluster_name%3D%22us-west2-calitp-airflow2-pr-88ca8ec6-gke%22%0Aresource.labels.pod_name%3D%22airflow-worker-dbvzs%22%0Aresource.labels.location%3D%22us-west2%22%0Atimestamp%3D%222025-04-02T14:04:40.417279904Z%22%0AinsertId%3D%225oorsbwmdu6eu4t7%22;startTime=2025-04-01T13:35:10.426Z?project=cal-itp-data-infra) (on the right use "Pin and show in context" to show the kubernetes container logs). The structure of the loop is:
```
Traceback (most recent call last): 
File "/opt/python3.11/lib/python3.11/site-packages/airflow/utils/log/secrets_masker.py", line 256, in _redact
    return self.replacer.sub("***", item)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3769, opened 2025-03-11
Update calitp-data-infra package to support pydantic 1.10

mjumbewu:
# Description

The `composer-2.8.3-airflow-2.6.3` image relies on `pydantic==1.10.15`, so we need to allow for it here.

Dependency for #3765 

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

- Ran `pytest` against the package under Python 3.8 and 3.11
- Built the airflow image including this latest code with a modified `Dockerfile.composer`

## Post-merge follow-ups

- [x] No action required


vevetron (approved):
LGTM!

Issue #3666, opened 2025-01-27
Add lifecycle rules to all non-production GCS buckets and BigQuery tables


erikamov:
## User story / feature request

As a Developer,
When I create test data on Google Cloud Storage buckets or BigQuery datasets or tables designated for testing or development,
I want the test data deleted after 30 days.

As part of the Cal-ITP Cost Reduction Plan, we need to clear test buckets to reduce Goggle Cloud monthly costs.
[Cal-ITP Test Buckets Cleanup.pdf](https://github.com/user-attachments/files/18564990/Cal-ITP.Test.Buckets.Cleanup.pdf)

### Acceptance Criteria

When I list GC Buckets designated for testing or development the should not contain data older than 30 days.

### Notes

**Make sure Autoclass disabled on those Buckets.**

erikamov:
Post message about the cleanup and attached PDF on slack channel calitp-data-infra. Should I post in any other channel?

tiffanychu90:
* One things I'm noticing for my set of BQ tables for dbt (sheet `Pivot Table 1`), I always follow the instructions [here](https://github.com/cal-itp/data-infra/tree/main/warehouse/README.md) to run `poetry run dbt run --full-refresh` if you haven't been creating new tables for a couple weeks.
   * I only really work in one section of the warehouse though (`mart_gtfs`), so I probably could have a lot of tables in `tiffany_*` portions of `cal-itp-data-infra-staging` deleted. What would be the command I should run to just bring in `mart_gtfs`? 
   * If the analysts can always just bring in the specific portion of the tables they need for adding new tables in dbt, which seems fairly infrequent nowadays except for me + Vivek, we can delete `christian`, `eric`, `mine` (and I'll repopulate with the portion I need).
* Let's clean up other users who aren't working on adding warehouse tables in the foreseeable future: `soren`, anyone else?

erikamov:
> * One things I'm noticing for my set of BQ tables for dbt (sheet `Pivot Table 1`), I always follow the instructions [here](https://github.com/cal-itp/data-infra/tree/main/warehouse/README.md) to run `poetry run dbt run --full-refresh` if you haven't been creating new tables for a couple weeks.
>   
>   * I only really work in one section of the warehouse though (`mart_gtfs`), so I probably could have a lot of tables in `tiffany_*` portions of `cal-itp-data-infra-staging` deleted. What would be the command I should run to just bring in `mart_gtfs`?
[...]

PR #3822, opened 2025-04-10
Add lifecyle rule to delete old data on dev and test buckets

erikamov:
# Description

This PR adds lifecyle rule to delete data older than 30 days on dev and test buckets (using Terraform).

I am deleting these three obsolete buckets instead of adding the rule since they are old and not in use anymore: `cold-storage-outputs-gtfs-data-test-charlie-test`, `cold-storage-outputs-gtfs-data-test` and `calitp-jamesl-gcp-components-tfstate`.

Resolves [#3666]

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

Tested running `Terraform plan` locally

## Post-merge follow-ups

- [ ] No action required

Check if data older than 30 days were deleted from dev and test buckets.


ohrite (approved):
🍐 pairing w/ @erikamov

PR #3826, opened 2025-04-11
Update external_airtable_transit_tech_stacks_services.yml

fsalemi:
# Description

Adding `contracts` field to external_airtable_transit_tech_stacks_services. `Contracts` filed is needed in order to create a contract to services bridge table.

## Type of change

- [X] New feature
- [ ] Documentation

## Post-merge follow-ups

- [] No action required

Bridge table will be defined after this step.


vevetron:
If you are still developing these airtables and this schema, consider just not including a schema at all.  Just have the yaml file with none of the specific fields specified.  If you don't include a schema on this yaml, everything will go through and I think usually the formats come out correct.

Whenever you feel like the schema is final, then include the schema into this yaml file.

vevetron (approved):

PR #3842, opened 2025-04-16
Update external_airtable_transit_tech_stacks_services.yml

fsalemi:
# Description

The `Services` table in the **Transit Technology Stacks** base (Airtable) is a synced table from the `Services` table in the **California Transit** base. During the syncing process, all lookup fields are transformed into plain text fields. This leads to multiple type mismatch errors when creating the external table via the related DAG.
For example:
```
Error while reading table: cal-itp-data-infra.external_airtable.transit_technology_stacks__services, error message: JSON parsing error in row starting at position 51112: Could not convert value 'string_value: "279, 4"' to integer. Field: itp_id__from_provider_; Value: 279, 4 File: gs://calitp-airtable/transit_technology_stacks__services/dt=2022-08-16/ts=2022-08-16T14:33:57.365242Z/services.jsonl.gz
```
To address this, I'm explicitly adding the `schema_fields` and including only the required fields to prevent these parsing errors.

## Type of change

- [X] New feature
- [ ] Documentation


fsalemi:
@vevetron it is ready now.

evansiroky (approved):

Issue #3848, opened 2025-04-18
NTD: resolve competing mart tables and schemas for NTD data


charlie-costanzo:
Currently, we unintentionally have two sets of mart tables focusing on NTD ridership: 
* One is currently used for analysis (these are found in `warehouse/models/mart/ntd`, although this schema name is not ideal)
  * A lot of work has been done to make these mature and useful, we want to preserve this logic and structure
* One was created alongside the other NTD data domains while we were ingesting all of the NTD endpoints in bulk (these are found in `warehouse/models/mart/ntd_ridership`)
  * These tables pretty much just expose the staging tables as mart (very very simple, almost 1:1 with staging tables, probably shouldn't be used anywhere and don't need to exist, but have the preferable schema name of `mart/ntd_ridership`)
 
We would like to keep the tables that are found in the first location (`mart/ntd`), but locate them in the schema of the second location (`mart/ntd_ridership` –  a much preferred schema name)

To do this, we will:
* Copy the `mart/ntd` tables to `mart/ntd_ridership`
* Make the old tables found in `mart/ntd` just views of the new ones in `mart/ntd_ridership`
* Add a deprecation message on the old tables
* Then share in analyst channels in slack and give analysts time to replace the old schema name with the new schema for the tables they have been using, probably ping Christian about this directly as well

We will also need to
* Enrich the tables in `mart/ntd` with `caltrans_district` (this has so far only been done in the `mart/ntd_ridership` tables)
* Go back and remove table materializations of ntd_ridership staging models that was done in #3847 (this was only done temporarily while both mart schemas were referencing the same staging tables)
  * stg_ntd__complete_monthly_ridership_with_adjustments_and_estimates__upt
  * stg_ntd__complete_monthly_ridership_with_adjustments_and_estimates__voms
  * stg_ntd__complete_monthly_ridership_with_adjustments_and_estimates__vrh
  * stg_ntd__complete_monthly_ridership_with_adjustments_and_estimates__vrm

PR #3870, opened 2025-04-27
NTD: iterative mart cleanup, consistency, enrichment

charlie-costanzo:
# Description
This PR does a lot of end-stage cleanup, consistency, and  and enrichment of the NTD endpoints that we have ingested, detailed below. 

* Applies appropriate schemas to new mart tables in `dbt_project.yml`
* Creates unioned intermediate and marts for the few tables that are ingested as single years
  * In the case of `ntd_annual_reporting/int_ntd__unioned_agency_information`, this actually creates a new mart table using code created by analysts in the appropriate schema, while still pointing their table (`ntd/dim_annual_agency_information`) at the new one until we complete #3848 
* Creates intermediate pivot tables for the few reamining that were yet to be pivoted in marts
* Provides some consistency in column naming, intuitive ordering
* Enriches a few tables with some of the missing grouping columns from agency information
* Updates appropriate `yml` files

Small follow-up PR to-do, and then these tables will be in a very solid place:
* primary/synthetic keying
* consistency in state/agency name inherited from agency dim table
* Agency dim table handling for (dimensional) caltrans district

## Type of change
- [x] New feature

## How has this been tested?
`poetry run dbt run -s +mart.ntd_annual_reporting +mart.ntd_assets +mart.ntd_funding_and_expenses +mart.ntd_ridership +mart.ntd_safety_and_security +mart.ntd`

Screenshot 2025-05-05 at 18 22 51

## Post-merge follow-ups
- [x] No action required


vevetron:
Hmm.  Why is the sync metabase failling.

erikamov:
> Hmm. Why is the sync metabase failling.

It is failing because Metabase tries to find the new models on the database.
When the pull request is created it uses `cal-itp-data-infra-staging` to test and check and it expects to find under `staging` schema (ex: `staging_staging.int_ntd__unioned_agency_information` instead of `charlie_staging.int_ntd__unioned_agency_information`).

I am working on it to run those models so it won't fail again for this reason.

erikamov (approved):
Great work!!

Issue #3784, opened 2025-03-18
Analyst sees that Metabase reflects the latest schema changes


ohrite:
## User story / feature request

As an Analyst,
When dbt models or dependencies change,
And when I view Metabase,
I can see that Metabase reflects those changes

### Acceptance Criteria

This will be visible in Metabase.

PR #3905, opened 2025-05-07
Fix sync Metabase and Visualize on Deploy dbt workflow

erikamov:
# Description

This PR fixes errors when generating and visualizing the model change report for #3784.

When `Deploy dbt` workflow runs, the job `Pull Request visualization` is failing with the message `Got a state selector method, but no comparison manifest`.
To visualize a model change it needs to compare the artifacts (ex. manifest.json) with the latest artifacts to compare the difference.
The artifacts in latest should be based on the models in the Main branch that is uploaded when `transform_warehouse` DAG runs. So I removed uploading artifacts to latest from `Deploy dbt` workflow.

I am also preparing to add a new step on `Sync Metabase` (that is part of Deploy dbt workflow).
In order to `Sync Metabase` to pass it needs the new or changed dbt models to be created on `cal-itp-data-infra-staging` using `staging` schema. Since locally we use our names as schema name, I am adding a `dbt run` for `staging` as part of the workflow to avoid another step to developers.

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Tested through this PR fixing errors that were happening on another PR https://github.com/cal-itp/data-infra/pull/3870.

## Post-merge follow-ups

- [ ] No action required

After merging this PR we will rebase PR https://github.com/cal-itp/data-infra/pull/3870 in order to fix  `Pull Request visualization` and enable the `dbt run` part on the sync Metabase.

### DIFF
--- a/.github/workflows/deploy-dbt.yml
+++ b/.github/workflows/deploy-dbt.yml
@@ -205,6 +205,17 @@ jobs:
         working-directory: warehouse
         run: poetry run dbt debug --target ${{ env.DBT_TARGET }}
 
+      # TODO: To be enabled when permission Terraform applies bigquery.tables.create
+      # - name: Download latest artifacts from GCS
+      #   working-directory: warehouse
+      #   if: ${{ env.DBT_TARGET  == 'staging' }}
+      #   run: gsutil cp -r gs://${{ env.DBT_ARTIFACTS_BUCKET }}/latest/ ./target/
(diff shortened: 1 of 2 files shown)


charlie-costanzo (approved):
Looks great, thank you so much for taking care of this!

vevetron (approved):
LGTM

PR #3928, opened 2025-05-22
Populate Metrolink `shape_id` to `fct_scheduled_trips`

### LINKED ISSUE #2194: Bug: Metrolink shape_array_key is None
**Describe the bug**
`Metrolink Schedule`'s `shape_array_key` values are all None because all their `shape_id` values are missing. In the `fct_daily_scheduled_trips` table, a fix can be done because there's `route_id` and `direction_id` also present. 

Should Metrolink have a different way of getting its `shape_array_key`? In v1, it seemed like we were still able to join `trips` and `shapes` for Metrolink because it was only missing `shape_id` values in `trips` but not `shapes`.

**To Reproduce**
1. [notebook for metrolink](https://github.com/cal-itp/data-analyses/blob/metrolink-shape-array/high_quality_transit_areas/metrolink_shape_array.ipynb)

**Expected behavior**

**Additional context**

tiffanychu90:
# Description

Within `mart_gtfs.fct_scheduled_trips`, use `route_id` and `direction_id` to fill in Metrolink's rows to get `shape_id`. 
Resolves #2194 - we've known about this for awhile, and a manual fix seems to be the only way for this to work in downstream analysis.
* The current bug is that `shape_id` is NaN in the trips table, but it's populated in `dim_shapes`. This merge won't link any of Metrolink's shapes in `fct_daily_scheduled_shapes` because of the trips that are counted filter out for null shapes [here](https://github.com/cal-itp/data-infra/blob/6821ffd9cbb07ca9527400278c41964e4b44a74d/warehouse/models/mart/gtfs/fct_daily_scheduled_shapes.sql#L39)

```
trips_counted AS (

    SELECT
        *
    FROM fct_scheduled_trips
    WHERE shape_id IS NOT NULL
    GROUP BY 1, 2, 3, 4, 5
)
```
* **TODO**: A query to check whether there were other missing shape_ids did show some in 2024, for Amtrak, a smattering for San Diego Schedule in May 2024 (but seems like they fixed it after that), and some other operators (~150ish rows for `year-gtfs_dataset_name-route_id-direction_id`. Potentially follow-up with these and add it to the seed table.

**Solution**:
* Added a seed table to populate these for Metrolink. Provides ability for us to add more in the future too, without us writing a super long CASE WHEN statement.
* A full refresh would fill in those missing shape_ids, prevent those from getting dropped, and the query in  `fct_daily_scheduled_shapes` should bring those shapes / shape_geometry in. 

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

```
jovyan@jupyter-tiffanychu90 ~/data-infra/warehouse (add-metrolink) $ poetry run dbt run -s fct_scheduled_trips --full-refresh
20:41:38  Running with dbt=1.5.1
20:41:41  Found 610 models, 1078 tests, 0 snapshots, 0 analyses, 853 macros, 0 operations, 13 seed files, 223 sources, 4 exposures, 0 metrics, 0 groups
20:41:41  
20:41:57  Concurrency: 8 threads (target='dev')
20:41:57  
20:41:57  1 of 1 START sql table model tiffany_mart_gtfs.fct_scheduled_trips ............. [RUN]
20:42:18  1 of 1 OK created sql table model tiffany_mart_gtfs.fct_scheduled_trips ........ [CREATE TABLE (135.2m rows, 57.9 GiB processed) in 21.20s]
20:42:18  
20:42:18  Finished running 1 table model in 0 hours 0 minutes and 36.51 seconds (36.51s).
20:42:18  
20:42:18  Completed successfully
20:42:18  
20:42:18  Done. PASS=1 WARN=0 ERROR=0 SKIP=0 TOTAL=1
```

Running `poetry run dbt run -s fct_scheduled_trips+` works, but `+fct_scheduled_trips+` will not, since it hits a query limit. Adjusted up from `2000000000000` and still got `500 Query exceeded limit for bytes billed: 3000000000000. 3003980447744 or higher required.`

## Post-merge follow-ups

- [ ] No action required
- [x] Actions required (specified below) - check this after the weekend's full refresh

### DIFF
--- a/warehouse/models/mart/gtfs/fct_scheduled_trips.sql
+++ b/warehouse/models/mart/gtfs/fct_scheduled_trips.sql
@@ -34,6 +34,12 @@ stop_times_grouped AS (
     SELECT * FROM {{ ref('int_gtfs_schedule__stop_times_grouped') }}
 ),
 
+-- use seed to fill in where shape_ids are missing
+derived_shapes AS (
+    SELECT *
+    FROM {{ ref('gtfs_optional_shapes') }}
+),
+
 gtfs_joins AS (
     SELECT
         {{ dbt_utils.generate_surrogate_key(['service_index.service_date', 'trips.key', 'stop_times_grouped.iteration_num']) }} AS key,
@@ -163,7 +169,7 @@ fct_scheduled_trips AS (
         gtfs_joins.agency_id,
         gtfs_joins.network_id,
         gtfs_joins.shape_array_key,
-        gtfs_joins.shape_id,
+        COALESCE(gtfs_joins.shape_id, derived_shapes.shape_id) AS shape_id,
         gtfs_joins.contains_warning_duplicate_trip_primary_key,
         gtfs_joins.num_distinct_stops_served,
         gtfs_joins.num_stop_times,
@@ -206,6 +212,10 @@ fct_scheduled_trips AS (
         AND gtfs_joins.service_date = daily_feeds.date
     LEFT JOIN dim_gtfs_datasets
         ON daily_feeds.gtfs_dataset_key = dim_gtfs_datasets.key
+    LEFT JOIN derived_shapes
+    ON dim_gtfs_datasets.name = derived_shapes.gtfs_dataset_name
+        AND gtfs_joins.route_id = derived_shapes.route_id
+        AND gtfs_joins.direction_id = derived_shapes.direction_id
     -- drop trips that no one can actually ride
     WHERE has_rider_service
 )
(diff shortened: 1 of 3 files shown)


vevetron (approved):
LGTM!

vevetron:
I don't know how to do this - should we have a test to make sure this is working right?

tiffanychu90:
> I don't know how to do this - should we have a test to make sure this is working right?
> 
> Maybe it isn't necessary if Metrolink fixes this in the next few months.

I materialized the table to do take this table and use it, and the `shape_ids` were there, and `fct_daily_scheduled_shapes` was built successfully, so I *think* it'll work?

PR #4004, opened 2025-06-24
[packages/calitp-data-analysis] Auth helpers for GCS geospatial data access

### LINKED ISSUE #3935: Feature: Add decorator to simplify sso-related geopandas changes

As an analyst, I want to be able to read geoparquet and [GDAL/OGR](https://geopandas.org/en/stable/docs/user_guide/io.html) supported spatial files from the warehouse. With SSO, this requires adding additional code, so I would like to have decorators in the `calitp_data_analysis` package I can add to `gpd.read_parquet` and `gpd.read_file` to make it more convenient.

### Acceptance Criteria

The following two blocks are equivalent:
```
from calitp_data_analysis import *
import geopandas as gpd

@placeholder_decorator_name
gpd.read_parquet("gs://test/file.parquet")
```

```
from calitp_data_analysis import *
import geopandas as gpd
import google.auth

credentials, _ = google.auth.default()
gpd.read_parquet("gs://test/file.parquet". storage_options={"token": credentials.token})
```

As are the following:
```
from calitp_data_analysis import *
import geopandas as gpd

@placeholder_decorator_name
gpd.read_file("gs://test/file.geojson")
```

```
from calitp_data_analysis import *
import geopandas as gpd

fs = get_fs()
with fs.open("gs://test/file.geojson") as f:
    gpd.read_file(f)
```

### Notes

To avoid slowing load times, this should ideally cache the relevant credentials so they don't need to be repeatedly fetched. Ideally, one decorator would be possible, but investigation is required to determine that. The decorator(s) should also work on `to_parquet`.

raebot:
# Description

Auth helpers for GCS geospatial data access

Resolves #3935 

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

See unit tests added in this PR's diff.

## Post-merge follow-ups

- [ ] No action required

### follow-ups
* Update https://github.com/cal-itp/data-infra/blob/main/images/jupyter-singleuser/pyproject.toml to use the new package version

### DIFF
--- a/packages/calitp-data-analysis/pyproject.toml
+++ b/packages/calitp-data-analysis/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.poetry]
 name = "calitp-data-analysis"
 requires-poetry = "~=2.0"
-version = "2024.12.6.post1"
+version = "2025.6.24"
 description = "Shared code for querying Cal-ITP data in notebooks, primarily."
 package-mode = true
 authors = ["Andrew Vaccaro"]


packages/calitp-data-analysis/calitp_data_analysis/gcs_geopandas.py
```diff
@@ -0,0 +1,40 @@
+import gcsfs  # type: ignore
+import geopandas  # type: ignore
+import google.auth  # type: ignore
```
raebot, line 3: @vevetron How have you all typically been handling mypy errors like [`Skipping analyzing *: module is installed, but missing library stubs or py.typed marker  [import-untyped]`](https://github.com/cal-itp/data-infra/actions/runs/15861357983/job/44719054487#step:6:199)?  There are type/stub packages somewhat related to these, but none looked exactly right. The [types-geopandas](https://pypi.org/project/types-geopandas/1.0.1.20250510/) package is only compatible with geopandas version 1.0.1 and above and we use version 0.14. The stub packages for the google packages are not actually maintained by google. Would you prefer I try to use some of those instead of ignoring here?

.pre-commit-config.yaml
```diff
@@ -21,7 +21,7 @@ repos:
         # Suppress SyntaxWarning about invalid escape sequence from calitp-data-infra dependency without modifying source
         entry: env PYTHONWARNINGS="ignore::SyntaxWarning" flake8
   - repo: https://github.com/psf/black
-    rev: 23.1.0
+    rev: 24.10.0
```
raebot, line 24: I made this match the Black version specified in toml files. Let me know if that's ok or not.

vevetron: I prefer we should leave this at 23.1.0 so the tests will pass.  I made a separate issue, we can update it there and then update the codebase. 

Alternatively we can fix all the linting issues here.

packages/calitp-data-analysis/calitp_data_analysis/gcs_geopandas.py
vevetron: I'm okay with ignoring them for now.  In this case.

We really need to update geopandas!

packages/calitp-data-analysis/calitp_data_analysis/gcs_geopandas.py
```diff
@@ -0,0 +1,40 @@
...
+    def gcs_filesystem(self, **kwargs):
+        """Returns a Google Cloud Storage Filesystem"""
+        return gcsfs.GCSFileSystem(token=self.token, **kwargs)
+
+    def read_parquet(self, path, *args, **kwargs):
+        """Delegates to geopandas.read_parquet with storage option token
+
+        Passes the auth credentials from Google auth as storage option token
+        """
+        storage_options = kwargs.get("storage_options", {}) | {"token": self.token}
```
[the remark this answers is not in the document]
raebot: You're right! I meant to use `pop` here instead. Just made that change.

packages/calitp-data-analysis/calitp_data_analysis/gcs_geopandas.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+
+    def __init__(self):
+        """Sets instance token to credentials returned from Google auth request"""
+        credentials, _ = google.auth.default()
+        self.token = credentials
+
+    def gcs_filesystem(self, **kwargs):
+        """Returns a Google Cloud Storage Filesystem"""
+        return gcsfs.GCSFileSystem(token=self.token, **kwargs)
```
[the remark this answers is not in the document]
mrtopsyt: That makes sense, although if we're providing a bespoke function for `read_parquet` I think it makes sense to also have one for `read_file`, since the solution you described also works like this, we just hadn't been using it:
```
fs = get_fs()
with fs.open("gs://test/file.geojson") as f:
    gpd.read_parquet(f)
```

I don't have a super strong opinion though, happy to defer to you and @vevetron. I'd also have to defer to @vevetron for whether specifying `"google_default"` or `"cloud"` is necessary, since I'm not very familiar with gcsfs.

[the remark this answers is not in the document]
vevetron: I'm not sure but I think google_default is the way to go.

vevetron (approved):

mrtopsyt (approved):

Issue #3780, opened 2025-03-18
Analyst sees payments update daily


ohrite:
## User story / feature request

As an Analyst,
When I view the latest Payments data,
I see it update daily at 1400h (UTC)

### Acceptance Criteria

This will be visible in the Composer dashboard.

### Blockers

Currently blocked on #3766: Upgrade to composer-2.8.6-airflow-2.7.3

PR #4018, opened 2025-06-27
Fix target name on Cosmos payment DAG

erikamov:
# Description

This PR updates Cosmos payment DAG to use `DBT_TARGET` (prod or staging) as the target name to run correcly on Production.

Related to #3780

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Tested running `cosmos_payment` DAG on Staging.

## Post-merge follow-ups

- [ ] No action required

Run Cosmos payment DAG after the change to check if it uses the correct target.

### DIFF
--- a/airflow/dags/cosmos_payment_dag.py
+++ b/airflow/dags/cosmos_payment_dag.py
@@ -1,7 +1,10 @@
+import os
 from datetime import datetime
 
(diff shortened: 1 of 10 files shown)


ohrite (approved):
Agree, let’s use this variable!

Issue #4028, opened 2025-07-02
Update dbt-docs site references to the new url `https://dbt-docs.dds.dot.ca.gov`


erikamov:
## User story / feature request

Update dbt-docs site references to point to the new url `https://dbt-docs.dds.dot.ca.gov`.

### Acceptance Criteria

All links and references for dbt-docs site points to the new url `https://dbt-docs.dds.dot.ca.gov`.

### Notes

For staging if needed use `https://dbt-docs-staging.dds.dot.ca.gov`.

PR #4041, opened 2025-07-03
Remove single quotes from dbt docs content-type files

erikamov:
# Description

The [new dbt docs site](https://dbt-docs.dds.dot.ca.gov) is triggering a file download instead of displaying the index.htm page. This PR removes single quotes around the content-type so the site works correctly.

[#4028]

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Removing quotes manually from content-type on files in the dbt-docs bucket made the site works.

## Post-merge follow-ups

- [ ] No action required

Check if all files in the bucket have the correct content-type without quotes.


ohrite (approved):
🍐 w/ @erikamov

PR #4044, opened 2025-07-03
Fix conversion error on `external_airtable.california_transit__ntd_agency_info`

erikamov:
# Description

This PR adds schema definition for `external_airtable.california_transit__ntd_agency_info` to specify column types.
Previous json files generated contain INTEGER fields with a decimal zero `.0`, so when the AUTO definition tries to create numeric columns as INTEGER it generates the conversion error `Could not convert value 'number_value: 	 "96120.0"' to integer`.

Also converted `original_due_date` and `fy_end_date` to STRING due to conversion error `400 Invalid date: '10/31/18' Field: original_due_date; Value: 10/31/18`

Resolves [#4021]

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Running external table DAG locally.

## Post-merge follow-ups

- [ ] No action required

Monitor DAGs to confirm the error were fixed.

### DIFF
--- a/.github/workflows/deploy-dbt.yml
+++ b/.github/workflows/deploy-dbt.yml
@@ -491,12 +491,13 @@ jobs:
         run: poetry run python scripts/visualize.py ci-report --latest-dir='./target/latest/'
 
       - name: Archive CI report
+        working-directory: warehouse/target
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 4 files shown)


ohrite (approved):
So good!

PR #4085, opened 2025-07-12
fix(ci): install hologit through NPM instead of Habitat/Biome

themightychris:
# Description

Habitat, originally by Chef, was acquired some years ago and we were using the community fork Biome but it still used Habitat's public package repository and the new owners recently shut down public use of the package repository

This has recently caused the Kubernetes CI jobs to start failing as they used Biome to install Hologit for generating a working tree of all Helm chart sources

This PR should fix this by installing Hologit through NPM instead of Biome—which only offered the advantage of guaranteeing the `git` version but current GitHub runners come with a new enough git version stock

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Opening this PR will be the test, if the kicked-off `preview-kubernetes` action succeeds we're goo to go

## Post-merge follow-ups

- [ ] No action required
  - [ ] Ensure the `deploy-kubernetes` action succeeds too

### DIFF
--- a/.github/workflows/deploy-kubernetes.yml
+++ b/.github/workflows/deploy-kubernetes.yml
@@ -39,14 +39,9 @@ jobs:
           location: ${{ env.GKE_REGION }}
       - run: curl -sSL https://install.python-poetry.org | python -
       - name: Set up hologit
-        env:
-          BIO_RELEASE: 1.6.821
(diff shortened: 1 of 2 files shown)


vevetron (approved):
Thank you!

PR #4187, opened 2025-08-12
dbt: create new env variable override for local development on prod external tables

charlie-costanzo:
# Description
During our infrastructure refactor, we determined the need to preserve local warehouse development on production external tables as sources of truth **WITH THE EXCEPTION OF RT EXTERNAL TABLES**. This PR introduces environment variable `DEV_SOURCE_GOOGLE_CLOUD_PROJECT` which, when added locally, allows for dbt to reference production external tables at runtime **WITH THE EXCEPTION OF RT EXTERNAL TABLES**.

Resolves: #4188 

## Type of change
- [x] New feature

## How has this been tested?
Screenshot 2025-08-12 at 14 58 10

## Post-merge follow-ups
- [x] Action required
To use the production external tables in local warehouse development, the following line needs to be added to your environment: 

`export DEV_SOURCE_GOOGLE_CLOUD_PROJECT='cal-itp-data-infra'`

### DIFF
--- a/warehouse/models/staging/gtfs/_src_gtfs_rt_external_tables.yml
+++ b/warehouse/models/staging/gtfs/_src_gtfs_rt_external_tables.yml
@@ -3,7 +3,7 @@ version: 2
 sources:
   - name: external_gtfs_rt
     description: Hive-partitioned external tables reading GTFS RT data and validation errors from GCS.
-    database: "{{ env_var('GOOGLE_CLOUD_PROJECT', var('GOOGLE_CLOUD_PROJECT')) }}"
+    database: "{{ env_var('GOOGLE_CLOUD_PROJECT', var('EXTERNAL_TABLE_SOURCE')) }}"
     schema: external_gtfs_rt_v2
     tables:
       - name: service_alerts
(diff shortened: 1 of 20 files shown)


lauriemerrell:
@erikamov sorry maybe something changed or I linked the wrong line, I meant to link: https://github.com/cal-itp/data-infra/blob/main/warehouse/dbt_project.yml#L31 -- basically, I think we need to document how this new variable interacts with that existing variable that already points at a project. It looks like the comments in `dbt_project.yml` have been updated appropriately, not sure if that was you or Charlie.

warehouse/models/staging/gtfs/_src_gtfs_rt_external_tables.yml
```diff
@@ -3,7 +3,7 @@ version: 2
 sources:
   - name: external_gtfs_rt
     description: Hive-partitioned external tables reading GTFS RT data and validation errors from GCS.
-    database: "{{ env_var('GOOGLE_CLOUD_PROJECT', var('GOOGLE_CLOUD_PROJECT')) }}"
+    database: "{{ env_var('GOOGLE_CLOUD_PROJECT', var('EXTERNAL_TABLE_SOURCE')) }}"
```
lauriemerrell, line 6: PR description says that GTFS RT data is meant to be excluded from this change, won't this update mean that it applies to GTFS-RT too? Might remove this change so that GTFS-RT is excluded, or update PR description to be clear that it affects all sources

erikamov:
@charlie-costanzo  I am closing this PR since no changes are needed in order to use external_tables in staging.

PR #4166, opened 2025-08-06
Feat/benefits extra claims

thekaveman:
# Description

Extracts the existing `event_properties.extra_claims` JSON column from raw Benefits events data for inclusion in the final mart table. This will allow users to e.g. filter and group on this field in Metabase.

Resolves #4059 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

## How has this been tested?

_If making changes to dbt models, please run the command `poetry run dbt run -s CHANGED_MODEL` and `poetry run dbt test -s CHANGED_MODEL`, then include the output in this section of the PR._

```console
$ poetry run dbt run -s fct_benefits_events
01:00:24  Running with dbt=1.10.1
01:00:28  Registered adapter: bigquery=1.10.0
01:00:29  Found 610 models, 1236 data tests, 14 seeds, 223 sources, 4 exposures, 995 macros
01:00:29  
01:00:29  Concurrency: 8 threads (target='dev')
01:00:29  
01:00:33  1 of 1 START sql table model kegan_mart_benefits.fct_benefits_events ........... [RUN]
01:00:36  1 of 1 OK created sql table model kegan_mart_benefits.fct_benefits_events ...... [CREATE TABLE (0.0 rows, 0 processed) in 3.14s]
01:00:36  
01:00:36  Finished running 1 table model in 0 hours 0 minutes and 6.24 seconds (6.24s).
01:00:36  
01:00:36  Completed successfully
01:00:36  
01:00:36  Done. PASS=1 WARN=0 ERROR=0 SKIP=0 NO-OP=0 TOTAL=1
```

```console
$ poetry run dbt test -s fct_benefits_events
01:00:57  Running with dbt=1.10.1
01:01:01  Registered adapter: bigquery=1.10.0
01:01:02  Found 610 models, 1236 data tests, 14 seeds, 223 sources, 4 exposures, 995 macros
01:01:02  Nothing to do. Try checking your model configs and model specification args
```

## Post-merge follow-ups

_Document any actions that must be taken post-merge to deploy or otherwise implement the changes in this PR (for example, running a full refresh of some incremental model in dbt). If these actions will take more than a few hours after the merge or if they will be completed by someone other than the PR author, please create a dedicated follow-up issue and link it here to track resolution._

- [x] No action required
- [ ] Actions required (specified below)

Issue #4059, opened 2025-07-08
Add the `extra_claims` column to Benefits event data in Metabase.


indexing:
## Feature request

Add the `extra_claims` column to Benefits event data in Metabase. This gets added to [Amplitude events](https://github.com/cal-itp/benefits/blob/main/benefits/enrollment/analytics.py#L22) returned enrollment on the `event_properties`, but is not defined on the [warehouse mart model](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/benefits/_mart_benefits.yml) or processed in its [pipeline script](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/benefits/fct_benefits_events.sql).

### Acceptance Criteria

- The `extra_claims` parameter is available when structuring question queries in Metabase. 
- We can build a question in Metabase that shows the percentage of Medicare cardholders enrollments in Cal-ITP Benefits categorized as `disabled`. 

### Notes

We are tracking several metrics in Amplitude that have been well-received by stakeholders at Cal-ITP. Thus, we would like to add them to the [Cal-ITP Benefits dashboard](https://dashboards.calitp.org/dashboard/42-cal-itp-benefits) in Metabase so stakeholders can track them regularly.

Here is a screenshot of the chart we need to recreate with the `extra_claims` attribute highlighted. All other attributes in this query besides `extra_claims` are currently available in Metabase.

<img width="1559" height="844" alt="Image" src="https://github.com/user-attachments/assets/2497e861-98a3-425d-9846-f6e4952366ac" />

PR #4191, opened 2025-08-13
Fix: Benefits extra_claims is a JSON array

thekaveman:
# Description

This column is a JSON array inside a JSON object, so we need to use the `json_extract_flattened_column` macro to parse it correctly.

Follow up to #4166 / #4059

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

@erikamov tested with the correct function call that this macro expands to: https://cal-itp.slack.com/archives/C050ZNDUL21/p1755101944951569?thread_ts=1754936604.833429&cid=C050ZNDUL21

## Post-merge follow-ups

- [x] No action required


erikamov:
I think compiled failed because the column has no name due to `no_alias = true`. Try to remove this param, the line should be like this: `{{ json_extract_flattened_column('event_properties', 'extra_claims') }},`

erikamov (approved):
Looks great!

Issue #4048, opened 2025-07-08
Payments: configure SFTP data ingest of Enghouse data feed to GCS bucket


charlie-costanzo:

jparr:
One question I have is how do we want to maintain a stable endpoint for the enghouse users? Currently the ingress ip is dynamic so can change when the load balancer is reset. Do we want to set up a mapping or fix the ip somehow?

PR #4355, opened 2025-09-30
Mount sftp data directory with ownership, Assign static ip

jparr:
ref: #4048

Mount the sftp data directory such that it is owned by the sftp-user. This will allow `setstat` from sftp clients.
Assign a static ip to the deployment service so that it doesnt change when redeployed.

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Documentation


ohrite (approved):
🍐 w/ @jparr

Issue #4426, opened 2025-10-24
Consider upating GTFS-RT test environment references to `AIRFLOW_ENV` env var to use new values


lauriemerrell:
## User story / feature request

As a developer, I just ran into an issue testing the GTFS schedule validator because this line: https://github.com/cal-itp/data-infra/blob/5de1d56859f74f6f36a5ae1bf3bbf31f58d7d827/airflow/dags/dags.py#L59 checked for `AIRFLOW_ENV = development` but [`AIRFLOW_ENV=cal-itp-data-infra-staging` now](https://github.com/cal-itp/data-infra/blob/main/airflow/.development.env#L7) in non-prod environments.

It seems like `AIRFLOW_ENV` is checked in a few other places in the GTFS-RT archiver: https://github.com/search?q=repo%3Acal-itp%2Fdata-infra%20AIRFLOW_ENV&type=code. Not sure if there is any functional impact to the current discrepancy, but it might be worth checking it out.

### Acceptance Criteria

Investigation of functional impact that GTFS RT  test archiver instance seems to be checking for an env var value that it won't find (`development`)

### Notes

Again, not sure of the functional impact here. Didn't look deeply into the archiver part

lauriemerrell:
@erikamov thanks sorry to clarify, I am fixing the schedule validator part in my PR #4350, my question is about the big picture -- consolidating with the other variable might make sense, but I can take a look at your PR! Thank you for jumping on this

erikamov:
I need to review with Laurie about the change needed here again since there were some work to upgrade the Validator 7.1 that was reverted.

PR #4439, opened 2025-10-29
Use `AIRFLOW_ENV=development` on `.development.env` file

erikamov:
# Description

Based on researching the past commits, the only place that may need to be changed is on `.development.env` file. Since there was not a staging environment when it was created, there are tests only for two options:`development` or `cal-itp-data-infra`.
 
This PR sets `AIRFLOW_ENV=development` on `.development.env file`.

Resolves [#4426]

## Type of change

- [ ] New feature
- [ ] Documentation

## Post-merge follow-ups

- [x] No action required


lauriemerrell (commented):
One comment, otherwise LGTM. I had made the fix in dags.py but I can just rebase on this

kubernetes/apps/overlays/gtfs-rt-archiver-v3-test/archiver-channel-vars.yaml
lauriemerrell: As noted in the ticket, I am just not sure what this env var does in this context, not sure if there's been a functional issue on the test archiver?

lauriemerrell (commented):
Ope, missed a save -- here's the comment

kubernetes/apps/overlays/gtfs-rt-archiver-v3-test/archiver-channel-vars.yaml
erikamov: @lauriemerrell  I reviewed my previous changes, and I think they were wrong.
The only place that looks like we would need to change is on this .development.env file.
Can you test your test archiver with this change?

PR #4454, opened 2025-11-03
NTD: deprecate legacy XLSX ridership tables in favor of API tables

### LINKED ISSUE #4434: Analysts no longer see deprecated NTD ridership tables
Now that we're utilizing the NTD API for ridership data, we can remove the tables that relied on the XLSX file ingest

charlie-costanzo:
# Description
Since we've merged in #4272 and ensured parity with the legacy XLSX tables, this PR deprecates the legacy tables in favor of those living in our API-derived marts.

Resolves #4434 

## Type of change
Analysts / @csuyat-dot have been notified of this breaking change

## How has this been tested?
Parity between the tables has been monitored over the past several months

## Post-merge follow-ups
Manually delete the tables in the warehouse


charlie-costanzo:
> I added a comment about deleting this model on PR #4455. But I think still need to remove the related buckets and env variables, and Post Merge should remove those models from Bigquery (Staging and Production).

Hey @erikamov – totally. The table deletion is captured in the post-merge follow-ups, I'll make sure that gets done. As for the containing buckets for the XLSX ridership data, they're actually still being used for some other files, but I can remove the ridership data from both the XLSX raw and clean buckets. The only env variables are for the XLSX scrape operations as a whole, which are still being used for other tables, so there shouldn't be a need to do anything there.

PR #4455, opened 2025-11-03
NTD: deprecate legacy XLSX ridership dag tasks in favor of API dag tasks

### LINKED ISSUE #4457: Developer no longer sees deprecated NTD XLSX ridership dag tasks
Following the migration from using the XLSX scrape for NTD ridership data, we now need to deprecate the dag tasks responsible for ingesting the XLSX data.

charlie-costanzo:
# Description
Since we've merged in #4272 and ensured parity with the legacy XLSX tables, this PR deprecates the legacy XLSX dag tasks in favor of those utilizing the API.

Resolves #4457 

## Type of change
Analysts / @csuyat-dot have been notified of this breaking change

## How has this been tested?
Parity between the tables has been monitored over the past several months

## Post-merge follow-ups
- [x] No actions required


erikamov:
How about removing the downstream models?

If those models are not need anymore, we also need to remove related buckets and env variables (managed by Terraform for `cal-itp-data-infra` and `cal-itp-data-infra-staging`), and tables on BigQuery (Production and Staging).

PR #4458, opened 2025-11-03
Payments: deprecate Littlepay feed v1 dag tasks now that all agencies have migrated to v3

### LINKED ISSUE #4456: Developer no longer sees Littlepay feed v1 dag tasks
Following the migration of all agencies from Littlepay feed v1 to feed v3, we need to deprecate the no-longer-relevant feed v1 dag tasks

charlie-costanzo:
# Description
Following the migration of all agencies to LIttlepay feed v3, we are now removing the previously-used feed v1 dag tasks.

Resolves #4456 

## Type of change

## How has this been tested?
Data parity has been ensured over the past several months

## Post-merge follow-ups
- [x] No action required


ohrite:
Also need to remove the operators for littlepay v1

PR #4508, opened 2025-11-13
Update helm charts to use bitnamilegacy while keeping the version the same

### LINKED ISSUE #4371: Update helm values to capture image overrides applied to cluster
To get Sentry back up last night, I went through the deployments/statefulsets for all the Sentry dependencies that were using `bitnami` images and swapped them to `bitnamilegacy`. With `zookeeper-clickhouse` I additionally needed to change the tag to `:3.8.0-debian-11-r9` as the `:3.8.0-debian-10-*` tag wasn't available under `bitnamilegacy`

We need to update the Sentry helm values to override the container image repo/tag values to match these so they don't get reverted next time we deploy changes to the cluster. This is a quick fix vs something more drastic like upgrading to the latest Sentry helm chart that we should do regardless of future plans for Sentry as it should only take a few minutes and without it we're liable to accidentally bring Sentry back down if we need to deploy any changes to anything else in the cluster

vevetron:
# Description

Based on https://github.com/sentry-kubernetes/charts/blob/develop/charts/sentry/CHANGELOG.md, the bitnami legacy issue is resolved in more recent sentry versions.

But Chris suggests we just override the tag values.

A recent deployment accidentally took down sentry again.

Resolves #4371 at least for sentry

## Type of change

## How has this been tested?

Not tested
## Post-merge follow-ups

- [ ] No action required
Watch the cluster

### DIFF
--- a/kubernetes/apps/charts/postgresql/.helmignore
+++ b/kubernetes/apps/charts/postgresql/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
--- a/kubernetes/apps/charts/postgresql/Chart.yaml
+++ b/kubernetes/apps/charts/postgresql/Chart.yaml
@@ -1,0 +1,10 @@
+apiVersion: v2
+name: postgresql
+description: Postgres for Kubernetes
+type: application
+version: 0.1.0
+appVersion: "1.4.2"
+dependencies:
+  - name: postgresql
+    version: 12.1.14
+    repository: https://charts.bitnami.com/bitnami


vevetron:
Should we fix https://github.com/cal-itp/data-infra/blob/main/kubernetes/apps/charts/postgresql/Chart.yaml as well?

themightychris:
> Should we fix https://github.com/cal-itp/data-infra/blob/main/kubernetes/apps/charts/postgresql/Chart.yaml as well?

It sounds like [the charts aren't getting moved](https://github.com/bitnami/charts/issues/36215#issuecomment-3233843307), just new versions won't be posted—so I think we're good to leave that as-is until we are looking at bigger upgrades

themightychris (approved):
I pushed up the needed tweaks, this is good to merge and deploy now

vevetron:
I don't think the zookeeper configs went through.

themightychris:
@vevetron we have 4 postgres instances running within the cluster:

- `metabase`: Deployed by [the Metabase chart](https://github.com/cal-itp/data-infra/blob/main/kubernetes/apps/charts/metabase/templates/database.yaml)
- `metabase-test`: Deployed by [the Metabase chart](https://github.com/cal-itp/data-infra/blob/main/kubernetes/apps/charts/metabase/templates/database.yaml)
- `sentry`: Deployed by [the Sentry chart](https://github.com/cal-itp/data-infra/blob/main/kubernetes/apps/charts/sentry/values.yaml#L22)
- `grafana`: Doesn't appear to be deployed by any chart. Grafana doesn't come with an embedded PostgreSQL option, but the currently running version `15.1.0` doesn't match what that chart I attempted to delete would install and the labels on the StatefulSet don't reflect it being created by any Helm chart

It appears that Grafana`s PostgreSQL isn't currently managed by anything in CI. It must have been manually created at some point and never versioned in the repo. 

In any case, deleting that chart isn't wired up to any process that could delete anything running in the cluster, and its for Postgres 12.1 which is ancient and not running anywhere in the cluster—it's definitely just cruft

Issue #4445, opened 2025-10-29
MST sees `fct_kuba__device_properties` updated daily


erikamov:
## User story / feature request

As a MST user
When visualizing Kuba Dashboard
I want to see the data updated daily

### Acceptance Criteria

See the data on `fct_kuba__device_properties` updated daily.

PR #4520, opened 2025-11-14
Add Kuba and Payment models to dbt_daily DAG

erikamov:
# Description

This PR adds Kuba device models to dbt_daily DAG in order to run every day and merges payment models from dbt_payments.

Since some payment models references GTFS models, it is causing to run in duplication on both DAGs (ex: stg_gtfs_schedule__unzip_outcomes), so merging would remove this duplication and simplify having only dbt_all and dbt_daily DAGs.

Resolves [#4445]

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

Tested running airflow locally and visualized new models:
image

## Post-merge follow-ups

- [x] No action required


ohrite (approved):
🍐 w/ @erikamov

Issue #4381, opened 2025-10-06
Bug: `download_gtfs_schedule_v2.email_download_failures` fails to send email


erikamov:
**Describe the bug**
When `download_gtfs_schedule_v2` DAG runs
It return `[Errno 99] Cannot assign requested address` error on `email_download_failures` task.

It was temporarily disabled to reduce all the noise on errors in the Pipeline.

**To Reproduce**
1- On local or Staging Airflow, trigger a new run on `download_gtfs_schedule_v2`.
2 - See the log

**Expected behavior**
Expected to see no errors.

**Additional context**

Full Log:
```
default-hostname
*** Reading remote logs from Cloud Logging.
[2025-10-06, 03:05:55 UTC] {taskinstance.py:1160} INFO - Dependencies all met for dep_context=non-requeueable deps ti=<TaskInstance: download_gtfs_schedule_v2.email_download_failures scheduled__2025-10-05T03:00:00+00:00 [queued]>
[2025-10-06, 03:05:55 UTC] {taskinstance.py:1160} INFO - Dependencies all met for dep_context=requeueable deps ti=<TaskInstance: download_gtfs_schedule_v2.email_download_failures scheduled__2025-10-05T03:00:00+00:00 [queued]>
[2025-10-06, 03:05:55 UTC] {taskinstance.py:1362} INFO - Starting attempt 2 of 2
[2025-10-06, 03:05:55 UTC] {taskinstance.py:1383} INFO - Executing <Task(PythonOperator): email_download_failures> on 2025-10-05 03:00:00+00:00
[2025-10-06, 03:05:55 UTC] {standard_task_runner.py:57} INFO - Started process 2823 to run task
[2025-10-06, 03:05:55 UTC] {standard_task_runner.py:84} INFO - Running: ['airflow', 'tasks', 'run', 'download_gtfs_schedule_v2', 'email_download_failures', 'scheduled__2025-10-05T03:00:00+00:00', '--job-id', '3025448', '--raw', '--subdir', 'DAGS_FOLDER/dags.py', '--cfg-path', '/tmp/tmpt560a_yo']
[2025-10-06, 03:05:55 UTC] {standard_task_runner.py:85} INFO - Job 3025448: Subtask email_download_failures
[...]

Issue #4364, opened 2025-10-01
Staff should be able to get airflow alert notifications in teams by email


vevetron:
## User story / feature request

Staff and analysts should be able to get teams alerts of any airflow job failures.

It'll help to have this alongside the slack alerts to make sure they are working and are useful.

If we want to use another platform that is better than teams, that would be good as well.

It would be useful if the alerts have multiple levels of severity.  One channel for critical issues, another for regular issues and another channel for non-important issues.

### Acceptance Criteria
There is a non-slack way of getting airflow alerts that "works".  It might not be optimal.

jparr:
It appears that airflow is sending the email but it is not showing up at dds.app.notify@dot.ca.gov. Recommend looking at postmark to see if we can get any hints as to whats happening.

PR #4539, opened 2025-11-25
Add default `to` email for failure notifications

jparr:
ref: #4364
enable outputs for download_gtfs_schedule #4381

Adds default mail to: `dds.app.notify@dot.ca.gov`


airflow/dags/ntd_report_from_blackcat/METADATA.yml
```diff
@@ -11,6 +11,7 @@ default_args:
     email:
       - "christian.suyat@dot.ca.gov"
       - "katrina.kaiser@dot.ca.gov"
```
erikamov: Katrina's email could be probably removed since she is not on Caltrains anymore.

PR #4350, opened 2025-09-30
Update to GTFS schedule validator v7.1 (#3763)

lauriemerrell:
# Description

_Describe your changes and why you're making them. Please include the context, motivation, and relevant dependencies._

* Upgrades GTFS validator to v7.1: adds rules etc.
* Updates the build & push GitHub workflow to push to development image when on a branch to facilitate testing of PodOperator
* Moves to uv-based workflow

Resolves #3763

Reviewer notes:
* May need to drop 93c20570d82c28d1b52598f8dc28d6ef56a37410 and 602b943c796856163c97f687d9ede632c74861b9 if testing locally (sets effective date to 12/1) -- if merging after that date, this should be bumped back
* Refactor on top of #4532? Defer to @ohrite and @erikamov as to whether this should be merged in current form

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

## How has this been tested?

_Include commands/logs/screenshots as relevant._

_If making changes to dbt models, please run the command `poetry run dbt run -s CHANGED_MODEL` and `poetry run dbt test -s CHANGED_MODEL`, then include the output in this section of the PR._

* Runs in local Composer for 10/25/25 (picked as random date)
* Can see validations in `laurie_staging.stg_gtfs_schedule__validation_notices` but they don't filter into the fct model because the job times out in local Composer (I think it's just an issue with local k8s resources?) so a validation outcomes file isn't generated, which is required for notices to get into the fct model

## Post-merge follow-ups

_Document any actions that must be taken post-merge to deploy or otherwise implement the changes in this PR (for example, running a full refresh of some incremental model in dbt). If these actions will take more than a few hours after the merge or if they will be completed by someone other than the PR author, please create a dedicated follow-up issue and link it here to track resolution._

- [ ] No action required
- [ ] Actions required (specified below)

Issue #3763, opened 2025-03-11
Update to GTFS Schedule Validator v7.1


evansiroky:
## User story / feature request

As a transit data quality analyst,
I want to use the most up-to-date GTFS Schedule Validator (currently v5.0),
So that I can have the latest validation results

### Acceptance Criteria

Data pipeline is updated to use GTFS Schedule Validator v7.1. No backfill needed.

### Notes
 
MobilityData discussion: https://github.com/MobilityData/gtfs-validator/discussions/2011

Previous update to our pipeline: https://github.com/cal-itp/data-infra/issues/3305

lauriemerrell:
Note that in #4548 or a follow up we will need to cherry pick back some commits to get the dbt stuff working and we'll want to change the effective date for the v7.1 upgrade

vevetron:
https://dashboards.calitp.org/question/4652-gtfs-schedule-validator-versions-by-date

<img width="1476" height="645" alt="Image" src="https://github.com/user-attachments/assets/a1ab4dbb-d682-422a-914b-ddb865250058" />

Seeing new v7.1 validations!

PR #4559, opened 2025-12-04
Revert "Update to GTFS schedule validator v7.1 (#3763)" 

### LINKED ISSUE #4546: Bug: Hourly GTFS unzip/validate failing since 11/24
**Describe the bug**

GTFS unzip/validate Airflow job has been failing in production since early last week ~11/24

**To Reproduce**

Look at production logs

**Expected behavior**

Should succeed every day

**Additional context**

Probably related to #3763

lauriemerrell:
Reverts cal-itp/data-infra#4350 to resolve #4546

@ohrite proposed leaving the dbt changes in, but they don't really make sense if v7.1 is not actually running. The min date test will fail and the seed will not be used. Instead would recommend making a separate follow up PR to add those once we have a set date for v7.1 to start actually running to everything can be aligned.

### DIFF
--- a/jobs/gtfs-schedule-validator/pyproject.toml
+++ b/jobs/gtfs-schedule-validator/pyproject.toml
@@ -3,6 +3,7 @@ name = "gtfs-schedule-validator"
 version = "0.1.0"
 description = ""
 authors = ["Andrew Vaccaro <atvaccaro@gmail.com>"]
+package-mode = false
 
 [tool.poetry.dependencies]
 python = ">=3.8,<3.10"


lauriemerrell:
openjdk:11 image is no longer found 😩 https://github.com/cal-itp/data-infra/actions/runs/19946071540/job/57195701716?pr=4559. 

So the build image GH Action will fundamentally not succeed, which means that we can't get the reverted change pushed to [our published image](https://github.com/cal-itp/data-infra/pkgs/container/data-infra%2Fgtfs-schedule-validator) automatically -- I don't think I have permissions on our package registry to force push our :latest image tag to a previous version, but I think we'll need to do that as a manual step to get this working in the short term until the refactor merges

PR #4562, opened 2025-12-05
Deprecate `bq-transform-svcacct` service account

### LINKED ISSUE #4478: Investigate deprecating service account `bq-transform-svcacct`
Recently, when investigating issues with the payments date spine and downstream reliability tables in #4453 / #4477, we noticed that it may be an issue with the service account used by composer not having permissions to query production tables to fulfill the `create_row_access_policy` macro.

In #4477, we introduce the new service account to the macro (we still need to monitor it's success), but were unsure if the existing service account (`bq-transform-svcacct`) is still needed for this or other functions, and whether it can be deprecated.

We need to investigate how (if at all) this service account is still being used, and if not, deprecate that service account and remove it from the `create_row_access_policy` macro.

Service account details:
```
resource "google_service_account" "tfer--114731815246487683241" {
  184:   account_id   = "bq-transform-svcacct"
  185    description  = "Grants Airflow pod operators access to GCS and bigquery"
  186    disabled     = "false"
```

charlie-costanzo:
# Description
While investigating unexpected behavior in the `create_row_access_policy` dbt macro, we discovered a previously-used but potentially no-longer-needed service account (`bq-transform-svcacct`) responsible for granting Airflow pod operators access to GCS and bigquery was still in place.

Upon performing an investigation into this service account's usage metrics, it appears as this service account is no longer used and safe to remove.

Resolves #4478 

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?
Bigquery service account usage data metrics in GCP console, filtered for the past two years (12/5/23-12/5/25)
Screenshot 2025-12-05 at 14 07 55

## Post-merge follow-ups
- [x] No action required

### DIFF
--- a/iac/cal-itp-data-infra/iam/us/outputs.tf
+++ b/iac/cal-itp-data-infra/iam/us/outputs.tf
@@ -442,10 +442,6 @@ output "google_service_account_tfer--113512472366715495471_id" {
   value = google_service_account.tfer--113512472366715495471.id
 }
 
-output "google_service_account_tfer--114731815246487683241_id" {
-  value = google_service_account.tfer--114731815246487683241.id
-}
-
 output "google_service_account_tfer--114755160399419974306_id" {
   value = google_service_account.tfer--114755160399419974306.id
 }


vevetron:
Very cool!

PR #4644, opened 2026-01-06
Add GCS helper functions for Pandas

raebot:
# Description

This PR introduces helper functions to make managing data in Google Cloud Storage with Pandas easier. It's very similar to the helpers we added for GeoPandas. This is needed because Pandas is migrating towards `pyarrow.fs` as the default filesystem interface instead of `fsspec`. The `gcsfs` package we use is `fsspec`. As of pandas 2.1.0, functions like `read_parquet` are defaulting to pyarrow. To get the old behavior for `fsspec` you need to pass the `filesystem` keyword arg. These helper functions handle passing that arg for you.

Considering the similarities between the GeoPandas and Pandas interfaces for functions like `read_parquet`, `to_parquet`, and the different file format readers, I considered converting `GCSGeoPandas` into a more general GCS helper class that would surface helper functions with combined handling for geospatial and non-geospatial data. In other words, there would just be none set of helper functions, you'd pass it a file path and it would internally resolve whether to handle it with either GeoPandas or Pandas with out the caller having to specify. This would be swanky, but it would require updating a bunch of code that's already using `GCSGeoPandas` and it's not clear if GeoPandas and Pandas interfaces will remain as similar as they are. We get a little bit more flexibility keeping the helpers separated and it's an easier setup path to using them, so I kept them separate.

Related to https://github.com/cal-itp/data-analyses/issues/1848

## Type of change

- [x] New feature
- [x] Documentation

## How has this been tested?

I added automated tests for the helpers and also did some manual testing of the code in JupyterHub.

## Post-merge follow-ups

- [ ] No action required
- [ ] Communicate these changes to analysts
- [ ] Use these new helpers in data-analyses repo where needed

### DIFF
--- a/docs/analytics_tools/python_libraries.md
+++ b/docs/analytics_tools/python_libraries.md
@@ -24,7 +24,7 @@ The following libraries are available and recommended for use by Cal-ITP data an
 1. [shared utils](#shared-utils)
 2. [calitp-data-analysis](#calitp-data-analysis)
    <br> - [Query SQL](#query-sql)
-   <br> - [Accessing Geospatial Resources](#accessing-geospatial-resources)
+   <br> - [Accessing Google Cloud Storage Data](#accessing-google-cloud-storage-data)
 3. [Querying Data in BigQuery](#querying-data-in-bigquery)
    <br> - [SQLAlchemy](#sqlalchemy)
    <br> - [pandas](#pandas-resources)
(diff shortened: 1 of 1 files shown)


edasmalchi (approved):
thanks Rae, looks good to me! (looks like we'll also need a review from a code owner though)

PR #4657, opened 2026-01-07
Payments: set Enghouse bucket env variables

charlie-costanzo:
# Description
This PR sets the production and staging environment variables for the Enghouse data feed sync.

Resolves: #4051

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?
n/a

## Post-merge follow-ups
ensure these variables persist in the environments and that the create_external_tables dag tasks succeed for Enghouse


erikamov (approved):
We need to fix prod and staging bucket names, but it can be in a different PR.
Approving using current names to fix dbt tables failing.

Issue #4632, opened 2025-12-31
Add a bridge table linking `schedule_gtfs_dataset_name` to `analysis_name` and `ntd_id`


tiffanychu90:
This bridge table provides the labeling needed for majority of `data-analyses` data products. This table moves away from the versioned values to provide consistent labeling across time-series data for visualizations.

Be able to label all GTFS schedule names to support visualizations that stretch back before we added `analysis_name` to a dimension table. At the same time, do the joins across several commonly used `mart_transit_database` tables into this, such as getting `caltrans_district, `county`, `ntd_id`, `rtpa`, and `organization_source_record_id`.

* `analysis_name` was added to `mart_transit_database.dim_gtfs_datasets` in #4268. As is, it appears at some point in Sep 2025, but any `schedule_gtfs_dataset_name` prior to that would not get a label if we continue with versioned values.
* Do a grouping by `schedule_source_record_id` so we are able to get that `analysis_name` filled for all dates. 
* Work the complicated joins done in [gtfs_funnel crosswalk script](https://github.com/cal-itp/data-analyses/blob/e9c8be5f66f1487d572bf7bb5b6813107b319786/gtfs_funnel/crosswalk_gtfs_dataset_key_to_organization.py) over.

Issue #4450, opened 2025-10-30
GTFS RT vs schedule rollups


tiffanychu90:
## aggregate existing models to create new grains or build out the RT corollary to what exists in schedule
* route-direction (new grain)
* new time grain: monthly, day_type (weekday/Sat/Sun to correspond with how ridership data is)

| grain | schedule; RT |
|---|---|
| daily trips | fct_scheduled_trips; <br>fct_observed_trips <br> #4558 |
| daily route-<br>direction | fct_daily_schedule_rt_route_direction_summary <br> #4777 |
| daily feed | fct_daily_feed_scheduled_service_summary<br>scheduled trips -> aggregate to feed;<br><br>fct_daily_rt_service_summary<br>observed trips -> aggregate to feed;<br><br>**fct_daily_schedule_rt_operator_summary** <br>(left join, keep rows even if no RT) <br> #4558, #4777 |
| monthly route-<br>direction | fct_monthly_schedule_route_direction_summary <br> #4526  |
| monthly feed | fct_monthly_operator_summary<br>fct_daily_schedule_rt_operator_summary -> aggregate <br> #4558, #4777 |

 **Decision for operator aggregation**: will not bring in `analysis_name`, as this causes too many rows to drop off in ways that decrease visibility. `mart_gtfs`, `mart_gtfs_rollup` will always keep all the rows, feeds possible. Always bring in bridge table to label and remove rows -> put decision making in hands of users doing the analyses

PR #4777, opened 2026-02-05
Daily route-direction summary + refactor operator summaries

tiffanychu90:
# Description

* Add a daily schedule + RT route-direction summary and refactor downstream models 
   * Be able to query specific routes for certain operators
   * Hermosa example: https://github.com/cal-itp/data-analyses/issues/1883
* #4450
* Noticing that `fct_scheduled_trips` and `fct_observed_trips` may result in missing matches on one side or another, switch to full outer join to debug where this happens
   * Affects where `analysis_name` is being used, because that assumes the record is found in scheduled trips
   * But this misses out on observations that are present in `fct_observed_trips`, but for whatever reason we are not finding in `fct_scheduled_trips` or bridge table?  
   * https://github.com/cal-itp/data-infra/issues/4632

## Type of change

- [x] Bug fix - fix missing values for NTD ID and RTPA and MPO in bridge table, drop those 
- [x] New feature - new daily schedule + RT route-direction table
- [ ] Documentation

## How has this been tested?
* `poetry run dbt run -s fct_daily_schedule_rt_route_direction_summary`
* `poetry run dbt run -s fct_daily_rt_service_summary`
* `poetry run dbt run -s fct_daily_schedule_rt_operator_summary`
* `poetry run dbt run -s fct_monthly_operator_summary`
* `poetry run dbt run -s bridge_gtfs_analysis_name_x_ntd`

## Post-merge follow-ups

- [x] No action required

### DIFF
--- a/warehouse/models/mart/gtfs/fct_daily_rt_service_summary.sql
+++ b/warehouse/models/mart/gtfs/fct_daily_rt_service_summary.sql
@@ -5,53 +5,59 @@
     )
 }}
 
-WITH fct_observed_trips AS (
+WITH rt_route_summary AS (
     SELECT *
-    FROM {{ ref('fct_observed_trips' )}}
-),
-
-observed_trips AS (
-    SELECT
-        *,
-        -- num_distinct_message_keys = num_distinct_extract_ts and
-        -- num_distinct_message_keys / extract_duration_minutes = messages per minute this entity was present for, how continuously this trip was updated.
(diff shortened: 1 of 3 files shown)


tiffanychu90:
The daily RT service summary, daily schedule + RT route-direction summary and daily schedule + RT operator summary have metrics like `vp_messages_per_minute`, `pct_vp_trips`, etc relevant for #4777

Issue #4814, opened 2026-02-19
Deprecate unneeded dags now that we have a new gtfs download dag


vevetron:
As a developer
I no longer need these old download gtfs dags
so we should remove them.

vevetron:
Initially I only wanted to fix our data from 1/1/2025 onwards but @lauriemerrell has shown that possible unzip failures in the past are killing dbt tests.

We should identify those errors and re-run for those dates as well if possible.  If we need to lose these dags due to airflow upgrade that is okay as well.

erikamov:
There is a PR https://github.com/cal-itp/data-infra/pull/5090 deleting those deprecated GTFS DAGs and old validator.
It also removes the deprecated `sync_ntd_data_xlsx` DAG.

vevetron:
Hopefully we can fix those data problems and then merge in that PR!  At that point we can really close this issue.

PR #4815, opened 2026-02-19
Make `unzip_and_validate_gtfs_schedule_hourly` and `download_gtfs_schedule_v2` DAGs run only Manually

erikamov:
# Description

As part of the deprecation of the previous GTFS DAGs (#4814), we are turning off schedules for `unzip_and_validate_gtfs_schedule_hourly` and `download_gtfs_schedule_v2` so it will only run when manually triggered.

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

Tested on Airflow locally.

## Post-merge follow-ups

- [x] No action required


vevetron (approved):
Paired with @erikamov

PR #4863, opened 2026-03-03
feat: microbatch strategy for dim stop arrivals (#4783)

### LINKED ISSUE #4783: Enhancement: dim_stop_arrivals is fairly inefficient
**Describe the Enhancement**
dim_stop_arrivals is costing 33$ a run.  I think it's because dim_stop_times is not optimized for this use case.

Someone should investigate and find ways to better optimize both tables.

If the cost goes down, the data is probably more optimized!

**To Reproduce**
https://dashboards.calitp.org/question/4060-how-much-do-dbt-runs-cost-per-table-per-day

**Expected behavior**
Less costly.

**Additional context**
Why isn't dim_stop_times even materialized / incremental?  strange.

Might also be worth checking out the bigquery jobs explorer to see what is happening.

lauriemerrell:
# Description

Resolves #4783 

Implements microbatch incremental strategy on `dim_stop_arrivals`.

`lookback=5` so that on Monday `dbt_all` run it will refresh data back to prior Thursday. 

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

Ran multiple times in staging. Verified that it ended up with identical data as in prod. 

_If making changes to dbt models, make sure they were created or update on Staging. Please run the command `poetry run dbt run -s CHANGED_MODEL --target staging` and `poetry run dbt test -s CHANGED_MODEL --target staging`, then include the output in this section of the PR._

## Post-merge follow-ups

- [ ] No action required

Monitor executions

### DIFF
--- a/warehouse/models/mart/gtfs/_mart_gtfs_dims.yml
+++ b/warehouse/models/mart/gtfs/_mart_gtfs_dims.yml
@@ -844,6 +844,8 @@ models:
       Each row is a cleaned row from a trips.txt file.
       Definitions for the original GTFS fields are available at:
       https://gtfs.org/reference/static#tripstxt.
+    config:
+      event_time: _feed_valid_from
     columns:
       - *schedule_dim_pk
       - *schedule_dim_gtfs_key_with_dupe_flag


vevetron (approved):
You say it's tested!  This might break things but we need to try.  Though there is some merge conflict going on.

tiffanychu90 (commented):
Yes, just try! Fairly limited in children. I did make a copy of our existing prod fct_daily_scheduled_stops (with the copy thing in the UI) so we can sanity check the results of this downstream.

Issue #4899, opened 2026-03-09
Bug: Elavon parse failing


lauriemerrell:
**Describe the bug**
_Please enter a clear and concise description of the bug._

Elavon parse Airflow task is failing without a meaningful error message on 3/9/26. Have tried clearing it and it continues to fail silently.

**To Reproduce**
_Please enter step by step instructions to reproduce the behavior. For example: Query column A for organization Y on date Z._

Try to clear Airflow execution for 3/9/26 for the `parse_elavon` DAG task.

**Expected behavior**
_Please enter a clear and concise description of what you expected to happen._

Should run

**Additional context**
_Please enter any other context about the problem here._

See also #4872 #4420 #2462

lauriemerrell:
Failed again today 3/13, cleared it.

Issue #4912, opened 2026-03-11
Add `_FILE_NAME` to Enghouse staging dbt models


lauriemerrell:
## User story / feature request

_Please describe your need, outlining the key users, the feature being requested, and the goal that that the feature will facilitate. For example: **As a \[user or stakeholder type\], I want \[software feature\] so that \[some business value\]**_

As an analytics engineer, I want visibility into the source file for data in the warehouse to enable troubleshooting when issues arise. Specifically, we should add `_FILE_NAME` to the staging dbt models for Enghouse data so we can see the source filename. (Until Enghouse has a parse job, we can surface this data directly in this way because the external tables are pointed at the location where data lands into the warehouse.)

### Acceptance Criteria

_Please enter something that can be verified to show that this user story is satisfied. For example:  **I can join table X with table Y.** or **Column A appears in table Z in Metabase.**_

Add `_FILE_NAME` to staging model. See: https://docs.cloud.google.com/bigquery/docs/query-cloud-storage-data#query_the_file_name_pseudo-column

### Notes

_Please enter any additional information that will facilitate the completion of this ticket. For example: Are there any constraints not mentioned above? Are there any alternatives you have considered?_

lauriemerrell:
@mrtopsyt -- unassigning myself, not 100% sure of next step given the updated parse job. Might merit a quick review to see whether this needs to be prioritized.

vevetron:
Lets review this one properly for sure - make sure the line numbers get through.

mrtopsyt:
Confirmed that _line_numbers are visible in mart tables and correct (`rides` maps to the correct line `ticket_results` and `settlements` maps to the correct line in `transactions`.)

PR #4933, opened 2026-03-16
Make elavon operations sequential and in memory

vevetron:
# Description

The Elavon dag parse job is failing since it copies all the files locally (into airflow) then unzips and makes one giant transactions zip before uploading.  

This unzips into memory, cleans, and rezips into individual files and uploads.

Work towards #4899

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Tested once in staging but should be tested more.

- ran dag locally
- uv run dbt run -s stg_elavon__transactions
--> Produced rows in staging

## Post-merge follow-ups

- [x] No action required


lauriemerrell (approved):
I cleaned up comments a little bit, confirmed that I can run in local Composer and that the resulting files can be read through the existing external table definition. Think this is worth a shot as a quick fix. 

Can do a follow up to add `_FILE_NAME` to Elavon staging tables a la #4912 for observability

Issue #5035, opened 2026-04-03
Remove old Elavon k8 infra after they are switched.


jparr:
Checklist of things we can remove once the parent is complete:

- [ ] terraform infra for old endpoint
- [ ]  K8s overlays (prod-sftp-ingest-elavon, dev-sftp-ingest-elavon)
- [ ] workload
- [ ] old credentials and env vars
- [ ] sync_elavon DAG
- [ ] Buckets? the prod bucket is huge since its daily syncs of the same data for years #4872

vevetron:
Heads-up while auditing `data-infra-apps` compute: the **live K8s workload is still running** even though #5068 removed its source.

- ✅ **K8s overlays** — deleted from source by #5068 (`kubernetes/apps/overlays/prod-sftp-ingest-elavon/`, incl. `ns.yaml`) and its entry dropped from `ci/kubernetes-workloads.yaml`.
- ❌ **workload — still live.** The `sftp-server-0` StatefulSet (fedora:latest, ~3.7 yr old) + its `data-sftp-server-0` **50 GiB PVC** are still running in the `prod-sftp-ingest-elavon` namespace on the `apps-v2` pool. Removing the CI entry stopped CI from reconciling it — which *orphaned* it rather than deleting it. Nothing in git references it anymore.

The switch looks complete: the new `elavon-sftp-deployment` (gcsfuse → `calitp-elavon-raw-v2`, in the `sftp-endpoints` cluster) has been live ~104 days, and the old server's logs show only kubelet health probes (`Connection closed by 10.138.0.84`) — zero authenticated SFTP sessions.

To check off **workload**, after a final confirmation that no external uploads still hit the old endpoint:

```
kubectl delete namespace prod-sftp-ingest-elavon   # removes the StatefulSet + reclaims the 50Gi PVC
```

Reclaims ~$2/mo (the disk); the pod reserves no CPU/memory, so no node savings. The remaining boxes (terraform for old endpoint, old creds/env vars, `sync_elavon` DAG, buckets per #4872) look partly handled by #5068 — worth a quick pass to tick them off or note what's left.

vevetron:
workload removed — ran `kubectl delete namespace prod-sftp-ingest-elavon` (StatefulSet + 50Gi PVC + `sftp-internet` LB, IP `34.145.56.125` released). Source cleanup in #5556.

HaroldBooker:
@jparr unassigning until the project management team figures out scoping of this item.

PR #5068, opened 2026-04-13
Remove old elavon infra

### LINKED ISSUE #4872: Stop syncing all Elavon history every day
Currently, we re-sync all files from Elavon every day. This causes unnecessary data duplication and perhaps counterintuitively is a bit complicated for troubleshooting purposes because we don't know exactly when specific data arrived.

* We should only sync new files, similar to what we do in Littlepay.
* We could also consider changing the partition scheme on the parsed data to add a level for source file (related to #4871) -- this would allow it to be an attribute without having to hard-code it in the file 
* Instead of creating one enormous JSON for all data, we should create one JSON per source file

jparr:
# Description

This PR is here when we are ready to switch elavon to the new endpoint. Dont merge before then

ref: #5035


mrtopsyt (approved):
Removes sync_elavon job - confirmed ok to merge in slack with @jparr and @vevetron

PR #4932, opened 2026-03-14
Fix dbt artifacts update timing

erikamov:
# Description

This PR changed the order when dbt artifacts are uploaded to buckets and Terraform changes are applied to ensure the information is updated correctly.

The correct order is:

1. Run model changes (create or update)
2. Compile dbt
3. Upload dbt artifacts
4. Check if there are Airflow Terraform changes, since it is changing dbt artifacts
5. Create Airflow Terraform Plan for Staging and Production
6. If branch starts with `staging/`, apply Airflow Terraform changes to Staging
7. When PR is merged to main, apply Airflow Terraform Plan for Staging and Production

Fix race condition issue on  #4929.

I changed Kuba models to microbatch in order to run and test the deploy-dbt workflow (requested on issue https://github.com/cal-itp/data-infra/issues/5064).

Other minor fixes:
* Upgraded `actions/checkout` to v6 on workflows touched by this PR
* Added configuration to cancel concurrent running on `lint` and `test-airflow` workflows

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

## How has this been tested?

Tested workflows using this PR https://github.com/cal-itp/data-infra/actions/runs/24419118852.
image

Kuba model tested locally:
```
❯ uv run dbt run -s models/staging/kuba/stg_kuba__device_properties.sql+
01:58:39  Found 627 models, 178 data tests, 16 seeds, 226 sources, 4 exposures, 1088 macros
01:58:39
01:58:39  Concurrency: 8 threads (target='dev')
01:58:39
01:58:44  1 of 2 START sql view model erikap_staging.stg_kuba__device_properties ......... [RUN]
01:58:46  1 of 2 OK created sql view model erikap_staging.stg_kuba__device_properties .... [CREATE VIEW (0 processed) in 2.15s]
01:58:46  2 of 2 START sql microbatch model erikap_mart_kuba.fct_kuba__device_properties . [RUN]
01:58:46  Batch 1 of 2 START batch 2026-04-14 of erikap_mart_kuba.fct_kuba__device_properties .. [RUN]
01:58:54  Batch 1 of 2 OK created batch 2026-04-14 of erikap_mart_kuba.fct_kuba__device_properties  [SCRIPT (0 processed) in 7.85s]
[...]

Issue #3711, opened 2025-02-19
Chore: Review production Buckets


erikamov:
### Chore

* Generate a list of production Buckets that has not been accessed within the last year to define if the data should be deleted or moved to a cold storage.

* Delete test buckets that are not in use anymoer.
   Most of buckets stating with `test-` on production `cal-itp-data-infra` were replaced by staging buckets on `cal-itp-data-infra-staging`, so now we can remove them from production.

erikamov:
PR https://github.com/cal-itp/data-infra/pull/5239 removes most of the deprecated Test buckets from production.
I left the buckets bellow because it was mentioned in some other parts that would need more time to review and replace by staging buckets:
* `test-calitp-airtable`
* `test-calitp-dbt-python-models`
* `test-calitp-gtfs-download-config`
* `test-calitp-gtfs-rt-raw`
* `test-calitp-gtfs-schedule-raw`
* `test-calitp-gtfs-schedule-raw-v2`
* `test-calitp-sentry`
* `dev-calitp-gtfs-rt-raw`

More buckets to review:
* `rt-parsed` and `rt-parsed-deprecated` contains old data. Could be deleted if there is no more need for that old data.
* `littlepay-data-extract-prod`
* `gtfs-data`, `gtfs-data-reports`, and `gtfs-data-test`

The `cal-itp-data-infra-enghouse-raw` can also be removed since it was just replaced by `calitp-enghouse-raw` because of the wrong name style.

vevetron:
ICEBOX - maybe someone can create a new issue based of of this as sort of a yearly review.

PR #5077, opened 2026-04-14
Convert remaining GTFS RT incremental models to microbatch (#5063)

lauriemerrell:
# Description

_Describe your changes and why you're making them. Please include the context, motivation, and relevant dependencies._

Resolves #5063

Converts the following models to microbatch
- [x] mart.gtfs.fct_daily_rt_feed_files
- [x] mart.gtfs.fct_hourly_rt_feed_files
- [x] mart.gtfs.fct_hourly_rt_feed_files_success
- [x] mart.gtfs.fct_service_alerts_messages_unnested
- [x] mart.gtfs.fct_trip_updates_no_stop_times
- [x] mart.gtfs.fct_vehicle_locations
- [x] intermediate.gtfs_quality.int_gtfs_quality__rt_validation_notices
- [x] intermediate.gtfs_quality.int_gtfs_quality__rt_validation_outcomes
- [x] intermediate.gtfs.int_gtfs_rt__distinct_download_configs
- [x] intermediate.gtfs.int_gtfs_rt__unioned_parse_outcomes

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

## How has this been tested?

_Include commands/logs/screenshots as relevant._

_If making changes to dbt models, make sure they were created or update on Staging. Please run the command `uv run dbt run -s CHANGED_MODEL --target staging` and `uv run dbt test -s CHANGED_MODEL --target staging`, then include the output in this section of the PR._

GitHub Actions runs confirming dbt runs as expected. 

## Post-merge follow-ups

_Document any actions that must be taken post-merge to deploy or otherwise implement the changes in this PR (for example, running a full refresh of some incremental model in dbt). If these actions will take more than a few hours after the merge or if they will be completed by someone other than the PR author, please create a dedicated follow-up issue and link it here to track resolution._

- [ ] No action required
- [x] Actions required (specified below)

* Run repartitioning notebook for `fct_daily_rt_feed_files` (`date` column)
* Monitor data volumes in prod after merge

PR #5078, opened 2026-04-15
Review deployment workflows

### LINKED ISSUE #4929: Clarify CICD GitHub Actions for dbt deployment -- possible race condition?

As an analytics engineer, I want to understand the purpose of the different CI/CD automations that affect the deployment of dbt artifacts and changes.

Specifically, I am confused about these three actions:

* Build dbt image: https://github.com/cal-itp/data-infra/actions/runs/22956890742/workflow
* Deploy dbt: https://github.com/cal-itp/data-infra/blob/main/.github/workflows/deploy-dbt.yml
* Apply Terraform for DAG and warehouse changes: https://github.com/cal-itp/data-infra/blob/main/.github/workflows/composer-apply-files.yml

Questions:
1. Why do we need a dbt image? Is that actually used for any deployment?
2. What is the intended relationship between the second two -- it seems like the locations that dbt is deployed to (`calitp-dbt-artifacts` bucket) do not correspond to the Terraform code that is deployed (in `calitp-composer` bucket)? 
3. Why do we run a full refresh as part of CI? Historically, we would not run a full refresh on CI, only manually after merge if it was needed. See for ex: https://github.com/cal-itp/data-infra/actions/runs/22956890726/job/66636950226 -- ran a full refresh of an 8TB model, it was not needed

See notes section for more details on issue that arose with #4894

### Acceptance Criteria

* Clarify ordering/dependencies -- does something need to be refactored to prevent a race condition here to ensure that the code is deployed as intended?
* Ensure that the version of dbt on the branch is compiled and deployed to Composer
* Consider deprecating any step that is not needed
* Consider stopping full refresh by default

### Notes

Issues specifically arose with the merging of #4894.

GitHub Actions:
* Deploy dbt - run: https://github.com/cal-itp/data-infra/actions/runs/22956890726, results in [this manifest file](https://storage.cloud.google/calitp-dbt-artifacts/manifest.json/dt%3D2026-03-11/ts%3D2026-03-11T14%3A18%3A05.997000%2B00%3A00/manifest.json) (in GCS) which correctly contains the new `dim_stop_times_orig` 
* Apply Terraform for DAG and warehouse changes - run: https://github.com/cal-itp/data-infra/actions/runs/22956890753/job/66636106172, results in [this manifest](https://caltrans.sharepoint.com/:u:/s/DOTPMPHQ-DDSContractors/IQCPEqk9ePZURp5dWAMyNhdZAcxO4fAVF_v4abi0w6wdCPA?e=OshlOs) (saved in Sharepoint) which does not include the new `stop_times_orig` file -- and then `dim_stop_times_orig` was not reflected in `dbt_all` run on 3/13. 

It seems like there was some kind of race condition or ordering issue. The [updated SQL did get added to the deployed warehouse bucket](https://storage.cloud.google/calitp-composer/data/warehouse/models/mart/gtfs/dim_stop_times_orig.sql) but not the deployed manifest file, and that seemingly resulted in it not getting added to `dbt_all`. (Which to me is also confusing... It seems wrong that the manifest in `calitp-composer/data/warehouse/target` does not correspond to what you'd get if you ran `dbt compile` on the contents of that folder....)

erikamov:
# Description

Follow up work to resolve #4929.

When merging [PR 4932](https://github.com/cal-itp/data-infra/pull/4932) to fix race condition for dbt artifacts, I noticed that the Terraform plan running again. It should run only during PR since Terraform Apply runs when merging.
The workflow ran successfully but there is no need for the plan to run again and I want to avoid possible issues.

As requested, this PR removes the step that uploads dbt artifacts to `calitp-staging-dbt-artifacts` on every `deploy-dbt` run and to `calitp-dbt-artifacts` when PRs are merged.

Fixes CKAN documentation that still mentioning dbt artifacts bucket but it actually uses the dbt-docs bucket.

Listing all dbt-artifacts buckets they current contains:
16,123,002,955 bytes  `gs://calitp-dbt-artifacts`
  3,315,682,497 bytes  `gs://calitp-staging-dbt-artifacts`
                        0 bytes  `gs://test-calitp-dbt-artifacts` (it was replaced by `calitp-staging-dbt-artifacts`)

The only data needed is:
         3,534,085 bytes  `gs://calitp-staging-dbt-artifacts/latest`

To save those extra storage costs for data that are not needed, we can delete all the content from the production `calitp-dbt-artifacts` bucket and most of the content from `gs://calitp-staging-dbt-artifacts` (besides `latest/` folder).

This PR will also remove the old and empty `test-calitp-dbt-artifacts` bucket, as part of the work for issue https://github.com/cal-itp/data-infra/issues/3711.

To complete the issue I am removing `Build dbt image` (build-warehouse-image.yml) since the image is deprecated.

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Tested on this PR, there is no more `upload to dbt artifact bucket`:
image

## Post-merge follow-ups

- [ ] No action required

Delete all data from `gs://calitp-dbt-artifacts` and most data from `gs://calitp-staging-dbt-artifacts` (besides `latest/` folder).


lauriemerrell (commented):
Thank you Erika!! This is exciting to get it cleaned up. However I am still wondering about two things:

 1. Why is the staging bucket (`calitp-staging-dbt-artifacts/latest`) still needed? Like why is that one still used if prod is not? (Just want to understand what is different between the staging and prod workflows.)
 2. Regarding not doing a second TF plan -- Just to confirm -- in the scenario where we have:
    a. Person A opens PR 1. Actions all run and succeed, TF plan runs.
    b. Person B opens PR 2. Actions all run and succeed, TF plan runs.
    c. PR 2 merges.
    d. The expectation is that PR 1 would need to be rebased before it can merge and this would trigger a new TF plan?  Is that a hard enough requirement that there is *no* chance that PR 1 could apply a stale plan that doesn't account for PR2 2? 

Basically I just want to confirm that you can't have the situation that someone is running a `tf apply` based on a stale plan.

erikamov:
> 2\. PR 2 merges.

@lauriemerrell 
Answering your questions:
1. There is this model image report that shows on PRs with model changes that adds comments on PRs, like this:
image

It may be from your time.
So it compares what is in the latest manifest with the manifest generated from the PR... it is how it knows which models are different to build the report. The report helps to inform if you need to change models to incremental or materialized tables.

The code is very old, but since it works I never changed. The only change I had to make is on the workflow itself, to add a link when the report is too big to post a comment on the PR.
Exactly what is happening on you latest [PR](https://github.com/cal-itp/data-infra/pull/5077), since you are changing a lot of models it is too big to add to the comment but it shows a comment with a link and you can download the [zip file to review](https://github.com/cal-itp/data-infra/actions/runs/24462999235/artifacts/6454202698)
image

The report only runs every time you create or update PRs, when the PR is merged, the workflow uploads the compiled dbt artifacts to the staging bucket as the latest files. It does not use prod buckets, because it just comparing changes while on PRs and the changes are only on staging at that time.

2. If the Plan is stale it won't be able to apply, but we always need to rebase a PR.
On your example, once the First PR is merged, the second PR will need to rebase. Then the rebase will build a new plan.

You may see some staging apply failing, because even though it has a plan, someone applied changes (probably manually or through a `staging/` branch) and made the plan to be outdated.

For production is very rare to fail, but it may happen if there were manual changes to it.
If the Terraform Apply fails, depending on the error you can re-run, or the next plan and apply will catch what is missing and apply it.

lauriemerrell (approved):

PR #5090, opened 2026-04-15
Remove deprecated DAGs

### LINKED ISSUE #4814: Deprecate unneeded dags now that we have a new gtfs download dag

erikamov:
# Description

This PR removes deprecated GTFS DAGs and Validator replaced by `download_parse_and_validate_gtfs`:
* download_gtfs_schedule_v2
* unzip_and_validate_gtfs_schedule_hourly
* jobs/gtfs-schedule-validator

Resolves #4814

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Applied changes to Staging Airflow.

## Post-merge follow-ups

- [x] No action required


lauriemerrell (commented):
Broadly looks good but two questions:

1. Did we confirm whether we can run the new job to re-validate historical data? Or maybe people would just need to do that from the command line or something? (Like say we want to re-validate a zipfile that failed in December 2025 -- can we do that with the new Airflow job somehow)?

2. Could we do a search of the repo docs for these DAG names and confirm that references to these DAGs are updated/removed? E.g. in https://github.com/cal-itp/data-infra/blob/main/airflow/dags/README.md

erikamov:
> Broadly looks good but two questions:
> 
> 1. Did we confirm whether we can run the new job to re-validate historical data? Or maybe people would just need to do that from the command line or something? (Like say we want to re-validate a zipfile that failed in December 2025 -- can we do that with the new Airflow job somehow)?
> 2. Could we do a search of the repo docs for these DAG names and confirm that references to these DAGs are updated/removed? E.g. in https://github.com/cal-itp/data-infra/blob/main/airflow/dags/README.md

I created this PR for when @vevetron is ready to remove these old DAGs and to make it easier if we are not "here".

lauriemerrell:
Ope, fair enough. Questions are for him to consider at that point then :)

PR #5106, opened 2026-04-17
Review `deploy-dbt` workflow

erikamov:
# Description

This is another attempt to deploy dbt changes in the correct sequence without erroring on Terraform Apply for Production.

I am running Plan before Apply.

https://github.com/cal-itp/data-infra/issues/4929

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Running this PR.

## Post-merge follow-ups

- [ ] No action required

Monitor workflows.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -22,10 +22,18 @@ This workflow builds a visual comment about model changes and uploads latest dbt
 
 ## composer-plan-files.yml
 
+> [!NOTE]
+> This workflow creates plans to show DAG changes.
+> If there are also changes to `iac/.../airflow/` or `iac/.../composer/` files, only the [terraform-plan.yml](#terraform-plan.yml) workflow will create the plan.
+
 When creating a PR that changes any file in the `airflow` directory, this workflow runs `Terraform plan` to create a comment with an execution plan to preview the changes that will be made to Production and Staging Airflow DAGs.
 
 
 ## composer-apply-files.yml
+
+> [!NOTE]
+> This workflow applies DAG changes.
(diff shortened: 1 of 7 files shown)


lauriemerrell:
@stevenschrayer might be a good reviewer here since he's been getting familiar with the CI/CD workflows (even though he's not a code owner)

stevenschrayer:
Nothing jumping out to me as blockers, though two small two questions:

- would it be useful a `concurrency` configuration to the different plan/apply actions? I'm wondering if that may help branches merging too close together or as a follow-up if there are still race conditions. Though thinking about it this may be unnecessary with terraform and locking state.
- have the new action(s) run against this branch? I didn't see them in run history

erikamov:
* I added concurrency only on Plan, because when it is just a PR we are constantly changing or rebasing, so it could be cancelled. For Apply, since it is based on the plan, it shouldn't cancel. It is a cycle: Once a PR is merged, the Terraform plan is applied, all pending PRs will need to be rebased and a new plan will be generated for each one. Then the cycle repeats.
* For the triggered workflow, it needs to be in main in order to run. I tried the other approach first because of it. I will need to merge the new workflow to really test it. (Updated the test part on the PR)

stevenschrayer:
@erikamov sounds good to me, thanks! also realized that the new actions and updated actions should get added to the README in .github/workflows/README

erikamov:
> @erikamov sounds good to me, thanks! also realized that the new actions and updated actions should get added to the README in .github/workflows/README

Good call! I am adding now.

ohrite (approved):

Issue #4922, opened 2026-03-12
Quarterly Maintenance: GKE Cluster


vevetron:
## Quarterly Maintenance: GKE Cluster

_Please Check for updated stable versions of the above software and relevant security patches._

Also address the following:
Received an email that:
- Cluster: data-infra-apps's certificate authority at us-west1 will expire on 2026-05-25 19:00:54+00.

### Prior Work

_Please Link to previously completed relevant tickets._

### Dependencies

_List any systems or components that might be affected by this maintenance job._

### Acceptance Criteria

- [ ] Describe the upgrade methodology and any specific steps involved (unless already documented in prior work).
- [ ] Document the upgrade process and any relevant changes.
- [ ] Create issues around any roadblocks encountered.
- [ ] Conduct thorough testing to verify functionality after the upgrade and patching.

### Rollback Plan

_Briefly outline the steps to revert to the previous version in case of issues._
(Probably not necessary for this)

### Notes

_Please enter any additional information that will facilitate the completion of this maintenance task. For example: Are there any constraints not mentioned above? Are there any alternatives you have considered?_

themightychris:
I'm going to start with the high priority ones and start staging PRs, but I won't be able to test/validate much without applying changes to the prod cluster and I'm assuming I should hold off on applying anything to the prod cluster until v is back?

Otherwise looking for thoughts on how far down this list I should go right now considering what we're planning to sunset and when

themightychris:
is there still a notion of retiring Grafana/Prometheus/Loki from the cluster in favor GCP native logging/monitoring? That seems imminently doable and maybe even less work than upgrading them all with the associated care and feeding

themightychris:
We can probably shut down pretty much everything in here but the RT archiver and JupyterHub, dropping in GCP native ingress to replace ingress-nginx and cert-manager. Slightly pricer but fully managed and ingress-nginx is entirely EOL this month

themightychris:
# Revised Plan: Mostly Decommission

After looking at the full picture, most of what's running on the cluster can be decommissioned rather than upgraded. Only three services warrant ongoing investment: the GTFS-RT Archiver, JupyterHub, and Sentry.

## Strategy

| Component | Strategy | Replacement / Notes |
|---|---|---|
| CA Certificate | **Rotate** | GKE credential rotation, before May 25, 2026 |
[...]

PR #5157, opened 2026-04-24
Add GKE Ingress alongside nginx for JupyterHub (test hostname)

themightychris:
## Summary

Phase 1 of migrating JupyterHub off `ingress-nginx` (which is EOL as of March 2026) onto GKE's built-in `gce` Ingress controller. This PR stands up the new path in parallel with the existing one — prod `notebooks.calitp.org` stays flowing through nginx, unchanged. The new Ingress is scoped to the test hostname `hubtest.k8s.calitp.jarv.us` for validation.

## Changes

New templates in the wrapper chart's `templates/` dir:

- **ManagedCertificate** — Google-managed TLS cert for `hubtest.k8s.calitp.jarv.us`. Starts provisioning automatically once the hostname resolves to the new GCLB IP (~15-30 min).
- **BackendConfig** — `timeoutSec: 86400` (default 30s kills long-running cells and WebSocket kernels) + health check at `/hub/health`.
- **Ingress** — `ingressClassName: gce`, attaches the ManagedCertificate, routes to `proxy-public`. Scoped to the test hostname during staging.

Plus a `jupyterhub.proxy.service.annotations` update in `values.yaml` to:
- Enable container-native load balancing (NEG) so the GCLB routes directly to pods rather than through kube-proxy
- Attach the BackendConfig

## Why test hostname first

The Managed Certificate won't issue until the hostname resolves to the GCLB IP. Staging on the test hostname `hubtest.k8s.calitp.jarv.us` lets us:
1. Let the cert provision in a zero-risk window
2. Validate JupyterHub behavior end-to-end on GCLB (WebSockets, long cells, auth)
3. Flip prod DNS in a separate, deliberate step after soak

## Already applied

All four resources are already applied to the cluster (service annotation, ManagedCertificate, BackendConfig, test Ingress). `helm template` confirms next helm deploy is idempotent with this PR's contents.

## What's NOT in this PR

- `notebooks.calitp.org` DNS change (later phase, coordinates with team)
- Removal of the nginx Ingress (later phase, after prod validation)
- ingress-nginx + cert-manager removal (later, after grafana + sentry ingresses also go away)

## Test plan

- [x] CI passes (lint, kubernetes preview)
- [x] `kubectl get managedcertificate jupyterhub-managed-cert -n jupyterhub` → status transitions from Provisioning → Active once DNS points at the GCLB
- [x] `kubectl get ingress jupyterhub-gce -n jupyterhub` gets a GCLB external IP within ~10 min of apply
- [x] Update `hubtest.k8s.calitp.jarv.us` DNS A record to point at the new GCLB IP
- [ ] JupyterHub login flow works end-to-end on the test hostname
- [ ] Spawn a notebook; run a long-running cell; verify kernel survives
- [ ] Page refresh during long kernel operation does not disconnect (WebSocket health)
- [ ] Helm upgrade is a no-op against the cluster post-merge

## References

- #4922 (cluster maintenance tracker)

### DIFF
--- a/kubernetes/apps/charts/jupyterhub/templates/gke-backend-config.yaml
+++ b/kubernetes/apps/charts/jupyterhub/templates/gke-backend-config.yaml
@@ -1,0 +1,13 @@
+apiVersion: cloud.google.com/v1
+kind: BackendConfig
+metadata:
+  name: jupyterhub-backend
+spec:
+  # Long notebook cells + WebSocket kernel sessions. Default 30s kills
+  # anything that takes longer than a trivial HTTP request.
+  timeoutSec: 86400
+  connectionDraining:
+    drainingTimeoutSec: 60
+  healthCheck:
+    requestPath: /hub/health
+    type: HTTP
--- a/kubernetes/apps/charts/jupyterhub/templates/gke-ingress.yaml
+++ b/kubernetes/apps/charts/jupyterhub/templates/gke-ingress.yaml
@@ -1,0 +1,24 @@
+# Second Ingress serving JupyterHub via GKE's built-in gce controller
+# (in parallel with the upstream chart's nginx-backed Ingress). Scoped
+# to the test hostname only during the staged migration off ingress-nginx
+# — prod notebooks.calitp.org continues flowing through nginx until
+# we're satisfied with the GKE Ingress behavior.
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: jupyterhub-gce
+  annotations:
+    networking.gke.io/managed-certificates: jupyterhub-managed-cert
+spec:
+  ingressClassName: gce
+  rules:
+    - host: hubtest.k8s.calitp.jarv.us
+      http:
+        paths:
+          - path: /
+            pathType: Prefix
+            backend:
+              service:
+                name: proxy-public
+                port:
+                  name: http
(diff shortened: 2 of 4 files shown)


themightychris:
This is already applied to the cluster with the test host name, so we should go ahead and get it merged so GitOps stay in sync and separately coordinate switching over the real hostname.

I'm not able to fully test JupyterHub with the test hostname because it would require messing with the OAuth configuration, but I think it's clear enough that the GKE Ingress controller is working correctly.

Next steps would be:

- [ ] Schedule a window where we can synchronize updating the DNS record and applying the new hostname to the cluster. There will be some downtime while SSL gets provisioned that we can't avoid. Possibly 30 minutes at MOST but hopefully far less
- [ ] Let a few days go by of users accessing JupyterHub through the new ingress and then tear down `ingress-nginx` and `cert-manager` if no issues are observed

lauriemerrell (approved):

themightychris:
I'm going to wait until after hours to merge this because the config checksum updates make it slightly less of a NO-OP as they'll force JuptyerHub to restart even though there is no effective configuration change

themightychris:
Instead of coordinating a switchover of the calitp.org hostname, we're going to assign `jupyterhub.dds.dot.ca.gov` in parallel and then switch over. Awaiting the DNS record creation

PR #5164, opened 2026-04-24
Remove mart payments schema from DBT Metabase exclusion (#5061)

### LINKED ISSUE #5061: Bug: Table & column docs not fully syncing to Metabase
**Describe the bug**

Checking in `Metabase > [left sidebar] Databases > Payments [Agency] databases > fct_payments_rides_v2` we don't see table and column level docs for all agencies. We see them for CCJPA but not Humboldt or El Dorado

**To Reproduce**

See above for the path 

**Expected behavior**

All tables in Metabase should have table and column descriptions from dbt 

**Additional context**

Per @csuyat-dot this is of increasing interest to agencies

jamalk23:
# Description

Completed with @mjumbewu's direction and guidance in a KT session.

This PR removes the payments mart schemas from the dbt-metabase sync exclusion list in `deploy-dbt.yml`, so table and column level docs now sync to Metabase for all payments agency databases (not just CCJPA). It also routes the sync at the staging Metabase instance when the workflow runs from against staging, mirroring the branch-conditional pattern already used for `GOOGLE_CLOUD_PROJECT`, `METABASE_DESTINATION_DATABASE`, and `DBT_DOCS_URL` in the same env block.

A new GitHub secret `METABASE_STAGING_API_KEY` was created, holding a new API key generated in staging Metabase under the display name `GitHub Actions - dbt-metabase`.

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Opening this PR triggers the `Deploy dbt` workflow which will synchronize the payments mart schemas.

Update post workflow run: 

The `Deploy dbt` workflow ran on this PR, but the `Sync Metabase` job was skipped because it is gated behind a `dbt-changed` check that only flips to `true` when files under `warehouse/models/**/*.sql` or `warehouse/seeds/*.csv` change. This PR touches neither, so the new branch-conditional staging URL/key logic was not exercised on this PR's run.

The same gate applies to the on-merge `Deploy dbt` run on `main`, so `Sync Metabase` will likely also skip at merge time. The new logic will be first exercised on the next PR that modifies a dbt model or seed, against staging Metabase. I will need to further discuss this with @mjumbewu to understand how we can cause the sync to occur when the merge to `main` occurs.

## Post-merge follow-ups

- [] No action required

Must confirm that `Sync Metabase` is not skipped and causes the doc sync to occur successfully.

### DIFF
--- a/.github/workflows/deploy-dbt.yml
+++ b/.github/workflows/deploy-dbt.yml
@@ -218,9 +218,79 @@ jobs:
         working-directory: warehouse
         run: uv run dbt debug --target ${{ env.DBT_TARGET }}
 
+      - name: Discover Payments Metabase databases
+        id: payments-databases
+        run: |
+          set -euo pipefail
+
+          # Each agency has its own Metabase database connection named
+          # "Payments - <Agency Name>" (per the create-metabase-dashboards
+          # runbook). Column-level docs from dbt have to land on every one
+          # of those connections, not just the central warehouse connection,
+          # so this step enumerates them via the Metabase API.
+          PAYMENTS_DBS=$(curl -fsS \
+            -H "x-api-key: ${METABASE_API_KEY}" \
+            "${METABASE_URL}/api/database" \
+            | jq -r '.data[] | select(.name | startswith("Payments - ")) | .name')
+
+          if [ -z "${PAYMENTS_DBS}" ]; then
+            echo "No 'Payments - *' databases found. Will sync only the central database."
+          else
+            echo "Discovered Payments databases:"
+            printf '  - %s\n' ${PAYMENTS_DBS}
+          fi
+
+          {
+            echo "names<<EOF"
+            echo "${PAYMENTS_DBS}"
+            echo "EOF"
+          } >> "${GITHUB_OUTPUT}"
+
       - name: Synchronize Metabase
         working-directory: warehouse
-        run: uv run dbt-metabase models -v --manifest-path=target/manifest.json --skip-sources --exclude-schemas=staging,payments --docs-url="${{ env.DBT_DOCS_URL }}" --metabase-url="${{ env.METABASE_URL }}" --metabase-database="${{ env.METABASE_DESTINATION_DATABASE }}" --metabase-api-key="${{ env.METABASE_API_KEY}}"
+        env:
+          PAYMENTS_DBS: ${{ steps.payments-databases.outputs.names }}
+        run: |
+          set -euo pipefail
+
+          sync_metabase_metadata() {
+            local db="$1"
+            echo "::group::dbt-metabase sync — ${db}"
+            uv run dbt-metabase models -v \
+              --manifest-path=target/manifest.json \
+              --skip-sources \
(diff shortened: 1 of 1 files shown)


lauriemerrell (changes requested):
Might be misunderstanding something here, but I don't think that as written this will actually address the issue of agencies being able to see the docs because they're looking at different database instances that still aren't covered here (it's not just that `mart_payments` is being skipped for the `Data Marts (formerly Warehouse Views)` database connection, it's that the Payments database connections aren't having docs pushed at all)

.github/workflows/deploy-dbt.yml
```diff
@@ -26,58 +26,11 @@ env:
   CALITP_BUCKET__PUBLISH: ${{ github.ref == 'refs/heads/main' && 'calitp-publish' || 'calitp-staging-publish' }}
   GOOGLE_CLOUD_PROJECT: ${{ github.ref == 'refs/heads/main' && 'cal-itp-data-infra' || 'cal-itp-data-infra-staging' }}
   METABASE_DESTINATION_DATABASE: ${{ github.ref == 'refs/heads/main' && 'Data Marts (formerly Warehouse Views)' || '(Internal) Staging Warehouse Views' }}
```
lauriemerrell, line 28: I don't think that the current alone will address the ticket because I think part of the issue is that we need the docs to sync for all the Payments databases, not just `Data Marts (formerly Warehouse Views)`... I think we need to iterate through all the Payments databases and sync docs for all of them? I.e. we need to handle multiple `METABASE_DESTINATION_DATABASE` values

lauriemerrell (dismissed):
Ope sorry meant to submit this inline comment alongside topline comment above

lauriemerrell (commented):
Changes LGTM! Would recommend an approval from @mjumbewu since he was more in the weeds on the mechanics here

.github/workflows/deploy-dbt.yml
```diff
@@ -217,9 +218,79 @@ jobs:
...
+        run: |
+          set -euo pipefail
+
+          sync_metabase_metadata() {
+            local db="$1"
+            echo "::group::dbt-metabase sync — ${db}"
+            uv run dbt-metabase models -v \
+              --manifest-path=target/manifest.json \
+              --skip-sources \
+              --exclude-schemas=staging,payments \
```
lauriemerrell: One question -- not sure if this action supports an `--include-schemas` instead of just `exclude` and if so if we should only include the `mart_payments` schema for the Payments connections to be conservative

jamalk23: Confirmed `--include-schemas` is supported in dbt-metabase 1.6.1 — pushed 3dce6061991d0909bda036a17eee953d48008990 that uses an allowlist for the per-agency loop while keeping the exclude approach on the central DB. Will review with @mjumbewu at tomorrow's meeting.

erikamov (approved):

jamalk23:
Post merge follow up:
- Metabase sync was not skipped, and it looks like docs are now in prod Metabase.

Issue #5143, opened 2026-04-23
Bug: Byte order mark preventing correct processing of OCTA stops.txt


lauriemerrell:
**Describe the bug**
_Please enter a clear and concise description of the bug._

@stevenschrayer reported this bug

The current OCTA GTFS schedule feed is showing as having all null `stop_id` values, seemingly because of a special character in the `stop_id` column name. 

Inspect file here: `gs://calitp-gtfs-schedule-parsed-hourly/stops/dt=2026-04-23/ts=2026-04-23T03:00:01.253275+00:00/base64_url=aHR0cHM6Ly9vY3RhLm5ldC9jdXJyZW50L2dvb2dsZV90cmFuc2l0LnppcA==/stops-001.jsonl.gz` and we see that the stop ID column name is rendering as `\ufeff\"stop_id\"`:

<img width="653" height="197" alt="Image" src="https://github.com/user-attachments/assets/3920fb49-2830-472a-bf55-15da08d7d966" />

**To Reproduce**
_Please enter step by step instructions to reproduce the behavior. For example: Query column A for organization Y on date Z._

Shows count of null stop IDs:

```sql
select _feed_valid_from, countif(stop_id is null) as ct_null, count(*) as ct_total
from `cal-itp-data-infra.mart_gtfs.dim_stops`
where base64_url='aHR0cHM6Ly9vY3RhLm5ldC9jdXJyZW50L2dvb2dsZV90cmFuc2l0LnppcA=='
group by 1
order by 1 desc
```

**Expected behavior**
_Please enter a clear and concise description of what you expected to happen._

Should parse this file and ignore the byte order mark 

**Additional context**
_Please enter any other context about the problem here._

Assume we need to add something in here: https://github.com/cal-itp/data-infra/blob/main/airflow/plugins/operators/gtfs_csv_to_jsonl_operator.py to ignore these

**Acceptance criteria:**
* Update the parse job to ignore these
* Re-process this OCTA feed (looks like it's the 2/17 feed)
* Re-run relevant tables in dbt (can probably just re-run 2/17 in GTFS schedule dim, then need to figure what RT tables might need to get re-run) -- missing stop ID is consequential

erikamov:
@ohrite  and I can take a look on that.

lauriemerrell:
Thanks @erikamov , I thought maybe @stevenschrayer should take a look because he has more availability right now? But defer to y'all

stevenschrayer:
[...]

PR #5184, opened 2026-04-29
Payments(Enghouse): create new parse operator and associated changes

### LINKED ISSUE #5198: Payments(Enghouse): Create parse step in pipeline for improved handling and resiliency
As a data engineer I want the Enghouse data pipeline to handle schema changes in the files that we receive from Enghouse gracefully, and use these process changes to include hive partitioning and table enrichment with columns populated with data from pipeline runs

charlie-costanzo:
# Description

This PR adds a CSV-to-JSONL parse step to the Enghouse payments pipeline to make it resilient to schema changes from Enghouse, who have changed their file format (column renames, drops) without advance notice in the past.

Previously, Enghouse data was ingested via fixed-position CSV external tables directly from the SFTP drop bucket. Any column rename or addition would silently corrupt or break downstream models. JSONL encodes column identity per-record, so schema changes are handled gracefully.

**What changed:**

- New `EnghouseCSVToJSONLOperator` scans all CSV files in the raw bucket (`cal-itp-data-infra-enghouse-raw`), converts them to gzipped JSONL, and writes to a new parsed bucket (`calitp-enghouse-parsed`) using `agency={agency}/dt={date}/` Hive partition structure. Files already present in the parsed bucket are skipped, making the operator safe to re-run. To re-parse a specific file, delete it from the parsed bucket and re-trigger the DAG.
- New `parse_enghouse` Gusty DAG (daily at 14:00 UTC, no catchup) runs the operator. The scan-all approach means the first run after merge will automatically pick up all historical files.
- All four external table configs (`taps`, `transactions`, `ticket_results`, `pay_windows`) updated from CSV to `NEWLINE_DELIMITED_JSON`, pointing at the parsed bucket with CUSTOM hive partitioning on `agency` (STRING) and `dt` (DATE).
- All four staging models updated to expose `agency` and `dt` as columns. Deduplication now uses `ORDER BY dt ASC` to keep the earliest delivery of any duplicate content.
- `stg_enghouse__taps` updated to reference `gps_longitude` instead of `ID` following an upstream column rename from Enghouse.
- Enghouse onboarding runbook and troubleshooting guide updated to document the parse step.

**Dependencies (already merged):**

- GCS buckets PR — `calitp-enghouse-parsed` and `calitp-staging-enghouse-parsed` buckets created
- Composer env var PR — `CALITP_BUCKET__ENGHOUSE_PARSED` added to both prod and staging `environment.tf`

Resolves #5198

## Type of change

- [x] New feature
- [x] Documentation

## How has this been tested?

`parse_enghouse`
Screenshot 2026-04-30 at 14 48 13

`create_external_tables`
Screenshot 2026-04-30 at 14 49 08

`uv run dbt run -s +fct_payments_rides_enghouse`
Screenshot 2026-04-30 at 14 53 46

Also lots of spot checking in the parsed data, external tables, staging tables, and mart tables.

## Post-merge follow-ups

First prod run: after merge, `parse_enghouse` must run before `create_external_tables`, which must run before any dbt runs that touch Enghouse models. The first parse run will process all historical files automatically.


airflow/plugins/operators/enghouse_csv_to_jsonl_operator.py
```diff
@@ -0,0 +1,148 @@
...
+                    self.log.warning("Unexpected path structure, skipping: %s", path)
+                    stats["errors"] += 1
+                    continue
+
+                _, agency, filename = parts
+                date = _date_from_filename(filename, execution_date)
+                stem = os.path.splitext(filename)[0]
+                dest_path = f"{entity}/agency={agency}/dt={date}/{stem}.jsonl.gz"
+
+                if self._hook().exists(
```
lauriemerrell, line 114: question about this -- we just had to re-parse some old OCTA data for #5143 -- will this prevent re-parsing no matter what? are we sure that's desirable?

lauriemerrell (approved):
I have a question about stopping re-parsing but defer to you on how to handle.

Also, seems like it might be preferable if it parsed a specific upstream file? Like if the Airflow date had a specific correspondence to the date of the file name? Just worried this might be kind of brittle if any kind of re-run/fix/backfill is needed. Might be cleaner if there was a more specific correspondence between the Airflow job and the files to be processed.

In that case might want to set the start date back to the data start and let it backfill when it merges? 

But, defer to you on this, you are definitely more familiar with it all than I am

charlie-costanzo:
Hey yeah good questions @lauriemerrell. Let me know what you think about the responses below, I'm more than happy to reconsider if you think a different approach is worthwhile

> Also, seems like it might be preferable if it parsed a specific upstream file? Like if the Airflow date had a specific correspondence to the date of the file name?

Yeah, this will prevent re-parsing a file that's already been written to the parsed bucket, to re-parse you'd need to manually delete the destination file(s) first. That was intentional, the operator is designed to be idempotent across repeated DAG runs without redundant work. For a case like the OCTA re-parse in #5143, you'd delete the relevant files from the parsed bucket and re-trigger the DAG. Source raw files are never touched, so re-parsing is always recoverable.

On date correspondence, I considered date-tied execution but decided against it because we can't guarantee when Enghouse delivers files relative to the parse run time. If a file arrives after the DAG runs for that day, date-tied execution would silently miss it with no automatic recovery. Scan-all ensures any late delivery is picked up on the next run. I guess there are other ways to configure this if it's worthwhile?

> In that case might want to set the start date back to the data start and let it backfill when it merges?

The prod parsed bucket is currently empty, so on the first run after merge the operator will pick up and parse all historical raw files regardless of start date or catchup setting, the skip logic only skips files already in the parsed bucket. Setting the start date back further won't change what gets processed.

lauriemerrell:
@charlie-costanzo ok that's fine with me... Might be worth commenting or documenting somewhere the idea of deleting parsed data to allow re-parse just because I think different of our operators handle this differently

charlie-costanzo:
> @charlie-costanzo ok that's fine with me... Might be worth commenting or documenting somewhere the idea of deleting parsed data to allow re-parse just because I think different of our operators handle this differently

Yeah good call I updated `airflow/dags/README.md` and `airflow/dags/parse_enghouse/README.md`

PR #5191, opened 2026-04-30
Redirect HTTP to HTTPS on JupyterHub GKE Ingress

themightychris:
## Summary

Plain HTTP to `jupyterhub.dds.dot.ca.gov` is currently served (200) instead of redirected. Adds a GKE `FrontendConfig` with `redirectToHttps.enabled: true` and references it from the Ingress, so GCLB issues a 301 from `http://` to `https://`.

`responseCodeName: MOVED_PERMANENTLY` (HTTP 301) — appropriate since we want browsers and bookmarks to remember the upgrade.

## Test plan

- [ ] CI passes
- [ ] After merge & deploy, `curl -I http://jupyterhub.dds.dot.ca.gov/` returns 301 with `Location: https://jupyterhub.dds.dot.ca.gov/`
- [ ] HTTPS continues to work normally

### DIFF
--- a/kubernetes/apps/charts/jupyterhub/templates/gke-frontend-config.yaml
+++ b/kubernetes/apps/charts/jupyterhub/templates/gke-frontend-config.yaml
@@ -1,0 +1,8 @@
+apiVersion: networking.gke.io/v1beta1
+kind: FrontendConfig
+metadata:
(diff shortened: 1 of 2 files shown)


vevetron (approved):
I don't understand but this should be okay.

PR #5092, opened 2026-04-15
Create new incremental macro and handle models at the `dt`/`service_date` incremental interface (#5053)

lauriemerrell:
# Description

_Describe your changes and why you're making them. Please include the context, motivation, and relevant dependencies._

Resolves #5053

* Creates new `ranged_incremental_min_date` and `ranged_incremental_max_date` macros for use in non-microbatch incremental models
* Sets `GTFS_SCHEDULE_START_DATE` to be just 7 days in the past in non-prod environments to limit huge (slow) microbatch runs 
* Renames `PROD_GTFS_RT_START` dbt variable to `GTFS_RT_START` for consistency / clarity (since variable is not just used in prod)
* Removes redundant incremental where clauses on `fct_daily_rt_feed_files`, `fct_trip_updates_no_stop_times`, and `message age` models that I missed when migrating previously
* Sets `full-refresh` to false on dbt manual models to prevent them getting full refreshed by GitHub Actions; per offline conversation with @vevetron **not** setting `full-refresh` to false on all models 
* Updates models to use new macro:

- [x] mart.gtfs.fct_daily_scheduled_stops -- also migrated to `insert_overwrite` strategy 
- [x] mart.gtfs.fct_observed_trips
- [x] mart.gtfs.fct_stop_time_metrics
- [x] mart.gtfs.fct_trip_updates_stop_metrics
- [x] mart.gtfs.fct_trip_updates_trip_metrics
- [x] mart.gtfs.fct_trip_updates_trip_summaries
- [x] mart.gtfs.fct_vehicle_locations_path
- [x] intermediate.gtfs.int_gtfs_rt__vehicle_positions_trip_stop_day_map_grouping -- adds an extra day of buffer for the service date filter on the upstream stops model 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

## How has this been tested?

_Include commands/logs/screenshots as relevant._

_If making changes to dbt models, make sure they were created or update on Staging. Please run the command `uv run dbt run -s CHANGED_MODEL --target staging` and `uv run dbt test -s CHANGED_MODEL --target staging`, then include the output in this section of the PR._

GitHub Action confirms dbt models run. Ran and spot checked locally. 

## Post-merge follow-ups

[...]

Issue #4865, opened 2026-03-04
Epic: Adopt microbatch incremental strategy


lauriemerrell:
## User story / feature request

_Please describe your need, outlining the key users, the feature being requested, and the goal that that the feature will facilitate. For example: **As a \[user or stakeholder type\], I want \[software feature\] so that \[some business value\]**_

As a warehouse maintainer, I want to simplify our use of incremental models to reduce the code complexity associated with them and reduce the risk of adding new incremental models or changing existing ones.

### Acceptance Criteria

_Please enter something that can be verified to show that this user story is satisfied. For example:  **I can join table X with table Y.** or **Column A appears in table Z in Metabase.**_

Migrate existing incremental models to use the microbatch incremental strategy. 

### Notes

_Please enter any additional information that will facilitate the completion of this ticket. For example: Are there any constraints not mentioned above? Are there any alternatives you have considered?_

See [per-model tracking notes here](https://caltrans.sharepoint.com/:x:/s/DOTPMPHQ-DDSContractors/IQBQuvE9wwP6S724KS-Zb3NsAQ2oCAC3T0gfE0g91JPD3P0?e=1cDtIC) (Caltrans Sharepoint)

See [general notes from migration process here](https://caltrans.sharepoint.com/:f:/s/DOTPMPHQ-DDSContractors/IgD9o-1P4maJSrdpCZhqAwINAc7KnxZts9UTek4pxkDgyWQ?e=krhIRf) (Caltrans Sharepoint)

lauriemerrell:
Per meeting yesterday 4/13/26 with @tiffanychu90, @vevetron, @jiaxiliu95, @fsalemi, and @vivekdandugula83 -- consensus is to go with option 2b.

Rationale & notes:
* Well-said by @tiffanychu90 -- "It's like we need to take a right turn, option 3 feels like saying 'don't turn left' and option 2 is just saying 'turn right'" (i.e., option 2 is clearer / more explicit).
* Still want to simplify the incremental where macro -- think this can be achieved pretty easily. Ideally can take a date range. 
* Want to use a non-microbatch strategy for all models that are partitioned on `service_date` -- i.e., just have a cutover point in the DAG where `dt`-based models are microbatch and non-`dt`-based use another incremental strategy.

Issue #4839, opened 2026-02-25
Build a validator for those TIDES files


HaroldBooker:

erikamov:
In conversation with Vivek, this is low priority for now.

vevetron:
I don't like how the validator is a ton of python - it'd be an unmaintainable mess in its current format.  Open to new ideas but low priority for now.

Issue #4700, opened 2026-01-15
Outside Developers have access to a bucket that contains Cal-ITP data


vevetron:
## User story / feature request

As an outside developer, I might want to request access to GTFS and GTFS-RT data from Caltrans. Since this might be a significant amount of data and and downloading, there should be a way to get this data where the user pays for data delivery costs.

This is terraform work, probably a combination of a separate bucket + a type of service account that allows for user pays access to our data.

It might need to also conform to CDT rules if possible. If not, we should discuss further.

This ticket is separate from the workflow that produces such data.
### Acceptance Criteria
We can invite (or they can invite themselves) and use data in a specific bucket but they pay for access.

### Notes

Hermosa Beach is making requests for raw GTFS-RT data. In the past we've offered samples in Tides format at the calitp analysis bucket which is public.

ohrite:
We've revisited this with @vevetron and we're implementing a requester-pays public bucket for the TIDES dataset.

PR #5216, opened 2026-05-01
Add fct_tides_vehicle_locations dbt model (closes #4837)

### LINKED ISSUE #4837: Build a process to convert GTFS-RT Vehicle Positions data > TIDES Vehicle Locations using Hermosa Beach data
The source info is in dim_provider_gtfs_data where is public_customer_facing_or_regional_subfeed_fixed_route
Files saved in a new Bucket

chrisyamas:
# Description

Resolves #4837

Adds `mart_tides.fct_tides_vehicle_locations`, the first TIDES-conformant model in the warehouse. Reshapes `fct_vehicle_locations` into the TIDES `vehicle_locations` schema and filters to public, customer-facing or regional-subfeed fixed-route GTFS-RT feeds via `dim_provider_gtfs_data`. The model produces the BigQuery table only; per-agency parquet export (#4693), CDN-fronted public bucket (#4700), and file validator (#4839) are tracked separately.

A few design decisions worth flagging:

- Materialized as `incremental` with `incremental_strategy='microbatch'`, `event_time='dt'`, `batch_size='day'`, `lookback=var('DBT_ALL_INCREMENTAL_LOOKBACK_DAYS')`, `begin=var('GTFS_RT_START')`, `partition_by` on `dt`, and `cluster_by='base64_url'`. Aligns with the team's [4/13 policy on dt-vs-service_date materialization](https://github.com/cal-itp/data-infra/issues/4865#issuecomment-4245166778) and PR #5092 ranged macros. Mirrors [int_gtfs_rt__vehicle_positions_trip_day_map_grouping](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/intermediate/gtfs/int_gtfs_rt__vehicle_positions_trip_day_map_grouping.sql) line-for-line. Tagged `tides_product`.
- `fct_vehicle_locations` drops NULL `trip_id` rows upstream, so deadhead and layover pings are not in the export. TIDES doesn't require `trip_id_performed`, so a future change could source from `fct_vehicle_positions_messages` to keep them.
- `dim_provider_gtfs_data` records multiple organization rows per VP feed when a feed is shared across agencies (govcbus.com is shared by 7 cities; the SD MTS feed is shared with the airport). Org info isn't part of the TIDES spec and isn't carried through; org metadata for the publish flow lives separately.
- `fct_vehicle_locations.key` is documented as "almost unique" upstream. The model adds a defensive `QUALIFY ROW_NUMBER` per partition with `_extract_ts DESC` as the tie-breaker.
- Bounds-enforcement CASEs (lat/lon, heading, speed, odometer) are kept and tagged with a brief comment grounding them in TIDES schema compliance and noting upstream RT data quality belongs to the TDQ team.

Scope is narrowed to the Hermosa Beach traversing operator set (Beach Cities Transit, LA Metro Bus, Torrance Transit) via the `tides_publication_keys` seed. Phase 2 expansion is additive on the seed.

TIDES = Transit Integrated Data Exchange Specification, https://tides-transit.org/main/.

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

```bash
uv run dbt seed -s tides_publication_keys --target staging
uv run dbt run -s +fct_tides_vehicle_locations --target staging --full-refresh
uv run dbt test -s fct_tides_vehicle_locations --target staging
```

Materialized in `cal-itp-data-infra-staging.christopher_mart_tides.fct_tides_vehicle_locations`. Row counts under the Hermosa-narrow seed match the prior view-materialized run (the materialization change reshapes storage, not the data):

- ~18.3M rows over the rolling 24-day window across the three feeds in scope
- Zero NULL on `location_ping_id`, `event_timestamp`, `vehicle_id`, `trip_id_performed`
- Zero violations on TIDES bounds (lat/lon, heading, speed, odometer, trip_stop_sequence >= 1)
- `current_status` enum mapping correct (no raw GTFS-RT values leak through)

## Post-merge follow-ups

- [ ] No action required

- First incremental run will backfill from `GTFS_RT_START` per the microbatch `begin=` setting. `full_refresh` is disabled per team microbatch convention; coordinate with team on any narrowed initial-load via `--event-time-start` / `--event-time-end` flags if needed.

Two follow-up PRs ready locally and stacked on this one:

- `feat/tides-trips-performed` (#5220) - second TIDES table sourced from `fct_observed_trips`
- `feat/tides-publishing` (#5229) and `feat/tides-publishing-infra` (#5230) - publish DAG, operators, and staging GCS bucket
- `chore/tides-validation-harness` (#5219) - Frictionless validator under `validation/tides/` (closes part of #4839)


vevetron:
regional-subfeed fixed-route feeds. Closes #4837. - remove from model text

vevetron:
- should be it's own grouping of queries, not in the "mart" space.  maybe - mart_tides?
- I think it should be a view and not materialized.  Access pattern is likely to be incredibly infrequent and storing the processed data for this is probably unncecessary.

chrisyamas:
@vevetron thanks, addressing your three comments above and the two follow-ups from the call (HAVING clauses on Slack, the in-code TIDES `#22` / `#252` reference on screen-share). I just force-pushed commit which:
- moved `Closes #4837` out of the model description and lives in the PR body now. Moved the model out of `mart_gtfs` into its own `mart/tides/` folder with a `mart_tides` schema, so it materializes in `<user>_mart_tides` rather than `<user>_mart_gtfs`. Same destination for the two stacked follow-up PRs (`chore/tides-validation-harness` and `feat/tides-trips-performed`), which carry the same shape and will go up for review once this lands
- converted from incremental microbatch to a view, agreed on the access-pattern reasoning. Dropped `partition_by`, `cluster_by`, `event_time`, `batch_size`, `begin`, `lookback`, `full_refresh`, and `on_schema_change` config keys that don't apply to a view
- added a model-level `meta: { publish.product: tides }` in the new yml, going with that key to fit the existing `publish.*` / `ckan.*` dotted-namespace convention I see on the CKAN-published models (e.g., `dim_gtfs_datasets_latest`). Happy to switch to a literal dbt `tags: ['tides_product']` instead if you'd prefer — I noticed the warehouse doesn't currently use dbt `tags:` anywhere, which is why I went with the meta key
- for the HAVING clauses (re: your Slack follow-up), I refactored the `public_subfeed_agencies` CTE from `ANY_VALUE(... HAVING MIN organization_name)` + `GROUP BY 1` to `QUALIFY ROW_NUMBER() OVER (PARTITION BY ... ORDER BY organization_name ASC) = 1`. So this should match the pattern used elsewhere in the warehouse for picking one canonical row per group?
- for the in-code TIDES issue reference you flagged on screen-share, dropped the parenthetical and replaced it with the full URL `https://github.com/TIDES-transit/TIDES/issues/252` in both the SQL comment and the yml description

warehouse/models/mart/tides/fct_tides_vehicle_locations.sql
```diff
@@ -0,0 +1,112 @@
...
+),
+
+tides_vehicle_locations AS (
+    SELECT
+        vp.key AS location_ping_id,
+        vp.service_date,
+        DATETIME(vp.location_timestamp, vp.schedule_feed_timezone) AS event_timestamp,
+
+        vp.trip_id AS trip_id_performed,
+        -- trip_id_scheduled left NULL for MVP; deriving requires a reliable
```
lauriemerrell, line 61: not sure if this is true -- [per GTFS spec](https://gtfs.org/documentation/realtime/reference/#message-tripdescriptor), `trip_id` in VP should reference schedule unless the `schedule_relationship` is one of a few specific values.... so maybe this should be based on that?

also, this is in the intent of the `trip_instance_key` identifier (to allow joins of a specific trip across feed types), so can use that for lookup if desired

warehouse/models/mart/tides/fct_tides_vehicle_locations.sql
```diff
@@ -0,0 +1,112 @@
...
+        vp.base64_url,
+        vp.gtfs_dataset_key,
+        agencies.organization_name,
+        agencies.organization_ntd_id
+    FROM source_vehicle_locations AS vp
+    INNER JOIN public_subfeed_agencies AS agencies
+        USING (gtfs_dataset_key)
+),
+
+-- TIDES requires location_ping_id strictly unique; the upstream key is
```
lauriemerrell: noting that `base64_url` is part of `key` so if two results have same value they need to have same URL... so that won't actually be a substantive tie break

warehouse/models/mart/tides/_mart_tides.yml
```diff
@@ -0,0 +1,123 @@
...
+
+models:
+  - name: fct_tides_vehicle_locations
+    description: |
+      Vehicle location pings reshaped to the TIDES (Transit Integrated Data
+      Exchange Specification) `vehicle_locations` schema
+      (https://tides-transit.org/main/). Sourced from `fct_vehicle_locations`
+      and filtered via `dim_provider_gtfs_data` to public, customer-facing or
+      regional-subfeed fixed-route feeds.
+    meta:
```
lauriemerrell: don't think we use `publish.product` anywhere else, what is intent for that as meta? can/should we be defining a dbt exposure? that is our standard for published items. see for example https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/gtfs_schedule_latest/_gtfs_schedule_latest.yml#L1270-L1314 and https://github.com/cal-itp/data-infra/blob/main/airflow/dags/publish_gtfs.py and https://github.com/cal-itp/data-infra/blob/main/airflow/plugins/operators/dbt_manifest_to_metadata_operator.py#L88-L90 for the GTFS --> CKAN publish flow

warehouse/models/mart/tides/fct_tides_vehicle_locations.sql
```diff
@@ -0,0 +1,112 @@
+{{ config(materialized='view') }}
+
+WITH source_vehicle_locations AS (
+    SELECT *
+    FROM {{ ref('fct_vehicle_locations') }}
+),
+
+-- dim_provider_gtfs_data fans out: a single vehicle_positions feed can be
```
lauriemerrell: I am a little confused by the logic here -- my instinct would be to either group by VP URL where it meets these criteria (public facing etc.) and `array_agg` the organization info so that all orgs can be used later in the publish process OR just select distinct on the non-org columns and ignore the organization parts.

Basically, not sure how ending up with a VP feed tagged with one of its organizations meets future needs -- if we need orgs, we should keep all of them and handle unnesting or whatever in the publish process. If we don't need all the orgs then let's just drop and publish under the VP URL and publish org related metadata separately.

lauriemerrell (commented):
A couple comments, think broad strokes look ok.

erikamov:
`Hermosa Beach` was selected to be the first candidate to share their data in previous meetings. For other agencies they would need to check if they agree to share the data. So we would need to filter the results to generate only for specific agencies.
@evansiroky and @vevetron, should we keep filtering only for `Hermosa Beach` and ignore this `customer_facing is FALSE`?

evansiroky:
We were working with the City of Hermosa Beach which is interested in studying on-time-performance. I believe the thought was to narrow the initial output to just the agencies that traverse Hermosa Beach for development purposes only to check on costs and implementation before expanding statewide.

erikamov:
> We were working with the City of Hermosa Beach which is interested in studying on-time-performance. I believe the thought was to narrow the initial output to just the agencies that traverse Hermosa Beach for development purposes only to check on costs and implementation before expanding statewide.

Yeah, it is what I remember too. :)

chrisyamas:
thanks both, the scope has been narrowed! went ahead and implemented it as MVP this morning rather than wait for our sync since the seed-based mechanism is small enough that landing it gives us something concrete to react to.

scope is now three feeds via a new `tides_publication_keys` seed:

- Beach Cities Transit (Hermosa local operator, BCT JPA): `9edf45e373638700ca420b1e588efdaf`
- LA Metro Bus (south bay routes 102, 130, 232, 344): `1745f7d9b9fa48cdbc8ea282e60602bd`
- Torrance Transit (Swiftly feed): `46b00e5c738a0ebf93522371d9899627`

filter is an INNER JOIN on the seed inside both `fct_tides_vehicle_locations` and `fct_tides_trips_performed`. the existing `public_customer_facing_or_regional_subfeed_fixed_route` filter stays in place; the seed is additive narrowing on top of it. row count drops to 18.3M (from ~96M) on the 9-day vehicle_locations window and 99,960 (from 590,253) on the 8-day trips_performed window.

PR description on this one is updated to match. PR 5220 (trips_performed) inherits the same seed; PRs 5229 and 5230 (the publishing pipeline + staging bucket, drafts up now) inherit the narrowing automatically since they consume the views.

one open question worth flagging for our sync today is whether it would be good to formalize a `tides_publication_consent` flag on `dim_provider_gtfs_data` long-term (defaulting to the existing public-customer-facing flag, overridable per agency), or keep the seed as the publication-list mechanism going forward. seed is the lightest mvp; flag would be more durable. happy to take either direction.

warehouse/models/mart/tides/fct_tides_vehicle_locations.sql
chrisyamas: good catch, agreed. per the GTFS-RT spec the VP `trip_id` does reference the schedule whenever `trip.schedule_relationship` is one of the in-schedule values, so deriving `trip_id_scheduled` from that conditional is the right shape. trip_instance_key is also viable as the join key.

leaving this PR's `trip_id_scheduled` as NULL for MVP and filing a follow-up to wire up the conditional + the join. happy to bump that into scope here if you'd prefer it land together.

warehouse/models/mart/tides/fct_tides_vehicle_locations.sql
chrisyamas: right, both ordering columns were degenerate at this grain since location_timestamp and base64_url are components of the upstream `key`. fixed in a fixup: pulled the dedup up to the source CTE and ordered by `_extract_ts DESC` (most-recently-extracted wins), which differs across the duplicates. trailing `deduped` CTE collapsed into the source. verified 0 dups on a sampled service_date in staging.

warehouse/models/mart/tides/_mart_tides.yml
chrisyamas: good call, exposure is the right shape. dropping the `publish.product` meta in this PR. adding a single `california_tides` exposure as part of PR 5220 (so both `fct_tides_vehicle_locations` and `fct_tides_trips_performed` ref()s resolve in the same checkout), modeled on the GTFS `california_open_data` block. owner / methodology fields filled in; `meta.destinations` now filled in via PR 5229 (publishing pipeline).

warehouse/models/mart/tides/fct_tides_vehicle_locations.sql
chrisyamas: yeah agreed the fan-out collapse is doing more than it should. going with your second framing: dropping `organization_name` / `organization_ntd_id` from both fact tables since orgs aren't part of the TIDES spec anyway.

the `public_subfeed_agencies` CTE shrinks to a `public_subfeed_keys` CTE that's just SELECT DISTINCT on dataset keys matching the public-customer-facing-or-regional-subfeed criterion, no QUALIFY needed. govcbus / SD MTS multi-org reality moves to publish-side metadata in PR 5229, separate from the TIDES tables themselves.

chrisyamas:
@ohrite @erikamov, quick note on the bounds-enforcement CASE statements (latitude / longitude / heading / speed / odometer) on `fct_tides_vehicle_locations`.

Given the 5/5 walkthrough on defensive coding belonging upstream with the TDQ team, I considered stripping the CASEs and letting raw upstream values through. The challenge is that the TIDES schema requires bounds-conformant rows for spec validation; out-of-range values fail Frictionless on the published parquet, not just on the warehouse table. So nulling spec-violators in this view preserves the row while dropping the unusable value, which keeps the downstream parquet validation green without modifying upstream `fct_vehicle_locations`.

I added a brief comment above the CASEs grounding them in TIDES spec compliance and pointing to TDQ as the upstream RT data quality owner. If you'd rather move bounds-checking to the validator step (post-export), I'm happy to flip that. The trade-off is the validator would then surface row-level rejects instead of the warehouse silently passing the spec-conformant subset, which depending on the dashboard story might be the right answer.

Let me know which way you'd prefer; either is straightforward to switch.

warehouse/seeds/tides_publication_keys.csv
lauriemerrell, line 1: Just noting that `key` is a versioned identifier so this will only ever pull this version of the GTFS dataset. If you want the persistent identifier this should be `source_record_id` which might be a two step lookup (not on the RT data itself)

warehouse/models/mart/tides/fct_tides_vehicle_locations.sql
```diff
@@ -0,0 +1,125 @@
...
+    ) = 1
+),
+
+-- Filter to feeds tagged public-customer-facing or regional-subfeed
+-- fixed-route. Org info isn't part of the TIDES spec and isn't carried
+-- through; org metadata for the publish flow lives separately.
+public_subfeed_keys AS (
+    SELECT DISTINCT vehicle_positions_gtfs_dataset_key AS gtfs_dataset_key
+    FROM {{ ref('dim_provider_gtfs_data') }}
+    WHERE _is_current = TRUE
```
lauriemerrell, line 42: Not totally sure about this -- if you were to re-run a historical window for this model (if it were transitioned to microbatch) you wouldn't want this `_is_current` flag).... Instead this should join based on the record's valid to/valid from dates corresponding to the `dt` of the RT data

lauriemerrell (approved):
Approving given limited runway but would at least call out follow up tickets for the items I mentioned, particularly microbatch. This model does not conform to our overall incremental model approach

warehouse/models/mart/tides/fct_tides_vehicle_locations.sql
```diff
@@ -0,0 +1,125 @@
...
+-- Upstream `key` is "almost unique" (documented unique_proportion at_least
+-- 0.999); TIDES requires location_ping_id strictly unique. `key` is composed
+-- of base64_url + location_timestamp + vehicle_id + ..., so those columns are
+-- all constant within a duplicate set and can't tie-break. `_extract_ts` is
+-- the upstream extract timestamp and differs across the duplicates, so pick
+-- the most-recently-extracted row per key.
+WITH source_vehicle_locations AS (
+    SELECT *
+    FROM {{ ref('fct_vehicle_locations') }}
+    {% if is_incremental() %}
```
lauriemerrell: Sorry meant to have a comment that this should be microbatch incremental strategy because both it and its immediate parent are partitioned on `dt`. See: https://github.com/cal-itp/data-infra/issues/4865#issuecomment-4245166778

chrisyamas:
@vevetron heads up, after digging into the team's incremental conventions saw the [4/13 policy on dt-vs-service_date materialization](https://github.com/cal-itp/data-infra/issues/4865#issuecomment-4245166778) and PR #5092's ranged macros. switching this one to microbatch to align with the dt-partition standard (matches [int_gtfs_rt__vehicle_positions_trip_day_map_grouping](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/intermediate/gtfs/int_gtfs_rt__vehicle_positions_trip_day_map_grouping.sql)). 5220 follows the service_date-partitioned shape with ranged macros + a +1 on the max date when sourcing from this model, per the pattern in `int_gtfs_rt__vehicle_positions_stops_by_service_date` (PR #5237).

PR #5219, opened 2026-05-02
Add Frictionless TIDES validation harness

chrisyamas:
# Description

Resolves part of #4839.

Adds a Python validator under `validation/tides/` that exports one feed's slice of a Cal-ITP TIDES warehouse table to local parquet, validates it against the official TIDES JSON schema via `frictionless`, and emits a JSON Lines outcome record. Lives outside the dbt project on purpose: this is a tooling concern, not a warehouse model.

The outcome envelope mirrors the shape used by `download_parse_and_validate_gtfs.py` so TIDES validation rows can land alongside GTFS validation rows in the same external-table layout. The pattern came out of the 2026-05-05 walkthrough with Doc and Erika.

The validator covers what the spec defines for `vehicle_locations`: required-not-null, type, bounds, enum, and primary-key uniqueness. Cross-table foreign-key checks (vehicle_locations -> trips_performed) follow once `fct_tides_trips_performed` lands.

Stacked on `feat/tides-vehicle-locations`. Rebases to `main` after #5216 merges.

## Outcome record shape

Each outcome line is one JSON object with eight top-level fields, identical in shape to the GTFS pipeline:

```jsonc
{
  "success": true,
  "exception": null,
  "step": "validate",
  "extract": {
    "filename": "<base64-of-feed-url>",
    "ts": "...",
    "config": {
      "feed_type": "tides_vehicle_locations",
      "name": "Beach Cities Transit",
      "url": "...",
      "service_date": "2026-04-30",
      "gtfs_dataset_key": "...",
      "warehouse_table": "fct_tides_vehicle_locations",
      "tides_schema_url": "..."
    },
    ...
  },
  "header": {
    "row_count": 12345,
    "feed_count": 1,
    "first_event_timestamp": "...",
    "last_event_timestamp": "...",
    "frictionless_valid": true,
    "error_counts_by_type": {}
  },
  "aggregation": {
    "filename": "validate_<base64>.jsonl.gz",
    "step": "validate",
    "first_extract": { ...recursive... }
  },
  "blob_path": "gs://calitp-tides-staging/validation_outcomes/...",
  "process_stderr": null
}
```

The `aggregation.filename` and the GCS blob name embed the base64 of the feed URL, matching the join-key convention used across the GTFS pipeline.

`run_validation(...)` is exposed as the library entry point; #5229 wraps it in `ValidateTIDESToGCSOperator` and slots that into the publish DAG.

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

```bash
cd validation/tides
uv sync

uv run validate_tides.py \
  --dataset christopher_mart_tides \
  --table fct_tides_vehicle_locations \
  --schema <path-to-your-TIDES-clone>/spec/vehicle_locations.schema.json \
  --service-date 2026-04-30 \
  --gtfs-dataset-key 9edf45e373638700ca420b1e588efdaf \
  --feed-name "Beach Cities Transit" \
  --feed-url "warehouse://mart_tides/fct_tides_vehicle_locations/9edf45e373638700ca420b1e588efdaf"
```

The outcome file at `exports/validate_<base64>.jsonl` carries one JSON line per validated feed, conforming to the eight-field envelope described above.

## Post-merge follow-ups

- [ ] No action required

- Cross-table FK checks via a Frictionless data-package descriptor follow once `fct_tides_trips_performed` lands.
- Production usage is via the operator wrapper landing in #5229; this PR is the standalone CLI for sandbox runs and the shared library entry point.


validation/tides/README.md
```diff
@@ -0,0 +1,58 @@
+# TIDES validation harness
+
+Local validator that exports a sample of a Cal-ITP TIDES table to parquet and runs Frictionless validation against the official TIDES JSON schema. Used to confirm Cal-ITP's TIDES output conforms to the open spec before publishing files to the public open-data bucket.
+
+Lives outside the dbt project on purpose: this is a tooling concern, not a warehouse model.
+
+## Setup
+
+```bash
+python -m venv .venv
```
lauriemerrell: Can this use `uv` like our other venvs? For consistency

validation/tides/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Usage
+
+Validate vehicle_locations for one service date in your sandbox:
+
+```bash
+python validate_tides.py \
+  --dataset christopher_mart_tides \
+  --table fct_tides_vehicle_locations \
+  --schema <path-to-your-TIDES-clone>/spec/vehicle_locations.schema.json \
```
lauriemerrell, line 36: This should be mentioned in setup, how is this populated?

validation/tides/validate_tides.py
```diff
@@ -0,0 +1,233 @@
...
+    """Run Frictionless validation of a parquet file against a TIDES schema."""
+    resource = Resource(
+        path=str(parquet_path),
+        schema=schema,
+        format="parquet",
+    )
+    return validate(resource)
+
+
+def render_report(report, table_name: str, service_date: str, row_count: int) -> str:
```
lauriemerrell: curious about intent here -- if this is running in an Airflow job, do we want this to output structured data that can be ingested into BQ (e.g., JSON, CSV, etc.) instead? Not sure what the workflow would be if this is emitting human readable .txt files (useful for local testing but not sure how these get used in the production flow)

lauriemerrell (commented):
Curious about the performance on this -- how long does it take to run for a day + agency? 

Also wondering about the path to refactor this for use in Airflow -- should this just be written as an Airflow operator from jump?

chrisyamas:
on performance, two timings against staging on the TIDES `vehicle_locations` view, service_date 2026-04-29:

- single agency (BART), 7,279 rows: BQ extract 4.4s, frictionless validate 5.9s, total 10.3s
- single agency (LA Metro, largest), 2,442,485 rows: BQ extract 592s, frictionless validate 12s, total 604s

frictionless validation against parquet stays cheap regardless of size; BQ extract through the REST endpoint dominates on big agencies and doesn't scale linearly. installing google-cloud-bigquery-storage and switching to the BQ Storage API path would shrink the extract time materially when that matters.

on the Airflow-operator question: agreed it makes sense to wrap this once the public-bucket publishing flow lands. PR 5229 (the publish DAG) shapes the operator surface, and the validator naturally slots in either as a downstream operator on the same DAG or as a post-export check before promote-to-public. keeping it as a local-dev script in this PR; refactor falls out of PR 5229.

validation/tides/README.md
chrisyamas: done, switched to uv. dropped requirements.txt, added a pyproject.toml + uv.lock under `validation/tides/` matching the warehouse / airflow / gtfs-rt-archiver shape. README setup is now `uv sync`.

validation/tides/README.md
chrisyamas: documented in a Prerequisites section: clone the upstream `https://github.com/TIDES-transit/TIDES.git` repo somewhere local, then point `--schema` at `<TIDES-clone>/spec/<table>.schema.json`. pinned to the spec's main branch; bump the local clone with `git pull` to pick up new schema versions.

validation/tides/validate_tides.py
chrisyamas: intent for this PR is local-dev: a human-readable .txt report so a developer can eyeball pass/fail and see the error breakdown when iterating on a TIDES model. the structured-output path makes sense for the Airflow operator version that PR 5229 unlocks. when the operator wraps this, swapping `render_report` for a JSON / CSV emitter that lands in BQ is a small change.

leaving the .txt as the local-dev surface here; the structured emitter slots in alongside (or replaces) it when the operator lands.

chrisyamas:
@erikamov heads up, the validator now emits JSON Lines outcomes instead of the human-readable text report it had before. envelope matches the 8-field shape from `download_parse_and_validate_gtfs.py` (`success`, `exception`, `step`, `extract`, `header`, `aggregation`, `blob_path`, `process_stderr`), and `aggregation.filename` plus the GCS object name both embed the base64 of the feed URL so the records join cleanly against the schedule and rt outcomes.

`run_validation(...)` is the library entry point now, and PR 5229 wraps it as `ValidateTIDESToGCSOperator` slotted after the parquet export in the publish DAG. the `header` block carries row count, the `frictionless_valid` flag, and `error_counts_by_type` so a Metabase dashboard can summarize without re-reading the underlying report.

standalone CLI is the sandbox path with `--outcome-output` and `--blob-path` flags so the same code runs the same way locally. happy to rename fields or shuffle keys around if the layout doesn't match what TDQ is expecting.

PR #5220, opened 2026-05-02
Add fct_tides_trips_performed dbt model

chrisyamas:
# Description

Adds `mart_tides.fct_tides_trips_performed`, sourced from `fct_observed_trips` and joined to `fct_scheduled_trips` for route metadata and to `fct_tides_vehicle_locations` for canonical vehicle_id per trip. Filtered to public, customer-facing or regional-subfeed fixed-route GTFS feeds via `dim_provider_gtfs_data`.

Stacked on `feat/tides-vehicle-locations`. Lives alongside `fct_tides_vehicle_locations` in the new `mart/tides/` folder. Includes a `relationships` test asserting that every `vehicle_id` here resolves to at least one row in `fct_tides_vehicle_locations`. Also adds the `california_tides` exposure on `_mart_tides.yml`.

A few design decisions worth flagging:

- Materialized as `incremental` with `incremental_strategy='insert_overwrite'`, `partition_by` on `service_date`, and `cluster_by=['service_date', 'base64_url']`, mirroring [fct_observed_trips](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/gtfs/fct_observed_trips.sql). Tagged `tides_product`. Per the team's [4/13 policy on dt-vs-service_date materialization](https://github.com/cal-itp/data-infra/issues/4865#issuecomment-4245166778) and PR #5092, the incremental window uses `ranged_incremental_min_date` / `ranged_incremental_max_date` macros on every upstream ref. When sourcing from `fct_tides_vehicle_locations` (microbatch on `dt` per #5216), the upper bound on `dt` is offset by `+1` day to ensure overlap with `service_date`, per the pattern in `int_gtfs_rt__vehicle_positions_stops_by_service_date` (PR #5237).
- Filtered to `appeared_in_vp = TRUE` upstream so every row has a derivable vehicle_id. Trips that only appeared in trip_updates (no VP) are excluded.
- Each trip is assigned its most-frequent vehicle_id during VP coverage (`APPROX_TOP_COUNT(vehicle_id, 1)`). Vehicles that change mid-trip get the dominant one.
- `schedule_trip_start` / `schedule_trip_end` cast to DATETIME using `feed_timezone` from `fct_scheduled_trips`. Falls back to `America/Los_Angeles` when the feed timezone is NULL, defensible default for California, flag if you'd rather a different fallback.
- Cross-feed dedup partition is `(service_date, trip_id_performed, gtfs_dataset_key)` so trips that share `trip_id_performed` across different feeds both survive.

Scope is narrowed to the Hermosa Beach traversing operator set (Beach Cities Transit, LA Metro Bus, Torrance Transit) via the `tides_publication_keys` seed inherited from #5216. Row count post-narrowing: ~100k rows over the 8-day window (down from ~590k unnarrowed).

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

```bash
uv run dbt run -s +fct_tides_trips_performed --target staging --full-refresh
uv run dbt test -s fct_tides_trips_performed --target staging
```

Materialized in `cal-itp-data-infra-staging.christopher_mart_tides.fct_tides_trips_performed`. Row counts under the Hermosa-narrow seed match the prior view-materialized run (the materialization change reshapes storage, not the data):

- ~100k rows over the 8-day window across the three feeds in scope
- Zero PK duplicates on `(service_date, trip_id_performed, gtfs_dataset_key)`
- Zero NULL on `service_date`, `trip_id_performed`, `vehicle_id`
- `schedule_relationship` distribution matches the unnarrowed prototype (Scheduled / NULL / Canceled / Added / Duplicated)

## Post-merge follow-ups

- [ ] No action required

- Initial deployment populates per the standard ranged-incremental lookback (`DBT_ALL_INCREMENTAL_LOOKBACK_DAYS`, default 5 days). To narrow or extend the initial load, coordinate with the team on adjusting `GTFS_RT_START` or running with explicit date filters.
- Update the Frictionless validation harness to include trips_performed schema validation.
- Open follow-up issue for `stop_visits` model and for `trip_start_stop_id` / `trip_end_stop_id` derivation (deferred past MVP, requires stop_times join).

### DIFF
--- a/warehouse/models/mart/tides/_mart_tides.yml
+++ b/warehouse/models/mart/tides/_mart_tides.yml
@@ -112,3 +112,126 @@ models:
         description: '{{ doc("column_base64_url") }}'
       - name: gtfs_dataset_key
         description: '{{ doc("gtfs_schedule_gtfs_dataset_key") }}'
+
+  - name: fct_tides_trips_performed
+    description: |
+      Trips performed reshaped to the TIDES (Transit Integrated Data Exchange
+      Specification) `trips_performed` schema (https://tides-transit.org/main/).
+      Sourced from `fct_observed_trips`, joined to `fct_scheduled_trips` for
+      route metadata and to `fct_tides_vehicle_locations` for canonical
+      vehicle_id per trip. Filtered upstream to `appeared_in_vp = TRUE` so
+      every row has a derivable vehicle_id.
+    columns:
+      - name: service_date
+        description: TIDES PK (with trip_id_performed). From `fct_observed_trips.service_date`.
+        data_tests:
+          - not_null:
+              arguments:
+                config:
+                  where: 'service_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 2 DAY)'
+      - name: trip_id_performed
+        description: TIDES PK (with service_date). From `fct_observed_trips.trip_id`.
+        data_tests:
+          - not_null:
+              arguments:
+                config:
+                  where: 'service_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 2 DAY)'
+      - name: vehicle_id
+        description: |
+          Most-frequent vehicle_id observed in `fct_tides_vehicle_locations`
+          for the same (service_date, trip_id_performed). TIDES required.
+        data_tests:
+          - not_null:
+              arguments:
+                config:
+                  where: 'service_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 2 DAY)'
+          - relationships:
+              arguments:
+                to: ref('fct_tides_vehicle_locations')
+                field: vehicle_id
+                config:
+                  where: 'service_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 2 DAY)'
+      - name: trip_id_scheduled
+        description: |
+          Coarse MVP: equals trip_id_performed when the trip appeared in VP
+          or TU. A stricter join to `fct_scheduled_trips` is a follow-up.
+      - name: route_id
+        description: From `fct_scheduled_trips` via trip_instance_key. NULL when no schedule.
+      - name: route_type
+        description: GTFS route_type. From `fct_scheduled_trips`.
+      - name: ntd_mode
+        description: NULL in MVP.
+      - name: route_type_agency
+        description: NULL in MVP.
+      - name: shape_id
+        description: From `fct_scheduled_trips`.
+      - name: pattern_id
+        description: NULL in MVP.
+      - name: direction_id
+        description: From `fct_scheduled_trips`.
+      - name: operator_id
+        description: NULL; not derivable from GTFS-RT.
+      - name: block_id
+        description: From `fct_scheduled_trips`.
+      - name: trip_start_stop_id
+        description: NULL in MVP; needs stop_times join on min stop_sequence.
+      - name: trip_end_stop_id
+        description: NULL in MVP; needs stop_times join on max stop_sequence.
+      - name: schedule_trip_start
+        description: |
+          DATETIME in agency tz from `fct_scheduled_trips.trip_first_departure_ts`.
+      - name: schedule_trip_end
+        description: |
+          DATETIME in agency tz from `fct_scheduled_trips.trip_last_arrival_ts`.
+      - name: actual_trip_start
+        description: DATETIME in agency tz from `fct_observed_trips.vp_min_ts`.
+      - name: actual_trip_end
+        description: DATETIME in agency tz from `fct_observed_trips.vp_max_ts`.
+      - name: trip_type
+        description: |
+          Constant 'In service' in MVP because the model filters to VP-observed
+          trips. Future expansion could distinguish deadhead / layover / pull-in
+          / pull-out given supplemental signals.
+        data_tests:
+          - accepted_values:
+              arguments:
+                values: ['In service', 'Deadhead', 'Layover', 'Pullout', 'Pullin', 'Extra Pullout', 'Extra Pullin', 'Deadhead To Layover', 'Deadhead From Layover', 'Other not in service']
+                config:
+                  where: 'service_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 2 DAY)'
+      - name: schedule_relationship
+        description: |
+          TIDES enum (Scheduled / Added / Unscheduled / Canceled / Duplicated)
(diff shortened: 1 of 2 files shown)


warehouse/models/mart/tides/fct_tides_trips_performed.sql
```diff
@@ -0,0 +1,119 @@
...
+    SELECT
+        service_date,
+        trip_id_performed,
+        APPROX_TOP_COUNT(vehicle_id, 1)[OFFSET(0)].value AS vehicle_id
+    FROM {{ ref('fct_tides_vehicle_locations') }}
+    WHERE vehicle_id IS NOT NULL
+    GROUP BY 1, 2
+),
+
+-- Same shared-feed agency collapse as fct_tides_vehicle_locations.
```
lauriemerrell: same comment as there re: org logic

warehouse/models/mart/tides/fct_tides_trips_performed.sql
```diff
@@ -0,0 +1,119 @@
...
+        ON o.vp_gtfs_dataset_key = a.gtfs_dataset_key
+    LEFT JOIN scheduled s
+        ON s.trip_instance_key = o.trip_instance_key
+    LEFT JOIN vehicle_per_trip v
+        ON v.service_date = o.service_date
+       AND v.trip_id_performed = o.trip_id
+),
+
+-- TIDES requires (service_date, trip_id_performed) unique. fct_observed_trips
+-- can have multiple rows per PK when the same trip appears in multiple feeds;
```
lauriemerrell: This should only dedupe within a given feed (VP URL or VP GTFS dataset), don't dedupe across feeds. There is no reason to expect that trip ID performed should be unique across feeds and dropping like this will result in trips basically arbitrarily missing from some feeds just because they used the same ID as another feed. TIDES isn't really designed for this cross-agency use-case (all IDs within TIDES should only be assumed to be unique within a given feed, not across feeds/agencies) and honestly maybe this is something that needs to get surfaced at the TIDES Spec level -- should there be an agency or feed identifier in this table for cross-agency use cases?

warehouse/models/mart/tides/_mart_tides.yml
```diff
@@ -121,3 +121,111 @@ models:
...
+
+  - name: fct_tides_trips_performed
+    description: |
+      Trips performed reshaped to the TIDES (Transit Integrated Data Exchange
+      Specification) `trips_performed` schema (https://tides-transit.org/main/).
+      Sourced from `fct_observed_trips`, joined to `fct_scheduled_trips` for
+      route metadata and to `fct_tides_vehicle_locations` for canonical
+      vehicle_id per trip. Filtered upstream to `appeared_in_vp = TRUE` so
+      every row has a derivable vehicle_id.
+    meta:
```
lauriemerrell: same comment I think this should be an exposure

warehouse/models/mart/tides/fct_tides_trips_performed.sql
chrisyamas: addressed alongside L4 on PR 5216: dropping `organization_name` / `organization_ntd_id` from this table too. the `public_subfeed_agencies` CTE shrinks to a `public_subfeed_keys` SELECT DISTINCT, no QUALIFY. orgs aren't part of the TIDES spec; metadata moves to publish-side in PR 5229.

warehouse/models/mart/tides/fct_tides_trips_performed.sql
chrisyamas: right, the partition was too coarse. fixed: added `vehicle_positions_gtfs_dataset_key` to the partition so dedup happens within feed, not across feeds. trips that share `trip_id_performed` across different feeds now both survive. base64_url tie-breaker dropped from the ORDER BY (redundant once partitioned by feed).

re-materialized with the fix on an 8-day window (2026-04-23 to 2026-04-30): 590,253 rows after, 562,378 distinct (service_date, trip_id_performed) pairs under the old grain. ~27,875 trips that were getting collapsed across feeds now survive as their own rows.

agreed on your TIDES-spec point: cross-feed uniqueness is not something the spec guarantees, so the assumption "trip_id_performed is unique within feed + service_date only" is the right framing. worth a thread in TIDES-community channels separately; happy to take that on.

warehouse/models/mart/tides/_mart_tides.yml
chrisyamas: added a single `california_tides` exposure on this PR (so both fct_tides_vehicle_locations and fct_tides_trips_performed refs resolve in the same checkout), modeled on the GTFS `california_open_data` exposure. publish.product meta block dropped from both yml entries. PR 5229 fills in `meta.destinations` for the public-bucket flow.

warehouse/models/mart/tides/fct_tides_trips_performed.sql
```diff
@@ -0,0 +1,132 @@
...
+        tags=['tides_product'],
+    )
+}}
+
+WITH observed AS (
+    SELECT *
+    FROM {{ ref('fct_observed_trips') }}
+    -- Drop TU-only trips (no VP) so every row has a derivable vehicle_id.
+    WHERE appeared_in_vp = TRUE
+    {% if is_incremental() %}
```
lauriemerrell: Since `fct_observed_trips` is not partitioned on `dt`, this should use our incremental ranged date macros as shown here: https://github.com/cal-itp/data-infra/pull/5092 so that it can have historical periods re-run in accordance with other models, etc.

lauriemerrell (approved):
Again approving for runway but ideally the incremental logic would follow warehouse pattern -- can defer to a follow up ticket if needed

chrisyamas:
@vevetron heads up, aligning this one with the team's [4/13 policy on dt-vs-service_date materialization](https://github.com/cal-itp/data-infra/issues/4865#issuecomment-4245166778) and PR #5092's ranged macros. service_date-partitioned + insert_overwrite stays; manual lookback replaced by `ranged_incremental_min_date` / `ranged_incremental_max_date` on every upstream ref (mirrors [fct_observed_trips](https://github.com/cal-itp/data-infra/blob/main/warehouse/models/mart/gtfs/fct_observed_trips.sql)). For the lookup against `fct_tides_vehicle_locations` (now microbatch on dt per #5216), filtering on dt with a +1 on the max date to handle the dt-vs-service_date overlap, per the pattern in `int_gtfs_rt__vehicle_positions_stops_by_service_date` (PR #5237).

Issue #5137, opened 2026-04-22
Bug: `int_gtfs_rt__vehicle_positions_trip_stop_day_map_grouping` has inconsistent date coverage due to incremental logic 


lauriemerrell:
**Describe the bug**
_Please enter a clear and concise description of the bug._

I think that `int_gtfs_rt__vehicle_positions_trip_stop_day_map_grouping` has incorrect handling in its initial ([this commit](https://github.com/cal-itp/data-infra/blob/131b34bf5a3268275c9347f182ec0bafa0efc011/warehouse/models/intermediate/gtfs/int_gtfs_rt__vehicle_positions_trip_stop_day_map_grouping.sql)) incremental logic such that it has inconsistent date coverage.

Here is the plot of what you get from this query (zoomed to last couple weeks):

```sql
select 
  dt,
  count(*)
from `cal-itp-data-infra.staging.int_gtfs_rt__vehicle_positions_trip_stop_day_map_grouping`
group by 1
order by 1 desc;
```

<img width="852" height="251" alt="Image" src="https://github.com/user-attachments/assets/d3fefb32-63c6-4843-8a09-481191a8c5af" />

Tuesdays, Wednesdays, and Fridays are higher than Mondays and Thursdays. 

I believe that this is happening because this model [uses the `insert_overwrite` incremental strategy ](https://docs.getdbt.com/docs/build/incremental-strategy?version=1.12#insert_overwrite) and filters the upstream stops model on `service_date` while filtering the upstream RT data on `dt` and then doing an inner join on the result.

What this means:

For the incremental run on 4/20, the model is filtering the upstream stops model on `service_date >= 4/16` and the upstream RT data on `dt >= 4/16`, doing an inner join on the results, and then dropping and replacing all affected `dt` partitions in the target model. This means that the `dt = 4/16` partition of `int_gtfs_rt__vehicle_positions_trip_stop_day_map_grouping` contains only the overlap between the service date and the UTC dt, which means it's missing 1/3 of the data for that day.

This issue specifically affects Mondays & Thursdays because that's when dbt_all runs so that's when the partition boundaries are strictest. The intervening days are fine because all the data for a `dt` is caught because multiple `service_date` values are included. 

**To Reproduce**
_Please enter step by step instructions to reproduce the behavior. For example: Query column A for organization Y on date Z._

See query above. 

**Expected behavior**
_Please enter a clear and concise description of what you expected to happen._

Should account for this boundary and update filters & joins accordingly.

[...]

PR #5237, opened 2026-05-05
Fix: Incremental logic for VP trip/stop/day grouping

lauriemerrell:
# Description

Additional follow up on #5137 post merge of #5092, which may not have fully fixed the issue -- see https://github.com/cal-itp/data-infra/issues/5137#issuecomment-4382256492

Leaving this for @tiffanychu90 to consult and ultimately take over the line. Recommended approach (partially implemented here, but not tested or fully validated):

* Rename this model -- it is not actually a `dt/service_date` mapping model, rename to model purpose
* As far as I can tell, the dependencies on the two VP trip models are not needed? It seems like we can join directly from schedule to vehicle locations using `trip_instance_key`?
* Related, it seems like the grain of this model is in fact just service_date, so it should be treated as such -- partition by service_date. I *think* that loading a couple models by `dt` should be ok in this fashion as noted in comments.
* This adds a new technique that we haven't used before but may be helpful in these large models going forward: pull a list of partition values using dbt's `call_statement` function and pass that to filter certain models (in this case stop times grouped, which is partitioned by `_feed_valid_from`, which we can't target effectively with a simple where statement/incremental logic.) We can't just do subqueries (`where (select thing from whatever)`) for these because BQ doesn't do partition elimination in those cases -- you need to pre-construct the list and pass it explicitly. 
* Adds a new start date variable in `dbt_project` just for these RT stop analysis models to limit how much history gets run for them while they're under construction. 

Remaining actions that would be needed to validate this approach:
* Compare results with what's in prod right now -- does this return equivalent information
* Run incrementally and confirm that we don't see variations at the date boundary between runs (i.e., should see pretty constant data volumes across all weekdays -- do some exploration to check the full universe of expected trips and make sure they are all present.) 
* If merged, delete the existing model.

In general, approach is: 
* Design the model you want (what columns), dependencies
* Then figure out how to make it incremental by constructing intentional batches to be inserted. You need to ensure that, when a date range is being processed, the affected date partitions will be covered in their entirety by the date ranges selected from the upstream parent models. So, if your batch has ANY data from a given service_date based on the selection criteria, you need to make sure that ALL data for that service_date will be present, because the partitions for all service date values will be dropped and overwritten based on what is present in the incoming batches. The one exception is the current_date() when the model is being run, which should be re-run in the subsequent dbt run and it's ok if it has one partial load as long as it gets a complete load on the next run. 
  * Note: As discussed offline, there is some challenge for `service_date` based models because there is sometimes late-arriving data for these (e.g., data for service_date 4/24 that arrives on 5/1, outside the lookback window). It could be possible to add a model that constructs dt/service_date combinations and then selects those (probably using the `call_statement` pattern) and passes those to filter the incoming models. However, the late-arriving data is not large in volume. Recommend that as a future initiative if needed but for now think that the number of affected trips is low enough to let it go.

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

_If making changes to dbt models, make sure they were created or update on Staging. Please run the command `uv run dbt run -s CHANGED_MODEL --target staging` and `uv run dbt test -s CHANGED_MODEL --target staging`, then include the output in this section of the PR._

Has not yet been tested, needs testing & validation

## Post-merge follow-ups

- [ ] No action required

### DIFF
--- a/warehouse/dbt_project.yml
+++ b/warehouse/dbt_project.yml
@@ -26,6 +26,7 @@ vars:
   GTFS_SCHEDULE_START: "{{ '2021-04-16' if target.name == 'prod' else (modules.datetime.datetime.now() - modules.datetime.timedelta(days=7)).strftime('%Y-%m-%d') }}"
   GTFS_RT_START: "{{ '2022-09-15' if target.name == 'prod' else (modules.datetime.datetime.now() - modules.datetime.timedelta(days=7)).strftime('%Y-%m-%d') }}"
   KUBA_DEVICE_START: '2025-08-11'
+  GTFS_RT_STOP_ANALYSIS_START: "{{ '2025-12-31' if target.name == 'prod' else (modules.datetime.datetime.now() - modules.datetime.timedelta(days=7)).strftime('%Y-%m-%d') }}"
   INCREMENTAL_MAX_DT: ''
   DBT_ALL_INCREMENTAL_LOOKBACK_DAYS: 5
   DBT_DAILY_INCREMENTAL_LOOKBACK_DAYS: 1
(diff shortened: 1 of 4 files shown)


erikamov:
I fixed merge conflicts just to keep branch up to date.

Issue #5161, opened 2026-04-24
Parallelize `dbt run` and `dbt compile / dbt docs gen` steps in `deploy dbt` GitHub Action


lauriemerrell:
## User story / feature request

_Please describe your need, outlining the key users, the feature being requested, and the goal that that the feature will facilitate. For example: **As a \[user or stakeholder type\], I want \[software feature\] so that \[some business value\]**_

As an analytics engineer, I want our deploy dbt GitHub Actions to run cleanly, efficiently, and with maximum chances of success.

Currently, if `dbt run` fails on merge, it blocks the `dbt compile --> `tf plan & apply` flow.

In the `deploy dbt` flow, there are two distinct concerns:

1. Getting the updated code into Airflow so that future dbt runs will use the new code
2. Updating the state of BigQuery to have the updated data

`dbt compile, dbt docs generate, tf plan, tf apply` are part of # 2. Running dbt is part of # 1. These can actually run in parallel and separately. Right now # 1 is gating # 2. 

### Acceptance Criteria

_Please enter something that can be verified to show that this user story is satisfied. For example:  **I can join table X with table Y.** or **Column A appears in table Z in Metabase.**_

`dbt run` is split out from `dbt compile`, at least within the internal dependency tree of the `deploy dbt` GHA. Maybe one can be a GH Action called `run dbt` and the other can be the `deploy dbt code` or something. 

### Notes

_Please enter any additional information that will facilitate the completion of this ticket. For example: Are there any constraints not mentioned above? Are there any alternatives you have considered?_

Motivating example is: https://github.com/cal-itp/data-infra/actions/runs/24849936790/job/72764426314. The `dbt run` fails because of a data quota issue, which is actually fine (we don't need to full refresh these models to get the changes, which are config only). But this is blocking the deploy steps to get the new code into Airflow.

PR #5232, opened 2026-05-05
Scope per-agency dbt-metabase sync by each connection's dataset filter

jamalk23:
# Description

Replaces the hardcoded `--include-schemas=mart_payments,mart_benefits` for per-agency Payments DBs with dynamic discovery of each connection's `dataset-filters-type` and `dataset-filters-patterns` from the Metabase API. The sync now passes exactly those patterns to dbt-metabase per agency, fixing the MetabaseStateError on connections (e.g., Humboldt, Lake Transit and some others) whose dataset filter doesn't include mart_benefits. Self-correcting against future per-agency config drift.

Change is being made as a follow up to PR #5164 for issue #5061 after @mrtopsyt found that several payment DBs docs were not synced correctly.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

## How has this been tested?

Opening this PR triggers the Deploy dbt workflow which will synchronize the payments mart schemas.

## Post-merge follow-ups

- [ ] No action required
- [X] Actions required (specified below)

Must confirm that Sync Metabase job is not skipped and causes the doc sync to occur successfully. Specifically, we need to check GitHub Actions logs -> Sync Metabase -> Synchronize Metabase task for any errors with an agency sync.

PR #5257, opened 2026-05-06
Run dbt models job only if there are model changes

erikamov:
# Description

This PR adds back the condition to only run models if there were changes to the job instead of in the command lines.
It reduces the extra workflow running time when there are no dbt model changes.

I also simplifying the Metabase to `needs: [compile, dbt-run]` since now it also needs `dbt-run` that already requires `[setup, dbt-changed]` so we don’t need to repeat it.

@jamalk23 needs `Sync Metabase` to run even without dbt models changes, for example there were a macro, or a documentation change. So I added the check `if: ${{ !failure() && !cancelled() }}` to run for those cases, but not when it fails or was cancelled.

Related to [#5161 and #4929] 

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Running workflow through this PR, it shows skipping the `dbt-run` job but still runs `Sync Metabase`:
image

Before it was running but skipping steps:
image
image

## Post-merge follow-ups

- [x] No action required

### DIFF
--- a/.github/workflows/deploy-dbt.yml
+++ b/.github/workflows/deploy-dbt.yml
@@ -215,6 +215,8 @@ jobs:
     runs-on: ubuntu-latest
     needs: [compile, dbt-run]
 
+    if: ${{ !failure() && !cancelled() }}
+
     permissions:
       contents: read
       id-token: write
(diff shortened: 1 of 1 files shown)


lauriemerrell (approved):

jamalk23:
@erikamov, if we can hold off on merging this until PR #5232 is reviewed/merged, that would ensure that all agency doc syncs occur. This would allow us to close #5061. Otherwise, as I understand it, we will have to wait until the next change in warehouse sql files for us to verify the doc-sync fix in main.

Issue #4786, opened 2026-02-09
Developers have access to dot.ca.gov emails from GCP - secret key


vevetron:
## User story / feature request

Developers in our GCP should be able to send emails from an @dot.ca.gov email address.  This would be helpful for more automated emails.

This ticket focuses on the secret key method.

To do this with Caltrans in their Tenant it could be a long process.  Caltrans has never done this before.

It involves working with [IMD.Enterprise.Services@dot.ca.gov](mailto:IMD.Enterprise.Services@dot.ca.gov)

ACS
	- https://azure.microsoft.com/en-us/products/communication-services/
	
	
	New subscription, new resource group
	
John K - payment licensing

The first time implemented within Caltrans
	- Sending us through the EAC proposal
	- Discuss Cost, billing
Goes through Enterprise Architecture Process
	- Solution
	- ACS to do it.
	- This would be evaluated - tied to something.  An Amendment - excemption - to this.
IT hardware and Software Standards
	- 

Go from GCP Get from SAAS cloud through … 

Mail relay from internal

https://it.onramp.dot.ca.gov/hardware-and-software-standards-exception-form

Requesting the ability to use ACS for our Email

	- Starts an internal discussion
	- Called to a review meeting
	- Give a presentation on why we should review your request

Get everyones buy in.

### Acceptance Criteria

Emails to the world via dot.ca.gov

### Notes

I'm not sure how much value this would provide or all the legitimate use cases where this is necessary over being able to send emails from within the VPN.

jparr:
I found another spot that we send emails via postmark during my security work. 
https://github.com/cal-itp/reports/blob/main/reports/generate_report_emails.py

jparr:
This one isn't on postmark but we should be able to solve for it too: https://github.com/cal-itp/data-infra/blob/main/cloud-functions/update-expired-airtable-issues/main.py

vevetron:
I received a secret key and tested it and put it into data-infra secrets.

PR #5259, opened 2026-05-06
Switch airflow notification to smtp provided by acs

jparr:
ref: #4786


jparr:
tested and confirmed with staging.

ohrite (approved):

PR #5245, opened 2026-05-06
Decommission self-hosted Sentry

themightychris:
> [!WARNING]
> **Do not merge until the Cal-ITP Benefits team confirms they've migrated off the self-hosted Sentry instance.** They're the last remaining user; merging before they're cut over will take their error visibility offline.

> [!NOTE]
> This PR is stacked on top of #5243 (decommission gtfs-rt-archiver-v3-test) and is targeted at that branch. Once #5243 merges, GitHub will automatically retarget this to `main`.

## Summary

Retires the in-cluster Sentry deployment (and the `postgresql-backup-sentry` release that backs up its metadata DB). Also strips the now-unused `SENTRY_DSN` config from the prod archiver overlay so nothing in source references the soon-to-be-gone `sentry.k8s.calitp.jarv.us` URL.

## Changes

Removes 2 release entries from `ci/kubernetes-workloads.yaml`:
- `sentry`
- `postgresql-backup-sentry`

Deletes Sentry's chart wrapper and supporting values:
- `kubernetes/apps/charts/sentry/` (Chart.yaml, values.yaml, .helmignore — the bundled `charts/sentry-20.0.0.tgz` is gitignored and will fall away naturally)
- `kubernetes/apps/values/postgresql-backup-sentry.yaml`

Strips dead Sentry references:
- `SENTRY_DSN` and `SENTRY_ENVIRONMENT` removed from `kubernetes/apps/overlays/gtfs-rt-archiver-v3-prod/archiver-channel-vars.yaml` (archiver is already at 0/0/0 replicas, but the ConfigMap reference shouldn't outlive Sentry itself). The test archiver's equivalent reference goes away with #5243 which deletes that overlay entirely.

Also drops obsolete docs and CI plumbing pointing at the self-hosted Sentry:
- `runbooks/workflow/sentry-triage.md` — every URL, search query, and the GitHub-issue integration in this runbook describes the self-hosted Sentry setup that's being torn down.
- `.github/workflows/sentry-release.yml` + the matching section in `.github/workflows/README.md` — release notification on merge to `main`; the receiving instance is gone.
- `kubernetes/README.md`: drops the now-stale Sentry mention from the workloads description (was already missing Grafana via #5242).

## After-merge cluster cleanup

Source-only change. The actual in-cluster teardown is a manual pass:

```bash
helm uninstall postgresql-backup -n sentry
helm uninstall sentry -n sentry

# Releases ~700+ Gi of PVCs across ClickHouse (3x 200Gi),
[...]

PR #5279, opened 2026-05-07
Remove Sentry SDK plumbing from Airflow / warehouse / gtfs-schedule-validator

themightychris:
Strips remaining references to the self-hosted Sentry instance that was decommissioned in #5245. The SDK calls left behind silently fail (DNS + retries on every DAG parse / RT parse / dbt run), so this is a real ongoing cost as well as a cleanup.

## Scope

**Env vars removed** (all references to a now-dead host):

- `SENTRY_DSN`, `SENTRY_ENVIRONMENT` — `airflow/.{development,staging,production,test}.env`
- `CALITP_BUCKET__SENTRY_EVENTS` — set in 4 places, audit confirms zero readers anywhere in the repo

**Composer Terraform** drops the matching passthroughs in:

- `iac/cal-itp-data-infra/composer/us/environment.tf`
- `iac/cal-itp-data-infra-staging/composer/us/environment.tf`

**Python deps** removed:

- `airflow/pyproject.toml` + `requirements.txt` + `requirements-dev.txt`
- `warehouse/pyproject.toml`
- `jobs/gtfs-schedule-validator/pyproject.toml`

Lockfiles regenerated: `airflow/uv.lock`, `warehouse/uv.lock`, `jobs/gtfs-schedule-validator/poetry.lock` (regenerated via Docker `python:3.9` since this project pins to Python 3.8–3.10).

## Source-code edits (review hot spots)

- **`airflow/plugins/scripts/gtfs_rt_parser.py`**: removes the sentry init, the `MAX_STRING_LENGTH` override, all `capture_exception` calls, the flush at the end of `main`, and the `scope` parameter on `ValidationProcessor.process` / `ParseProcessor.process`. The fingerprinting / `set_context` / `push_scope` blocks were Sentry-only; surrounding `logging.exception`, `outcomes.append`, and the exceptions accumulation + final `RuntimeError` remain unchanged so the parser still records and surfaces failures correctly.
- **`airflow/dags/download_gtfs_schedule_v2/download_schedule_feeds.py`**: drops sentry init + the `push_scope` / `HTTPError` fingerprinting block. Logging and outcomes-append preserved. Removes unused `from requests.exceptions import HTTPError` (only ref was inside the deleted block).
- **`airflow/dags/unzip_and_validate_gtfs_schedule_hourly/unzip_gtfs_schedule.py`**: drops sentry init + push_scope block.
- **`warehouse/scripts/run_and_upload.py`**: removes the `report_failures_to_sentry` function, the `report-failures` Typer subcommand, the 6 `Dbt*Error` classes, all 3 in-process call sites in `run()`, plus now-unused imports (`json`, `gcsfs`, `Manifest`, `RunResults`). The function and CLI command had no callers in `.github/`, `airflow/`, or `warehouse/` (grep-verified) and existed solely to ship dbt failures to Sentry.
- **`jobs/gtfs-schedule-validator/gtfs_schedule_validator_hourly.py`**: drops sentry init + push_scope. The fingerprinting `extract_hash = fs.stat(...)` call inside the `except` was Sentry-only, so the hoisted `fs = get_fs()` it referenced is also removed. An unrelated later `get_fs()` call in the same function is preserved.

## Out of scope (intentionally)

- **`services/gtfs-rt-archiver-v3/**`**: the in-cluster archiver is dormant at 0/0/0 replicas. Its Sentry plumbing is inert and would only matter if the archiver were ever revived. Left for a future "revive" PR if that ever comes up.

## Test plan

- [ ] CI green
- [ ] After merge: confirm Composer DAG-parse logs no longer show DNS retries against `sentry.k8s.calitp.jarv.us`
- [ ] Confirm an RT parser invocation still surfaces validator stderr correctly (no Sentry-side state lost) — easiest spot-check is the next gtfs-rt-validator hourly DAG run

## References

- #4922 (cluster maintenance tracker — see [latest comment](https://github.com/cal-itp/data-infra/issues/4922#issuecomment-4389027663))
- #5245 (Sentry decommission — sets the context for why these calls now fail silently)
- #5276 (parallel doc cleanup PR for the same stream)


tihuang02 (approved):

vevetron (approved):
Ran uv run pytest locally - this might be okay.

vevetron:
Breaks staging currently.  Actually it was fine.

PR #5244, opened 2026-05-06
Decommission ingress-nginx and cert-manager

themightychris:
> [!WARNING]
> **Do not merge until Sentry is decommissioned or migrated to GKE Ingress.** This PR disables the chart-managed nginx Ingress for \`sentry.calitp.org\` along with the system-level ingress controller; merging before Sentry has an alternate path will take its UI offline.

> [!NOTE]
> This PR is stacked on top of #5242 (decommission monitoring stack) and is targeted at that branch. Once #5242 merges, GitHub will automatically retarget this to \`main\`.

## Summary

Removes the system-level ingress controller (ingress-nginx) and certificate issuer (cert-manager) along with the chart-managed nginx-class Ingresses on the workloads that still referenced them. Also rolls in the docs hostname swap and the \`kubernetes/README.md\` updates that reflect the post-ingress-nginx world.

By the time this PR is mergeable:
- JupyterHub has already migrated to GKE Ingress with Google-managed certificates (#5157, #5191, #5192)
- The monitoring stack is gone (#5242)
- Sentry has either been retired or moved to its own GKE Ingress *(this is the merge gate)*

### Removed CI dependencies on nginx-class Ingress (chart values)
- \`kubernetes/apps/charts/jupyterhub/values.yaml\`: \`ingress.enabled: false\`
- \`kubernetes/apps/charts/sentry/values.yaml\`: \`ingress.enabled: false\`

### Removed bootstrap manifest references
- \`kubernetes/system/kustomization.yaml\` (only ever applied at original cluster bootstrap, never reconciled by CI; kept around for documentation)

### Doc updates that go with the ingress changes
- \`docs/**\` and \`kubernetes/README.md\`: hostname swap from \`notebooks.calitp.org\` to \`jupyterhub.dds.dot.ca.gov\` (was previously #5249, rolled in here)
- \`kubernetes/README.md\`: \"Domain Name Changes\" recipe rewritten — referenced \`ingress.hosts\` / \`ingress.tls.hosts\` which this PR removes; new recipe points at \`templates/gke-managed-certificate.yaml\` (\`spec.domains\`) and \`templates/gke-ingress.yaml\` (\`spec.rules[].host\`)
- \`kubernetes/README.md\`: \"System workloads\" wording updated since the cluster no longer runs an in-cluster ingress controller or certificate manager

## After-merge cluster cleanup

Source-only change. The actual cluster teardown is a manual pass:

\`\`\`bash
# ingress-nginx
[...]

PR #5242, opened 2026-05-06
Decommission in-cluster monitoring stack (Grafana/Prometheus/Loki/Promtail)

themightychris:
## Summary

Retires the in-cluster Grafana / Prometheus / Loki / Promtail stack now that Google Managed Service for Prometheus + the Cloud Monitoring dashboard for the GTFS-RT archiver have been running in parallel and proven out. The archiver was the only thing we still relied on this stack for.

## Changes

Removes 5 release entries from \`ci/kubernetes-workloads.yaml\`:
- \`grafana\`
- \`loki\`
- \`postgresql-backup-grafana\`
- \`prometheus\`
- \`promtail\`

Deletes the 6 values files that backed them:
- \`kubernetes/apps/values/grafana.yaml\`
- \`kubernetes/apps/values/grafana-prod.yaml\`
- \`kubernetes/apps/values/loki-prod.yaml\`
- \`kubernetes/apps/values/postgresql-backup-grafana.yaml\`
- \`kubernetes/apps/values/prometheus.yaml\`
- \`kubernetes/apps/values/promtail.yaml\`

(The chart definitions themselves come from the \`jarvus-cluster-template\` holosource and aren't tracked here.)

Also drops the now-stale Grafana mention from the workloads description in \`kubernetes/README.md\`.

## After-merge cluster cleanup

Source-only change — CI will simply stop reconciling these releases. The actual in-cluster resources need a manual cleanup pass:

\`\`\`bash
helm uninstall grafana            -n monitoring-grafana
helm uninstall postgresql-backup  -n monitoring-grafana
helm uninstall prometheus         -n monitoring-prometheus
helm uninstall loki               -n monitoring-loki
helm uninstall promtail           -n monitoring-loki

# PVCs: Grafana PG (dashboard config), Prometheus TSDB, Loki log data —
# all replaceable by Cloud Monitoring / Cloud Logging.
kubectl delete pvc --all -n monitoring-grafana
kubectl delete pvc --all -n monitoring-prometheus
kubectl delete pvc --all -n monitoring-loki

kubectl delete namespace \\
  monitoring-grafana monitoring-prometheus monitoring-loki monitoring
\`\`\`

## What this unblocks

Prerequisite for #5244 (ingress-nginx + cert-manager teardown) — removes the \`monitoring.calitp.org\` Ingress, one of the three remaining nginx-class Ingresses on the cluster.

## Test plan

- [ ] CI passes (kubernetes preview shows the 5 releases dropping from the deploy plan)
[...]

PR #5280, opened 2026-05-07
Hologit + CI cleanup after the cluster decommission stream

themightychris:
Small cleanup of orphan plumbing left behind by the cluster decommission stream (#4922). All targets reference paths that were deleted in earlier PRs in the stream and are now dead glob entries / dead projections.

### Hologit projections — `.holo/branches/kubernetes-workspace/`

- **Deleted** `kubernetes/apps/charts/{grafana,loki,prometheus,promtail}.toml` — these projected the upstream charts from `jarvus-cluster-template` into `kubernetes/apps/charts/`. The in-cluster monitoring stack was decommissioned in #5242, so the destination directories no longer exist and the projections produce nothing useful.
- **Edited** `_data-infra.toml`: drops `kubernetes/system/**` from the local-files glob. That directory was deleted in #5244.

### Workflow path filters

- `.github/workflows/deploy-kubernetes.yml` and `.github/workflows/preview-kubernetes.yml`: remove `kubernetes/system/**` from the `paths:` trigger filter. The directory can never appear in a diff again.

### Empty directory

- **Deleted** `kubernetes/apps/values/.gitignore` (which causes the now-empty `kubernetes/apps/values/` directory to fall away in git). The single line in the ignore (`sentry_sensitive.yaml`) was for a Sentry chart that no longer exists. Every chart that consumed `kubernetes/apps/values/*.yaml` was removed across #5242 / #5245 / #5246.

## Test plan

- [ ] CI green
- [ ] No drift in next preview/deploy run on a real Kubernetes-touching PR

## References

- #4922 (cluster maintenance tracker — see [latest comment](https://github.com/cal-itp/data-infra/issues/4922#issuecomment-4389027663))
- #5242 (monitoring stack decommission — context for the four `.toml` deletions)
- #5244 (ingress-nginx + cert-manager — context for the `kubernetes/system/**` removals)


tihuang02:
is `hologit` still in use? why not completely removing `.holo`?

tihuang02 (approved):
looks good.

tihuang02:
#5458 for re-evaluation of `.holo/`

PR #5287, opened 2026-05-08
Refactor TIDES models for SCD Type 2 historical correctness + persistent source_record_id seed

chrisyamas:
# Description

Follow-up to #5216 and #5220 addressing two compounding SCD-related correctness issues that surfaced during their production deploys, plus a bounded microbatch backfill window.

## What was wrong

1. **Versioned-key seed.** `tides_publication_keys.csv` keyed on `vehicle_positions_gtfs_dataset_key` (the versioned MD5 hash). When operators rotate GTFS feed URLs, the hash changes, so historical `fct_vehicle_locations` rows tagged with old hashes silently fall out of the publication set.

2. **`_is_current = TRUE` on `dim_provider_gtfs_data`.** The prior `public_subfeed_keys` CTE filtered to the CURRENT public-customer-facing flag bound to the CURRENT key. Historical RT rows tagged with old keys couldn't match.

3. **`begin=var('GTFS_RT_START')`** on `fct_tides_vehicle_locations`. On `target.name == 'prod'` that resolves to `'2022-09-15'`, generating 1332 daily microbatches from scratch on first deploy. Most scan upstream daily partitions and filter to 0 rows due to issues 1 + 2 above.

The 1+2 combination was flagged in @lauriemerrell's inline review on #5216 ([discussion_r3204968948](https://github.com/cal-itp/data-infra/pull/5216#discussion_r3204968948), [discussion_r3204978905](https://github.com/cal-itp/data-infra/pull/5216#discussion_r3204978905)). Issue 3 manifested when the post-merge prod deploys were observed running ~90 min on empty batches before being cancelled.

## What this PR does

- Replaces the `_is_current = TRUE` + versioned-key filter with a single **SCD Type 2 join** to `dim_provider_gtfs_data` using `_valid_from` / `_valid_to`. Each VP row's join uses the dim record valid at that row's `_extract_ts` (or `vp_min_ts` for the trips_performed equivalent). Recovers historical correctness for both the public-fixed-route flag and the operator identity.
- Swaps the seed from `gtfs_dataset_key` to `vehicle_positions_source_record_id` (persistent Airtable record IDs from `dim_provider_gtfs_data`). Stable across upstream feed-key rotations.
- Adds a new `TIDES_PRODUCT_START` dbt var (mirrors `KUBA_DEVICE_START` shape) bounding the prod backfill to `'2025-12-16'`, aligning with when current dim records for our 3 publication operators became valid. Both fct models now use this var instead of `GTFS_RT_START`.

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Tested locally (target=dev, 7-day window):

```bash
uv run dbt seed --select tides_publication_keys --target dev
uv run dbt run --select fct_tides_vehicle_locations --target dev
uv run dbt run --select fct_tides_trips_performed --target dev
```

`fct_tides_vehicle_locations` produced **264,250 rows for 2026-05-01 across all 3 Hermosa feeds (1,562 distinct vehicles)**. SCD Type 2 join correctly resolves the dim record valid at each row's `_extract_ts`. Other batches in the local sandbox produced 0 rows because `christopher_mart_gtfs.fct_vehicle_locations` only has May 1 data — the model logic itself is verified.

`fct_tides_trips_performed` compiled and ran cleanly with the symmetric SCD Type 2 join on `vp_min_ts`.

## Post-merge follow-ups

- [ ] No action required

The cancelled prior deploys of #5216 and #5220 left empty partitions in `cal-itp-data-infra.mart_tides.fct_tides_vehicle_locations` (and an empty `fct_tides_trips_performed` if 5220's deploy ran far enough). To get a clean post-merge state, recommend dropping both prod tables before this PR's merge triggers the new deploy:

```sql
DROP TABLE IF EXISTS `cal-itp-data-infra.mart_tides.fct_tides_vehicle_locations`;
DROP TABLE IF EXISTS `cal-itp-data-infra.mart_tides.fct_tides_trips_performed`;
```

If the drop doesn't happen, the deploy will still succeed — just with empty pre-2025-12-16 partitions persisting (cosmetic-only, no impact on query correctness for `dt >= '2025-12-16'`).

cc @vevetron @lauriemerrell

### DIFF
--- a/warehouse/models/mart/tides/fct_tides_trips_performed.sql
+++ b/warehouse/models/mart/tides/fct_tides_trips_performed.sql
@@ -19,7 +19,7 @@ WITH observed AS (
     -- Drop TU-only trips (no VP) so every row has a derivable vehicle_id.
     WHERE appeared_in_vp = TRUE
       AND service_date
-        BETWEEN {{ ranged_incremental_min_date(default_lookback=var("DBT_ALL_INCREMENTAL_LOOKBACK_DAYS"), data_earliest_start=var("GTFS_RT_START")) }}
+        BETWEEN {{ ranged_incremental_min_date(default_lookback=var("DBT_ALL_INCREMENTAL_LOOKBACK_DAYS"), data_earliest_start=var("TIDES_PRODUCT_START")) }}
             AND {{ ranged_incremental_max_date() }}
 ),
 
@@ -36,7 +36,7 @@ scheduled AS (
         trip_last_arrival_ts
     FROM {{ ref('fct_scheduled_trips') }}
     WHERE service_date
-        BETWEEN {{ ranged_incremental_min_date(default_lookback=var("DBT_ALL_INCREMENTAL_LOOKBACK_DAYS"), data_earliest_start=var("GTFS_RT_START")) }}
+        BETWEEN {{ ranged_incremental_min_date(default_lookback=var("DBT_ALL_INCREMENTAL_LOOKBACK_DAYS"), data_earliest_start=var("TIDES_PRODUCT_START")) }}
             AND {{ ranged_incremental_max_date() }}
 ),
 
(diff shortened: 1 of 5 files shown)


warehouse/models/mart/tides/fct_tides_trips_performed.sql
```diff
@@ -50,25 +50,34 @@ vehicle_per_trip AS (
...
-    FROM {{ ref('tides_publication_keys') }}
+-- SCD Type 2 join to dim_provider_gtfs_data: resolve the dim record valid
+-- at vp_min_ts (the earliest VP timestamp for the trip), not the current
+-- state. Filter to the publication-list source_record_ids and to the
+-- public-customer-facing or regional-subfeed fixed-route flag as it stood
+-- when the trip was observed.
+filtered_observed AS (
+    SELECT o.*
+    FROM observed AS o
+    INNER JOIN {{ ref('dim_provider_gtfs_data') }} AS d
```
lauriemerrell: This is probably fine but I might simplify/clarify for readability by filtering `dim_provider_gtfs_data` first (inner join it to the seed in a prior CTE so that only relevant records are present) and then just use that in this inner join rather than this inner join with a subquery on the seed in the where clause.

lauriemerrell (approved):

warehouse/models/mart/tides/fct_tides_trips_performed.sql
chrisyamas: ah okay yes. i applied it to both models for symmetry. pulled the seed-filter + public-fixed-route check into a `publication_dim_records` CTE so the SCD Type 2 join is just the time-window match

Issue #5332, opened 2026-05-20
Bug: Unused Kubernetes Nodes


vevetron:
**Describe the bug**
We have some kubernetes nodes still running and doing nothing since we took down the archiver and sentry and the old way of running airflow

https://console.cloud.google/kubernetes/clusters/details/us-west1/data-infra-apps/details?project=cal-itp-data-infra&csesidx=1253265761

vevetron:
<img width="1861" height="1005" alt="Image" src="https://github.com/user-attachments/assets/295cedae-35f1-47c3-907a-b72868a55489" />

vevetron:
Related - https://github.com/cal-itp/data-infra/pull/5245, and PRs that removed the archiver.

vevetron:
```
⏺ Bash(gcloud compute instances list --project=cal-itp-data-infra --filter="name~gke-data-infra-apps"
      --format="table(name,zone,status,creationTimestamp)" 2>&1)
  ⎿  NAME                                                ZONE        STATUS   CREATION_TIMESTAMP
     gke-data-infra-apps-apps-v2-2729c0c0-z7fe           us-west1-a  RUNNING  2026-05-17T03:05:48.434-07:00
     gke-data-infra-apps-jupyterhub-users-b57e08f4-gxky  us-west1-a  RUNNING  2026-05-16T05:12:22.816-07:00
     … +7 lines (ctrl+o to expand)

⏺ Zero nodes in jobs-v1 and gtfsrt-v4 right now.

⏺ Bash(gcloud logging read 'resource.type="gce_instance" AND
      (protoPayload.methodName="v1.compute.instances.insert" OR
      protoPayload.methodName="v1.compute.instances.de…)
  ⎿  TIMESTAMP                    METHOD_NAME                  RESOURCE_NAME
     2026-05-20T23:38:12.986810Z  v1.compute.instances.delete  projects/1005246706141/zones/us-west1-a/insta
     nces/gke-data-infra-apps-jobs-v1-cd18666b-arx4
     … +16 lines (ctrl+o to expand)

⏺ Yes — node pools have spun down and stayed down.

  No scale-ups in the last 2 days, so they've stayed at 0.
```

I can delete in a week or two.

PR #5333, opened 2026-05-20
Shrink unused kubernetes node pools - scale jobs-v1 and gtfsrt-v4 node pool floors to 0

vevetron:
# Description

Drops `min_node_count` from `1` → `0` on the `jobs-v1` and `gtfsrt-v4` node pools in the `data-infra-apps` cluster (us-west1). Both pools are orphaned — Composer Airflow KPO jobs now target the `airflow-jobs` autopilot cluster in us-west2 (`POD_CLUSTER_NAME=airflow-jobs`, `POD_LOCATION=us-west2`), so nothing schedules onto these pools. `max_node_count` is preserved (2 and 3 respectively) so autoscale-up still works if anything ever lands on them again.

Works toward #5332.  After confirmed dead we can delete them.

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Ran `terraform plan` locally in `iac/cal-itp-data-infra/gke/us/`. Plan showed only the two expected `min_node_count` diffs on `jobs-v1` and `gtfsrt-v4`, no other drift.

## Post-merge follow-ups

- [ ] No action required

After the apply workflow runs, verify the pools drain:

- `kubectl get nodes -l cloud.google.com/gke-nodepool=jobs-v1` → 0 nodes within ~10 min
- `kubectl get nodes -l cloud.google.com/gke-nodepool=gtfsrt-v4` → 0 nodes within ~10 min
- Spot-check Composer DAGs for any failures referencing these pools

### DIFF
--- a/iac/cal-itp-data-infra/gke/us/container_node_pool.tf
+++ b/iac/cal-itp-data-infra/gke/us/container_node_pool.tf
@@ -65,7 +65,7 @@ resource "google_container_node_pool" "tfer--data-infra-apps_gtfsrt-v4" {
   autoscaling {
     location_policy      = "BALANCED"
     max_node_count       = "2"
-    min_node_count       = "1"
+    min_node_count       = "0"
     total_max_node_count = "0"
     total_min_node_count = "0"
   }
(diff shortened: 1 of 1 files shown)


tihuang02 (approved):
looks good

PR #5339, opened 2026-05-22
Pin urllib3<2.0 to fix Composer 3 gzip decode crash

vevetron:
# Description

Composer 3's image (`composer-3-airflow-2.10.5-build.36`) ships a Google-patched
urllib3 whose `_GzipDecoder.decompress()` does not accept the `max_length` kwarg
that urllib3's own `_decode()` passes in. This crashes every GCS download of a
gzip-compressed response — including `manifest.json` reads via `GCSHook.download()`,
which is the first task in `publish_gtfs` (`dbt_manifest_to_metadata`). The DAG
aborts before any downstream CKAN task runs.

Pinning `urllib3<2.0` resolves to `1.26.20`, which Composer leaves alone instead
of overlaying its broken patched build. The pin is applied in all three files
Composer/uv read:

- `airflow/requirements.txt` — consumed by `iac/.../composer/us/variables.tf` (Composer 2)
- `airflow/requirements-c3.txt` — consumed by the same module for Composer 3
- `airflow/pyproject.toml` — local dev via `uv`

### Why `<2.0` and not a newer bound

| Constraint | Resolved | Tests passed | Tests failed |
|---|---|---|---|
| `urllib3<2.0` (this PR) | 1.26.20 | 113 | 5 (pre-existing, #5338) |
| `urllib3<2.6` | 2.5.0 | 93 | 25 (**+20**) |
| `urllib3>=2.7` | 2.7.0 | 88 | 25 (**+20**) |

urllib3 2.x is broadly incompatible with our current dependency chain
(ckanapi, requests, soda hook, ntd xlsx hook, airtable, BigQuery, mobility
database, manifest operators). Moving to 2.x is a separate, larger effort —
tracked under the dependency epic #5231; this PR documents `urllib3<2.0` as a
known constraint there.

Resolves #5334

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

## How has this been tested?

**Local Composer 3 environment** (built from `composer-3-airflow-2.10.5-build.36`):

1. Added `urllib3<2.0` to `requirements-dev.txt`
2. `make teardown && make setup && make sync && make start`
3. `docker exec ... pip show urllib3` → **1.26.20** (downgraded from C3's broken build)
4. Triggered `publish_gtfs` via Airflow API
5. `dbt_manifest_to_metadata`: **success** (previously crashed on C3 staging)
[...]

Issue #5334, opened 2026-05-21
Bug: ckan api push is failing in composer 3


vevetron:
**Describe the bug**
^^^^^^^^^^^^^
  File "/opt/python3.11/lib/python3.11/site-packages/urllib3/response.py", line 642, in _decode
    data = self._decoder.decompress(data, max_length=max_length)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: _GzipDecoder.decompress() got an unexpected keyword argument 'max_length'
[2026-05-19, 01:22:02 UTC] {taskinstance.py:1227} INFO - Marking task as FAILED. dag_id=publish_gtfs, task_id=dbt_manifest_to_metadata, run_id=scheduled__2026-05-12T00:00:00+00:00, execution_date=20260512T000000, start_date=20260519T012201, end_date=20260519T012202
[2026-05-19, 01:22:02 UTC] {connectionpool.py:868} WARNING - Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError("HTTPSConnection(host='sentry.k8s.calitp.jarv.us', port=443): Failed to establish a new connection: [Errno 111] Connection refused")': /api/2/envelope/
[2026-05-19, 01:22:02 UTC] {connectionpool.py:868} WARNING - Retrying (Retry(total=1, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError("HTTPSConnection(host='sentry.k8s.calitp.jarv.us', port=443): Failed to establish a new connection: [Errno 111] Connection refused")': /api/2/envelope/
[2026-05-19, 01:22:02 UTC] {base.py:84} INFO - Retrieving connection 'smtp_acs'
[2026-05-19, 01:22:02 UTC] {email.py:272} INFO - Email alerting: attempt 1
[2026-05-19, 01:22:02 UTC] {connectionpool.py:868} WARNING - Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError("HTTPSConnection(host='sentry.k8s.calitp.jarv.us', port=443): Failed to establish a new connection: [Errno 111] Connection refused")': /api/2/envelope/
[2026-05-19, 01:22:03 UTC] {email.py:283} INFO - Sent an alert email to ['dds.app.notify+staging@dot.ca.gov']
[...]

PR #5349, opened 2026-05-28
Revert urllib3<2.0 pin that breaks the Composer 3 staging deploy

vevetron:
# Description

Reverts #5339 (commit 2c60aa9e6). That PR pinned `urllib3<2.0` in `airflow/requirements-c3.txt`, but the pin is unsatisfiable on Composer 3: its image (`composer-3-airflow-2.10.5-build.36`) ships `apache-airflow==2.10.5+composer`, which requires `urllib3>=2.0`. The in-cluster PyPI build fails to resolve, so the staging deploy errors out:

> Cannot install apache-airflow==2.10.5+composer and urllib3<2.0 because these package versions have conflicting dependencies.

The pin passed local testing because `composer-dev` uses public `apache-airflow==2.10.5` (no urllib3>=2.0 constraint) and reads `requirements-dev.txt`, not `requirements-c3.txt` — so the conflict only surfaces in the real Composer 3 build.

This reverts to unblock deploys. The original gzip decode crash is back open in #5334 and needs a fix that works within urllib3 2.x.

Works toward #5334

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Commit-for-commit revert of #5339; no new logic. The conflict it removes is proven by the failed staging deploy ([run 26591265897](https://github.com/cal-itp/data-infra/actions/runs/26591265897/job/78350452943)). No dbt model changes. Real verification is the staging deploy succeeding on this branch's merge.

## Post-merge follow-ups

- [ ] No action required

- #5334 used to design a urllib3-2.x-compatible fix for the Composer 3 gzip decode crash (the deleted `test_urllib3_gzip_regression.py` logic should be salvaged for that fix).


tihuang02 (approved):
we need to address urllib3. it has high vulnerability.

https://github.com/cal-itp/data-infra/security/dependabot?q=is%3Aopen+severity%3Acritical%2Chigh+urllib3

PR #5382, opened 2026-06-04
Local composer3 airflow

### LINKED ISSUE #5355: Local composer environment should move from composer 2 to composer 3
Right now our local composer environment is composer 2

### Acceptance Criteria
We have a local way to test Composer 3

### Notes
Bonus - make it work for Windows and Mac

tihuang02:
# Description

Use composer 3 to orchestrate AirFlow in local dev environment.

Resolves #5355

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Tested on WSL -- All DAGs executed w/o errors
Not working on Windows -- failed to install `openlineage-sql`
Not tested on MacOS

image

- [ ] No action required
open an issue for `openlineage-sql` on Windows platform.


tihuang02:
need to break into smaller PRs to update many outdated dependencies first.

vevetron:
The publish_gtfs CKAN push (ckan_push) is failing on staging Composer 3. Root cause: the image ships google-cloud-storage==3.1.1,
  whose vendored _media gzip decoder drops urllib3 2.x's max_length arg → TypeError on any gzip-served GCS download. (gcs sits at 3.1.1 because firebase-admin 7.4.0 floors it >=3.1.1.)

  Fix: pin google-cloud-storage==2.19.0 in requirements-c3.txt — the 2.x line uses standalone google-resumable-media (no broken decoder) and forces firebase-admin back down.

  Also in here:
  - Consolidated requirements: make sync now installs requirements-c3.txt (so local dev mirrors staging-c3), deleted the near-duplicate
  requirements-dev.txt, and made the container host-user mapping per-OS (the host-user mapping breaks the entrypoint on macOS).
  - Dropped the decommissioned Sentry env vars from .development.env / .staging_c3.env (complements #5279, which misses the C3 files).
  - Test env: the branch had stripped main's pyproject pins, so a fresh uv.lock drifted onto urllib3 2.x / gcs 3.x and the suite was red at import. Added the 4 deps the operators import (providers-google, providers-cncf-kubernetes, humanize, sentry-sdk) and restored urllib3<2.0.0 + gcs==2.19.0 (+ aiohttp<3.14). Suite now collects and runs 110 pass / 5 fail — the 5 are TestGTFSValidatorHook/TestGtfsRtParser shelling out to the validator JAR, which exits with a different code under my local JDK (1 vs the asserted 255); should be green on CI.

  Follow-ups I'd flag: confirm those 5 validator tests on CI (or loosen the hardcoded 255); finish Sentry removal alongside #5279; reconcile this branch's pyproject more fully with main's pins.

vevetron (approved):
This doesn't fix everything but gets us closer to a resolution!

Issue #5412, opened 2026-06-10
Upgrade prod to composer-3-airflow-2.10.5-build.36


tihuang02:
As an Admin viewing the Google Cloud Composer dashboard,
When I view the version number,
I see that it is composer-3-airflow-2.10.5-build.36
And I see that each DAG runs without errors

--- 

Notes: Repeat #5311 on prod

PR #5425, opened 2026-06-11
Disable catchup_by_default on prod composer3

vevetron:
# Description

Set `core-catchup_by_default = false` on the prod `calitp-composer3` environment. The Gusty DAGs (built from `METADATA.yml`) put `catchup: False` under `default_args`, where Airflow ignores it (catchup is a DAG-level setting), so on the fresh Composer 3 metadata DB they all fall back to the default `catchup_by_default = true` and backfill from their start dates. This flips the global default to `false` so they don't. The two standalone Python DAGs that set catchup explicitly were already handled in #5424.

Works toward #5412

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Not tested.

## Post-merge follow-ups

- [ ] No action required

Existing backfill DAG runs that were already created before this applies must be cleared by hand; this only prevents new catchup.

### DIFF
--- a/iac/cal-itp-data-infra-staging/composer/us/environment.tf
+++ b/iac/cal-itp-data-infra-staging/composer/us/environment.tf
@@ -41,6 +41,7 @@ resource "google_composer_environment" "calitp-staging-composer3" {
 
       airflow_config_overrides = {
         celery-worker_concurrency                  = 4
+        core-catchup_by_default                    = false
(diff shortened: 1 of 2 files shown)


jparr (approved):

vevetron:
This change did not achieve the desired objective.  Not sure why not.

PR #5429, opened 2026-06-11
Set gusty DAG catchup at metadata top level and remove dead catchup override

vevetron:
# Description

The gusty DAGs (built from `METADATA.yml`) had `catchup: False` nested under `default_args`, where Airflow/gusty ignore it — so on the fresh Composer 3 metadata DB they fell back to the global default `catchup_by_default = true` and backfilled from their start dates. This moves `catchup: False` to the top level of each `METADATA.yml` so gusty passes it directly to `DAG(catchup=False)`. It also removes the dead / incorrect `core-catchup_by_default` override added in #5425, which targeted the wrong airflow.cfg section (`[core]` instead of `[scheduler]`) and was a no-op.

Works toward #5412

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Validated locally and on staging composer3:
- Local A/B in the composer-local-dev runtime: a gusty DAG with `catchup` under `default_args` backfilled 10 runs under the global default `true`; moving `catchup: False` to the top level produced 1 run.
- Deployed to staging composer3 via the `staging/5412-gusty-catchup` branch.

## Post-merge follow-ups

- [] No action required
Keep working on #5412

### DIFF
--- a/iac/cal-itp-data-infra-staging/composer/us/environment.tf
+++ b/iac/cal-itp-data-infra-staging/composer/us/environment.tf
@@ -41,7 +41,6 @@ resource "google_composer_environment" "calitp-staging-composer3" {
 
       airflow_config_overrides = {
         celery-worker_concurrency                  = 4
-        core-catchup_by_default                    = false
         core-dag_file_processor_timeout            = 600
         core-dagbag_import_timeout                 = 600
         core-dags_are_paused_at_creation           = true
--- a/iac/cal-itp-data-infra/composer/us/environment.tf
+++ b/iac/cal-itp-data-infra/composer/us/environment.tf
@@ -154,7 +154,6 @@ resource "google_composer_environment" "calitp-composer3" {
 
       airflow_config_overrides = {
         celery-worker_concurrency                  = 4
-        core-catchup_by_default                    = false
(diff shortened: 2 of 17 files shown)


iac/cal-itp-data-infra-staging/composer/us/environment.tf
```diff
@@ -41,7 +41,6 @@ resource "google_composer_environment" "calitp-staging-composer3" {
 
       airflow_config_overrides = {
         celery-worker_concurrency                  = 4
-        core-catchup_by_default                    = false
```
tihuang02, line 44: are we sure about changing the global setting?

iac/cal-itp-data-infra/composer/us/environment.tf
```diff
@@ -154,7 +154,6 @@ resource "google_composer_environment" "calitp-composer3" {
 
       airflow_config_overrides = {
         celery-worker_concurrency                  = 4
-        core-catchup_by_default                    = false
```
tihuang02, line 157: are we sure about changing the global setting?

tihuang02 (approved):
looks good

PR #5489, opened 2026-06-24
modify lifecycle_rule of staging buckets to save money

### LINKED ISSUE #5430: Tier the staging GTFS-RT raw bucket to cold storage
Add `SetStorageClass` lifecycle rules (Coldline at 5 days Archive at 365 days) to the staging GTFS-RT raw bucket, to validate the change before prod. Old raw is read only by `parse_and_validate_rt` for the current hour, so cold tiering is safe.

Done when `terraform plan` is an in-place update (no replace), objects transition, and `parse_and_validate_rt` stays green.

tihuang02:
# Description

modify lifecycle_rule of staging buckets to save money
Resolves #5430 

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

See CI terraform jobs

## Post-merge follow-ups

- [ ] No action required

observe the cost trend on Google console.

### DIFF
--- a/iac/cal-itp-data-infra-staging/gcs/us/storage_bucket.tf
+++ b/iac/cal-itp-data-infra-staging/gcs/us/storage_bucket.tf
@@ -215,18 +215,32 @@ resource "google_storage_bucket" "calitp-staging" {
   storage_class               = "STANDARD"
   uniform_bucket_level_access = "true"
 
+  soft_delete_policy {
+    retention_duration_seconds = 0
+  }
+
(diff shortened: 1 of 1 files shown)


vevetron:
We don't want to change all the files in all the buckets, just specific ones. I think there is a terraform way to only hit specific buckets.

vevetron:
We decided not to go down this path because of unforeseen consequences of modifying lots of tiny files seen in Daniel's Analyses.

Issue #5493, opened 2026-06-26
Bug: Airflow dags failing to parse - likely due to removing parsed elavon bucket


vevetron:
**Describe the bug**

<img width="2124" height="864" alt="Image" src="https://github.com/user-attachments/assets/1fa234cc-9445-401f-a306-c1beb2a23169" />

<img width="3878" height="2276" alt="Image" src="https://github.com/user-attachments/assets/a636bb00-4f62-4757-9b86-1c1226727dc2" />

PR #5494, opened 2026-06-26
restore elavon-parsed and elavon-raw buckets

tihuang02:
# Description

restore elavon-parsed and elavon-raw buckets on terraform files. Note that the #5068 meant to delete the buckets but it never succeeded because the buckets were not empty.  With this fix, it brings terraform back in sync with the state.

Work towards #5493 

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

No changes in `terraform plan`

## Post-merge follow-ups

- [ ] No action required

Once this is applied, need to fix the composer environment to restore environment variable for elavon-parsed bucket.


vevetron (approved):
Good first step!

PR #5483, opened 2026-06-24
feat(metabase): Cloud Armor for production in preview mode

jamalk23:
# Description

Follows #5275 / #5468 (staging Cloud Armor policy + tuning).
Resolves #5480 (deploy prod Cloud Armor in preview mode).
Enables #5481 (review prod preview logs before enforcing).

Bring the Cloud Armor policy to **production** Metabase (`metabase.dds.dot.ca.gov`) in **preview mode**, per #5480. This mirrors the fully tuned staging policy — so prod starts with the staging-derived false-positive fixes already baked in — but every enforcing rule is `preview = true`, meaning **nothing is blocked**; matches are logged only. That produces the preview logs to review per #5481 before flipping to enforce (a later PR).

**Policy contents (mirror of tuned staging, #5275 + #5468)**
- **Geo-restriction (rule 500):** would deny non-US/CA traffic. Caltrans egress (`149.136.0.0/16`) is US, unaffected.
- **OWASP WAF (rules 1000–1700):** SQLi/XSS/RCE/LFI/RFI/scanner/protocol-attack/session-fixation, with the staging-derived signature opt-outs (`942420` cookie SQLi, `921170` HPP) and the `/api/dataset*` + `/api/card*` exclusions for SQL-bearing endpoints.
- **Rate limits (rules 9000/9100):** per-IP login `rate_based_ban` and global throttle, with the Caltrans `149.136.0.0/16` exemption.
- **Default allow** (max priority).

**Logging:** LB request logging enabled (`sample_rate = 1.0`) so the preview soak captures all would-be matches.

This satisfies #5480's done criteria: the prod backend has the policy attached in preview mode and Cloud Armor logs show matches without blocking.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation

Preview mode is non-blocking, so there is no behavior change to prod traffic on merge.

## How has this been tested?

- Policy structure/expressions are identical to the tuned, soak-validated staging policy (#5275 + #5468), where these rules ran under enforcement against real usage.
- `terraform plan` (CI) shows the new `google_compute_security_policy.metabase`, logging enabled, and the policy attached to `metabase-backend-metabase`; no other resources change.
- Post-merge verification is the review itself (#5481): `jsonPayload.previewSecurityPolicy.outcome = "DENY"` entries are would-be blocks to vet.

_No dbt changes; dbt run/test steps N/A._

## Post-merge follow-ups

[...]

PR #5499, opened 2026-06-30
feat(metabase): geo-restrict to US/CA/GB/NL/HU (staging + prod)

### LINKED ISSUE #5481: Review Cloud Armor preview-mode logs for Metabase in production
Review the prod Cloud Armor preview logs to find legitimate Metabase requests the
rules would block (false positives), and list the exemptions to apply before
enforcing. Staging hit this — see #5468.

Done when we have a summary of what would be blocked and the exemptions to apply.

jamalk23:
# Description

Part of #5481. Follows #5483 (prod Cloud Armor preview deploy).

**Enforce** geo-restriction on Metabase — allow only **US, CA, GB, NL, HU** (rule priority 500) — in staging and prod. Requests geolocated outside these countries receive a `403`.

- **Prod:** geo rule moves from **preview → enforce**, and the allowlist widens from US/CA to US/CA/GB/NL/HU.
- **Staging:** was already enforcing geo; the allowlist widens to match prod.

The allowlist covers where Metabase's users are — US-based staff/analysts plus confirmed UK/EU users (GB/NL/HU added on that basis). Published dashboards are intentionally allowlist-scoped: there are **no path carve-outs** for public/embedded dashboards, so viewers outside these countries are blocked. This is the deliberate access policy.

## Type of change

- [x] New feature
- [ ] Documentation

Geo-restriction now actively blocks out-of-allowlist access, including public dashboard viewers outside US/CA/GB/NL/HU — an intentional access-policy change, flagged as breaking because it removes access that previously worked for out-of-region viewers.

## How has this been tested?

- **#5481 preview soak** (prod, `sample_rate = 1.0`): the geo rule ran in preview against real traffic. Its would-be denials were overwhelmingly bots/scanners, plus out-of-region public-dashboard views that this policy intentionally blocks. The rule only matches non-allowlist regions, so in-allowlist traffic (US/CA/GB/NL/HU) is unaffected.
- `terraform plan` shows only the rule-500 change in each workspace (prod: `preview` removed + allowlist widened; staging: allowlist widened); no other resources.

## Post-merge follow-ups

- [ ] No action required

- On apply, prod begins returning `403` to non-US/CA/GB/NL/HU requests. Watch Cloud Logging (`jsonPayload.enforcedSecurityPolicy.priority=500` / `outcome="DENY"`) for denials from *expected* regions (misgeolocation, or a newly-relevant country) and extend the allowlist if needed.
- The WAF + rate-limit rules are enforced separately in the `enforce-waf` PR; when both land, ensure the final policy keeps this **widened + enforced** geo rule (that branch still carries the older US/CA-preview geo rule).

### DIFF
--- a/iac/cal-itp-data-infra-staging/metabase/us/cloud_armor.tf
+++ b/iac/cal-itp-data-infra-staging/metabase/us/cloud_armor.tf
@@ -5,10 +5,10 @@ resource "google_compute_security_policy" "metabase-staging" {
   rule {
     priority    = 500
     action      = "deny(403)"
-    description = "Geo-restrict to US and Canada"
+    description = "Geo-restrict to US/CA/GB/NL/HU"
     match {
       expr {
-        expression = "!(origin.region_code == 'US' || origin.region_code == 'CA')"
+        expression = "!(origin.region_code == 'US' || origin.region_code == 'CA' || origin.region_code == 'GB' || origin.region_code == 'NL' || origin.region_code == 'HU')"
       }
     }
   }
--- a/iac/cal-itp-data-infra/metabase/us/cloud_armor.tf
+++ b/iac/cal-itp-data-infra/metabase/us/cloud_armor.tf
@@ -6,10 +6,10 @@ resource "google_compute_security_policy" "metabase" {
     priority    = 500
     action      = "deny(403)"
     preview     = true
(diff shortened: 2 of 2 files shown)


vevetron (approved):
From the analysis - iff most of the traffic outside of this was scanners / bots, it's worth doing this.

PR #5538, opened 2026-07-14
Provide `organization_source_record_id` in Littlepay Tables to Disambiguate Regional Agencies

### LINKED ISSUE #5294: Feat(Payments): Add seed table for determining Agency ID for SLORTA/SLO City

As an analyst, I would like to differentiate between Littlepay data for SLORTA and SLO City.

SLO City and SLORTA share a `participant_id` in Metabase. The current warehouse matches `participant_id` to `agency_id` based exclusively on `participant_id`, which leads to incorrect results in this case. We know what `device_id`s map to which agency, so we could use a seed table to let us join on both `participant_id` and `device_id` and get correct results.

### Acceptance Criteria

- [ ] A seed table is created to map device ids to agency id for SLORTA / SLO City
- [ ] The Littlepay tables are updated to use this new agency id.
- [ ] Documentation exists for how to update this table when we receive info about new device IDs

### Notes

- SLORTA and SLO city currently accept that they can access each other's payments data, so this does not require changing row level permissions. If this was an issue for future agencies, or becomes an issue in the future, we would have to consider how to implement row level permissions in this case. I am also unsure if I would consider this approach sufficient if data access issues exist, because we may experience delays in receiving updates on what devices are associated with each agency
- Once this change is made, SLORTA / SLO City dashboards would need to be updated to use agency_id rather than participant_id

mrtopsyt:
# Description

Implement regional agency disambiguation using `device_id` for `fct_payments_rides_v2` and `fct_payments_aggregations`. A regional agency is an agency that has aggregation, benefits, and/or capping shared between multiple agencies using the same Littlepay participant id. 

- Add `organization_source_record_id` and `organization_name` to `fct_payments_rides_v2` based on `particpant_id` for non-regional agencies and `participant_id`  + `device_id` for regional agencies.
- Add `first_tap_organization_source_record_id` and `first_tap_organization_name` for regional agencies only.

Resolves #5294

## Type of change

- [x] New feature
- [ ] Documentation

### Run models and tests

```
uv run dbt run --full-refresh -s +fct_payments_rides_v2 +fct_payments_aggregations
uv run dbt test -s fct_payments_rides_v2 fct_payments_aggregations int_payments__regional_agencies_first_tap_by_aggregation int_payments__agency_joined_device_transactions
```

Note that some failing tests are lingering from Littlepay ambiguities - these are not introduced by this issue.

### Check that organization name and source record id match expectations

```
SELECT organization_name, organization_source_record_id, device_id, count(*) AS ct
FROM cal-itp-data-infra-staging.ap_mart_payments.fct_payments_rides_v2
WHERE participant_id = 'slorta'
GROUP BY  organization_name, organization_source_record_id, device_id
ORDER BY device_id DESC
```

Expect these to match the device mapping in Airtable

### Check that the join between rides and aggregations works

```
WITH expected_first_tap AS (
    SELECT
        aggregation_id,
        organization_source_record_id AS expected_first_tap_organization_source_record_id
    FROM `cal-itp-data-infra-staging.ap_mart_payments.fct_payments_rides_v2`
    WHERE participant_id = 'slorta'
        AND aggregation_id IS NOT NULL
        AND transaction_date_time_utc IS NOT NULL
    -- pick the org on the earliest tap in each aggregation, matching the model's tie-break
    QUALIFY ROW_NUMBER() OVER (
        PARTITION BY aggregation_id
        ORDER BY transaction_date_time_utc, littlepay_transaction_id
    ) = 1
),

actual AS (
    SELECT
        aggregation_id,
        first_tap_organization_source_record_id
    FROM `cal-itp-data-infra-staging.ap_mart_payments.fct_payments_aggregations`
    WHERE participant_id = 'slorta'
)

SELECT
    COALESCE(actual.aggregation_id, expected.aggregation_id) AS aggregation_id,
    actual.first_tap_organization_source_record_id           AS actual_value,
    expected.expected_first_tap_organization_source_record_id AS expected_value,
FROM actual
FULL OUTER JOIN expected_first_tap AS expected USING (aggregation_id)
WHERE actual.first_tap_organization_source_record_id
    IS DISTINCT FROM expected.expected_first_tap_organization_source_record_id
ORDER BY aggregation_id
```

Expect an empty table - note the slorta filter, because we are interested in it as the only regional agency

### Examine regional agency aggregations rows with null first tap id

```
select aggregation_id, first_tap_organization_source_record_id, participant_id, latest_micropayment_transaction_datetime_pacific, reconciliation_category
from `cal-itp-data-infra-staging.ap_mart_payments.fct_payments_aggregations`
where participant_id = 'slorta' and first_tap_organization_source_record_id is null
```

Expect an empty table

## Post-merge follow-ups

- [ ] No action required

@amandaha8 to implement this in Metabase with the following recommended actions
- Use organization name for the agency filter in slorta's contactless metrics dashboard for questions based on `fct_payments_rides_v2`
- Remove any agency filters for other tables in the contactless metrics dashboard that do not use `fct_payments_rides_v2` - the joins currently used are causing fan out
- Create a new dashboard using `fct_payments_aggregations` to provide revenue splits between the two agencies
  - I think the correct approach is to take the sum of `elavon_net_amount`, filtered for `reconciliation_category = 'Settled non-zero sales (with Elavon match)'`  by `first_tap_organization_name`. Experimentation might be required 
  - Expect this to provide a different value to the existing split in the metrics dashboard - that is a gross revenue figure that does not account for unsettled sales, and the merge seems to be introducing a fan out.

### Notes
This introduced an explicit dependency of fct_payments_rides on fct_payments_aggregations. We have avoided this in the past. I think it's fine for this limited case (a field that is null except for one participant), and because the alternative would be to create a new table with a Littlepay <> Elavon join. I would prefer to avoid adding that complexity, so I implemented this way, but am open to changing.


vevetron (approved):
LGTM.  Might need to be reworked later.

Issue #3713, opened 2025-02-19
Chore: Review Kubernetes configuration


erikamov:
### Chore

Review Kubernetes configuration.

While Kubernetes itself isn't one of the biggest cost drivers, most of the fees it causes are reflected in the Compute Engine charges and the configuration of Kubernetes can be a significant driver of that cost. We identified that the data-infra-apps cluster (where users' JupyterHub environments run) is configured to be multi-zone.
We recommend scale the non-airflow cluster down to 1 zone so there is a single pool of nodes that everyone shares, and we could potentially cut the min number of nodes in half while also tuning the autoscaling to be more responsive.

vevetron:
Current-state check on `data-infra-apps` (July 2026): the `jupyterhub-users` pool sits at a 6-node floor (`e2-highmem-2`, max 12, regional). Measured usage is light — ~1.5 avg concurrent notebooks (peak 6, zero on weekends), so the floor is 2–3× over-provisioned. Both halves of the recommendation apply: lower the floor and tune the reclaim.

**1. Lower the floor 6 → 3** — where the savings are. Reversible, doesn't evict running notebooks. It's `total_min_node_count` on the `jupyterhub-users` pool in `iac/cal-itp-data-infra/gke/us/container_node_pool.tf`, applied via `terraform-apply.yml` (same path as #5332, which scaled the other pools' floors):

```
    total_min_node_count = 3   # was 6
```

3 fewer nodes ≈ ~$70/mo each (node + boot disk) → realistically **~$130–200/mo (~$1,600–2,400/yr)**, since the pool floats to 3–4 nodes during work hours.

**2. Explicit 3h idle cull** — a UX guardrail (chart default is 1h). Long enough not to kill a kernel over lunch, short enough to free abandoned notebooks same-day. PR-reviewed change in `kubernetes/apps/charts/jupyterhub/values.yaml`:

```yaml
jupyterhub:
  cull:
    enabled: true
    timeout: 10800   # 3h idle (was 1h)
    every: 600
```

**3. (Optional) `OPTIMIZE_UTILIZATION` profile** — the cluster is on the conservative `BALANCED`; this reclaims emptied nodes faster so the pool settles to the floor. It's cluster-wide (affects the `apps-v2`/`gtfsrt-v4`/`jobs-v1` pools too), so better as a separate follow-up than bundled here.

[...]

PR #5542, opened 2026-07-20
Lowers jupyterhub-users node floor 6 to 3

vevetron:
# Description

Lowers the `jupyterhub-users` node-pool floor from 6 to 3 and sets an explicit 3h idle cull for JupyterHub.

The pool sat at a 6-node floor while measured usage is low (peak 6, zero on weekends), so the floor was 2–3× over-provisioned. Dropping it to 3 keeps headroom, is reversible, and evicts no running notebooks — the autoscaler just stops holding empty nodes. The 3h cull (up from the chart default of 1h) is what empties an idle notebook's node so the autoscaler can reclaim it, while still being long enough not to kill a kernel over lunch or a long meeting.

The floor change applies via `terraform-apply.yml` (same pipeline as #5332, which scaled the other pools' floors); the cull change deploys via `deploy-kubernetes.yml`.

Works toward #3713.

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

- Not tested, will test with terraform and watch

## Post-merge follow-ups

- Floor change auto-applies on merge via `terraform-apply.yml`; cull auto-deploys via `deploy-kubernetes.yml` — no manual action.
- Separate, out of scope here: right-size the `apps-v2` control-plane pool (3× n1-standard-4, near-empty), gated on removing the orphaned Elavon SFTP pod (#5035) which currently pins a node against scale-down.
- Optional: cluster `OPTIMIZE_UTILIZATION` autoscaling profile (cluster-wide) to reclaim emptied nodes faster.

### DIFF
--- a/iac/cal-itp-data-infra/gke/us/container_node_pool.tf
+++ b/iac/cal-itp-data-infra/gke/us/container_node_pool.tf
@@ -67,7 +67,7 @@ resource "google_container_node_pool" "tfer--data-infra-apps_jupyterhub-users" {
     max_node_count       = "0"
     min_node_count       = "0"
     total_max_node_count = "12"
-    total_min_node_count = "6"
+    total_min_node_count = "3"
   }
 
   cluster            = google_container_cluster.tfer--data-infra-apps.name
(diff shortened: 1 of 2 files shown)


vevetron:
There is some terraform drift happening due to unresolved #5332 .  Merge after https://github.com/cal-itp/data-infra/pull/5543

tihuang02 (approved):
looks good.

vevetron:
I can't verify the cull is being adjusted in the kubernetes previews but I think it's okay.

Issue #5527, opened 2026-07-09
Support `trip_modifications` feed type in GTFS dataset configuration and RT Archiver


fsalemi:
## Summary

Extend the GTFS dataset configuration so the GTFS-RT Archiver recognizes Trip Modifications as a supported realtime feed type and downloads the protobuf into the raw bucket.

## Description

The current GTFS dataset transformation only maps the following realtime feed types:

- `service_alerts`
- `trip_updates`
- `vehicle_positions`

Trip Modifications is not currently recognized, preventing the archiver from downloading published Trip Modifications feeds.

This issue covers the minimum changes required for the archiver to ingest Trip Modifications into the raw bucket.

## Proposed Changes

- Update the configuration staging table to recognize `trip_modifications`.
- Add `trip_modifications` to the active GTFS-RT archiver heartbeat service so those configuration records are published to the download topic.
- Add or update unit tests to verify that Trip Modifications feeds are included while unsupported feed types remain excluded.
- Verify that the active archiver downloads the protobuf and stores it under the `trip_modifications` prefix in the raw bucket.

## Acceptance Criteria

- The configuration staging table supports `trip_modifications`.
- The GTFS-RT Archiver successfully downloads and stores Trip Modifications protobuf files in the raw bucket.
- Existing Service Alerts, Trip Updates, and Vehicle Positions continue to function without regression.

vevetron:
SC metro might have trip modifications.

PR #5601, opened 2026-07-31
fix(gtfs-rt-archiver): use terraform service account

fsalemi:
## Description

Updates the GTFS-RT Archiver GitHub Actions workflow to use the dedicated Terraform service account for Terraform plan and apply.

The workflow was previously using `github-actions-service-account`, which does not have permission to update the GTFS-RT Archiver Cloud Functions and resulted in:

`cloudfunctions.functions.update` permission denied.

This change uses `github-actions-terraform` for the Terraform deployment steps, consistent with the existing Terraform workflows.

Ref #<https://github.com/cal-itp/data-infra/issues/5527>

## Changes

- Add `TERRAFORM_SERVICE_ACCOUNT` for the staging Terraform service account.
- Update `staging-plan` to authenticate with `TERRAFORM_SERVICE_ACCOUNT`.
- Update `staging-apply` to authenticate with `TERRAFORM_SERVICE_ACCOUNT`.

## How has this been tested?

The issue was identified when the GTFS-RT Archiver staging deployment attempted to update the Cloud Functions and failed due to insufficient permissions.

### DIFF
--- a/.github/workflows/gtfs-rt-archiver.yml
+++ b/.github/workflows/gtfs-rt-archiver.yml
@@ -20,6 +20,7 @@ env:
   PYTHON_VERSION: '3.11'
   UV_VERSION: '0.11.5'
   SERVICE_ACCOUNT: github-actions-service-account@cal-itp-data-infra-staging.iam.gserviceaccount.com
+  TERRAFORM_SERVICE_ACCOUNT: github-actions-terraform@cal-itp-data-infra-staging.iam.gserviceaccount.com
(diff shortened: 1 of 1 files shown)


tihuang02 (approved):
looks right. 
I wonder why we didn't hit this until now.

fsalemi:
@tihuang02 
Yeah, I was wondering the same thing. My guess is that previous production/IaC changes were going through the generic Terraform workflow using github-actions-terraform, and this service-specific staging path just hadn’t needed to update the Cloud Functions since the auth setup changed.

PR #5606, opened 2026-08-03
fix(gtfs-rt-archiver): deploy service changes to production

fsalemi:
## Description

Adds a production Terraform apply job to the GTFS-RT Archiver workflow.

The workflow currently deploys changes under `services/gtfs-rt-archiver/**` to staging, but does not deploy the same source changes to the production GTFS-RT Archiver. This caused the recently added `trip_modifications` support to be available in staging but not production.

The new production job:

- Runs only after a successful staging deployment.
- Runs only on `main`.
- Uses the production Terraform service account and Workload Identity provider.
- Applies the production GTFS-RT Archiver Terraform configuration at `iac/cal-itp-data-infra/gtfs-rt-archiver/us`.

ref: https://github.com/cal-itp/data-infra/issues/5527

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## How has this been tested?

- Confirmed the existing workflow successfully deploys service-code changes to staging.
- Confirmed the workflow currently has no production apply job.
- Confirmed production did not receive the recent `trip_modifications` source-code update.

### DIFF
--- a/.github/workflows/gtfs-rt-archiver.yml
+++ b/.github/workflows/gtfs-rt-archiver.yml
@@ -19,10 +19,15 @@ concurrency:
 env:
   PYTHON_VERSION: '3.11'
   UV_VERSION: '0.11.5'
+
   SERVICE_ACCOUNT: github-actions-service-account@cal-itp-data-infra-staging.iam.gserviceaccount.com
   TERRAFORM_SERVICE_ACCOUNT: github-actions-terraform@cal-itp-data-infra-staging.iam.gserviceaccount.com
   WORKLOAD_IDENTITY_PROVIDER: projects/473674835135/locations/global/workloadIdentityPools/github-actions/providers/data-infra
   PROJECT_ID: cal-itp-data-infra-staging
+
+  PRODUCTION_TERRAFORM_SERVICE_ACCOUNT: github-actions-terraform@cal-itp-data-infra.iam.gserviceaccount.com
+  PRODUCTION_WORKLOAD_IDENTITY_PROVIDER: projects/1005246706141/locations/global/workloadIdentityPools/github-actions/providers/data-infra
+  PRODUCTION_PROJECT_ID: cal-itp-data-infra
 
 jobs:
   test:
@@ -56,7 +61,7 @@ jobs:
       - name: Install uv
         uses: astral-sh/setup-uv@v7
         with:
-          version:  ${{ env.UV_VERSION }}
+          version: ${{ env.UV_VERSION }}
           python-version: ${{ env.PYTHON_VERSION }}
           enable-cache: true
           cache-python: true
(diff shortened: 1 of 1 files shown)


tihuang02 (commented):
The change looks good in general. But I am reluctant to start pulling in the new data type.

Should we first discuss as a team about:
have we see any data in staging yet? do we want to verify the data and plan for how to consume the data, also estimate the cost increase before taking this to production?

tihuang02:
Actually, I just spent the afternoon reviewing the github actions. `staging-plan` and `production-plan` are already covered by `terraform-plan.yml`.  `staging-apply` and `production-apply` are already covered by `terraform-apply.yml`.  The only useful job in `gtfs-rt-archiver.yml' is `test`, which runs pytest for the archiver python file.

I propose we clean this file up and leave only the `test` job. And work on `terrafrom-plan.yml` and `terraform-apply.yml` for any misbehaviors in staging or prod deployment.

fsalemi:
> Actually, I just spent the afternoon reviewing the github actions. `staging-plan` and `production-plan` are already covered by `terraform-plan.yml`. `staging-apply` and `production-apply` are already covered by `terraform-apply.yml`. The only useful job in `gtfs-rt-archiver.yml' is `test`, which runs pytest for the archiver python file.
> 
> I propose we clean this file up and leave only the `test` job. And work on `terrafrom-plan.yml` and `terraform-apply.yml` for any misbehaviors in staging or prod deployment.

@tihuang02 ,
That also works. However, since `terraform-plan.yml` and `terraform-apply.yml` currently determine their Terraform targets only from changed `iac/**` directories, how would you handle changes under `services/gtfs-rt-archiver/**`? Would you have the `targets` job also include `iac/cal-itp-data-infra-staging/gtfs-rt-archiver/us` and `iac/cal-itp-data-infra/gtfs-rt-archiver/us` when the service code changes?

fsalemi:
> The change looks good in general. But I am reluctant to start pulling in the new data type.
> 
> Should we first discuss as a team about: have we see any data in staging yet? do we want to verify the data and plan for how to consume the data, also estimate the cost increase before taking this to production?

@tihuang02 
Yes. I confirmed that Trip Modifications are now being archived in the staging raw bucket. Initially I only observed the 15-byte header-only feeds, but I have since confirmed that Long Beach is periodically publishing actual Trip Modifications with entities.

My goal was to get the raw feed into the pipeline first. The next step would be to verify the payloads and then discuss how we want to parse and expose Trip Modifications downstream.

I agree that we should estimate the storage and processing impact before enabling it in production.

tihuang02:
Note down the agreement in the meeting:
1. move the current `downloader.py` to a staging branch, so we can download the `trip_modifications` data to the staging buckets.
2. revert the main `downloader.py` to the version w/o `trip_modifications`.
3. `terraform_apply.yml` and `terraform_plan.yml` should re-deploy cloud functions if files in `services/gtfs-rt-archiver/` changed.

Will create issues for each of the above.

PR #5608, opened 2026-08-03
reduce the downloader timeout from 30 to 10 secs

### LINKED ISSUE #5565: Cap the GTFS-RT archiver feed read timeout at 10 seconds
Lower `REQUEST_READ_TIMEOUT` from `30` to `10` in both
`iac/cal-itp-data-infra/gtfs-rt-archiver/us/service.tf` and the staging
service.tf, and keep the value as configuration rather than a hardcoded literal.

This puts a ceiling on how long a single slow feed can hold an archiver instance.
It does not touch steady state — normal fetches finish in ~1.1s (p90 ~1.4s), well
under both 30s and 10s — so it is **spike insurance**: on the slow-feed days that
roughly doubled spend (late May / early June 2026), a 10s cap bounds the worst
case at `~feeds × 10s` instead of letting hung feeds run open-ended. A 10s cap
also sits below the 20s poll cadence, so a feed finishes or is dropped inside its
own window.

`service.py` already wraps the download in try/except and logs `Failed …`, so a
tripped timeout is a clean skip — no crash, no partial object written.

Verify: a deliberately slow feed is abandoned at ~10s with a logged exception and
no object written; normal feeds are unaffected.

tihuang02:
# Description

reduce the downloader timeout from 30 to 10 secs.

Resolves #5565 

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

`terraform plan`

## Post-merge follow-ups

- [x] No action required

### DIFF
--- a/iac/cal-itp-data-infra-staging/gtfs-rt-archiver/us/.terraform.lock.hcl
+++ b/iac/cal-itp-data-infra-staging/gtfs-rt-archiver/us/.terraform.lock.hcl
@@ -4,6 +4,7 @@
 provider "registry.terraform.io/hashicorp/archive" {
   version = "2.7.1"
   hashes = [
+    "h1:62VrkalDPMKB9zerCBS4iKTbvxejwnAWn/XXYZZQWD4=",
     "h1:A7EnRBVm4h9ryO9LwxYnKr4fy7ExPMwD5a1DsY7m1Y0=",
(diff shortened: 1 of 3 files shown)


vevetron (approved):
We did some analysis and this should have minimal impact.  This is a defensive move against cost overruns when a vendor's api calls get really slow.

vevetron:
I don't totally understand but some analysis shows the timeout change didn't achieve the expected outcome.

PR #5530, opened 2026-07-10
refactor(metabase): unify Dockerfile + add test-instance runbook; publish image tags only from main (takes over #5277)

jamalk23:
# Description

Takes over the still-applicable half of #5277. Chris Alfano's commits are cherry-picked with authorship preserved; two follow-up commits update the work for what has landed on `main` since his branch was created in May.

Refs #4864. Supersedes #5277.

**Carried over from #5277 (cherry-picked):**

- **`refactor(metabase)`** — unify `Dockerfile.production` / `Dockerfile.staging` into a single `Dockerfile`. The Cloud SQL Unix-socket symlink is now created at runtime by `entrypoint.sh` from `CLOUD_SQL_INSTANCE_CONNECTION_NAME` (auto-detected when exactly one instance is mounted at `/cloudsql/`, which covers the existing prod and staging services). The build workflow produces one image tagged both `:staging` and `:production`. This is what lets an operator deploy any Metabase instance from the same image — the capability the new runbook depends on. Existing services keep their pinned digests; the unified image is picked up at their next intentional redeploy.
- **`docs(runbooks)`** — add `metabase-test-instance-from-prod.md`: provision a disposable Metabase in `cal-itp-data-infra-staging` seeded with a prod dump, for upgrade rehearsal against real data, isolated investigation of prod data anomalies, and the periodic #4864 recovery exercise. Includes measured timings from the 2026-05-07 validation run.

**New in this PR:**

- **`docs(metabase)`** — update the runbook and `services/metabase/README.md` for post-May `main`: #5386 added nightly automated exports to the same `exports/` prefix the runbook's manual export used, so the runbook now prefers reusing the nightly `.sql.gz` dump, names manual exports `-adhoc`, and guards teardown so it can never delete a nightly backup. README's references to the split Dockerfiles (added by #5385 after #5277 branched) updated for the unified one.
[...]

PR #5647, opened 2026-08-12
fix(metabase): set Cloud SQL connection name on production (deploys v0.58.24, CVE-2026-72898)

### LINKED ISSUE #4928: Formalize cloud-run metabase updates

As a self-running software in the cloud called Metabase
I shouldn't depend on humans
to get my updates

See if we can configure dependabot to handle metabase upgrades.  Especially critical security ones.

### Acceptance Criteria

There is an automatic system to initialize metabase updates in staging and/or prod.  If that doesn't work, good documentation on a process.

jamalk23:
# Description

Refs #4928.

Production half of the `CLOUD_SQL_INSTANCE_CONNECTION_NAME` fix, split out so staging could be validated first. Same root cause, diagnosed on staging revision `metabase-staging-00008-mzj` — see the staging PR for the full writeup.

`entrypoint.sh` falls back to `ls /cloudsql | head -1` when the variable is unset, and that cannot work on Cloud Run: the socket is connectable but the directory is not enumerable, so the container exits 1.

Production is currently unaffected **only** because it hasn't been redeployed since before #5530 and is still running an image built from the old split Dockerfile, which baked the symlink in at build time. Its next redeploy — for any reason — picks up the unified image and fails identically.

Pushed 6eb45b660 after approval —

> Widens the startup probe from 110s to 300s (48 × 5s probing + 60s initial delay;
> Cloud Run caps the probing product at 240s, so this is the maximum available).
> 
> Metabase doesn't serve / until Liquibase migrations finish. If the probe expires
> mid-migration, Cloud Run kills the container holding DATABASECHANGELOGLOCK and the
> next boot hangs on a lock nobody owns — a manual fix, on prod, mid-incident.
> Staging's migration took 65s of the old 110s budget; prod's database is larger.
> 
> failure_threshold is a ceiling, not a wait — a healthy container goes Ready on its
> first successful probe, so normal starts and autoscaling are unchanged. The plan
> now shows 2 changes instead of 1.

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

Validated on staging via the staging PR before merging this. The change is identical apart from the instance reference.

## Post-merge follow-ups

- [ ] No action required

> [!IMPORTANT]
> Merging triggers `terraform-apply` on the production module, creating a new revision that re-resolves `:production` and deploys Metabase **v0.58.24** — the CVE-2026-72898 fix. This is the production upgrade. Do not merge until the staging PR is merged and validated.

1. Watch the migration: `gcloud beta run services logs tail metabase --region=us-west2 --project=cal-itp-data-infra`. Prod's database is larger than staging's, so expect it to take longer.
1. Confirm `/api/health` returns `ok` and the revision digest is `sha256:5d9e1044…`.
1. Then the remaining CVE-2026-72898 work: IOC search in the LB logs, and the advisory's post-upgrade cleanup (clear `core_session`, audit API keys and admin accounts, rotate connected-database credentials).

### DIFF
--- a/iac/cal-itp-data-infra/metabase/us/cloud_armor.tf
+++ b/iac/cal-itp-data-infra/metabase/us/cloud_armor.tf
@@ -13,6 +13,23 @@ resource "google_compute_security_policy" "metabase" {
     }
   }
 
+  # Interim mitigation for CVE-2026-72898 (GHSA-vwf4-m7j8-wcjf): unauthenticated
+  # SQL injection via /api/session/reset_password leading to full admin access.
+  # Added out-of-band on 2026-08-11 and codified here so that a terraform apply
+  # cannot silently drop it while the service is still on a vulnerable version.
+  # Remove only after v0.58.24 is confirmed serving. Blocks self-serve password
+  # reset while in place.
+  rule {
+    priority    = 600
+    action      = "deny(403)"
+    description = "TEMP: block CVE-2026-72898 pending v0.58.24 upgrade"
+    match {
+      expr {
+        expression = "request.path.startsWith('/api/session/reset_password')"
+      }
+    }
+  }
+
   rule {
     priority    = 1000
     action      = "deny(403)"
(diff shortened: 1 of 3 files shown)


tihuang02 (approved):
looks good. same was applied to staging by #5648

PR #5646, opened 2026-08-11
4928 upgrade metabase + update runbook

jamalk23:
# Description

Rewrites the Metabase upgrade runbook for Cloud Run, and applies it to bump Metabase to the fix for an actively exploited critical CVE.

Refs #4928.

**`docs(runbooks)` — rewrite `upgrade-metabase.md`.** It still described the pre-Cloud-Run Kubernetes deployment (Lens, restic, `kubectl -n metabase-test`, edits to `kubernetes/apps/values/metabase.yaml` — a path that no longer exists), so it was unusable. Rewritten against the current deployment, and it documents three things that weren't written down anywhere:

- Merging a version bump publishes images but **deploys nothing** — `terraform-apply` is path-filtered to `iac/**/*.tf` and the image string in `service.tf` is an unchanging mutable tag, so Terraform sees no diff. The deploy is an explicit `gcloud run services update` per environment.
- `metabase-staging` is a small test environment, not a prod snapshot, so it isn't an upgrade rehearsal. Minor-version jumps and the eventual Enterprise cutover route through `metabase-test-instance-from-prod.md` first; patch bumps don't.
- Schema migrations are irreversible, so a pre-upgrade Cloud SQL backup is required and rollback splits into the migrated / not-yet-migrated cases.

**`chore(metabase)` — bump `v0.58.7` → `v0.58.24`.**

> [!WARNING]
> Patch for **CVE-2026-72898** / `GHSA-vwf4-m7j8-wcjf` — CVSS **10.0**, unauthenticated SQL injection via `/api/session/reset_password` giving full admin access. Metabase has confirmed active exploitation.

Prod runs `v0.58.7` and is publicly reachable, so it's in scope. `v0.58.24` is the vendor-designated fix and the highest patch on the line prod already runs, so there's no minor-version migration risk. A `deny(403)` rule blocking the vulnerable endpoint was added out-of-band to both Cloud Armor policies on 2026-08-11 as interim mitigation.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation

## How has this been tested?

- CI builds the image on this PR without publishing (the gate from #5530) — that's what confirms the new base still works with our layers.
[...]

PR #5648, opened 2026-08-12
fix(metabase): set Cloud SQL connection name on staging (unblocks v0.58.24 deploy)

### LINKED ISSUE #4928: Formalize cloud-run metabase updates

jamalk23:
# Description

The v0.58.24 deploy to staging failed its startup probe. Revision `metabase-staging-00008-mzj` crash-looped with:

ERROR: CLOUD_SQL_INSTANCE_CONNECTION_NAME must be set, or a single Cloud SQL instance must be mounted at /cloudsql/

`entrypoint.sh` falls back to `ls /cloudsql | head -1` when that variable is unset, and **that cannot work on Cloud Run** — the socket at `/cloudsql/<connection-name>/.s.PGSQL.5432` is connectable, but the directory is not enumerable, so `ls` returns empty and the container exits 1. Nothing else was misconfigured: the instance was attached and `volume_mounts` declared `mount_path = "/cloudsql"`.

Never exercised until now — the old split Dockerfiles baked the symlink at build time, and #5530's validation set the variable explicitly. Staging hadn't been redeployed since 2026-05-29, so this was the unified image's first run on a real service.

Sets the variable explicitly on staging from the instance's own `connection_name`, restoring what the old image did. Also corrects `README.md`, which claimed the fallback covered the existing deployments.

Production needs the same fix and is handled in a follow-up PR, deliberately held so staging can be validated first.

Refs #4928.

## Type of change

- [ ] New feature
- [x] Documentation

## How has this been tested?

Diagnosed from the failed revision's logs; staging stayed up on revision 00007 throughout and traffic never shifted. The fix restores the exact connection name the previous working image had baked in. Not yet verified running — `terraform-apply` on merge is what deploys it.

## Post-merge follow-ups

- [ ] No action required

> [!IMPORTANT]
> Unlike the Dockerfile bump in #5646, this is a `.tf` change, so `terraform-apply` **will** trigger and create a new staging revision — which re-resolves `:staging` and deploys v0.58.24. Merging this upgrades staging. Production is untouched.

1. Confirm staging: `/api/health` returns `ok`, revision digest is `sha256:5d9e1044…`, dashboards and a BigQuery-backed question work.
1. Soak, then merge the production PR.

### DIFF
--- a/iac/cal-itp-data-infra-staging/metabase/us/cloud_armor.tf
+++ b/iac/cal-itp-data-infra-staging/metabase/us/cloud_armor.tf
@@ -13,6 +13,23 @@ resource "google_compute_security_policy" "metabase-staging" {
     }
   }
 
+  # Interim mitigation for CVE-2026-72898 (GHSA-vwf4-m7j8-wcjf): unauthenticated
+  # SQL injection via /api/session/reset_password leading to full admin access.
+  # Added out-of-band on 2026-08-11 and codified here so that a terraform apply
+  # cannot silently drop it while the service is still on a vulnerable version.
+  # Remove only after v0.58.24 is confirmed serving. Blocks self-serve password
+  # reset while in place.
+  rule {
+    priority    = 600
+    action      = "deny(403)"
+    description = "TEMP: block CVE-2026-72898 pending v0.58.24 upgrade"
+    match {
+      expr {
+        expression = "request.path.startsWith('/api/session/reset_password')"
+      }
+    }
+  }
+
   rule {
     priority    = 1000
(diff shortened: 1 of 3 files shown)


tihuang02 (approved):
looks good

tihuang02 (approved):

Issue #5548, opened 2026-07-21
Stop re-scanning the full Elavon history on every warehouse build


vevetron:
Every warehouse build reprocesses and re-scans the **entire** Elavon history, not just the new day's data.

- `sync_elavon` (daily) mirrors the full contents of the Elavon SFTP server into each timestamped GCS partition, and `parse_elavon` re-parses that whole partition to JSONL — the DAG's own README notes this "handles all of history" and that the implementation is "meant to be temporary."
- Downstream, `stg_elavon__transactions` selects across all accumulated dumps and `QUALIFY DENSE_RANK`s to keep the latest, so a single run scans **~7.5 TiB**. It runs on every daily payments build (`dbt_daily` / `dbt_all`), making Elavon one of the larger recurring BigQuery scan costs.

This is the unchecked work already listed on the parent epic #2462 — filing it as a standalone, actionable ticket so it can be scheduled on its own.

### Proposed approach (from #2462)

1. Detect changes in source files using **file hashes**, and only parse changed/new files to JSONL (skip unchanged history).
2. Make `stg_elavon__transactions` **incremental** rather than filtering on `execution_ts` over the full-history partition.
3. (Optional, enabler) Turn `elavon_to_gcs_jsonl.py` into an Airflow **operator** so runs isolate by context date instead of wall-clock — supports the incremental approach.

### Done when

- A routine daily payments build no longer scans the full Elavon history — bytes scanned for `stg_elavon__transactions` drops from ~7.5 TiB toward a single increment.
- `fct_payments_chargeback_transactions` and the other Elavon-derived marts produce identical results to the current full-history build (no rows lost, deposits/billing/adjustments/chargebacks reconcile).
- A one-time full rebuild path is documented (for backfills / source corrections).

### Notes

- Parent epic: #2462 (Future Improvements for Elavon Data Pipeline). Cross-link: #5442 (Production cost controls).
- Sibling context already done: #4872 (stop syncing all history every day), #4420 (Terraform SFTP). Still open: #4871, #4909, #5035.
- Source: `airflow/dags/{sync_elavon,parse_elavon}`, `warehouse/models/staging/payments/elavon/stg_elavon__transactions.sql`.

vevetron:
### Measured monthly cost this could save

[...]

PR #5653, opened 2026-08-13
Add a lifecycle rule to elevon-parsed

jparr:
Saves money on stg_elavon__transactions and storage

# Description

`QUALIFY DENSE_RANK() OVER (ORDER BY execution_ts DESC) = 1` 
(https://github.com/cal-itp/data-infra/blob/8c74d92c0c02a22e37f8dfb9fbec86bd6d13910d/warehouse/models/staging/payments/elavon/stg_elavon__transactions.sql#L14)

costs about $4 each day because it has to full scan the whole table to find the last execution. If we just drop old executions the table is much smaller and this line is pennies.

This is a simple alternative to the path suggested in the issue that would require rewriting the parse logic resulting in a flat output and then changing the staging model. This would have dropped execution history anyway so this seems harmless.

14 day retention is an arbitrary value here and open for change. I could see an argument for 30 days? or 3 days?

ref: #5548

## Type of change

cost reduction

### DIFF
--- a/iac/cal-itp-data-infra-staging/gcs/us/storage_bucket.tf
+++ b/iac/cal-itp-data-infra-staging/gcs/us/storage_bucket.tf
@@ -161,6 +161,22 @@ resource "google_storage_bucket" "calitp-staging-elavon-parsed" {
   requester_pays              = "false"
   storage_class               = "STANDARD"
   uniform_bucket_level_access = "true"
+
+  lifecycle_rule {
+    action {
+      type = "Delete"
+    }
+
+    condition {
+      age                        = "14"
+      created_before             = ""
+      days_since_custom_time     = "0"
+      days_since_noncurrent_time = "0"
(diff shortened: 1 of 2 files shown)


erikamov (approved):

mrtopsyt:
Discussed on call with Jeff. I think this is lower risk, but to mitigate risks as occurred with #5068 I think we should do the following before merging. 

- Delete parsed data > 12 days old from GCP, so it can be restored from a backup if necessary - the bucket has a soft delete policy
- Confirm that the table runs correctly
- Merge this PR, and monitor

tihuang02 (approved):
looks good. thx!

jparr:
The lifecycle rule will also soft delete so that data is still recoverable within the soft delete's retention policy if the bucket has soft delete enabled.

mrtopsyt:
Oh cool, I didn't realize that - I'm just comfy merging in that case

Issue #5567, opened 2026-07-23
Poll GTFS-RT service alerts every 5 minutes instead of every 20 seconds


vevetron:
Service-alerts feeds change rarely, but today they are polled on the same 20s
cadence as vehicle positions — ~15× more often than they need to be. Make the
poll cadence per feed type and set `service_alerts` to ~5 minutes.

**This is the biggest measured cost lever in the epic.** Service alerts are 32.8%
of requests (154 of 470 feeds) and — measured from paired archiver logs — take
~1.1s to fetch, about the same as any other feed (fetch time is network-bound,
not payload-bound). So they are ~31% of instance-seconds. Dropping to 5 minutes
removes 14/15 of them → ~29% of archiver instance-time, **~$4.25/day ≈ ~$128/mo**
(against a ~$441/mo baseline). It also removes ~31% of the raw objects/writes the
archiver produces — a direct assist to Reduce GCS bucket storage cost (#5435).

- Cadence comes from the clock fan-out (`services/gtfs-rt-archiver/clock.yaml`
  publishes 3 batches/minute) combined with
  `services/gtfs-rt-archiver/gtfs_rt_archiver/heartbeat.py::messages()`, which
  currently publishes **all** feed types on **every** batch. service_alerts needs
  to be published on ~1 of every 15 batches.

Verify: service_alerts feeds are fetched ~once per 5 minutes; `trip_updates` and
`vehicle_positions` remain at 20s.

Note: confirm 5 minutes is acceptable to downstream RT consumers before shipping.

fsalemi:
Investigating the current clock fan-out and heartbeat logic to implement per-feed polling cadence for service_alerts, while keeping trip_updates and vehicle_positions at the existing 20-second interval...

vevetron:
@tihuang02 please communicate about this and plan this together.  We want to compile all the requirements from the epic.

vevetron:
Daniel to review

PR #5670, opened 2026-08-19
fix: lower GTFS-RT write alert threshold

fsalemi:
# Description

Follow-up to #5567.

Polling service_alerts every 5 minutes reduced the expected GTFS-RT raw GCS writes by ~30%, causing the existing low-write alert threshold of 5,000 writes/5 min to trigger under normal operation.

This lowers the threshold from 5,000 to 4,000 writes/5 min to account for the new expected baseline.

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

No test is necessary

## Post-merge follow-ups

- [X] No action required


tihuang02 (approved):
looks good

PR #5685, opened 2026-09-02
feat(gtfs-rt-archiver): add high-frequency cohort archiver with 3s cadence

### LINKED ISSUE #5566: Developer can run a selected subset of feeds on a separate archiver schedule

As a Developer or Analyst,
When I want a few chosen agencies archived on a different cadence than the
standard 20s run — for example a small transit-signal-priority cohort captured
every few seconds for a study,
I can name that subset of feeds and have the archiver run them on their own
schedule, separate from the main run.

### Acceptance Criteria

1. A configuration selects a subset of feeds (by name / agency / feed type) that
   the clock/heartbeat job can target independently of the full feed list.
2. The selected subset can be run on its own cadence without changing the cadence
   of everything else.
3. The selection is config-driven, so a cohort can be turned on or off without a
   code change.

### Notes

- This is about carving out an analysis cohort, **not** about dropping slow or
  broken feeds — it's the mechanism the TSP high-frequency story builds on.
- The full feed list is assembled today in
  `services/gtfs-rt-archiver/gtfs_rt_archiver/heartbeat.py::messages()` from the
  gzipped download configs, filtered only by `feed_type`; this adds a subset
  selector to that.

tihuang02:
# Description

Adds an optional **high-frequency lane** to the GTFS-RT archiver: a parallel copy of the existing clock → heartbeat → fetch pipeline that archives a named cohort of feeds at a 3-second cadence for time-boxed studies (e.g. transit signal priority work needing sub-20s vehicle position samples).

It runs the same code from the same source zip and differs only in configuration. It is entirely off by default — `high_frequency_cohort` defaults to `[]`, which sets `count = 0` on every resource in the lane, so there is no scheduler, no workflow execution, and no idle instance until someone deliberately turns it on.

- **Configuration-driven toggle.** `high_frequency_cohort` (a list of download config names) gates the whole lane. Terraform variable validation caps it at 2 feeds and enforces a cadence floor of 3 seconds that divides evenly into 60. Names match the download config `name` field case-insensitively and tolerant of whitespace; an unmatched name logs a warning and archives nothing rather than silently selecting everything.
- **Fixed 3s grid from a dedicated Cloud Workflow clock** (`clock_high_frequency.yaml`). Each tick sleeps until an absolute batch time, so a slow publish does not accumulate drift, and consecutive ticks stay exactly one cadence apart across execution boundaries. The clock stamps `batch_at` onto each message; `Configuration` then derives `dt`/`hour`/`ts` from that single instant instead of `publish_time`, which is what keeps 20 ticks per minute in 20 distinct object paths instead of collapsing into 3. Production messages carry no `batch_at` and keep the existing 20-second floor unchanged.
- **Separate destination bucket** — `calitp-gtfs-rt-raw-high-frequency` (staging: `calitp-staging-gtfs-rt-raw-high-frequency`), STANDARD class, 180-day delete lifecycle rule, no retention policy. Kept separate on purpose: cadence is not recorded downstream, so mixing 3-second samples into the standard bucket would shift the published GTFS quality scores for that agency (`rt_20sec_vp` measures feed staleness at scrape time, which polling faster improves without the producer changing anything).
- **Blast-radius guardrails.** The lane gets its own Cloud Functions instance pool (`max_instance_count = 5`), so a misconfigured cohort cannot starve the production archiver's pool. `REQUEST_READ_TIMEOUT` is pinned below the cadence so requests cannot stack against a slow feed. `CALITP_GTFS_RT_FAIL_ON_OVERWRITE` turns a colliding object path into a logged `PreconditionFailed` rather than silent data loss — set only on this lane, since production relies on benign Pub/Sub redelivery rewriting the same object. Cohort feeds also keep running on the standard 20s clock, so the production RT dataset never develops a gap for a study agency.
- **Separate production and staging configurations** (two `high_frequency.tf` files, both defaulting to off).
- **README runbook** covering agency consent, enabling/disabling, the `{"limit": 1}` canary via the Cloud Scheduler payload, where the data lands, and how to stop the lane immediately by pausing the scheduler.

Resolves #5566

## Type of change

- [x] New feature
- [ ] Documentation

## How has this been tested?

**Unit tests (CI `test` job — passing).** 20 new tests, no new VCR cassettes required:

- `tests/test_configuration.py::TestConfigurationBatchAt` — `ts()` is exact and unfloored when `batch_at` is present; `dt`/`hour`/`ts` stay consistent across an hour boundary (published 12:59:59, batched 13:00:02), which the RT parser cross-checks; and the core regression guard: 20 ticks on a 3-second grid produce 20 distinct `destination_prefix()` values, where the `publish_time` path produces 3.
- `tests/test_configuration.py::test_json_omits_batch_at` — `Configuration.json()` is embedded in `PARTITIONED_ARTIFACT_METADATA`, so its shape must not change.
- `tests/test_heartbeat.py::TestHeartbeatCohort` — the three cohort states are distinct (env var absent → no filtering; `"[]"` → fail closed and publish nothing; a list → publish only those feeds), name matching is case- and whitespace-insensitive, a typo selects nothing rather than everything, the cohort bypasses the 5-minute service-alert throttle, `limit` applies after the cohort filter, and production messages still omit `batch_at`.

Regression guards on the production path are deliberate throughout: without `batch_at`, every accessor and every message shape is byte-identical to `main`.

**Terraform plan (CI, both environments — passing).** With the cohort empty, the plan is what it should be — the lane itself does not materialize:

```
Plan: 3 to add, 2 to change, 1 to destroy.

  # google_storage_bucket.gtfs-rt-raw-high-frequency will be created
  # google_storage_bucket_iam_member.gtfs-rt-raw-high-frequency-writer will be created
  # google_storage_bucket_object.gtfs-rt-archiver must be replaced
  # google_cloudfunctions2_function.gtfs-rt-archiver will be updated in-place
  # google_cloudfunctions2_function.gtfs-rt-archiver-heartbeat will be updated in-place
```

The two in-place function updates and the object replacement are the existing production functions picking up the new source zip hash; no behavior change, since every new code path is gated on `batch_at` / the cohort env var being present.

**Not yet tested end-to-end against a live feed.** The lane has not been enabled in staging or production, so the clock workflow, the Eventarc wiring, and real 3-second archiving are unexercised. That is intentional — enabling requires agency agreement first (see the runbook) — but it means the first enablement should be treated as the real test, and canaried with `{"limit": 1}`.

No dbt models were changed.

## Post-merge follow-ups

- [ ] No action required

Merging deploys the lane **off**. The only production side effects are the creation of the `calitp-gtfs-rt-raw-high-frequency` bucket and its IAM binding, plus a source-zip redeploy of the two existing archiver functions. Production `terraform apply` is not auto-approved, so this needs the apply to be approved after merge.

Before actually running a study, in order:

1. **Get and record the agency's agreement** for the increased poll rate. A 3-second poll is roughly a seven-fold request-rate increase against their server, and the archiver has no rate-limit handling or backoff. If an agency blocks us in response, we lose that feed for the standard pipeline too.
2. **Enable the cohort** via a follow-up PR setting `high_frequency_cohort` in `iac/cal-itp-data-infra/gtfs-rt-archiver/us/variables.tf`. Canary with the Cloud Scheduler payload set to `{"limit": 1}` for the first hour, and check the logs for the unmatched-name warning rather than assuming silence means success.
3. **Turn it off when the study ends** — set the cohort back to `[]` and merge. Every lane resource is `count`-gated and will be destroyed; the bucket and collected data remain, subject to the 180-day lifecycle rule.

Steps 2 and 3 will happen well after this merge and are tracked separately from this PR; the runbook in `services/gtfs-rt-archiver/README.md` has the full procedure including how to pause the scheduler out of band for an immediate stop.


iac/cal-itp-data-infra-staging/gtfs-rt-archiver/us/high_frequency.tf
```diff
@@ -0,0 +1,258 @@
...
+  # retention period would leave a misconfiguration's objects stranded until it
+  # expired. Note the consequence -- with nothing blocking replacement, a
+  # colliding object path overwrites silently, which is why the archiver function
+  # below runs with CALITP_GTFS_RT_FAIL_ON_OVERWRITE.
+  lifecycle_rule {
+    action {
+      type = "Delete"
+    }
+    condition {
+      age = 180
```
tihuang02: longer than 180 days, or remove it to keep until manually removed.

vevetron:
In discussion with @mrtopsyt, I don't think it's justified to make the gtfs-rt archiver more complex for this use case.  Can we just duplicate the archiver python code for this problem/project.

mrtopsyt:
I agree that we should avoid making changes to the production archiver. Could we achieve that without having to make a copy of the existing production arcchiver by creating a new bucket similar to `calitp-gtfs-download-config` and pointing the environment variable `CALITP_BUCKET__GTFS_DOWNLOAD_CONFIG` to that bucket on the high frequency archiver only? Then we could just populate this new bucket manually for now, avoid making any changes to the archiver python code since it seems riskier, and have a separate PR to write something simple to put hardcoded config values in there?

I'm not opposed to creating a new copy of the archiver as an interim though - it could just be the simpler approach here and we can do something else later.

The variable is currently set here

https://github.com/cal-itp/data-infra/blob/465663ab3148a283419b826db2956dc05ab2c2ce/iac/cal-itp-data-infra-staging/gtfs-rt-archiver/us/service.tf#L53

And then used in the archiver here

https://github.com/cal-itp/data-infra/blob/e5c0eedf04d8cd21536a0651d5e8461e2299aca1/services/gtfs-rt-archiver/gtfs_rt_archiver/heartbeat.py#L70

(I think) we define it separately for Composer and for archiver tests.

PR #5708, opened 2026-09-10
Fix Airflow failure Slack log link for Caltrans SSO users

### LINKED ISSUE #5040: Bug: Airflow failure logs link in Slack notification gives `403` or prompts for login
**Describe the bug**

In the notifications for an Airflow failure in the #alerts-data-airflow Slack channel, there's a `Check Log` link. When @mrtopsyt and I tried to use one of those links today, it gave her a 403 error and for me it prompted me to log in (from a screen that doesn't allow SSO login flow -- have to enter a Google account, not a federated URL). In my case I was already authenticated in the given browser and confirmed I can access the logs directly within the UI. 

**To Reproduce**

[See example Slack message](https://jarvus.slack.com/archives/C09SNFWCU23/p1775400276341109) -- try to click the `Check Logs` link and see if it works. 

**Expected behavior**

The link should take you to an accessible screen and not prompt for login if you're already auth'd 

**Additional context**

Suspect this is low priority since it seems like we can in fact access the relevant screen by clicking around in the UI. Wanted to document though in case it affects Teams notifications too or anything else

jamalk23:
# Description

Resolves #5040

The `Check Log` link in Airflow failure alerts in `#alerts-data-airflow` gave Caltrans users a `403` or a Google-accounts login screen they can't get through, even when they could reach the same logs by navigating the Airflow UI.

**Root cause:** Composer serves each environment's Airflow UI on two hosts that differ only in auth — `*.composer.googleusercontent.com` (Google accounts only) and `*.composer.byoid.googleusercontent.com` (workforce/external identities). The alert built its link from `ti.log_url`, which always emits the Google-account host. Caltrans staff are Workforce Identity Federation users (`dot-ca-gov` pool) with no Google account, so on that host they can only ever get a 403 (browser holds an unauthorized Google session) or a dead-end `accounts.google.com` prompt (no Google session). This is a [documented Composer limitation](https://docs.cloud.google.com/composer/docs/composer-3/access-environments-with-workforce-identity-federation): Airflow-generated links only carry the Google-account URL and "must be adjusted" for external identities.

**Change:** `log_failure_to_slack` now posts a single **Check Log (Caltrans SSO)** link: the byoid deep link, wrapped in the environment's `_signin` endpoint, wrapped in `https://auth.cloud.google/signin/locations/global/workforcePools/dot-ca-gov/providers/dot-gcp?continueUrl=…`. Pre-filling the provider means users go through the normal Caltrans Microsoft login (or straight through if already signed in) and land directly on the failing task's logs tab — a bare byoid link instead strands signed-out users on an unexplained "Provider name" prompt. The `_signin` middle hop exists because `auth.cloud.google` rejects byoid URLs as `continueUrl` (400) but accepts `*.composer.cloud.google`.

The link intentionally routes only through Caltrans SSO. The few remaining Google-account users with Airflow UI access are slated for removal; in the meantime they can reach task logs through the Airflow UI directly.

The unique ID, region, and byoid host are all derived from `ti.log_url` at runtime, so the same code serves prod and staging, and non-Composer hosts (local dev) pass through unchanged. If Google ever changes the `_signin` URL shape, behavior degrades to the provider-name prompt, not a broken link.

Related observations from investigating (not addressed here):
- Failure **emails** (default Airflow template → `dds.app.notify@dot.ca.gov`) embed the same Google-account-only URL and have the same problem for federated users.

## Type of change

- [ ] New feature
- [ ] Documentation

## How has this been tested?

- Verified unauthenticated redirect behavior of each hop with `curl`:
  - the Google-account URL bounces to `accounts.google.com` (dead end for federated identities);
  - the byoid URL bounces to the federated `_signin` with the log-page destination preserved;
  - the chained sign-in URL 302s directly into the Caltrans Azure AD SAML flow with no provider prompt.
- Click-tested the full chained URL end to end in a browser: Caltrans SSO → logs tab of the intended task instance (`dbt_all.warehouse_test`).
- Asserted the new `caltrans_sso_log_url()` output is byte-for-byte identical to the manually click-tested URL, and that non-Composer hosts (e.g. `localhost`) pass through unchanged.
- `black` (23.1.0, repo pin) and `isort` (5.12.0) pass; file compiles.

## Post-merge follow-ups

- [x] No action required

(DAGs deploy through the normal sync on merge; the next task failure alert exercises the change. The email-link limitation above can be spun into a separate issue if we want to fix Teams/email notifications too.)


fsalemi (approved):
Looks good to me. Approved!
One minor thought: since `caltrans_sso_log_url()` is a new standalone URL transformation function, would it be worth adding a couple of unit tests for the Composer URL conversion and non-Composer pass-through, so the expected behavior is automatically verified if this function changes in the future?

Issue #5705, opened 2026-09-09
Bug: GTFS-RT parse/validate tasks page on-call for a transient crash that recovers on retry


vevetron:
**Describe the bug**

Whenever the hourly GTFS-RT parse/validate pipeline (`parse_and_validate_rt`) hits a
transient gRPC crash, on-call is paged even though the crash recovers on its own. The
`parse_rt_*` / `validate_rt_*` tasks abort with a SIGABRT out of the gRPC event engine:

```
ev_epoll1_linux.cc:373] (event_engine) Epoll1Poller encountered epoll_wait error: Bad file descriptor
airflow.exceptions.AirflowException: Bash command failed. The command returned a non-zero exit code -6.
```

This is ~96% of all crashes on the DAG over the last 30 days (2,886 of ~3,000 attempts).
Because the DAG sets **no explicit `retries` / `retry_delay`** (`airflow/dags/parse_and_validate_rt.py`),
when the crash recurs a few times within one hour the task exhausts the default budget,
is marked FAILED, and fires a failure email + Slack alert. It peaked mid-August
(11 terminal failures on Aug 18) and is now a sporadic tail — Aug 31, Sep 1, Sep 9.

**To Reproduce**

Intermittent; not deterministically reproducible. Evidence from prod (`calitp-composer3`):

```
gcloud logging read 'resource.type="cloud_composer_environment"
  resource.labels.environment_name="calitp-composer3"
  labels."workflow"="parse_and_validate_rt"
  textPayload:"exit code -6"' --project=cal-itp-data-infra --freshness=30d
```

**Expected behavior**

Transient SIGABRT crashes are absorbed by task retries and do not page on-call. On-call
only hears about it when the *same* task fails repeatedly across runs.

**Additional context**

Known transient on Composer 3 (also seen in `copy_production_to_staging`). The August wave
lines up with unpinned google/gRPC dependency drift. Likely fix: add explicit `retries`
(e.g. 3–4) + a short `retry_delay` to the mapped `parse_rt_*` / `validate_rt_*` tasks;
longer-term, pin the google/gRPC deps.

PR #5716, opened 2026-09-15
Increase GTFS-RT parse and validate retries

fsalemi:
## Description

The `parse_and_validate_rt` DAG occasionally encounters transient gRPC crashes that cause mapped `parse_rt_*` and `validate_rt_*` tasks to exit with code `-6` (`SIGABRT`). These failures generally recover on retry or in a subsequent hourly DAG run, but can currently exhaust the inherited retry budget and trigger unnecessary email and Slack notifications.

This change adds an explicit retry policy to the mapped parse/validate tasks, increasing their tolerance for these transient failures before they are marked as failed and on-call notifications are triggered.

Ref: https://github.com/cal-itp/data-infra/issues/5705 

## Changes

- Add 4 explicit retries to the mapped `parse_rt_*` and `validate_rt_*` tasks.
- Retain the existing 5-minute retry delay between attempts.
- Limit the retry configuration to the mapped parse/validate `BashOperator` tasks rather than applying it to all tasks in the DAG.

This is intended as a minimal first step to reduce unnecessary notifications from transient crashes. 

## Type of change

- [ ] New feature
- [ ] Documentation

## How is this PR tested?

- Deployed the changes to the staging Airflow environment.
- Ran the `parse_and_validate_rt` DAG successfully in staging.
- Verified that the mapped parse/validate tasks are configured with `retries = 4` and `retry_delay = 0:05:00`.

image

## Post-merge follow-ups

Monitor the `parse_and_validate_rt` DAG after deployment to determine whether the additional retries sufficiently reduce unnecessary notifications from transient crashes. If excessive notifications continue for failures that recover in subsequent hourly runs, consider suppressing alerts until the failure persists across runs.

- [ ] No action required


vevetron (approved):
WIP Test.

Issue #5727, opened 2026-09-17
Epic: Compact the GTFS-RT raw archive into cold storage


vevetron:
The prod bucket `calitp-gtfs-rt-raw-v2` holds ~2.2 billion tiny objects (~44 TiB)
on Standard storage — ~$1,036/mo and rising ~$24/mo every month — almost all of it
written once and never read again. Compacting each `(feed_type, hour)` into one
compressed file in Archive-class storage cuts that to ~$35/mo, saving ~$12k in the
first year with a ~$1,000–1,500 one-time cost (payback ~5–6 weeks).

The sequence is inherently safe — nothing is deleted until the final step, after
the entire backlog is verified, so every earlier step is freely reversible. The
work neither waits on nor conflicts with the Pub/Sub archiver redesign (#5581):
the same compactor becomes that design's hourly rollup.

Relates to #5442, #5663, #5581.

Issue #5726, opened 2026-09-17
Delete the original tiny objects from the GTFS-RT raw bucket


vevetron:
Once the entire backlog is compacted and verified, purge the ~2.2 billion tiny
originals so `calitp-gtfs-rt-raw-v2` Standard storage drops to the ~14-day hot
window. Deletion itself is free (lifecycle `Delete` has no per-object charge and
Standard has no early-delete fee); the only avoidable cost is the ~$240
soft-delete tail.

In order:

1. Protect the archive copy first: Terraform an **unlocked 5-year
   `retention_policy`** onto the compacted bucket, so the data we are about to
   make the only copy cannot be deleted or replaced before it is 5 years old.
2. Terraform: remove the raw bucket's (unlocked) 1-year `retention_policy` —
   until it is gone, nothing younger than 1 year can be deleted, even by
   lifecycle.
3. Decide soft-delete: set retention to 0 before the purge to skip the ~$240
   tail, or keep the 7-day undo window as cheap insurance.
4. Terraform: add the lifecycle rule `Delete` at `age > 14d` — only after the
   full backlog has verified manifests, since an age rule is bucket-wide and
   would otherwise delete uncompacted days. This rule then stays on forever as
   the steady state (compactor archives at ~D−7, lifecycle purges at 14d,
   leaving a week of margin if compaction ever falls behind).

Done when `storage/object_count` for the bucket has dropped to the hot-window
level and raw-bucket storage spend sits at ~$12/mo (14-day window) and stays
flat, with all history in the compacted archive (~$28/mo).

Full mechanics and cost table: the spike's COSTS.md deletion section
(https://github.com/vevetron/gtfs-rt-compaction-spike).

Issue #5722, opened 2026-09-17
Create buckets for the compacted GTFS-RT archive in staging and prod


vevetron:
Terraform two new buckets (staging + prod, us-west2) to hold the compacted
GTFS-RT archive, and record the format decision so everything downstream builds
against it once.

Bucket settings:

- Lifecycle: `SetStorageClass → ARCHIVE` at 30 days. Files land Standard so
  verification and any near-term reprocessing read for free; with only ~72
  large files/day, the per-object transition cost is negligible.
- Lifecycle: `Delete` at age 5 years — the archive holds data 5 years, then it
  disappears automatically. (Age counts from object write, so backfilled history
  lives ~5 years from compaction; steady-state files written at ~D−7 track data
  age closely.)
- **No `retention_policy` at creation** — a retention policy blocks overwrites,
  and the pilot will rewrite files. After the archive is verified, add an
  **unlocked 5-year `retention_policy`** so verified files can't be deleted or
  replaced early; it then aligns exactly with the Delete rule (objects become
  deletable at the moment lifecycle purges them). Sequenced in the deletion
  issue: protect the archive copy before purging the originals.
- Uniform bucket-level access, soft-delete default is fine.

Format decisions:

- One file per `(feed_type, hour)` — hourly files in daily directories:
  `{feed_type}/dt=YYYY-MM-DD/hour=HH.parquet`
- parquet + zstd; each record carries the raw protobuf bytes verbatim plus its
  `PARTITIONED_ARTIFACT_METADATA`, `base64_url`, and `ts`.

Done when both buckets exist via the normal Terraform workflow.

PR #5728, opened 2026-09-17
Add compacted GTFS-RT archive buckets in staging and prod

vevetron:
# Description

Creates the destination buckets for the compacted GTFS-RT archive — one
parquet-zstd file per feed type per hour instead of ~27k tiny objects (see
epic #5727 for the full plan and economics).

- prod `calitp-gtfs-rt-raw-compacted`: Standard on write (so verification
  reads are free), lifecycle `SetStorageClass → ARCHIVE` at 30 days,
  `Delete` at 5 years. 

  - Intentionally no `retention_policy` yet — it would
  block rewrites during the pilot; an unlocked 5-year policy is added after
  verification (#5726).
- staging `calitp-staging-gtfs-rt-raw-compacted`: added to
  `environment_buckets` (inherits staging's standard 90-day cleanup) plus a
  remote-state output for the compactor DAG.

Works toward #5722

## Type of change

- [x] New feature

## How has this been tested?

Not tested

## Post-merge follow-ups
There will be modifications to these buckets later.


iac/cal-itp-data-infra-staging/gcs/us/variables.tf
```diff
@@ -8,6 +8,7 @@ locals {
     "calitp-staging-gtfs-rt-archiver",
     "calitp-staging-gtfs-download-config",
     "calitp-staging-gtfs-download-config-test",
+    "calitp-staging-gtfs-rt-raw-compacted",
```
tihuang02, line 11: I think there is a good chance this bucket won't share the policies (iam, lifecycle) with others. Perhaps not to use `for_each` and rather creating it alone separately?

